PR #160, opened 2026-03-10
fix(dungeon): back navigation and hierarchical docs browser in crawl

lancekrogers:
## Summary

- **Consistent back navigation**: Each item now keeps a stable Step 1 action menu, and cancelling child pickers returns to that menu for the same item instead of accidentally advancing the crawl.
- **Root vs child exit behavior**: In child pickers (`Move` status picker and docs routing browser), `Esc` goes back and `Ctrl-C` aborts the crawl. `Skip` and `Quit` remain explicit Step 1 actions.
- **Dedicated docs destination browser**: Replaces the broken docs routing prompt with a Bubble Tea browser that supports list navigation, drill-down, backtracking, and typed filtering while showing the current path and visible child nodes.
- **Selective auto-commit for crawl moves**: Crawl auto-commit now commits the moved item paths and reports real commit failures separately from no-op commits.

## Docs Browser Controls

- `j/k` or arrow keys: move selection
- `Enter`: select the current directory as the destination
- `l` or `Right`: drill into the selected directory
- `h`, `Left`, or `Esc`: go back one level; at root, return to Step 1
- `Ctrl-C`: abort the crawl
- Typing or `/`: edit the current level filter

## Test plan

- [x] Run `go test ./internal/dungeon/...`
- [x] Run `go test ./internal/git/...`
- [x] Run `go test ./internal/git/commit/...`
- [x] Run `go test ./cmd/camp/...`
- [x] Manual: `camp dungeon crawl --triage` → `Move`, then `Esc` returns to Step 1 for the same item
- [x] Manual: `camp dungeon crawl --triage` → `Route to docs`, then `Esc` goes up one level and at root returns to Step 1
- [x] Manual: in docs routing, `Enter` selects the current directory, `l` drills deeper, and typed filtering remains editable after drill-down
- [ ] `just test all` currently has one remaining integration failure outside the stale review thread: `TestDungeonMove_TriageDirectToStatusWithCommit`

### DIFF
--- a/cmd/camp/dungeon_crawl.go
+++ b/cmd/camp/dungeon_crawl.go
@@ -176,7 +176,7 @@ func commitCrawlChanges(ctx context.Context, cmdCtx *dungeonCommandContext, tria
 		relDungeon = cmdCtx.Dungeon.DungeonPath
 	}
 
-	files := crawlMovedItemPaths(relDungeon, triage, inner)
+	files := crawlCommitPaths(relDungeon, triage, inner)
 	preStaged, err := stageTrackedCrawlSourceDeletions(
 		ctx,
 		cmdCtx.CampaignRoot,
@@ -293,6 +293,23 @@ func crawlMovedItemPaths(relDungeon string, summaries ...*dungeon.CrawlSummary) 
 
 	sort.Strings(paths)
 	return paths
+}
+
+func crawlCommitPaths(relDungeon string, summaries ...*dungeon.CrawlSummary) []string {
+	paths := crawlMovedItemPaths(relDungeon, summaries...)
+
+	logPath, ok := cleanCrawlCommitPath(filepath.Join(relDungeon, "crawl.jsonl"))
+	if !ok || !isSafeCrawlCommitPath(logPath) {
+		return paths
+	}
+
+	for _, path := range paths {
+		if path == logPath {
+			return paths
+		}
+	}
+
+	return append(paths, logPath)
 }
 
 func crawlMoveDestinationBase(relDungeon, status string) (string, bool) {
--- a/cmd/camp/dungeon_crawl_test.go
+++ b/cmd/camp/dungeon_crawl_test.go
@@ -230,6 +230,31 @@ func TestCrawlMovedItemPaths(t *testing.T) {
 	}
 }
 
+func TestCrawlCommitPaths(t *testing.T) {
+	summary := &dungeon.CrawlSummary{
+		MovedItems: map[string][]string{
+			"docs/api":  {"routed.md"},
+			"completed": {"finished.md"},
+		},
+	}
+
+	got := crawlCommitPaths("workflow/design/dungeon", summary)
+	want := []string{
+		"docs/api/routed.md",
+		"workflow/design/dungeon/completed/finished.md",
+		"workflow/design/dungeon/crawl.jsonl",
+	}
+
+	if len(got) != len(want) {
+		t.Fatalf("crawlCommitPaths() len=%d, want=%d (%v)", len(got), len(want), got)
+	}
+	for i := range want {
+		if got[i] != want[i] {
+			t.Fatalf("crawlCommitPaths()[%d]=%q, want=%q (full=%v)", i, got[i], want[i], got)
+		}
+	}
+}
+
 func TestCrawlSourceDeletionPaths(t *testing.T) {
 	summary := &dungeon.CrawlSummary{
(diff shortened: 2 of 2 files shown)


internal/dungeon/triage.go
```diff
@@ -55,134 +50,144 @@ func RunTriageCrawl(ctx context.Context, svc *Service, parentPath string) (*Craw
...
-		Placeholder("architecture/api").
+		Title(fmt.Sprintf("Route %s to docs/ subdirectory:", itemName)).
+		Description("Type or press Tab to browse available subdirectories.").
 		Value(&destination)
 
 	if len(suggestions) > 0 {
-		input = input.Suggestions(suggestions)
+		input = input.
+			Placeholder(suggestions[0]).
+			Suggestions(suggestions)
```
obey-agent: This assumes `huh.NewInput().Suggestions(...)` gives users a practical way to browse the existing docs destinations, but in the current CLI flow those suggestions are not actually surfaced. On the current branch, `camp dungeon crawl --triage` -> `Route to docs` only prints `Route note.md to docs/ subdirectory:` with no visible options, and a bad value like `bogus` then errors and skips the item. So this reintroduces the original problem: the user still doesn't have a safe, discoverable way to choose an existing docs path.

obey-agent (approved):
Re-reviewed at 7ce6a36. The latest docs-browser refactor fixes the runtime issue I was blocking on: the browser is now visible and stateful in the actual CLI flow, with working Tab/Shift+Tab cycling, Enter drill-down/selection, Esc back one level, and Esc at root returning to Step 1 of triage. I reran `go test -count=1 ./internal/dungeon/...`, rebuilt the CLI, and smoke-tested both nested selection (into `docs/business/pricing`) and nested/root Escape behavior in fresh temp campaigns. `go test -count=1 ./cmd/camp/...` still fails in the existing leverage tests, and I have already reproduced that same failure on `origin/main`, so I treated it as baseline noise rather than a regression from this PR.

obey-agent (changes requested):
P2: the latest PR head fixes several of the picker/runtime issues, but the core Step 1 cancel contract is still wrong. The PR summary and test plan both say Escape at Step 1 should skip the current item while explicit `Quit` stops the crawl. On the current head, cancelling the Step 1 action form still returns `ErrCrawlAborted`, and `cmd/camp/dungeon_crawl.go` still turns that into a whole-crawl cancellation. That means `Esc` and `Quit` are still collapsed together in the shipped flow. I would restore the earlier `summary.Skipped++` / `break itemLoop` behavior here and keep `ErrCrawlAborted` for true abort paths such as Ctrl+C from the TUI pickers.

Validation on the current head: `go test -count=1 ./internal/dungeon/... ./internal/git/... ./cmd/camp -run 'TestDungeonCrawl|TestRunDungeonCrawl|TestExecuteDungeonCrawl|TestCommit'` passed, and `go build -o /tmp/camp-pr160-open-current ./cmd/camp` passed.

obey-agent (commented):
Updated the PR body/test plan to match the current head.

The latest Step 1 `Esc` comment was valid relative to the old PR description. The branch no longer uses the earlier "Step 1 Esc skips item" contract; the intended/current model is:
- Step 1 remains the root action menu for the current item
- Child pickers (`Move` status picker and docs routing) use `Esc` to return to Step 1
- `Ctrl-C` aborts the crawl
- `Skip` and `Quit` remain explicit Step 1 actions

The older docs-browser blockers were on superseded commits and were later re-reviewed/approved at `535e65c` and `7ce6a36`, so I am treating those as stale review history rather than current-head regressions.

Separate from the review-thread mismatch above, `just test all` is still surfacing one real integration failure on the current head: `TestDungeonMove_TriageDirectToStatusWithCommit` leaves the tracked source deletion behind after the supposed auto-commit. I’m treating that as the remaining actual blocker rather than a spec-review issue.

obey-agent (approved):
Re-reviewed at `3df8090`.

The previous blocker is fixed. `commitCrawlChanges()` now builds its scope from `crawlCommitPaths(...)`, which explicitly adds `dungeon/crawl.jsonl` back into the auto-commit set alongside the moved paths and tracked source deletions. That closes the repo-dirty regression from the prior head.

Validation on the updated head:
- `go test -count=1 ./internal/dungeon/... ./internal/git/... ./cmd/camp/...`
- `just test all`
- `just test all` is green, including `123/123` integration tests

I did not find a remaining blocking correctness issue on the current head.

PR #168, opened 2026-03-13
feat: add quest scaffold and dev quest CLI

lancekrogers:
## Summary
- scaffold `.campaign/quests` during `camp init` and restore it through `camp init --repair`
- add the dev-only `camp quest` command group with create/list/show/edit/rename and lifecycle operations backed by `internal/quest`
- carry quest IDs into auto-commit tags and extend scaffold, manifest, release-profile, and quest lifecycle test coverage

## Testing
- just test unit
- go test -tags=dev ./...

### DIFF
--- a/cmd/camp/quest/create.go
+++ b/cmd/camp/quest/create.go
@@ -1,6 +1,6 @@
 //go:build dev
 
-package main
+package quest
 
 import (
 	"fmt"
@@ -38,7 +38,7 @@ Examples:
 }
 
 func init() {
-	questCmd.AddCommand(questCreateCmd)
+	Cmd.AddCommand(questCreateCmd)
 
 	flags := questCreateCmd.Flags()
 	flags.String("purpose", "", "Short purpose statement")
--- a/cmd/camp/quest/edit.go
+++ b/cmd/camp/quest/edit.go
@@ -1,6 +1,6 @@
 //go:build dev
 
-package main
+package quest
 
 import (
 	"fmt"
@@ -32,7 +32,7 @@ Examples:
 }
 
 func init() {
-	questCmd.AddCommand(questEditCmd)
+	Cmd.AddCommand(questEditCmd)
 	questEditCmd.Flags().Bool("no-commit", false, "Don't create a git commit")
 	questEditCmd.ValidArgsFunction = completeQuestSelector
 }
--- a/cmd/camp/quest/helpers.go
+++ b/cmd/camp/quest/helpers.go
@@ -1,6 +1,6 @@
 //go:build dev
 
-package main
+package quest
 
 import (
 	"context"
--- a/cmd/camp/quest/helpers_test.go
+++ b/cmd/camp/quest/helpers_test.go
@@ -1,6 +1,6 @@
 //go:build dev
 
-package main
+package quest
 
 import (
 	"context"
--- a/cmd/camp/quest/lifecycle.go
+++ b/cmd/camp/quest/lifecycle.go
@@ -1,6 +1,6 @@
 //go:build dev
 
-package main
+package quest
 
 import (
 	"fmt"
@@ -83,7 +83,7 @@ func init() {
 		questArchiveCmd,
 		questRestoreCmd,
 	} {
-		questCmd.AddCommand(sub)
+		Cmd.AddCommand(sub)
 		sub.Flags().Bool("no-commit", false, "Don't create a git commit")
 		sub.ValidArgsFunction = completeQuestSelector
 	}
--- a/cmd/camp/quest/list.go
+++ b/cmd/camp/quest/list.go
@@ -1,6 +1,6 @@
 //go:build dev
 
-package main
+package quest
 
 import (
 	"github.com/Obedience-Corp/camp/internal/quest"
@@ -25,7 +25,7 @@ var questListCmd = &cobra.Command{
 }
 
 func init() {
-	questCmd.AddCommand(questListCmd)
+	Cmd.AddCommand(questListCmd)
 	questListCmd.Flags().BoolVar(&questListAll, "all", false, "Include dungeon quests")
 	questListCmd.Flags().BoolVar(&questListDungeon, "dungeon", false, "Show only dungeon quests")
 	questListCmd.Flags().BoolVar(&questListJSON, "json", false, "Emit JSON output")
--- a/cmd/camp/quest/rename.go
+++ b/cmd/camp/quest/rename.go
@@ -1,6 +1,6 @@
 //go:build dev
 
-package main
+package quest
 
 import (
 	"fmt"
@@ -27,7 +27,7 @@ Examples:
 }
 
 func init() {
-	questCmd.AddCommand(questRenameCmd)
+	Cmd.AddCommand(questRenameCmd)
 	questRenameCmd.Flags().Bool("no-commit", false, "Don't create a git commit")
 	questRenameCmd.ValidArgsFunction = completeQuestSelector
 }
--- a/cmd/camp/quest/root.go
+++ b/cmd/camp/quest/root.go
@@ -1,15 +1,16 @@
 //go:build dev
 
-package main
+package quest
 
 import (
-	"github.com/Obedience-Corp/camp/internal/commands/release"
 	"github.com/spf13/cobra"
 )
 
-var questCmd = &cobra.Command{
-	Use:   "quest",
-	Short: "Manage working contexts within a campaign",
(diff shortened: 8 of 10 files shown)


internal/quest/service.go
```diff
@@ -0,0 +1,536 @@
...
+		return nil, err
+	}
+	if err := WriteActiveID(ctx, s.campaignRoot, q.ID); err != nil {
+		return nil, err
+	}
+
+	return &MutationResult{
+		Quest:     q,
+		Files:     []string{newDir, ActiveQuestPath(s.campaignRoot)},
+		PreStaged: []string{oldDir},
```
obey-agent: `PreStaged` is documented as "already staged externally", but nothing stages `oldDir` before `autoCommitQuest` hands this result to `commit.Quest`. `git add newDir` leaves the old tracked path as an unstaged delete, so the default auto-commit here records the restored quest in its new location while keeping the old path in the commit tree. I reproduced that with a minimal repo: `git add new && git commit --only new/file.txt` leaves both `old/file.txt` and `new/file.txt` in `HEAD`. Please either stage `oldDir` before returning or teach the commit helper to stage `PreStaged` paths itself.

obey-agent (commented):
## Follow-up Review (commit 1a5b7c95)

### Previous Critical Issue — Fixed

**Auto-commit staging bug**: `autoCommitQuest` now calls `git.RemoveCached()` on `PreStaged` paths before invoking `commit.Quest`. This correctly stages the deletion side of `os.Rename` moves for `complete`, `archive`, and `restore` lifecycle commands. Uses `--ignore-unmatch` for safety when paths aren't tracked.

### Previous Important Issue — Fixed

**Dead code in `moveToStatus`**: The redundant `StatusArchived && StatusCompleted` conditional block has been removed.

### Remaining Suggestions (Non-blocking)

1. **Integration test**: No test yet exercises the auto-commit path in a real git repo. Consider adding one to prevent regression.
2. **`fmt.Errorf` usage**: Several places in `quest.go` and `service.go` use `fmt.Errorf` instead of `camperrors` — minor consistency cleanup.
3. **`containsStatus` in service.go**: LSP flags this can be simplified with `slices.Contains`.

### Verdict: APPROVE

Critical and important items are resolved. The auto-commit staging fix is correct and safe. Remaining items are minor suggestions that can be addressed in follow-up commits.

cmd/camp/quest.go
```diff
@@ -0,0 +1,34 @@
...
+  camp quest pause qst_20260313_abc123
+  camp quest restore qst_20260313_abc123`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return cmd.Help()
+	},
+}
+
+func init() {
+	release.MarkDevOnly(questCmd)
+	questCmd.GroupID = "planning"
```
obey-agent: Grouping `quest` under `Planning Commands` conflicts with the semantics in this same file. If quests are explicitly "not planning structures," they should not show up to users as part of the planning command family in `camp --help`. Please move this to a non-planning group or introduce a more appropriate group for execution/workspace context.

obey-agent:
## Design Spec Review — Quest System

The quest design specification has been finalized at:
`docs/superpowers/specs/2026-03-13-camp-quest-system-design.md` (in the campaign repo)

Below are the areas where this PR diverges from the approved spec. These need to be resolved before merge.

---

### BLOCKER: Commit Tag Format

**Spec says:** Single bracket with quest ID appended after campaign ID:
```
[OBEY-CAMPAIGN-8deed8b4-qst_20260312_8f3k2m] Action: Subject
```

**PR implements:** Double brackets — separate `[campaign][quest]` tags:
```
[OBEY-CAMPAIGN-8deed8b4][qst_20260312_8f3k2m] Action: Subject
```

**Files:** `internal/git/campaign_tag.go` — `FormatQuestTag` and `FormatContextTags` produce double brackets.

**Required change:** Merge quest ID into the campaign bracket. `FormatCampaignTag` should accept an optional quest ID and produce `[OBEY-CAMPAIGN-{id}-{questID}]` when present. Remove `FormatQuestTag` as a separate function.

---

### BLOCKER: Active Quest File (.active)

**Spec says:** No `.active` file. No `camp quest use` command. Multiple quests can be open simultaneously. Quest context comes only from:
1. `--quest` flag (explicit)
2. `CAMP_QUEST` env var (set by obey/festival_app)

**PR implements:** `.active` file at `.campaign/quests/.active` with `ReadActiveID`/`WriteActiveID`. Every status change writes to `.active`. The scaffold creates `.active` during init.

**Rationale for spec decision:** Single-active-quest tracking assumes one quest at a time, but multiple quests can be open simultaneously. Active quest binding is an agent-session concern that belongs in obey/festival_app, not in camp. This is deferred until those systems are operational.

**Required change:** Remove `ActiveFileName`, `ReadActiveID`, `WriteActiveID`, and all `WriteActiveID` calls from service methods. Remove `.active` from scaffold. Simplify `ResolveContext` to: explicit flag → env var → no quest context.

---

### WARNING: TUI Approach

**Spec says:** Use the same vim-style BubbleTea editor from `internal/intent/tui/vim/` for the description field, matching the `camp intent add` experience. Multi-step flow: name → purpose → description (vim editor).

**PR implements:** `huh` forms for name/purpose input + external `$EDITOR` for YAML editing via `CreateWithEditor`.

**Discussion:** The spec calls for the vim editor because it provides a better inline terminal experience than bouncing to an external editor. The `huh` forms for name/purpose are fine, but the description entry should use the vim editor component, not external `$EDITOR` on raw YAML. This matches the UX quality bar of `camp intent add`.

---

### WARNING: Missing `tags` Field in Spec (spec will be updated)

The PR correctly implements a `Tags []string` field on the Quest struct. The spec omitted this — we'll update the spec to include it.

---

### WARNING: Missing Commit Actions in Spec (spec will be updated)

The PR correctly implements `QuestRename` and `QuestEdit` commit actions. The spec only listed 6 actions (Create, Pause, Resume, Complete, Archive, Restore). We'll update the spec to include all 8.

---

### WARNING: Missing `--no-commit` Flag in Spec (spec will be updated)

The PR adds `--no-commit` to lifecycle commands. Good addition — the spec will be updated to document this.

---

### NOTE: `completed → archived` Transition

The PR allows archiving a completed quest (`StatusCompleted` is a valid source for `Archive`). The spec didn't list this transition explicitly. This is correct behavior — we'll update the spec.

---

### NOTE: Work Item Linking (Not Yet Implemented)

The spec includes `camp quest link`/`unlink`/`links` commands and a `links` array in quest.yaml for tracking intents, designs, festivals, etc. linked to a quest. This is not in the PR yet and can be a follow-up.

---

### Summary

| Item | Verdict | Action |
|------|---------|--------|
| Commit tag format (single vs double bracket) | BLOCKER | PR must change to single bracket |
| `.active` file | BLOCKER | PR must remove; defer to obey/festival_app |
| TUI approach (vim editor vs huh + $EDITOR) | WARNING | PR should use vim editor for description |
| Tags field | WARNING | Spec will be updated to match PR |
| QuestRename/QuestEdit actions | WARNING | Spec will be updated to match PR |
| `--no-commit` flag | WARNING | Spec will be updated to match PR |
| `completed → archived` | NOTE | Spec will be updated |
| Work item linking | NOTE | Follow-up PR |

obey-agent (commented):
## Follow-up Review — commits `70304dbd`, `1bf19055`

### `repair.go`: derive dungeon paths from `dungeonscaffold.StandardStatuses` — Good

Replaces hardcoded `completed`, `archived`, `someday` strings with a loop over `dungeonscaffold.StandardStatuses`. This is the right fix — if a new dungeon status is added to the scaffold package, repair will automatically pick it up instead of silently missing it. Clean elimination of a drift risk.

### Quest examples reframed — Addresses prior review

Examples now use `q2-reliability`, `data-pipeline-rethink`, `customer-onboarding` instead of feature-sized names. Consistent with the "long-lived working context" framing from the earlier review.

No new issues.

obey-agent (commented):
## Follow-up Review — commit `179fbe35`

### Quest subpackage extraction — Clean

Moves quest commands from `package main` into `cmd/camp/quest/` with an exported `Cmd` pointer, following the exact same pattern as PR #170's package-ownership model. Registration shim (`quest_register.go`) handles `release.MarkDevOnly` and `rootCmd.AddCommand` from the parent package.

- `GroupID` moved from `init()` into the var declaration — consistent with #170's convention
- All subcommands register against `Cmd` instead of the old `questCmd`
- `//go:build dev` tag preserved on all files including the registration shim

No issues. Good alignment with the cmd reorganization direction.

PR #169, opened 2026-03-14
refactor(dungeon): reduce crawl commit path helper duplication (#165)

### LINKED ISSUE #165: refactor(dungeon): simplify crawl commit path helpers after PR 160
## Summary

Follow up on the crawl auto-commit scope hardening in PR #160 by refactoring the helper logic in `cmd/camp/dungeon_crawl.go`.

The current implementation is correct and fully green, but the crawl commit-path helpers accumulated duplicated path-cleaning, safety checks, and dedupe logic while fixing two regressions in the same area:
- staged rename source deletions dropping out of selective commit scope
- `dungeon/crawl.jsonl` dropping out of crawl auto-commit scope

## Problem

`cmd/camp/dungeon_crawl.go` now has overlapping guard logic across:
- `crawlMovedItemPaths`
- `crawlCommitPaths`
- `crawlMoveDestinationBase`
- `crawlSourceDeletionPaths`
- `cleanCrawlCommitName`
- `cleanCrawlCommitPath`
- `isSafeCrawlCommitPath`

This is serviceable but harder to reason about than it should be. The repeated `if !ok { return/continue }` and repeated path safety checks make future edits to crawl commit scope more error-prone.

## Goal

Refactor the crawl commit-path helpers to reduce duplication without changing behavior.

## Constraints

- Preserve current behavior exactly.
- Do not change the crawl UX or auto-commit contract.
- Keep protection against unsafe paths and traversal.
- Preserve inclusion of `dungeon/crawl.jsonl` in crawl auto-commit scope.
- Preserve rename/deletion correctness covered by existing tests.

## Suggested Direction

- Introduce one small helper for safe path normalization + dedupe append.
- Collapse repeated path validation branches where the helper already guarantees safety.
- Keep commit-scope construction readable by separating:
  - moved destination paths
  - tracked source deletions
  - crawl log inclusion

## Validation

At minimum rerun:
- `go test ./cmd/camp/...`
- `go test ./internal/git/... ./internal/git/commit/...`
- `just test all`

## Context

This should happen after PR #160 merges, not as part of reopening that PR’s already-validated behavior changes.

obey-agent:
## Summary

Closes #165.

- Introduce `crawlPathSet` with an `appendSafe` method to consolidate the repeated safe-path normalization + dedup pattern that was copy-pasted across `crawlMovedItemPaths`, `crawlCommitPaths`, and `crawlSourceDeletionPaths`.
- Collapse `crawlCommitPaths` to build paths directly via `crawlPathSet` rather than calling `crawlMovedItemPaths` and then manually checking for the log path; the log is now appended through the same `appendSafe` path, which handles the dedup naturally.
- Rename the inner closure in `crawlSourceDeletionPaths` to `appendSummarySourcePaths` to make its intent clearer.
- No behavior changes — all existing tests pass unchanged.

## Test plan

- [x] `go test ./cmd/camp/...` — all pass
- [x] `go test ./internal/git/... ./internal/git/commit/...` — all pass
- [x] `go test ./...` — full suite green
- [x] `go vet ./cmd/camp/...` — clean
- [x] `go fmt ./cmd/camp/...` — no changes (import reorder in gendocs.go from formatter included)

### DIFF
--- a/cmd/camp/dungeon_crawl.go
+++ b/cmd/camp/dungeon_crawl.go
@@ -271,18 +271,16 @@ func newCrawlPathSet() *crawlPathSet {
 }
 
 // appendSafe cleans path, verifies it is safe, and adds it if not already present.
-// Returns true when the path was added.
-func (s *crawlPathSet) appendSafe(path string) bool {
+func (s *crawlPathSet) appendSafe(path string) {
 	path = filepath.Clean(path)
 	if !isSafeCrawlCommitPath(path) {
-		return false
+		return
 	}
 	if _, exists := s.seen[path]; exists {
-		return false
+		return
 	}
 	s.seen[path] = struct{}{}
 	s.paths = append(s.paths, path)
-	return true
 }
 
 func (s *crawlPathSet) sorted() []string {
@@ -314,9 +312,11 @@ func crawlMoveDestinationBase(relDungeon, status string) (string, bool) {
 	return filepath.Join(cleanDungeon, cleanStatus), true
 }
 
-func crawlMovedItemPaths(relDungeon string, summaries ...*dungeon.CrawlSummary) []string {
-	ps := newCrawlPathSet()
-
+// populateMovedPaths appends the destination paths for all moved items in the
+// given summaries into ps. This is the single source of truth for the
+// summary → status → name iteration used by both crawlMovedItemPaths and
+// crawlCommitPaths.
+func populateMovedPaths(ps *crawlPathSet, relDungeon string, summaries ...*dungeon.CrawlSummary) {
 	for _, summary := range summaries {
 		if summary == nil || !summary.HasMoves() {
 			continue
@@ -335,7 +335,11 @@ func crawlMovedItemPaths(relDungeon string, summaries ...*dungeon.CrawlSummary) 
 			}
 		}
 	}
+}
 
+func crawlMovedItemPaths(relDungeon string, summaries ...*dungeon.CrawlSummary) []string {
+	ps := newCrawlPathSet()
+	populateMovedPaths(ps, relDungeon, summaries...)
 	return ps.sorted()
 }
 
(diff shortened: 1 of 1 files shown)


cmd/camp/dungeon_crawl.go
```diff
@@ -336,6 +314,62 @@ func crawlMoveDestinationBase(relDungeon, status string) (string, bool) {
...
+			if !ok {
+				continue
+			}
+			for _, name := range names {
+				cleanName, ok := cleanCrawlCommitName(name)
+				if !ok {
+					continue
+				}
+				ps.appendSafe(filepath.Join(base, cleanName))
+			}
```
obey-agent: **Medium: re-duplicated loop body**

This nested `summaries → status → names` loop is identical to the one in `crawlMovedItemPaths` above (lines 320-336). The old code delegated to `crawlMovedItemPaths(...)` and appended the log path — one call. This refactor replaces seen-map duplication with full-loop duplication.

Consider extracting a shared helper:
```go
func populateMovedPaths(ps *crawlPathSet, relDungeon string, summaries ...*dungeon.CrawlSummary) {
    for _, summary := range summaries {
        if summary == nil || !summary.HasMoves() {
            continue
        }
        for status, names := range summary.MovedItems {
            base, ok := crawlMoveDestinationBase(relDungeon, status)
            if !ok {
                continue
            }
            for _, name := range names {
                cleanName, ok := cleanCrawlCommitName(name)
                if !ok {
                    continue
                }
                ps.appendSafe(filepath.Join(base, cleanName))
            }
        }
    }
}
```

Then both `crawlMovedItemPaths` and `crawlCommitPaths` call it, keeping the iteration logic in one place.

cmd/camp/dungeon_crawl.go
```diff
@@ -259,57 +259,35 @@ func buildCrawlCommitMessage(campaignRoot, parentPath string, triage, inner *dun
...
+type crawlPathSet struct {
+	seen  map[string]struct{}
+	paths []string
 }
 
-func crawlCommitPaths(relDungeon string, summaries ...*dungeon.CrawlSummary) []string {
-	paths := crawlMovedItemPaths(relDungeon, summaries...)
+func newCrawlPathSet() *crawlPathSet {
+	return &crawlPathSet{seen: make(map[string]struct{})}
+}
```
obey-agent: **Minor: unused return value**

`appendSafe` returns `bool` but no call site checks the return. If this is intentionally fire-and-forget, consider either dropping the return value or adding a brief doc comment explaining why callers ignore it (e.g., "callers don't need to distinguish new vs. duplicate"). As-is it looks like an oversight.

PR #203, opened 2026-04-04
feat: add manual priority system to workitem TUI and CLI

lancekrogers:
## Summary

- Add a lightweight manual priority system to `camp workitem` so users can mark workitems as high/medium/low importance with instant visual feedback and persistent, self-cleaning state
- Priority store persists to `.campaign/settings/workitems.json` with atomic writes, lazy creation, and automatic cleanup of stale entries
- TUI supports interactive priority assignment via `P` key with single-keypress picker (h/m/l/0/Esc)
- Prioritized items sort into buckets (high > medium > low > unset) before recency-based ordering
- JSON schema bumped to `workitems/v1alpha3` reflecting the new sort order and `manual_priority` field

### Key changes

- **`internal/workitem/priority/`** — new package with `ManualPriority` type, `Store` struct, Load/Save/Set/Clear/Prune/Apply/SaveOrDelete operations
- **`internal/workitem/model.go`** — `ManualPriority string` field with `omitempty` JSON tag
- **`internal/workitem/sort.go`** — priority bucket as primary sort key
- **`internal/commands/workitem/workitem.go`** — pipeline: discover → load store → prune (full set) → apply → sort → filter → render
- **`internal/workitem/tui/`** — priority mode key handlers, colored H/M/L row badges, preview field, footer/help updates, cursor preservation after resort
- **`internal/scaffold/init.go`** — `settings/workitems.json` added to `.campaign/.gitignore` scaffold

Also includes a prior fix: restore intent explorer preview focus (`Tab` behavior, filter shortcuts, preview-focus reset).

### Design invariants

- Prune runs against the **full** discovery set before any view-level filters — running `camp workitem --type design` never erases priorities on intent items
- Empty stores are deleted (no empty artifacts on disk)
- Cursor follows the selected item by `WorkItem.Key` after priority-triggered resorts

## Test plan

- [x] `go test ./...` passes across full codebase (0 failures)
- [x] `go build ./...` and `go vet ./...` clean
- [x] Unit tests: ManualPriority Rank/Valid, Store CRUD, atomic write, nil-map init
- [x] Integration tests: filter safety (type/query/limit), stale pruning, full pipeline simulation
- [x] Sort tests: 8-scenario table-driven test for priority bucket ordering
- [x] TUI tests: preserveSelection (key found, not found, empty list), all existing TUI tests updated
- [ ] Manual TUI verification: priority assignment, badge rendering, persistence across restarts

### DIFF
--- a/internal/commands/workitem/workitem.go
+++ b/internal/commands/workitem/workitem.go
@@ -13,6 +13,7 @@ import (
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 	"github.com/Obedience-Corp/camp/internal/paths"
 	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
+	"github.com/Obedience-Corp/camp/internal/workitem/priority"
 	wktui "github.com/Obedience-Corp/camp/internal/workitem/tui"
 )
 
@@ -69,6 +70,26 @@ Examples:
 				return camperrors.Wrap(err, "discovering work items")
 			}
 
+			// Load priority store and prune stale entries against full discovery set.
+			storePath := priority.StorePath(campaignRoot)
+			store, err := priority.Load(storePath)
+			if err != nil {
+				return camperrors.Wrap(err, "loading priority store")
+			}
+			validKeys := make(map[string]bool, len(items))
+			for _, item := range items {
+				validKeys[item.Key] = true
+			}
+			if priority.Prune(store, validKeys) {
+				if err := priority.SaveOrDelete(storePath, store); err != nil {
+					return camperrors.Wrap(err, "saving pruned priority store")
+				}
+			}
+
+			// Apply priority overlay and re-sort with priority buckets.
+			items = priority.Apply(store, items)
+			wkitem.Sort(items)
+
 			items = wkitem.Filter(items, flagTypes, flagStages, flagQuery)
 			if flagLimit > 0 && flagLimit < len(items) {
 				items = items[:flagLimit]
@@ -85,9 +106,9 @@ Examples:
 				fmt.Println(items[0].AbsPath(campaignRoot))
 				return nil
 			case flagPrint:
-				return runTUI(ctx, items, true, campaignRoot, resolver)
+				return runTUI(ctx, items, true, campaignRoot, resolver, store, storePath)
 			default:
-				return runTUI(ctx, items, false, campaignRoot, resolver)
+				return runTUI(ctx, items, false, campaignRoot, resolver, store, storePath)
 			}
 		},
 	}
@@ -136,12 +157,12 @@ func outputJSON(campaignRoot string, items []wkitem.WorkItem) error {
 	return enc.Encode(payload)
 }
 
-func runTUI(ctx context.Context, items []wkitem.WorkItem, printOnly bool, campaignRoot string, resolver *paths.Resolver) error {
+func runTUI(ctx context.Context, items []wkitem.WorkItem, printOnly bool, campaignRoot string, resolver *paths.Resolver, store *priority.Store, storePath string) error {
 	if len(items) == 0 {
 		return fmt.Errorf("no work items found")
 	}
 
-	model := wktui.New(ctx, items, campaignRoot, resolver)
+	model := wktui.New(ctx, items, campaignRoot, resolver, store, storePath)
 	p := tea.NewProgram(model, tea.WithAltScreen())
 	result, err := p.Run()
 	if err != nil {
--- a/internal/intent/tui/explorer/layout.go
+++ b/internal/intent/tui/explorer/layout.go
@@ -85,6 +85,10 @@ func (m *Model) recalculateLayout() {
 		}
 	}
 
+	if !m.shouldShowPreview() {
+		m.previewFocused = false
+	}
+
 	// Notify user of layout mode change
 	if oldMode != m.layoutMode && m.ready {
 		switch m.layoutMode {
--- a/internal/intent/tui/explorer/preview_focus_test.go
+++ b/internal/intent/tui/explorer/preview_focus_test.go
@@ -1,0 +1,109 @@
+package explorer
+
+import (
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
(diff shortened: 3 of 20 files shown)


obey-agent (approved):
Verdict: Approve

Fixes broken preview focus flow in the intent explorer TUI. +147/-3 across 3 files. Root cause correctly identified: Tab was always routing to filter bar when preview wasn't focused, making it impossible to focus the preview pane.

**`update_keys.go` (+34/-3)**

Tab behavior reworked:
- When preview is NOT visible: Tab goes to filter bar (original behavior preserved)
- When preview IS visible: Tab toggles `previewFocused` between list and preview
- `previewFocused` cleared when routing to filter bar without preview

New `t` and `s` shortcuts directly open type (chip 0) and status (chip 1) filter dropdowns. This compensates for Tab no longer routing to the filter bar when preview is visible — filter access remains keyboard-reachable.

`focusFilterChip` helper: clears preview focus, sets filter bar focus, blurs all chips, focuses and opens the target chip. Bounds-checked with fallback to chip 0.

**`layout.go` (+4)**

`recalculateLayout` now clears `previewFocused` when `shouldShowPreview()` returns false. Prevents stale focus state when layout narrows and force-hides the preview.

**`preview_focus_test.go` (+109, new)**

Four regression tests using `makeTestModel`:
1. Tab toggles preview focus when visible (verifies focus stays on `focusList`, filter bar not activated)
2. `t`/`s` shortcuts open correct filter chips (verifies chip index, open state, focus transfer)
3. Explicit hide (`v` key) clears preview focus
4. Force-hide (narrow width) clears preview focus via `recalculateLayout`

Staff standard: Yes. Clean fix, correct root cause analysis, keyboard accessibility preserved through new shortcuts.

obey-agent (changes requested):
Verdict: Request Changes (re-review — PR grew from 1 commit to 7, +1333/-39, 20 files)

Previous review only covered commit 1 (preview focus fix). This covers the full PR including the new workitem priority system. All tests pass (`go test ./internal/workitem/... ./internal/intent/tui/explorer/...`), `go vet` clean.

## Bug (Must Fix)

**`clearPriority` does not zero out `ManualPriority` on the WorkItem** (`update.go:307-323`)

`clearPriority()` calls `priority.Clear(store, key)` (removes from store map), then `priority.Apply(store, items)`. But `Apply` only *sets* `ManualPriority` for items found in the store — since the key was just removed, `Apply` skips it, leaving the stale value (e.g., `"high"`) on the struct. The item continues sorting as high-priority and displaying the badge until a full refresh.

Fix: make `Apply` idempotent by also clearing items NOT in the store:
```go
func Apply(store *Store, items []workitem.WorkItem) []workitem.WorkItem {
    for i := range items {
        if entry, ok := store.ManualPriorities[items[i].Key]; ok {
            items[i].ManualPriority = string(entry.Priority)
        } else {
            items[i].ManualPriority = ""
        }
    }
    return items
}
```

## Should Fix

**Stale TODOs in `store_test.go:305-307`**: Three TODO comments reference "enable after WorkItem.ManualPriority field is added in sequence 02". That field was added in commit 3 of this same PR. The `Apply` tests should be written — especially given the bug above.

**Silenced errors on interactive save** (`update.go:297,315`): `_ = priority.SaveOrDelete(...)` in `assignPriority` and `clearPriority` means a failed save silently loses the user's priority assignment. Consider setting a status message on error so the user knows the save failed.

## Non-Blocking

**Duplicate `priorityRank`**: `sort.go` re-implements the same switch as `ManualPriority.Rank()` in `priority.go`. Could call `priority.ManualPriority(p).Rank()` instead.

**`fmt.Errorf` vs `camperrors.Wrap`**: `priority/store.go` uses `fmt.Errorf` while the rest of the workitem package uses `camperrors.Wrap`. Minor convention deviation.

**Missing `context.Context` on I/O**: `Load`, `Save`, `SaveOrDelete` perform file I/O without context. Per project Go standards, these should accept context. Minor in practice since operations are fast.

## What's Done Well

- **Atomic writes**: temp-file-then-rename with deferred cleanup. Correct.
- **Prune-before-filter**: command pipeline prunes against full unfiltered set, applies priorities, sorts, THEN filters/limits. Exactly right.
- **Cursor preservation**: `preserveSelection(key)` handles found/not-found/empty correctly with test coverage.
- **Key binding isolation**: priority mode checked first in Update dispatch. `1/2/3` maps to priority levels in priority mode, type filters in normal mode — no conflict.
- **Sort backward compatibility**: empty `ManualPriority` returns rank 4, preserving timestamp-based ordering for unprioritized items. Explicitly tested.
- **Schema version bump**: v1alpha2 → v1alpha3 with updated SortInfo.
- **Gitignore scaffold**: `settings/workitems.json` added to generated `.campaign/.gitignore`.
- **Test coverage**: 307 lines store tests, 231 lines integration tests, 89 lines sort tests. Table-driven with good edge cases.

Staff standard: Not yet — the `clearPriority` bug means a user action (clearing a priority) doesn't take effect until refresh. One-line fix in `Apply` resolves it.

obey-agent (approved):
Verdict: Approve (re-review after commit 8)

Commit 8 (`0dcda48`) addresses all three findings from the previous review:

**1. Bug fixed: `Apply` now idempotent**
Added `else { items[i].ManualPriority = "" }` — items not in the store have their stale priority cleared. `clearPriority` now works correctly without requiring a full refresh.

**2. Stale TODOs replaced with real tests**
Three TODO placeholders replaced with actual tests:
- `TestApply_DecoratesMatchingItems` — verifies matching items get priority, non-matching stay empty
- `TestApply_ClearsStaleValues` — the exact bug scenario: item has stale `"medium"` not in store, Apply clears it
- `TestApply_EmptyStore` — empty store clears all stale values

**3. Save errors surfaced in TUI**
Both `assignPriority` and `clearPriority` now check the `SaveOrDelete` error, exit priority mode, and show a status message (`"save failed: ..."`) instead of silencing.

All previous findings resolved. The non-blocking items (duplicate `priorityRank`, `fmt.Errorf` convention, missing context on I/O) remain as noted but are not blocking.

Staff standard: Yes.

PR #204, opened 2026-04-08
feat: add linked project support via symlinks

lancekrogers:
## Summary

Add support for linking external local projects into campaigns via symlinks, complementing the existing git submodule approach. This is a draft/starting point — several areas need further work before merge.

### Why this is needed

Campaigns are git monorepos where projects are added as submodules. This works well for portability but creates friction for a common use case: **users who already have projects cloned elsewhere on their machine**. Today they must either:

1. Re-clone into the campaign as a submodule (duplicating the repo)
2. Use `--local` which still clones and bakes an absolute path into `.gitmodules` (non-portable anyway)

Based on feedback, most users care more about "all my projects in one place that works" than strict git portability. They want to point their campaign at existing work, not reorganize their filesystem.

### Why symlinks

Symlinks are the simplest solution that gives filesystem transparency:
- `cd projects/my-app` just works — editors, tools, scripts all see real files
- `camp project list` finds them via `os.ReadDir`
- Git commands work inside the symlink target if it's a git repo
- No special mount permissions needed (unlike bind mounts)

The trade-off is they're not portable across machines, which is acceptable since that's the explicit contract: linked projects are local conveniences, submodules are for portability.

### No manifest — symlinks are the source of truth

~~The initial implementation included a `.campaign/linked-projects.json` manifest.~~ After discussion, this was identified as redundant. Everything the manifest stored is derivable from the filesystem at runtime:

- **Is it linked?** → `os.Lstat()` checks if entry in `projects/` is a symlink
- **Where does it point?** → `os.Readlink()` gives the target path
- **Is it a git repo?** → `os.Stat(target + "/.git")`

The manifest should be removed from this PR. The symlinks themselves are the source of truth for the campaign→project direction. For the reverse direction (project→campaign), see the `.camp` marker file discussion below.

### Architecture: two-way linking without a manifest

The linking relationship is established by two lightweight markers:

1. **Symlink in `projects/`** → the campaign knows about the project
2. **`.camp` file in the project directory** → the project knows about the campaign

No manifest, no duplicated state. Each side of the relationship is stored exactly once, at the location that needs it.

## What this PR implements (code changes)

- `Source` and `LinkedPath` fields on `Project` type with `submodule`, `linked`, `linked-non-git` constants
- ~~`.campaign/linked-projects.json` manifest~~ (to be removed — see above)
- `AddLinked()` / `UnlinkProject()` functions (`link.go`)
- `List()` updated to follow symlinks and annotate source type
- `Remove()` detects linked projects and unlinks cleanly (original files untouched)
- `ResolveFromCwd()` checks project list before git fallback (handles non-git linked projects)
- Table output shows `SOURCE` column when linked projects exist
- `--link` flag on `camp project add` CLI command

## What still needs to be done

See comments below for detailed discussion of:
1. Removing the manifest and using filesystem detection instead
2. All `camp project` subcommands that need linked project support
3. `.camp` marker file for campaign context detection (critical)
4. Fuzzy navigation (`cgo`/`camp go`) fixes

https://claude.ai/code/session_01MM4k12uBDhiw5sLHVzZV44

### DIFF
--- a/internal/project/link.go
+++ b/internal/project/link.go
@@ -1,0 +1,263 @@
+package project
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+)
+
+// LinkOptions configures how a local project is linked into the campaign.
+type LinkOptions struct {
+	// Name overrides the project name (defaults to directory basename).
+	Name string
+	// Path overrides the destination path (defaults to projects/<name>).
+	Path string
+}
+
+// LinkResult contains information about the linked project.
+type LinkResult struct {
+	// Name is the project name.
+	Name string
+	// Path is the relative path from campaign root (symlink location).
+	Path string
+	// Source is the original absolute path.
+	Source string
+	// Type is the detected project type.
+	Type string
+	// IsGit indicates whether the linked project is a git repository.
+	IsGit bool
+}
+
+// AddLinked links an external directory into the campaign via symlink.
+// The project can be a git repo or a plain directory. The symlink is created
+// inside projects/ and the project is recorded in the linked projects manifest.
+// A .gitignore entry is added so git does not track the symlink.
+func AddLinked(ctx context.Context, campaignRoot, localPath string, opts LinkOptions) (*LinkResult, error) {
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	// Resolve to absolute path
+	absLocal, err := filepath.Abs(localPath)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "failed to resolve local path")
+	}
+
+	// Verify the path exists and is a directory
+	info, err := os.Stat(absLocal)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("path does not exist: %s", localPath)
+		}
+		return nil, camperrors.Wrapf(err, "cannot access %s", localPath)
+	}
+	if !info.IsDir() {
+		return nil, fmt.Errorf("path is not a directory: %s", localPath)
+	}
+
+	// Determine project name
+	name := opts.Name
+	if name == "" {
+		name = filepath.Base(absLocal)
+	}
+	if err := ValidateProjectName(name); err != nil {
+		return nil, err
+	}
+
+	// Determine destination path
+	destPath := opts.Path
+	if destPath == "" {
+		destPath = filepath.Join("projects", name)
+	}
+
+	fullPath := filepath.Join(campaignRoot, destPath)
+
+	// Check if already exists
+	if _, err := os.Lstat(fullPath); err == nil {
+		return nil, &ErrProjectExists{Name: name, Path: destPath}
+	}
+
+	// Check if it's a git repo
+	isGit := isGitRepo(absLocal)
+
+	// Determine source type
+	source := SourceLinked
+	if !isGit {
+		source = SourceLinkedNonGit
+	}
+
+	// Ensure parent directory exists
+	if err := os.MkdirAll(filepath.Dir(fullPath), 0755); err != nil {
+		return nil, camperrors.Wrapf(err, "create parent directory")
+	}
+
+	// Create the symlink (absolute path for reliability)
+	if err := os.Symlink(absLocal, fullPath); err != nil {
+		return nil, camperrors.Wrapf(err, "create symlink")
+	}
+
+	// Add to manifest
+	entry := LinkedProjectEntry{
+		Source:  source,
+		AbsPath: absLocal,
+		IsGit:  isGit,
+		AddedAt: time.Now(),
+	}
+	if err := AddToManifest(campaignRoot, name, entry); err != nil {
+		// Clean up symlink on manifest failure
+		os.Remove(fullPath)
+		return nil, camperrors.Wrapf(err, "record linked project")
+	}
+
+	// Add gitignore entry so the campaign repo doesn't track the symlink
+	if err := addGitignoreEntry(campaignRoot, destPath); err != nil {
+		// Non-fatal - warn but continue
+		fmt.Fprintf(os.Stderr, "Warning: could not add .gitignore entry: %v\n", err)
+	}
+
+	// Detect project type
+	projectType := detectProjectType(absLocal)
+
+	return &LinkResult{
+		Name:   name,
+		Path:   destPath,
+		Source: absLocal,
+		Type:   projectType,
+		IsGit:  isGit,
+	}, nil
+}
+
+// UnlinkProject removes a linked project's symlink and manifest entry.
+// Returns true if the project was a linked project that was unlinked.
+func UnlinkProject(ctx context.Context, campaignRoot, name string) (bool, error) {
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+
+	isLinked, _, err := IsLinkedProject(campaignRoot, name)
+	if err != nil {
+		return false, err
+	}
+	if !isLinked {
+		return false, nil
+	}
+
+	projectPath := filepath.Join(campaignRoot, "projects", name)
+
+	// Verify it's actually a symlink before removing
+	info, err := os.Lstat(projectPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// Symlink already gone, just clean up manifest
+			RemoveFromManifest(campaignRoot, name)
+			return true, nil
+		}
+		return false, camperrors.Wrapf(err, "check linked project")
+	}
+
+	if info.Mode()&os.ModeSymlink != 0 {
+		if err := os.Remove(projectPath); err != nil {
+			return false, camperrors.Wrapf(err, "remove symlink")
+		}
+	}
+
+	// Remove from manifest
+	if _, err := RemoveFromManifest(campaignRoot, name); err != nil {
+		return false, err
+	}
+
+	// Remove gitignore entry
+	destPath := filepath.Join("projects", name)
+	if err := removeGitignoreEntry(campaignRoot, destPath); err != nil {
+		fmt.Fprintf(os.Stderr, "Warning: could not remove .gitignore entry: %v\n", err)
+	}
+
+	return true, nil
+}
+
+// isGitRepo checks whether a directory contains a .git directory or file.
+func isGitRepo(path string) bool {
+	_, err := os.Stat(filepath.Join(path, ".git"))
+	return err == nil
+}
+
+// addGitignoreEntry appends a path to the campaign root's .gitignore
+// if it's not already present.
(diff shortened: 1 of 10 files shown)


lancekrogers:
## TODO: All `camp project` subcommands must work with linked projects

The current implementation covers `add`, `list`, `remove`, and `resolve` (cwd detection). But every subcommand under `camp project` needs to be audited:

| Command | Status | Notes |
|---------|--------|-------|
| `add --link` | ✅ Implemented | Creates symlink + manifest + gitignore entry |
| `list` | ✅ Implemented | Follows symlinks, shows SOURCE column |
| `remove` | ✅ Implemented | Detects linked, removes symlink only |
| `commit` | ⚠️ Needs work | Should work for linked git repos (git commands run inside resolved path). Needs to skip/warn for non-git linked projects. |
| `run` | ⚠️ Needs work | Should work transparently — commands run in the project directory regardless of how it was added. |
| `remote` | ⚠️ Needs work | `remote add/remove/set-url` should work for linked git repos. Skip for non-git. |
| `worktree` | ⚠️ Needs work | Git worktrees for linked git repos should work since git commands resolve through the symlink. Skip for non-git. |
| `new` | ✅ N/A | Creates new repos, not relevant to linking existing ones. |

**General pattern:** Check `project.Source` before running git-specific operations. For `SourceLinkedNonGit`, either skip gracefully with a message or error with a clear explanation. For `SourceLinked` (git repos), most git operations should work since commands run inside the resolved symlink target.

---
_Generated by [Claude Code](https://claude.ai/code)_

lancekrogers:
## TODO (Critical): Campaign context detection for symlinked projects

**This is the biggest open problem with linked projects.**

### The problem

Campaign root detection (`campaign.Detect()`) works by walking up the directory tree looking for `.campaign/`. When you `cd` into a symlinked project that lives at e.g. `/home/user/code/my-app`, you're physically outside the campaign tree. Walking up from `/home/user/code/my-app` will never find `.campaign/`.

This means:
- `camp go my-symlinked-project` works fine (navigates via the campaign's `projects/` symlink)
- But once inside the resolved target path, **all camp commands fail** — no campaign detected
- `camp go docs`, `camp go p`, any navigation — broken
- The user loses all campaign context

This is especially bad with `cgo` (fuzzy `camp go`) which is a core navigation workflow.

### Proposed solution: `.camp` marker file

Place a small `.camp` or `.camp.json` file inside linked project directories:

```json
{
  "campaign_id": "8deed8b4-...",
  "campaign_root": "/home/user/my-campaign"
}
```

**How it would work:**

1. `campaign.Detect()` already walks up looking for `.campaign/`. Add a secondary check: if `.campaign/` is not found, also look for `.camp` file during the walk.
2. If `.camp` is found, read the `campaign_root` from it and verify that root still has `.campaign/`.
3. `camp project add --link` writes this file into the linked project directory automatically.
4. `camp project remove` cleans it up.
5. `.camp` should be added to common gitignore templates since it's machine-local.

**Additional UX opportunity:**

This marker file enables a powerful new interaction: running any `camp` command inside a project that has no `.camp` file and isn't inside a campaign could prompt:

> *"No campaign detected. Would you like to link this project to a campaign?"*

Then list registered campaigns from the global registry and let the user pick one. This turns camp into something you can invoke from anywhere, not just inside campaign trees.

### Considerations

- The `.camp` file should be gitignored by default (it's machine-local, like the manifest)
- Need to handle stale `.camp` files where the campaign root no longer exists
- Multiple campaigns could theoretically link the same project — `.camp` would point to the most recent one, or could store an array
- `CAMP_ROOT` env var already exists as an override — `.camp` file would be a middle ground between env var and directory-walk detection

---
_Generated by [Claude Code](https://claude.ai/code)_

lancekrogers:
## TODO: Fuzzy navigation (`cgo` / `camp go`) with linked projects

### Current behavior

`camp go` and its fuzzy variant `cgo` resolve navigation targets relative to the campaign root using the jumps config and project list. This works because the user is always physically inside the campaign directory tree.

### What breaks with linked projects

When a user jumps into a linked project, their shell's cwd resolves to the symlink target (e.g. `/home/user/code/my-app`), not the campaign-relative path (`/home/user/my-campaign/projects/my-app`). From that location:

1. Campaign detection fails (see context detection comment above)
2. Even if detection is solved via `.camp` file, the navigation index may not include the resolved external path in its jump targets
3. Jumping *back* from a linked project to campaign locations (e.g. `camp go docs`) needs the campaign root to be known

### What needs to happen

1. **Solve campaign detection first** (`.camp` marker file approach)
2. **Navigation index** needs to be aware of linked projects and their resolved paths, so fuzzy matching works in both directions
3. **Jump target resolution** in `camp go` should check if the current directory is a linked project (via manifest lookup) and use the campaign root from the `.camp` file or manifest for resolving relative targets
4. **`cgo` fuzzy finder** should include linked project names in its candidate list regardless of current directory, as long as campaign context is available

---
_Generated by [Claude Code](https://claude.ai/code)_

lancekrogers:
## TODO: Remove the manifest, use filesystem detection instead

The `.campaign/linked-projects.json` manifest in the current code is redundant and should be removed. Campaigns are isolated — they don't store machine-local state beyond cache. The manifest duplicates what the filesystem already provides.

### What to change

**Remove:**
- `internal/project/manifest.go` — entire file
- All `LoadManifest` / `SaveManifest` / `AddToManifest` / `RemoveFromManifest` / `IsLinkedProject` calls
- `linked-projects.json` from `.campaign/.gitignore` scaffold

**Replace with filesystem detection:**

```go
// Detect linked project by checking if the entry is a symlink
func isSymlink(path string) bool {
    info, err := os.Lstat(path)
    if err != nil {
        return false
    }
    return info.Mode()&os.ModeSymlink != 0
}
```

- `List()` already checks `entry.Type()&os.ModeSymlink` — this stays, just remove the manifest lookup inside `resolveLinkedProject()`
- `Remove()` should use `os.Lstat()` to detect symlinks instead of `IsLinkedProject()`
- `AddLinked()` just creates the symlink + gitignore entry (+ `.camp` marker file in the target, once that's implemented)
- `UnlinkProject()` just removes the symlink + gitignore entry (+ `.camp` marker file cleanup)

The `LinkedPath` field on `Project` can be populated at runtime via `os.Readlink()` instead of reading from a manifest.

---
_Generated by [Claude Code](https://claude.ai/code)_

lancekrogers:
## Revised: `.camp` marker file design

After further discussion, the `.camp` marker file is not just a nice-to-have for navigation — it's the foundational piece that makes linked projects work properly. Without it, any camp command run from inside a linked project's resolved path will fail because campaign detection can't walk up to `.campaign/`.

### The dependency chain

```
Symlink in projects/     →  campaign knows about the project
.camp file in project    →  project knows about the campaign
Campaign registry        →  maps campaign IDs to roots (already exists)
```

No manifest needed. Each side of the relationship is stored exactly once, at the location that needs it.

### How `.camp` file works

Written by `camp project add --link` into the linked project's directory:

```json
{
  "campaign_id": "8deed8b4-...",
  "campaign_root": "/home/user/my-campaign"
}
```

**Detection flow in `campaign.Detect()`:**

1. Walk up looking for `.campaign/` (existing behavior)
2. If not found, walk up looking for `.camp` file (new fallback)
3. If `.camp` found, read `campaign_root`, verify it still has `.campaign/`
4. Return that root

**Lifecycle:**
- Created by `camp project add --link`
- Removed by `camp project remove`
- Should be gitignored (machine-local) — add `.camp` to default gitignore patterns
- `camp doctor` can detect stale `.camp` files where the campaign root no longer exists

### UX opportunity

The `.camp` file also enables an interaction for projects not yet linked to any campaign:

> Running `camp` inside `/home/user/code/some-project` (no `.camp`, not inside a campaign):
> *"No campaign detected. Would you like to link this project to a campaign?"*
> → List campaigns from global registry, user picks one, symlink + `.camp` created automatically

This makes camp invocable from anywhere, not just inside campaign trees.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/project/manifest.go
```diff
@@ -0,0 +1,141 @@
...
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+)
+
+// LinkedProjectManifest stores metadata about projects linked into the campaign
```
obey-agent, line 13: This manifest creates a second source of truth for something the filesystem already knows. Whether a project is linked, where it points, and whether it's a git repo are all derivable from `projects/<name>` with `os.Lstat`/`os.Readlink`/`.git` checks. Keeping a machine-local JSON registry here guarantees drift on rename/remove/broken-link paths, so I'd drop the manifest entirely and let the symlink be authoritative.

internal/project/link.go
```diff
@@ -0,0 +1,263 @@
...
+		IsGit:  isGit,
+		AddedAt: time.Now(),
+	}
+	if err := AddToManifest(campaignRoot, name, entry); err != nil {
+		// Clean up symlink on manifest failure
+		os.Remove(fullPath)
+		return nil, camperrors.Wrapf(err, "record linked project")
+	}
+
+	// Add gitignore entry so the campaign repo doesn't track the symlink
```
obey-agent, line 118: Writing linked entries into tracked `.gitignore` turns a machine-local convenience into repo churn. Different contributors will have different local link sets, so this will create noisy diffs for a path that should never be shared. This belongs in `.git/info/exclude` instead.

internal/project/resolve.go
```diff
@@ -74,24 +74,46 @@ func ResolveFromCwd(ctx context.Context, campRoot string) (*ResolveResult, error
 		return nil, camperrors.Wrap(err, "failed to get working directory")
 	}
 
+	// Resolve symlinks in cwd so we can match against both symlink and target paths
+	resolvedCwd, _ := filepath.EvalSymlinks(cwd)
+	resolvedCamp, _ := filepath.EvalSymlinks(campRoot)
+
+	// Look up the project in the dynamic list first — this covers linked projects
```
obey-agent, line 81: This still doesn't solve the reverse-direction case: once the user `cd`s into the real external directory, `camp` has no way to recover the campaign root from there. Resolving against the project list only helps if the current process is already under the symlink path or a git repo we can discover. We need a project-side marker (for example `.camp`) or another reverse link before linked-project navigation is actually complete.

obey-agent (changes requested):
Verdict: Request Changes

Overview:
This draft found the right user problem, but the implementation still depends on a redundant manifest and does not yet support reverse campaign detection from the real external working directory. In its current form this should not be the merge target.

Key Findings:
- `internal/project/manifest.go`: the extra machine-local registry will drift from filesystem truth.
- `internal/project/link.go`: machine-local link state is written into tracked `.gitignore`.
- `internal/project/resolve.go`: reverse detection is incomplete without a project-side marker such as `.camp`.

What's Done Well:
- The draft correctly identified the need for a symlink-based model.
- It surfaced useful context that informed the stronger follow-up implementation.

Staff Standard:
Not yet. I would not ship this revision as the merge target.

PR #206, opened 2026-04-09
fix: --campaign bare flag triggers interactive picker

### LINKED ISSUE #205: camp intent add --campaign (bare flag) fails instead of opening interactive picker
## Bug Description

Running `camp intent add --campaign` (with no value after `--campaign`) should open an interactive picker to choose a campaign from the registry. Instead, Cobra rejects it:

```
Error: flag needs an argument: --campaign
```

The flag help text says "omit value to pick interactively" but this never works.

## Root Cause

In `cmd/camp/intent/add.go` line 62:

```go
flags.Lookup("campaign").NoOptDefVal = ""
```

Cobra's pflag library only treats a flag as "optional value" when `NoOptDefVal` is set to a **non-empty string**. Setting it to `""` has no effect — Cobra still requires an argument.

## How It Should Work

1. User runs: `camp intent add --campaign`
2. Cobra parses `--campaign` without a value, uses `NoOptDefVal` as default
3. `cmd.Flags().Changed("campaign")` returns `true`
4. The resolver in `resolve()` (line 195-239) detects empty campaign + `Changed == true` and calls `pickCampaign()` to show the fuzzy-finder UI
5. User selects a campaign interactively

The resolver logic at lines 195-239 is already correct — it properly handles the bare-flag case. The bug is that Cobra never lets execution reach the resolver.

## Fix

**File:** `cmd/camp/intent/add.go`

Change line 62 from:
```go
flags.Lookup("campaign").NoOptDefVal = ""
```

To a non-empty sentinel value, e.g.:
```go
flags.Lookup("campaign").NoOptDefVal = "__picker__"
```

Then in `runIntentAdd` (around line 72), normalize the sentinel back to empty before passing to the resolver:

```go
targetCampaign, _ := cmd.Flags().GetString("campaign")
if targetCampaign == "__picker__" {
    targetCampaign = ""
}
```

The existing resolver logic (`resolve()`) already handles `targetCampaign == ""` + `Changed("campaign") == true` by calling the interactive picker, so no further changes are needed there.

## Scope Check

Verify whether other commands use the same `NoOptDefVal = ""` pattern and fix them too if so.

## Tests

`TestIntentAddCampaignResolver_BareCampaignUsesPicker` in `add_test.go` already tests the picker path — it should pass once the Cobra parsing is fixed. Consider adding an integration-level test that exercises the actual flag parsing (not just the resolver).

lancekrogers:
## Summary
- `camp intent add --campaign` (no value) errored with "flag needs an argument" instead of opening the interactive campaign picker
- Root cause: `NoOptDefVal` was set to `""` — Cobra ignores empty values and still requires an argument
- Fix: set `NoOptDefVal` to a non-empty sentinel (`__picker__`), normalize it back to empty before passing to the resolver

Fixes #205

## Test plan
- [x] `TestIntentAddCampaignResolver_BareCampaignUsesPicker` passes
- [x] `TestIntentAddCampaignResolver_BareCampaignNonInteractiveFails` passes
- [x] `TestIntentAdd_TargetCampaignSmoke` passes
- [x] Full build clean
- [ ] Manual: `camp intent add --campaign` opens fuzzy picker
- [ ] Manual: `camp intent add --campaign obey-campaign` still resolves by name

### DIFF
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -23,6 +23,12 @@ import (
 	"github.com/Obedience-Corp/camp/internal/paths"
 )
 
+// noOptCampaign is the NoOptDefVal for the --campaign flag. Cobra requires a
+// non-empty string to allow a flag without a value (bare --campaign). The
+// resolver treats this sentinel identically to "" so no real campaign name is
+// reserved.
+const noOptCampaign = "\x00pick"
+
 var intentAddCmd = &cobra.Command{
 	Use:   "add [title]",
 	Short: "Create a new intent",
@@ -59,7 +65,7 @@ func init() {
 	flags.BoolP("full", "f", false, "Full TUI mode with body textarea")
 	flags.StringP("campaign", "c", "", "Target campaign by name or ID; omit value to pick interactively")
 	flags.Bool("no-commit", false, "Don't create a git commit")
-	flags.Lookup("campaign").NoOptDefVal = "__picker__"
+	flags.Lookup("campaign").NoOptDefVal = noOptCampaign
 }
 
 func runIntentAdd(cmd *cobra.Command, args []string) error {
@@ -70,9 +76,6 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 	useEditor, _ := cmd.Flags().GetBool("edit")
 	fullMode, _ := cmd.Flags().GetBool("full")
 	targetCampaign, _ := cmd.Flags().GetString("campaign")
-	if targetCampaign == "__picker__" {
-		targetCampaign = ""
-	}
 	noCommit, _ := cmd.Flags().GetBool("no-commit")
 
 	campaignResolver := newIntentAddCampaignResolver(cmd.ErrOrStderr())
@@ -196,6 +199,11 @@ func newIntentAddCampaignResolver(stderr io.Writer) intentAddCampaignResolver {
 }
 
 func (r intentAddCampaignResolver) resolve(ctx context.Context, targetCampaign string, targetChanged bool) (*config.CampaignConfig, string, error) {
+	// Normalize the Cobra NoOptDefVal sentinel — bare --campaign means "pick interactively".
+	if targetCampaign == noOptCampaign {
+		targetCampaign = ""
+	}
+
 	if !targetChanged {
 		cfg, campaignRoot, err := r.loadCurrent(ctx)
 		if err != nil {
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -59,7 +59,7 @@ func init() {
 	flags.BoolP("full", "f", false, "Full TUI mode with body textarea")
 	flags.StringP("campaign", "c", "", "Target campaign by name or ID; omit value to pick interactively")
 	flags.Bool("no-commit", false, "Don't create a git commit")
-	flags.Lookup("campaign").NoOptDefVal = ""
+	flags.Lookup("campaign").NoOptDefVal = "__picker__"
 }
 
 func runIntentAdd(cmd *cobra.Command, args []string) error {
@@ -70,6 +70,9 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 	useEditor, _ := cmd.Flags().GetBool("edit")
 	fullMode, _ := cmd.Flags().GetBool("full")
 	targetCampaign, _ := cmd.Flags().GetString("campaign")
+	if targetCampaign == "__picker__" {
+		targetCampaign = ""
+	}
 	noCommit, _ := cmd.Flags().GetBool("no-commit")
 
 	campaignResolver := newIntentAddCampaignResolver(cmd.ErrOrStderr())


cmd/camp/intent/add.go
```diff
@@ -70,6 +70,9 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 	useEditor, _ := cmd.Flags().GetBool("edit")
 	fullMode, _ := cmd.Flags().GetBool("full")
 	targetCampaign, _ := cmd.Flags().GetString("campaign")
+	if targetCampaign == "__picker__" {
```
obey-agent: This sentinel now collides with a real user input. `camp intent add --campaign __picker__` (or `--campaign=__picker__`) can no longer target a campaign whose name or ID is literally `__picker__`, because it is always normalized into the interactive-picker path and will even fail in non-interactive mode. Since campaign names/IDs come from registry data, the fix needs to record "bare flag was used" without reserving a valid string value.

obey-agent (commented):
Verdict: Request Changes

Overview:
This PR fixes the immediate Cobra/pflag problem behind bare `--campaign`, and the surrounding resolver/tests are otherwise in good shape. But the chosen sentinel introduces a new lookup regression: a real campaign whose name or ID is `__picker__` can no longer be targeted explicitly.

Key Findings:
- `cmd/camp/intent/add.go`: normalizing the user-visible string `__picker__` back to `""` makes `camp intent add --campaign __picker__` and `camp intent add --campaign=__picker__` indistinguishable from the bare-flag picker path. I left an inline comment on this line.

What's Done Well:
- The root cause analysis in the PR description is correct about Cobra ignoring `NoOptDefVal=""`.
- The resolver structure keeps the targeting logic testable, and the added bare-flag coverage improves confidence in the intended interactive path.
- I validated the existing targeted tests on the PR head: `go test ./cmd/camp/intent -run TestIntentAdd` passed.

Staff Standard:
Not yet. I would want the bare-flag detection to avoid reserving a valid campaign name before I would be comfortable submitting this as my own work.

cmd/camp/intent/add.go
lancekrogers: Good catch. Replaced `__picker__` with a NUL-prefixed sentinel (`\x00pick`) that cannot collide with any valid campaign name (registry names can't contain NUL bytes). Also moved the normalization out of `runIntentAdd` and into the resolver — it's the only consumer and it already has the `targetChanged` signal, so this keeps the sentinel handling contained in one place.

obey-agent (approved):
Verdict: Approve

Overview:
This is a small, correct Cobra fix. The sentinel value keeps `--campaign` bare-flag behavior distinct from the normal empty-string path without disturbing the existing resolver flow, and the tests cover both interactive and non-interactive behavior.

Key Findings:
- No blocking issues found.

What's Done Well:
- The root cause is handled in one place instead of adding branching at each call site.
- The behavior is documented in code and covered by focused tests.

Staff Standard:
Yes. I would be comfortable shipping this as-is.

obey-agent (commented):
## Staff-Level Review — Approve (with notes)

**Verdict**: Approve

**Overview**: Small, focused fix for a real Cobra UX gotcha. The NUL-prefixed sentinel (`"\x00pick"`) is defensible — NUL bytes cannot travel through a shell into a flag value, so the collision surface with real campaign names is effectively zero. The normalization is placed correctly in the resolver, before the `targetChanged` branch, so all three code paths (no flag / `--campaign <name>` / bare `--campaign`) resolve correctly. I traced the logic end-to-end and it behaves.

### Key Findings

1. **Mechanics are correct.** Bare `--campaign` now sets `targetCampaign = "\x00pick"` with `targetChanged = true`; the resolver normalizes to `""` and falls through to `pickCampaign(ctx, reg)`. Named `--campaign <x>` keeps `x`. Omitted flag keeps `targetChanged = false` and loads from cwd.

2. **Pattern is worth generalizing.** You've established a "bare flag = picker" invariant, plus a sentinel + normalize shim. If any other command wants the same UX, the pattern will get copy-pasted. A small helper (e.g. `cmdutil.InteractiveFlag`) that installs the NoOptDefVal and returns a normalize function would keep this in one place. Follow-up, not a blocker.

3. **Audit opportunity.** `grep -rn NoOptDefVal cmd/camp` shows only `intent/add.go` uses this. Several other commands have `--campaign` / `--project` flags that could benefit from the same UX but currently reject bare usage entirely. Worth filing a follow-up so `intent add` isn't the lone exception. See inline at `cmd/camp/intent/add.go:29`.

4. **Manual test items still unchecked.** The PR description lists two manual items:
   - `camp intent add --campaign` opens fuzzy picker
   - `camp intent add --campaign obey-campaign` still resolves by name
   
   Both should be smoke-tested on a real terminal before merge. The automated tests exercise the resolver directly, but the Cobra binding and TUI entrypoint are what the user actually hits.

### What's Done Well

- NUL sentinel is the right choice — clever and safe.
- Doc comment on the constant explains *why* the sentinel exists, which is exactly what future maintainers need.
- Normalization lives in `resolve()` rather than `runIntentAdd`, so every caller that goes through the resolver is safe.
- Three targeted tests cover the interactive picker, the non-interactive failure, and the smoke path — that's the right shape for this kind of fix.
- PR description links the root cause (`NoOptDefVal == ""` → cobra still requires an argument) to the fix, which makes the change self-documenting.

### Staff Standard

I'd sign off on this as written. The scope is appropriately small, the mechanism is well-chosen, and the test coverage exercises the real decision points. The two follow-ups (reusable helper, audit other commands) are value-adds rather than blockers — the fix itself is solid and I'd merge it after the two manual test items are ticked.

cmd/camp/intent/add.go
```diff
@@ -23,6 +23,12 @@ import (
 	"github.com/Obedience-Corp/camp/internal/paths"
 )
 
+// noOptCampaign is the NoOptDefVal for the --campaign flag. Cobra requires a
+// non-empty string to allow a flag without a value (bare --campaign). The
+// resolver treats this sentinel identically to "" so no real campaign name is
+// reserved.
```
obey-agent, line 29: **Solid pattern — worth codifying as a reusable helper**

The NUL-prefixed sentinel is a good choice: NUL bytes can't travel through a shell into a flag value, so the collision surface with real campaign names is effectively zero. I traced the three paths (`camp intent add`, `--campaign <name>`, bare `--campaign`) through `resolve()` and the logic is correct.

Two related observations, neither blocking:

1. **Consider extracting the pattern.** You now have a "bare flag = interactive picker" invariant, a sentinel constant, and a normalization shim. If any other flag in camp wants the same UX, copy-pasting the NUL sentinel + resolver normalize is a trap waiting to happen. A small helper like `cmdutil.InteractiveFlag(flag, &value)` that installs the NoOptDefVal and returns a `Normalize(string) string` function would keep this invariant in one place. Fine to do in a follow-up.

2. **Audit other commands for the same bug.** `grep -rn NoOptDefVal cmd/camp` shows only `intent/add.go` sets this, but several subcommands have `--campaign` / `--project` flags that *could* benefit from the same "omit value = pick" UX. They currently reject bare usage entirely. Worth filing a follow-up to apply the pattern uniformly, so users don't discover it works for `intent add` and nowhere else.

The fix itself is correct and minimal — LGTM on the mechanics.

PR #207, opened 2026-04-09
add linked project support

lancekrogers:
## What changed
- introduce linked workspaces as a first-class project type backed by campaign-side symlinks under `projects/` and machine-local `.camp` markers in external directories
- add dedicated `camp project link` / `camp project unlink` commands and keep `camp project add` focused on submodules; support targeting another campaign via `--campaign` and the shared picker when invoked outside current campaign context
- extend campaign detection and command dispatch so `camp go`, `camp run`, `camp project run`, `camp project commit`, and project remote/worktree flows work correctly from linked project directories
- distinguish linked git vs linked non-git behavior across list/resolve/remove/output and route git-only operations through explicit linked-non-git errors instead of ambiguous git failures
- reject unsafe link states including cross-campaign marker collisions, duplicate link targets in one campaign, and delete-style removal of external linked targets
- track linked symlink changes in campaign git history as distinct `Link` / `Unlink` actions rather than reusing submodule add/remove semantics
- harden marker and persistence behavior with atomic writes, marker ownership checks, restored `CAMP_ROOT` override semantics, and expanded integration / TTY coverage
- update CLI docs and generated reference pages for the revised linked-project workflow

## Why
The optional-submodule design now treats campaign membership as a filesystem relationship instead of a submodule-only management model. This implementation follows that design directly: campaigns can attach existing local workspaces through a tracked symlink in `projects/`, while the linked directory keeps only the machine-local metadata needed for reverse detection and command routing.

## User impact
- campaigns can attach existing local repos or plain directories without converting them into submodules
- linked workspaces participate in navigation, run, commit, remote, and worktree flows where that behavior makes sense
- linked non-git directories are surfaced clearly and git-only commands fail with explicit errors
- campaign git history distinguishes linked-workspace attach/detach from submodule add/remove
- remaining non-blocking cleanup is tracked separately in `#226` and `#221`, so it does not need to block this merge

## Validation
- `go test ./...`
- `go test -tags integration ./tests/integration`
- targeted container-backed linked-project integration coverage for link / unlink / run / go / remove / root-detection flows


internal/project/link.go
```diff
@@ -0,0 +1,226 @@
...
+	isGit := isGitRepo(absLocal)
+	marker := campaign.LinkMarker{
+		Version:      1,
+		CampaignRoot: campaignRoot,
+		ProjectName:  name,
+	}
+	if cfg, err := config.LoadCampaignConfig(ctx, campaignRoot); err == nil {
+		marker.CampaignID = cfg.ID
+	}
+	if err := campaign.WriteMarker(absLocal, marker); err != nil {
```
obey-agent, line 132: This writes a single mutable `.camp` marker into the external repo. Because `os.Getwd`/`EvalSymlinks` resolve a symlinked cwd to the physical target, linking the same local repo into a second campaign overwrites this marker and makes commands run from the first campaign's linked path resolve against the second campaign instead. I reproduced it with `camp run pwd` from the first link printing the second campaign root after the second link. We need to either reject "already linked to another campaign" here or store enough context to disambiguate multiple campaigns instead of a single global marker.

obey-agent (commented):
Verdict: Request Changes

Overview:
This PR does a solid job threading linked-project awareness through add/list/run/go/worktree flows, and the targeted linked integration cases I ran passed. The remaining blocker is a correctness regression in the marker model: the same external repo cannot be safely linked into more than one campaign.

Key Findings:
- `internal/project/link.go`: `AddLinked` overwrites a single `.camp` marker in the external repo. After linking that repo into campaign B, running `camp` from campaign A's linked path resolves against campaign B instead. I reproduced this with `camp run pwd` from the first link printing the second campaign root.

What's Done Well:
- Linked vs submodule/non-git source distinctions are carried through list/resolve/run/worktree surfaces consistently.
- The new linked-project integration coverage is directionally good and the targeted linked tests I ran passed.

Staff Standard:
Not yet. I would want the linked-project marker model to stop rerouting commands into the wrong campaign before I'd be comfortable submitting this as my own work.

obey-agent (approved):
Verdict: Approve

Overview:
This is the stronger linked-project implementation. It removes manifest-based state, adds reverse `.camp` detection, and carries linked-project behavior through the command surface with solid unit and integration coverage.

Key Findings:
- No blocking issues found in the current revision.

What's Done Well:
- Keeps machine-local state out of tracked files by using `.git/info/exclude`.
- Distinguishes linked git and linked non-git projects cleanly in resolution and UX paths.
- Extends the feature across the command surface instead of leaving partial support behind.

Staff Standard:
Yes. I would be comfortable shipping this revision.

obey-agent (approved):
Verdict: Approve

Follow-up:
The linked-project marker collision is now fixed on the current head. `camp project add --link` rejects linking an external repo that is already attached to another live campaign instead of overwriting its `.camp` marker, and the branch now includes unit plus integration coverage for that cross-campaign case.

obey-agent (commented):

**Verdict**: Request Changes

**Overview**: This is a substantial, well-scoped feature — linked projects as a first-class citizen alongside submodules is a real improvement and the filesystem-marker design (`.camp` JSON file + campaign-side symlink) is clean. The core architecture is right and the integration surface (detect, resolve, list, remove, run, go, worktrees) is covered thoughtfully. I can tell the implementation was thought through. That said, there are three issues in `internal/project/link.go` that I can't sign off on without changes: silent takeover of stale markers, a swallowed `LoadCampaignConfig` error, and non-atomic writes to `.git/info/exclude`. Individually minor, collectively they undermine the "data integrity" story the marker design depends on.

### Key Findings

1. **`ensureLinkMarkerAvailable` silently relinks over stale markers.** If an existing marker points to a campaign root that no longer passes `IsCampaignRoot`, the function returns `nil` without so much as a log line. The user's marker — possibly written on another machine in a Dropbox-synced setup — gets clobbered with no audit trail. At minimum log a warning, ideally require `--force` to override. See inline on `internal/project/link.go:176`.

2. **`LoadCampaignConfig` error swallowed in `AddLinked`.** The marker is written with `CampaignID = ""` on any load failure, with no stderr signal. Worse, `ctx.Err()` isn't re-checked after the load, so a user who hits Ctrl-C mid-link can still have `WriteMarker` run, leaving a half-linked state (symlink exists, marker missing campaign_id, info/exclude not updated). See inline on `internal/project/link.go:101`.

3. **Non-atomic `os.WriteFile` on `.git/info/exclude`.** `ensurePatternInFile` / `removePatternFromFile` truncate-then-write. A crash in the gap corrupts the user's git metadata; two concurrent link invocations on the same repo will last-writer-wins. Standard fix: CreateTemp + rename. The tests will also benefit — current behavior isn't parallel-safe. See inline on `internal/project/link.go:269`.

### Secondary Observations (no inline — acknowledge or defer)

- **`campaign/detect.go:detectCampaignByWalking`** silently drops marker read errors and treats an invalid `marker.CampaignRoot` as "keep walking." If a user's campaign got moved, linked projects become invisible with no diagnostic. A debug log at the skip path would help when users complain about detection.
- **`.camp` marker has no tamper detection** beyond a version integer. For a personal tool this is fine, but worth noting as a design choice rather than an oversight. A HMAC keyed off the campaign ID would be overkill; surfacing "marker version" mismatches in `camp status` would be enough.
- **`UnlinkProject`'s `targetPath == ""` branch** silently leaves orphaned markers in the linked directory. Consider resolving the symlink via `os.Readlink` before removing it, so the function can always find the marker target on its own.
- **Error wrapping consistency**: `ensureInfoExclude` / `removeInfoExclude` return raw errors; most other functions in this file use `camperrors.Wrap`. Minor stylistic drift.

### What's Done Well

- The `LinkMarker` struct is minimal, typed, and versioned — the JSON format is simple to debug and future-proofs extensions.
- Symlink path resolution using `filepath.EvalSymlinks` through `normalizeCampaignRoot` is the right call and matches the detection path, which means linked projects resolve consistently even when the user's cwd goes through symlinks (macOS `/var` → `/private/var`, iCloud Drive, etc.).
- `pathutil.ValidateBoundary` at `fullPath` creation is the right place to catch escape attempts, and using a shared boundary helper keeps this consistent with the sandbox package work.
- `ValidateProjectName` defensively guards against `filepath.Base` returning something odd.
- The `--local` vs `--link` mutual-exclusion check and help text make the UX clear.
- `detect.go` walks logical path first then resolved path, which is the right order and saves a latent bug where symlinked cwds would skip marker detection.
- Integration test `tests/integration/project_linked_test.go` is a good signal that this was exercised end-to-end, not just unit-mocked.

### Staff Standard

This is close — the architectural shape is right and most of the integration work is well-executed. I wouldn't ship it in this state because the three inline issues are each in the "quiet data-integrity bug" category, and linked projects are explicitly about users trusting a tool to manage filesystem state across boundaries. Once the marker takeover is surfaced, the config-load error is visible, and the exclude-file writes are atomic, I'd approve this with confidence. The follow-up is small and targeted.

internal/project/link.go
```diff
@@ -0,0 +1,293 @@
...
+	if sameCampaignRoot(marker.CampaignRoot, campaignRoot) {
+		return nil
+	}
+
+	markerRoot, err := normalizeCampaignRoot(marker.CampaignRoot)
+	if err != nil {
+		return camperrors.Wrap(err, "resolve existing marker campaign root")
+	}
+	if !campaign.IsCampaignRoot(markerRoot) {
+		return nil
```
obey-agent, line 312: **Silently relinks over stale markers**

If a linked project previously belonged to a campaign that has since been deleted or moved (so `IsCampaignRoot(markerRoot)` returns false), this function returns `nil` and `AddLinked` happily overwrites the marker. No log, no warning, no audit trail. The user who ran `camp project add --link` gets a clean success message and has no idea they just clobbered metadata from another machine or a previous workspace layout.

Two problems this causes:

1. **Ambiguity on multi-machine setups** — a directory synced via Dropbox/iCloud/etc. that's linked on machine A gets silently re-linked on machine B when the A path isn't currently valid. Neither user sees that this happened.
2. **No recovery signal** — when a user wonders later "why is this project linked to campaign X instead of Y?", there's nothing in the logs to explain it. The data's already overwritten.

Minimum fix: log a warning to stderr and surface it in the CLI result:

```go
if !campaign.IsCampaignRoot(markerRoot) {
    fmt.Fprintf(os.Stderr, "warning: project was previously linked to %s (no longer a campaign root); taking over marker\n", markerRoot)
    return nil
}
```

Better: require an explicit `--force` / `--relink` flag to override. "Silent takeover of existing state" is the kind of thing that burns users who trusted the tool to warn them.

internal/project/link.go
```diff
@@ -0,0 +1,293 @@
...
+		return nil, camperrors.Wrap(err, "create symlink")
+	}
+
+	isGit := isGitRepo(absLocal)
+	marker := campaign.LinkMarker{
+		Version:      1,
+		CampaignRoot: normalizedCampaignRoot,
+		ProjectName:  name,
+	}
+	if cfg, err := config.LoadCampaignConfig(ctx, normalizedCampaignRoot); err == nil {
```
obey-agent: **Silent swallow of `LoadCampaignConfig` error leaves marker with empty `CampaignID`**

If `LoadCampaignConfig` fails for any reason — transient I/O, malformed campaign.yaml, context cancellation en route — the marker is written with `CampaignID = ""` and the failure never reaches the user. A later cross-check that uses `marker.CampaignID` (e.g. to validate a link still refers to the same campaign after a rename) will be useless, and there's no trace in logs to explain why.

Two concerns stacked here:

1. **Error swallow** — this is exactly the class of mistake the project's error-handling standards warn against. If you genuinely don't want to fail the link when config can't be loaded, at minimum log a warning so it's diagnosable:
   ```go
   cfg, cfgErr := config.LoadCampaignConfig(ctx, normalizedCampaignRoot)
   if cfgErr != nil {
       fmt.Fprintf(os.Stderr, "warning: could not load campaign config; marker will lack campaign_id: %v\n", cfgErr)
   } else {
       marker.CampaignID = cfg.ID
   }
   ```
2. **Context cancellation** — `LoadCampaignConfig` takes `ctx`. If the user hits Ctrl-C between the symlink-create and the marker-write, this block ignores `ctx.Err()` and proceeds to `WriteMarker` anyway, leaving a half-linked state (symlink exists, marker written without campaign_id, ensureInfoExclude not run). Check `ctx.Err()` after the config load, or treat `LoadCampaignConfig` failure as a rollback-and-bail with the symlink removed.

The second point matters because CLI cancellation is the normal user-signal for "I changed my mind" and the linked-project state shouldn't be left partial when that happens.

internal/project/link.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	content := strings.TrimRight(string(data), "\n")
+	if content != "" {
+		content += "\n"
+	}
+	content += pattern + "\n"
+	return os.WriteFile(path, []byte(content), 0644)
+}
+
```
obey-agent, line 432: **Non-atomic write corrupts `.git/info/exclude` on crash or concurrent writer**

`os.WriteFile` is not atomic — it truncates-then-writes. Two concerns:

1. **Crash window.** If the process is killed (OOM, Ctrl-C at the wrong microsecond, panic elsewhere) between the truncate and the final bytes landing on disk, the user's `.git/info/exclude` is left as a partial file. For `.git/info/exclude` specifically this is annoying rather than fatal, but it's the user's git metadata and silently damaging it violates the "leave things at least as good as we found them" principle.
2. **Concurrent writes.** Two `camp project add --link` invocations from different shells racing on the same linked repo will last-writer-wins the exclude file, potentially dropping one of the pattern entries.

The standard fix is write-to-temp + rename:

```go
func ensurePatternInFile(path, pattern string) error {
    if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
        return err
    }

    data, err := os.ReadFile(path)
    if err != nil && !os.IsNotExist(err) {
        return err
    }
    for _, line := range strings.Split(strings.TrimRight(string(data), "\n"), "\n") {
        if strings.TrimSpace(line) == pattern {
            return nil
        }
    }

    content := strings.TrimRight(string(data), "\n")
    if content != "" {
        content += "\n"
    }
    content += pattern + "\n"

    tmp, err := os.CreateTemp(filepath.Dir(path), ".exclude-*")
    if err != nil {
        return err
    }
    tmpName := tmp.Name()
    defer os.Remove(tmpName) // best-effort cleanup if rename fails

    if _, err := tmp.WriteString(content); err != nil {
        tmp.Close()
        return err
    }
    if err := tmp.Close(); err != nil {
        return err
    }
    return os.Rename(tmpName, path)
}
```

Same treatment for `removePatternFromFile`. This also makes the unit tests stable under parallelism.

obey-agent (approved):
Verdict: Approve

Follow-up:
The current head closes the remaining linked-marker integrity gaps. `.camp` now stores only `active_campaign_id`, mismatched legacy markers are rejected instead of being silently overwritten, `LoadRegistry` is best-effort for ownership messaging instead of blocking the guard itself, and `.git/info/exclude` updates now use atomic temp-file renames. Unit and linked integration coverage were rerun on this revision.

obey-agent (commented):
## Staff-Level Re-Review — Approve (with one small follow-up)

**Verdict**: Approve

**Overview**: All three of the inline issues I flagged in the previous review were addressed at `174c5b6` and `1a57ac3b`. The fixes are more ambitious than the minimum I suggested — instead of just patching the symptoms, you reshuffled the marker schema around `ActiveCampaignID` as the primary identifier and added a registry lookup so stale-marker error messages can actually name the other campaign. That's the right direction.

### How each prior finding landed

1. **`ensureLinkMarkerAvailable` silently relinks over stale markers** — fixed, and better than my suggestion. The function now:
   - Matches by `EffectiveCampaignID()` (new/legacy aware) or `CampaignRoot` fallback via `markerMatchesCampaign`.
   - Errors explicitly on legacy markers that lack an ID (`linked project has an existing legacy .camp marker for another campaign; remove it before linking`).
   - Uses `config.LoadRegistry` to resolve the conflicting campaign's current path and includes it in the error message.
   - Never silently takes over — the user gets a clear error instead of a lie.

2. **`LoadCampaignConfig` error swallowed in `AddLinked`** — fixed cleanly. `LoadCampaignConfig` is now called upfront (not lazily inside the marker construction), the error is propagated via `camperrors.Wrap`, `cfg.ID == ""` is validated explicitly, and the new `ctx.Err()` check after `ensureLinkMarkerAvailable` closes the cancellation gap I was worried about. The marker is now constructed with `ActiveCampaignID` set directly rather than the old pattern of "maybe-set-campaign-id-in-the-happy-path."

3. **Non-atomic `os.WriteFile` in `ensurePatternInFile`/`removePatternFromFile`** — fixed via a new `writeFileAtomically` helper that: preserves existing file mode via `os.Stat().Mode().Perm()`, creates the temp file in the same directory (so `os.Rename` stays atomic across filesystems), writes + `Sync()` + `Close()` + `Rename()`, and defers cleanup of the temp file. Both helpers now use it. Good.

### Schema refactor — worth acknowledging

The move from `{CampaignRoot, ProjectName, CampaignID}` → `{ActiveCampaignID, + legacy fields}` with `EffectiveCampaignID()` and `LinkMarkerVersion = 2` is the right long-term shape. Using campaign ID as the primary identifier is more robust than path-based matching (paths move, Dropbox/iCloud syncs differ across machines, EvalSymlinks resolution varies). The legacy field preservation keeps old markers readable. The new `internal/campaign/registry_lookup.go` with `CAMP_REGISTRY_PATH` + `XDG_CONFIG_HOME` is a nice touch for testability and multi-host setups.

### One small follow-up (not blocking)

**`WriteMarker` is still non-atomic.** `internal/campaign/link_marker.go:WriteMarker` calls `os.WriteFile(MarkerPath(projectDir), data, 0644)` directly. The `.camp` marker is arguably *more* critical than `.git/info/exclude` — a corrupted marker breaks detection entirely and the user may not know why until they hit `ErrNotInCampaign` from a linked directory. Route this through `writeFileAtomically` too (or add a thin wrapper in the `campaign` package if you want the helper to live closer to the marker code). Small change, and it rounds out the "no critical file write is non-atomic" invariant.

### What's Still Done Well

- Integration test coverage at `tests/integration/project_linked_test.go` was updated alongside the refactor (test `+25/-2`), so the new registry-lookup path is exercised end-to-end.
- `link_marker_test.go` got 43 new lines covering the new ID/legacy semantics.
- Error messages that name the conflicting campaign by *path* (resolved via registry lookup) are genuinely helpful rather than cryptic.
- The legacy-marker refusal path (`remove it before linking`) is explicitly conservative — better than silent takeover, even if it's slightly less ergonomic. Correct default.

### Staff Standard

I'd sign off on this as my own work. The fixes went beyond what I asked for, the refactor is in the right direction, and the tests follow. The `WriteMarker` atomicity nit is small enough to handle as a follow-up PR. Ship it.

obey-agent (approved):
Verdict: Approve

Follow-up:
The remaining marker-write issue is now closed on the current head. `.camp` writes use the same temp-file-plus-rename pattern as the exclude-file updates, and campaign marker coverage now includes preserving existing permissions while round-tripping the new `active_campaign_id` schema.

obey-agent (commented):
## Staff-Level Re-Review — Still Approved

**Verdict**: Approve

The atomic-write follow-up I flagged in the previous review is fixed at `ae21785`. `WriteMarker` now routes through `writeFileAtomically`, and the new `TestWriteMarker_PreservesExistingPermissions` specifically asserts that a 0600 marker stays 0600 after a rewrite — exactly the subtle permission-preservation behavior the helper depends on. Nice targeted test.

### What landed

- `WriteMarker` (`internal/campaign/link_marker.go:71`) now uses the same write-temp → sync → close → rename pattern as `ensurePatternInFile`/`removePatternFromFile` in `internal/project/link.go`.
- New test covers the mode-preservation invariant, and the assertion on `LinkMarkerVersion` also confirms the round-trip through the schema upgrade path.

### One small follow-up (not blocking)

**`writeFileAtomically` is now duplicated in two packages.** It lives identically in `internal/campaign/link_marker.go` and `internal/project/link.go` — 38 lines each, byte-for-byte copies. That's fine for now (both are unexported, both are self-contained), but these will drift the moment someone tweaks one and forgets the other. When you next touch either file, consider extracting it to something like `internal/fsutil/atomic_write.go` so there's a single source of truth. Pure cleanup, no behavior change.

### Staff Standard

Still approve. The fix is correct, the test is targeted, and the "no critical file write is non-atomic" invariant now holds across the entire linked-project write surface. Ship it.

lancekrogers:
Extracted `writeFileAtomically` into `internal/fsutil/atomic_write.go` as exported `WriteFileAtomically` in b5a8b23. Both `internal/campaign/link_marker.go` and `internal/project/link.go` now call the shared helper; the two identical copies are gone.

Added focused tests on the new package:
- `TestWriteFileAtomically_NewFile` — new-file write + default mode 0644
- `TestWriteFileAtomically_PreservesExistingMode` — 0600 file stays 0600 (the subtle invariant the higher-level tests already depended on)
- `TestWriteFileAtomically_NoTempLeftBehindOnSuccess` — no `.tmp-*` leftover
- `TestWriteFileAtomically_UnwritableDir` — surfaces the error when the parent dir doesn't exist

`just test all` passes (2120 unit + 145 integration). Linter is clean on all four touched files.

internal/project/link.go
```diff
@@ -0,0 +1,335 @@
...
+	destPath := opts.Path
+	if destPath == "" {
+		destPath = filepath.Join("projects", name)
+	}
+	fullPath := filepath.Join(campaignRoot, destPath)
+	if err := pathutil.ValidateBoundary(campaignRoot, fullPath); err != nil {
+		return nil, camperrors.Wrap(err, "project path boundary violation")
+	}
+
+	if _, err := os.Lstat(fullPath); err == nil {
```
obey-agent, line 116: This only catches destination-path collisions. The same external directory can still be linked again under a different name/path in the same campaign, and the branch then gets into a broken state: both link commands succeed, but `project list` only shows one alias and `camp project run --project <second-alias>` fails with `project not found in campaign`. Please reject duplicate target paths within a campaign before creating the second symlink.

internal/project/link.go
```diff
@@ -0,0 +1,335 @@
...
+		return nil, camperrors.Wrap(err, "create parent directory")
+	}
+	if err := os.Symlink(absLocal, fullPath); err != nil {
+		return nil, camperrors.Wrap(err, "create symlink")
+	}
+
+	isGit := isGitRepo(absLocal)
+	marker := campaign.LinkMarker{
+		Version:          campaign.LinkMarkerVersion,
+		ActiveCampaignID: cfg.ID,
```
obey-agent, line 130: `camp project add --link` now writes only `active_campaign_id`, but nothing here guarantees the current campaign is present in the local registry. I reproduced `camp init --no-register`, then `camp project add --link`, then `camp go p review-test --print` from inside the linked repo: detection falls through to `ErrNotInCampaign` because `resolveMarkerCampaignRoot()` can only recover the root from the registry. Please either register/sync the campaign as part of linking or persist a same-machine root fallback in the marker.

obey-agent (changes requested):
Verdict: Request Changes

Overview:
The linked-project work here is directionally strong, and the branch still carries the earlier marker-integrity fixes cleanly. Re-reviewing with fresh eyes turned up two new correctness gaps in the `--link` flow that leave users in states the CLI cannot resolve reliably.

Key Findings:
- `internal/project/link.go:109`: linking into a `camp init --no-register` campaign writes a marker that linked-cwd detection cannot resolve, because v2 markers only store `active_campaign_id` and `resolveMarkerCampaignRoot()` can only recover the root from the local registry.
- `internal/project/link.go:95`: the same external directory can still be linked twice into one campaign under different names; both adds succeed, but only one alias remains discoverable and the second cannot be resolved by name.

What's Done Well:
- Linked git vs linked non-git handling is threaded through list/resolve/run/worktree surfaces consistently.
- The branch keeps the earlier marker/exclude-file integrity fixes intact while expanding integration coverage.

Staff Standard:
Not yet. I would want both broken-state cases closed before I'd be comfortable shipping this as my own work.

internal/project/link.go
lancekrogers: I want to push back on this one.

**The repro is correct**, but the conclusion treats the \`--no-register\` + \`--link\` interaction as a correctness gap when it's actually a deliberate consequence of two orthogonal opt-outs meeting.

### \`--no-register\` is a feature, not an oversight

\`camp init --no-register\` (and \`camp clone --no-register\`) exists for two legitimate use cases:

1. **Ephemeral testing of the CLI itself** — contributors running \`camp init\` in scratch dirs without polluting their registry. This is how a lot of manual testing of camp happens today.
2. **Users who don't want their campaigns discoverable** via the global registry (privacy / isolation). The flag help is literally "Don't add to global registry."

Both cases explicitly opt out of registry-backed discovery. Any feature that requires the registry will not work for those users, and that's the contract they chose when they passed the flag.

### Both suggested fixes break that contract

- **"Register/sync the campaign as part of linking"** — this silently re-enters the campaign into the registry on the user's behalf, defeating \`--no-register\`. Users who passed that flag get their campaign entered anyway the moment they use \`--link\`. That's worse than the current behavior, not better.

- **"Persist a same-machine root fallback in the marker"** — this duplicates state. The registry is supposed to be the single source of truth for \`campaign_id → root\`. Putting the root in every \`.camp\` marker means:
  - Moving the campaign → every linked marker is now stale and needs repair.
  - Two caches of the same fact that can drift (which is exactly why your earlier review on this PR correctly pushed us to the minimal v2 marker that stores only \`active_campaign_id\`). Re-adding \`CampaignRoot\` partially undoes the simplification you signed off on at \`174c5b6\`.

### What I'd accept instead, if the UX gap is the real concern

The legitimate complaint inside your finding is: when a user is in a linked dir of an unregistered campaign, they get \`ErrNotInCampaign\` with no hint about what's wrong. That's a diagnostic / docs issue, not a design issue.

If you want to close the UX loop without touching the design, options in order of cheapness:

1. **Warn at link time**: if \`camp project add --link\` is run from a campaign that isn't in the registry, print a warning stating that linked-cwd detection won't work until the campaign is registered, and point at \`camp registry add\` (or equivalent). One-line feature, preserves \`--no-register\`.
2. **Reject the link at link time** for unregistered campaigns with the same message. More defensive, but strictly-speaking blocks a valid workflow (the user might not need linked-cwd detection).
3. **Document it** in \`--no-register\`'s help text and the linking docs: "Campaigns created with \`--no-register\` cannot be resolved from linked project directories; register the campaign to enable that."

Happy to add #1 to this PR if you think the diagnostic is worth adding now. But I don't want to register campaigns behind users' backs or put the campaign root back in every marker. Neither of those is the right response to "the user opted out of the registry."

internal/campaign/registry_lookup.go
```diff
@@ -0,0 +1,59 @@
...
+
+type registrySnapshot struct {
+	Campaigns map[string]registeredCampaign `json:"campaigns"`
+}
+
+type registeredCampaign struct {
+	Path string `json:"path"`
+}
+
+func lookupRegisteredCampaignRoot(campaignID string) (string, bool, error) {
```
obey-agent, line 28: **Parallel implementation of registry loading.** This file reimplements logic that already exists in `internal/config`:
- `registryPath()` duplicates `config.RegistryPath()` (paths.go:31), and hardcodes the literal `"obey"`/`"campaign"` strings instead of using the existing `config.OrgName`/`config.AppName` constants. If those constants ever change, this code silently drifts.
- `lookupRegisteredCampaignRoot` reimplements registry loading + JSON unmarshal + path normalization that `config.LoadRegistry` plus `Registry.GetByID` already provide (and which `internal/project/link.go:184` already uses).
- The function takes no `context.Context` — every other I/O entry point in the codebase does. This violates the "context first on I/O" project standard.

I see *why* you reached for a parallel impl: `internal/config` already imports `internal/campaign`, so importing the other direction would be an import cycle. The right fix is to extract registry loading + paths into a leaf package (e.g. `internal/registry`, `internal/paths/registry`) that both `config` and `campaign` can import — not to duplicate it. Two implementations of the same on-disk format is a maintenance trap.

Also, the function is package-private but has no test exercising the registry-miss path on its own; coverage comes incidentally through `TestDetect_FromLinkedProjectMarker` which constructs JSON via string concat (link_marker_test.go:117). That's brittle.

internal/campaign/detect.go
```diff
@@ -47,17 +38,43 @@ func Detect(ctx context.Context, startDir string) (string, error) {
...
 
-	// Walk up directory tree
+	realDir, err := filepath.EvalSymlinks(dir)
+	if err == nil && realDir != dir {
+		if root, walkErr := detectCampaignByWalking(ctx, realDir); walkErr == nil {
+			return root, nil
+		} else if !errors.Is(walkErr, ErrNotInCampaign) {
+			return "", walkErr
+		}
+	}
```
obey-agent, line 66: **Breaking semantic change to `CAMP_ROOT` env var, undocumented.** Before this PR, setting `CAMP_ROOT=/path/to/campaign` was a hard override — `Detect` returned that root immediately if it had `.campaign/`, regardless of cwd. Now (per the new `Detect` flow at L41-L73) the env var only takes effect as a *fallback* when walk-based detection fails AND the cwd is within the env value (`isPathWithin(dir, envRoot)`).

This is what forced the test changes in `internal/skills/skills_test.go` (`t.Setenv(EnvCampaignRoot, dir)` → `os.Chdir(dir) + Setenv(EnvCampaignRoot, "")`) and the rename of `detect_test.go:60` from `"env var override"` → `"validated env var inside campaign"`.

Real-world impact: any CI script or external tool doing `CAMP_ROOT=/work/campaign-X camp project list` from `/tmp` (or any directory outside the campaign) used to work and now returns `ErrNotInCampaign`. There's no migration note, no docs change, and no compatibility shim.

Two ways to land this safely:
1. Restore the early hard-override path for `CAMP_ROOT` and only treat it as fallback when *unset*. Linked-project detection is the new motivation; making linked detection co-exist with the legacy override is straightforward.
2. Keep the new behavior but: (a) call it out explicitly in the PR description, (b) add a `CAMP_ROOT_STRICT=1` opt-in for the old override, and (c) update any docs that mention CAMP_ROOT.

(1) is the safer choice; the old override was a debugging escape hatch and removing it surprises users.

internal/nav/exec.go
```diff
@@ -56,6 +57,10 @@ func ExecInDir(ctx context.Context, dir string, command []string) (*ExecResult,
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+	cmd.Env = os.Environ()
+	if campRoot, err := campaign.Detect(ctx, dir); err == nil && campRoot != "" {
```
obey-agent, line 61: **Performance regression on every command exec — uncached `campaign.Detect` in a hot path.** Each `cgo -c <cmd>` invocation now pays for an uncached `campaign.Detect` call here. The new Detect (per detect.go in this PR) walks up the directory tree, calls `EvalSymlinks`, tries `os.Stat(.campaign)` at each level, AND on each level reads `.camp` markers (JSON unmarshal), AND if a marker is found, falls into `lookupRegisteredCampaignRoot` which reads + JSON-parses `registry.json`. Every navigation exec.

The exact same pattern was added to `cmd/camp/cmdutil/command.go:31`. Both call sites should:
1. Use `campaign.DetectCached` (which already handles cwd-based caching via `internal/campaign/cache.go`), OR
2. Better — accept the campaign root as a parameter from the caller. The caller (e.g. `runGo`, `runRun`) already detected it via `DetectCached` once. Threading it down avoids both the cache miss and the second walk.

Right now the same campaign is detected twice per invocation, and the second detection is uncached. For shell completion / fast-feedback flows where camp is invoked many times per second, this matters. A minimal fix is to swap `campaign.Detect` → `campaign.DetectCached` (the cwd is the same), but threading is the staff-grade answer.

cmd/camp/project/commit.go
```diff
@@ -77,8 +77,15 @@ func runProjectCommit(cmd *cobra.Command, args []string) error {
 
 	resolvedPath := result.Path
 
+	if result.Source == projectsvc.SourceLinkedNonGit {
+		return fmt.Errorf("project %q is a linked non-git directory and cannot be committed with camp", result.Name)
+	}
```
obey-agent, line 82: **Magic-string check duplicated 8+ times.** This same shape:
```go
if resolved.Source == project.SourceLinkedNonGit {
    return fmt.Errorf("project %q is a linked non-git directory and does not support …", resolved.Name)
}
```
appears in: this file (commit.go:80), `cmd/camp/project/run.go:89`, `cmd/camp/project/remote/{add,list,remove,rename,set_url}.go`, `cmd/camp/project/worktree/{add,list,remove}.go`, `cmd/camp/worktrees/create.go:89` — at least 9 sites.

Three things wrong here:
1. `fmt.Errorf` for production errors violates the project's stated error standard.
2. The error message is hand-rolled at every site, so every site has slightly different wording ("git remotes", "git worktrees", "committed with camp", "does not support …"). That's predictable typo / drift territory.
3. The string-comparison-to-source-constant is the kind of check that benefits from a typed helper.

Suggested replacement (in `internal/project`):
```go
type ErrNonGitOperation struct {
    Name      string
    Operation string
}
func (e *ErrNonGitOperation) Error() string { … }

func (p Project) RequireGit(operation string) error {
    if p.Source == SourceLinkedNonGit {
        return &ErrNonGitOperation{Name: p.Name, Operation: operation}
    }
    return nil
}
```
Then every call site is `if err := resolved.RequireGit("git commits"); err != nil { return err }`. One source of truth, errors.Is-friendly, no string drift.

Or — push the check into `Resolve` via a `RequireGit bool` option so the cmd layer doesn't need to know about `SourceLinkedNonGit` at all.

internal/fsutil/atomic_write.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"os"
+	"path/filepath"
+)
+
+// WriteFileAtomically writes content to path via a same-directory temp file
+// and an atomic rename. If the destination already exists, its permission
+// bits are preserved; otherwise defaultMode is used. The temp file is fsync'd
+// before the rename so crashes cannot leave behind half-written content.
+func WriteFileAtomically(path string, content []byte, defaultMode os.FileMode) error {
```
obey-agent, line 14: **New helper introduced, but existing parallel implementations are not migrated.** This is a quality improvement (fsync, perm preservation, deferred temp cleanup) over the ad-hoc atomic-write patterns scattered across the codebase. But the PR only uses it from new code (link.go x2 + link_marker.go x1). The existing implementations in:
- `internal/pins/pins.go:51` (Save)
- `internal/nav/index/cache.go:29` (cache write)
- `internal/workitem/priority/store.go:38` (Save)
- `internal/config/registry.go:67` (SaveRegistry, atomic write per the comment but I haven't verified)

…all stay on their inferior implementations. The project guideline is explicit: "Did you ensure that you improved existing code where appropriate rather than implementing parallel versions of the features you were assigned?" Adding a 5th implementation without consolidating to it leaves the codebase in a worse state than before.

Either:
1. Migrate the four existing call sites in this PR (mechanical change, ~5 lines each), or
2. Land this helper in a follow-up PR explicitly scoped to the consolidation, with the migration being the whole point.

(2) is fine if the scope creep concern is real — but please open the follow-up before this lands so the intent is on the record.

internal/project/link.go
```diff
@@ -0,0 +1,335 @@
...
+		_ = os.Remove(fullPath)
+		return nil, camperrors.Wrap(err, "write .camp marker")
+	}
+
+	if err := ensureInfoExclude(ctx, campaignRoot, filepath.ToSlash(destPath)); err != nil {
+		fmt.Fprintf(os.Stderr, "Warning: could not update campaign exclude file: %v\n", err)
+	}
+	if isGit {
+		if err := ensureGitInfoExclude(ctx, absLocal, campaign.LinkMarkerFile); err != nil {
+			fmt.Fprintf(os.Stderr, "Warning: could not ignore .camp in linked repo: %v\n", err)
```
obey-agent, line 139: **`fmt.Errorf` in production code, plus warnings being silently dropped to stderr.** Several issues compounded in this file:

1. `fmt.Errorf` at L41, L57, L69, L180, L193, L197, L219 — the project standard says "Always use the project's established error framework. Never use `fmt.Errorf` in production code." Several of these would be much better as typed sentinels: `ErrLinkPathRequired`, `ErrCampaignMissingID`, `ErrLegacyMarkerCollision`, `ErrAlreadyLinkedToOtherCampaign`. Then commands can branch on them with `errors.Is`.

2. `fmt.Fprintf(os.Stderr, "Warning: …")` at L117 (info/exclude update failed) and L121 (linked-repo .camp ignore failed). These are quietly downgraded from errors to log lines, but they have real consequences:
   - L117 failing means the next `git status` in the campaign root will show `projects/<name>` as untracked. The user has no way to detect this happened from the success path of `camp project add --link`.
   - L121 failing means the linked repo's owner will see `.camp` as a tracked-but-unstaged file in their *external* repo, which is exactly what this code was supposed to prevent.
   
   At minimum, surface these in the `LinkResult` (e.g., `Warnings []string`) and have the cmd layer print them with appropriate emphasis. Better: bubble them up as errors and have the caller decide; an unconfigured `info/exclude` is a genuine failure of the "linked projects don't pollute git" contract.

3. The same warning-to-stderr antipattern recurs in `UnlinkProject` at L156 — and there it can leave the linked repo with stale `.camp` in info/exclude permanently after unlink.

internal/campaign/link_marker_test.go
```diff
@@ -0,0 +1,120 @@
...
+		t.Fatalf("ActiveCampaignID = %q, want %q", marker.ActiveCampaignID, "campaign-123")
+	}
+	if marker.Version != LinkMarkerVersion {
+		t.Fatalf("Version = %d, want %d", marker.Version, LinkMarkerVersion)
+	}
+}
+
+func writeTestRegistry(path, campaignID, campaignRoot string) error {
+	data := []byte("{\n  \"version\": 2,\n  \"campaigns\": {\n    \"" + campaignID + "\": {\n      \"name\": \"test\",\n      \"path\": \"" + campaignRoot + "\"\n    }\n  }\n}\n")
+	return os.WriteFile(path, data, 0o644)
```
obey-agent: **Test fixture builds registry JSON via string concat.** `writeTestRegistry` hand-rolls the JSON payload as a Go string literal:
```go
data := []byte("{\n  \"version\": 2,\n  \"campaigns\": {…}")
```
This is a brittle test fixture. The canonical representation of the registry lives in `internal/config/Registry` with custom marshaling rules (e.g. `config.LoadRegistry` populates `ID` from the map key separately, see registry.go:53-57). If `config.SaveRegistry` ever changes the format — adds a field, changes a key name — this string-concat fixture won't update, and the test will produce JSON that the canonical loader rejects.

Use `config.SaveRegistry` (or `config.NewRegistry` + `Add` + `SaveRegistry`) to seed the test registry. That way the fixture and the production code stay in lockstep.

This is also the reason the parallel registry implementation in `internal/campaign/registry_lookup.go` *appears* to work in tests — the test bypasses the canonical write path entirely. If it used the real `SaveRegistry`, the parallel reader would have to match its output exactly and the duplication risk would be visible.

obey-agent (commented):
## Verdict: Request Changes

## Overview
This PR adds a meaningful new capability — non-submodule "linked" projects backed by a symlink + `.camp` marker file — and the integration test coverage for the user-facing flows is solid. The marker schema (v1 → v2 with backward-compat) is well-considered. But the implementation duplicates several existing patterns in the codebase, takes shortcuts on the project's stated error standards, and introduces a real performance regression and an undocumented breaking change to `CAMP_ROOT`. Several issues need to be addressed before this lands.

## Key Findings (must-fix)
- **Parallel registry implementation in `internal/campaign/registry_lookup.go`** — duplicates `config.LoadRegistry` and `config.RegistryPath`, hardcodes the `"obey"`/`"campaign"` literals instead of using the existing constants, and takes no `context.Context`. The right fix is to extract registry loading to a leaf package both `config` and `campaign` can import. See inline on registry_lookup.go:28.
- **Breaking semantic change to `CAMP_ROOT`** — used to be a hard override, now only a fallback that requires cwd to be inside it. Forces the test changes in `skills_test.go` and the rename of `detect_test.go:60`. Not called out in the PR description; no migration note. See inline on detect.go:60.
- **Performance regression in `internal/nav/exec.go` and `cmd/camp/cmdutil/command.go`** — uncached `campaign.Detect` on every command exec, which now reads JSON markers and potentially the registry on each filesystem walk. Should be `DetectCached` at minimum, ideally threaded through from the caller. See inline on exec.go:61.
- **`SourceLinkedNonGit` magic-string check duplicated 8+ times** across `commit.go`, `run.go`, `remote/*.go`, `worktree/*.go`, `worktrees/create.go` — each with its own hand-rolled `fmt.Errorf` message that drifts site-to-site. Should be a `Project.RequireGit(operation)` helper or a `Resolve` option. See inline on commit.go:82.
- **`fmt.Errorf` in production code** at multiple sites in `internal/project/link.go` (L41, L57, L69, L180, L193, L197, L219), violating the project standard. Should be typed sentinels (`ErrLinkPathRequired`, `ErrAlreadyLinkedToOtherCampaign`, etc.) so commands can branch on them. See inline on link.go:121.
- **Important warnings silently dropped to stderr** in `AddLinked` and `UnlinkProject` — failures to update `info/exclude` mean the linked repo will leak `.camp` into git status, which is exactly the failure mode this code was supposed to prevent. Should be in `LinkResult.Warnings` at minimum, or returned as errors. See inline on link.go:121 (item 2).

## Key Findings (should-fix)
- **`fsutil.WriteFileAtomically` is a strict improvement over the four ad-hoc atomic-write implementations in the codebase (pins, nav/index/cache, workitem/priority/store, config/registry), but the PR doesn't migrate any of them.** The codebase now has 5 atomic-write impls instead of 1. Either consolidate in this PR or open a follow-up. See inline on atomic_write.go:14.
- **Test fixture `writeTestRegistry` constructs registry JSON via string concat** instead of using `config.SaveRegistry`. Brittle to format drift, and incidentally hides the parallel-implementation problem above. See inline on link_marker_test.go:119.

## What's Done Well
- The marker schema versioning (v1 → v2) with `EffectiveCampaignID()` is the right shape — backward-compatible reads, forward-only writes. This is correct hard-mode design.
- Integration tests in `tests/integration/project_linked_test.go` cover the user-facing flows end-to-end: add/remove, multi-campaign rejection, `info/exclude` cleanliness, auto-detect from linked cwd, run dispatch in linked project. Comprehensive.
- The fall-through chain in `Detect` (logical → resolved → marker → env-var-fallback) handles the symlinked-cwd case explicitly with a separate code path; that's better than the EvalSymlinks-then-walk approach it replaces.
- `linkedProjectTarget` and `validateManagedProjectPath` keep the boundary check, even though the failure mode shifted from "path-resolution boundary error" to "linked project delete blocked". Net behavior for end users is the same.
- `LinkResult` correctly wraps the typed result instead of `(string, string, bool, ...)` — easy to extend.
- All unit tests pass; the build is green.

## Staff Standard
**No.** Not in this state. The new capability is a real product value, but landing 1940 lines of additions that include (a) a parallel registry implementation, (b) an undocumented breaking change to a public env var, (c) `fmt.Errorf` violations after the explicit project standard against them, (d) a duplicated 8-site magic-string check, and (e) silent stderr warnings on file operations that have correctness implications — that's not staff-grade delivery, even with a strong test suite.

Address the six must-fix items and this becomes a confident approve. The two should-fix items can land here or as immediate follow-ups, but please commit to the consolidation plan if you defer them.

internal/project/link.go
lancekrogers: Addressed in a targeted follow-up PR against this branch: [#214](https://github.com/Obedience-Corp/camp/pull/214).

Agree the symptom you reproduced is real — but the root cause turned out to be \`deduplicateByRemoteURL\` in \`internal/project/list.go:230\` silently keeping only one per repo URL, not a missing destination-path check. Rejecting duplicate targets up front fixes the user-visible brokenness (silent disappearance from \`list\` / \`run\` / \`leverage\`) while preserving the URL dedup — it still handles the submodule-plus-linked-clone case that was its original purpose.

The new \`ErrProjectAlreadyLinked\` names both the existing and attempted aliases and points at \`camp project unlink\` so re-linking under a new name is straightforward. Tests cover both the rejection and the unlink-then-relink path.

obey-agent:
Moved the linked-project history semantics into #207 because this is base feature behavior, not stacked follow-up polish. Linked project symlinks are now tracked in the campaign repo and auto-commit as distinct \ / \ actions so git-history reporting can tell them apart from submodule \ / \. The external repo’s \ marker still stays local/ignored.\n\nTests:\n- go test ./cmd/camp/project ./internal/project ./internal/git/commit\n- go test ./tests/integration -tags integration -run 'TestProject_(AddLink_|Remove_LinkedProject)|TestProjectRun_AutoDetectFromLinkedCwd|TestGo_FromLinkedProjectCwd|TestRun_ProjectDispatch_LinkedProject'

obey-agent:
Moved the linked-project history semantics into #207 because this is base feature behavior, not stacked follow-up polish. Linked project symlinks are now tracked in the campaign repo and auto-commit as distinct Link / Unlink actions so git-history reporting can tell them apart from submodule Add / Remove. The external repo’s .camp marker still stays local/ignored.

Tests:
- go test ./cmd/camp/project ./internal/project ./internal/git/commit
- go test ./tests/integration -tags integration -run 'TestProject_(AddLink_|Remove_LinkedProject)|TestProjectRun_AutoDetectFromLinkedCwd|TestGo_FromLinkedProjectCwd|TestRun_ProjectDispatch_LinkedProject'

cmd/camp/project/add.go
```diff
@@ -115,6 +193,171 @@ func runProjectAdd(cmd *cobra.Command, args []string) error {
...
+	}
+
+	if !targetChanged {
+		cfg, campaignRoot, err := r.loadCurrent(ctx)
+		if err == nil {
+			reg, regErr := r.loadRegistry(ctx)
+			if regErr != nil {
+				return nil, "", camperrors.Wrap(regErr, "load registry")
+			}
+			if err := ensureProjectAddCampaignRegistered(reg, cfg, campaignRoot); err != nil {
```
obey-agent: @/tmp/camp207-inline.md

obey-agent (changes requested):
Verdict: Request Changes
Risk: medium-high.

Overview:
This PR adds the linked-project model and wires it through campaign detection, run/navigation, and project resolution paths. The marker ownership work, linked cwd handling, and integration coverage are all materially stronger than the earlier version, but the new project-target resolver regresses an existing in-campaign add workflow.

Key Findings:
- [P1] `cmd/camp/project/add.go:271` now rejects the current campaign unless it is registered in the global registry, which breaks `camp project add` / `camp project add --link` inside campaigns created with `--no-register` or later unregistered. I reproduced the failure on the current head; `main` does not have this dependency.

What's Done Well:
- The linked marker format and reverse-detection path are much more defensible now than the earlier mutable-root approach.
- The linked-project integration coverage around add/list/run/go/remove is substantive and caught the intended workflows.
- The new `Link` / `Unlink` commit actions improve campaign history clarity.

Staff Standard:
Not yet. I would not be comfortable merging a change that makes an existing in-campaign project-add flow depend on registry state unless that product decision is explicit, documented, and consistently enforced across related commands.

obey-agent:
## Staff Review: Camp PR #207

**Verdict:** Approve

This review is **not** treating PR size as a problem. The branch is large because it carries a full feature slice with command wiring, internal plumbing, unit tests, integration tests, and docs. That is acceptable. The question is whether the feature is coherent, correctly integrated, and safe to ship.

### Correction

A prior version of this review treated registry-backed reverse detection for linked repos from unregistered campaigns as a blocker. That was the wrong product contract.

The clarified contract is:
- unregistered campaigns may still be used from their own cwd
- unregistered campaigns are intentionally **not** discoverable from outside that cwd

Linked-repo reverse detection is an external discovery path, so its dependence on registry state is consistent with that contract. I am withdrawing that prior finding.

### Overview

This is strong feature work. The linked-project model is coherent, the command surface is clearer than the earlier `project add --link` overloading, and the branch does the hard integration work instead of stopping at a narrow happy path: `project link` / `project unlink`, detection, resolution, list/remove, `camp go`, `camp run`, `project commit`, remotes, worktrees, unit tests, integration tests, and docs are all updated together.

Campaign-owned state stays in the campaign repo, machine-local reverse-detection state stays in the external repo, marker ownership is protected, and linked git vs linked non-git behavior is handled intentionally across the command surface. Given the clarified discoverability contract, I do not have remaining blocking correctness findings from this review pass.

### Architecture Snapshot

```mermaid
flowchart LR
  A["camp project link"] --> B["internal/project.AddLinked"]
  B --> C["projects/<name> symlink"]
  B --> D["write local .camp marker"]
  E["camp go / camp run / project commit from linked cwd"] --> F["campaign.Detect"]
  F --> G["read marker"]
  G --> H["resolve active_campaign_id through registry"]
  H --> I["recover campaign root"]
  I --> J["dispatch linked-project workflow"]
```

### What Is Done Well

- The feature is integrated rather than bolted on.
- `project link` / `project unlink` is a clearer CLI model than continuing to overload `project add`.
- Marker ownership protections are materially better than the earlier mutable-root approach.
- `internal/fsutil.WriteFileAtomically` is the right extraction.
- The linked git vs linked non-git distinction is carried consistently through remotes, worktrees, and command gating.
- The TTY integration helper gives the interactive flows real test coverage.

### Validation

I reviewed the full diff against `main` and saved the raw diff plus full staff review package locally in the campaign repo.

Validation on the current PR head:
- `go test -tags integration ./tests/integration` ✅
- `go test ./...` ⚠️ fails in `cmd/camp/leverage`, but the same failure occurs on `main`, so that is baseline noise rather than a regression from this PR
- linked-project command routing works across the intended surfaces on the current head

### Recommendation

Mergeable. Risk is medium because the feature touches a broad surface area, but I do not have remaining blocking findings from this review pass.

Updated local review artifacts:
- `workflow/code_reviews/camp-pr-207/REVIEW.md`
- `workflow/code_reviews/camp-pr-207/PR_DIFF.patch`

obey-agent (approved):
Verdict: Approve

Follow-up:
With product intent clarified that unregistered campaigns are intentionally not discoverable outside their own cwd, I am withdrawing the prior registry/discoverability blocker. I do not have remaining blocking findings on the current head.

Risk:
Medium, because the feature is broad, but the linked-project implementation is coherent and covered across the operational surfaces that matter.

obey-agent (approved):
Verdict: Approve

Follow-up:
With product intent clarified that unregistered campaigns are intentionally not discoverable outside their own cwd, I am withdrawing the prior registry/discoverability blocker. I do not have remaining blocking findings on the current head.

Risk:
Medium, because the feature is broad, but the linked-project implementation is coherent and covered across the operational surfaces that matter.

cmd/camp/project/commit.go
obey-agent: Addressed on ab83127. Added internal/project.NonGitOperationError plus ResolveResult.RequireGit(operation), replaced the repeated SourceLinkedNonGit + fmt.Errorf checks across project commit / remote / worktree command paths, and switched project commit cancellation to git.ErrCommitCancelled so the error chain now carries the camp cancellation sentinel instead of a raw string.

internal/campaign/detect.go
obey-agent: Revalidated on the current head (`0cf7545`). I built `/tmp/camp-pr207` from this branch and `/tmp/camp-main` from `origin/main`, then from an outside cwd ran `CAMP_ROOT=<tmp>/campaign camp root` against both. `origin/main` returned the campaign root with exit 0; this branch returned `Error: not inside a campaign directory` with exit 1. So this is still a real compatibility regression in the current revision, not just an early-diff concern.

obey-agent (changes requested):
Verdict: Request Changes

Overview:
The linked-project implementation is materially stronger than the early revisions and the targeted linked unit/integration flows I re-ran on the current head passed. The remaining blocker is a compatibility regression in campaign detection: this branch changes `CAMP_ROOT` from a hard override into a bounded fallback, which breaks existing out-of-tree invocations that still work on `origin/main`.

Key Findings:
- [P1] `internal/campaign/detect.go:62`: `CAMP_ROOT=/path/to/campaign camp root` from an outside cwd now fails with `not inside a campaign directory`. I verified this on the current head by building real binaries from both this branch and `origin/main`; `origin/main` exits 0 and prints the campaign root, while this branch exits 1. That is a real behavioral regression for external tooling and scripts unless the contract is being intentionally changed and migrated.

What's Done Well:
- The linked-project model is coherent now: marker ownership, linked git vs linked non-git behavior, and linked cwd resolution are all much more defensible than the earlier revisions.
- The branch updates the operational surfaces that matter instead of stopping at a narrow happy path: list, resolve, run, go, commit, remotes, worktrees, and linked integration coverage all moved together.
- The targeted validation I ran on the current head passed: `go test ./internal/campaign ./internal/project ./cmd/camp/project ./cmd/camp/project/remote ./cmd/camp/project/worktree ./internal/fsutil` and `go test -tags integration ./tests/integration -run '\''TestProject_(Link_|Unlink_|Commit_LinkedGitRepo)|TestProjectRun_AutoDetectFromLinkedCwd|TestGo_FromLinkedProjectCwd|TestRun_ProjectDispatch_LinkedProject'\''`.

Staff Standard:
Not yet. I would not merge an undocumented compatibility break to `CAMP_ROOT` without either restoring the prior hard-override semantics or explicitly changing and documenting that contract.

internal/campaign/detect.go
obey-agent: Addressed on fd9bba7. Restored CAMP_ROOT to its historical hard-override behavior before the linked-project walk logic runs, while still normalizing the override path through filepath.Abs/EvalSymlinks when possible. Added container-backed regression coverage in tests/integration/root_test.go that runs `cd /test && CAMP_ROOT=/campaigns/root-env /camp root` and asserts exit 0 plus the expected outside-campaign relative root. Validation: `go test ./internal/campaign` and `go test ./tests/integration -tags integration -run "^TestRoot_(AtCampaignRoot|FromNestedDirectory|JSON|NotInCampaign|RespectsCampRootOverrideOutsideCampaign)$"`.

obey-agent: Addressed on fd9bba7. Restored CAMP_ROOT to its historical hard-override behavior before the linked-project walk logic runs, while still normalizing the override path through filepath.Abs/EvalSymlinks when possible. Added container-backed regression coverage in tests/integration/root_test.go that runs `cd /test && CAMP_ROOT=/campaigns/root-env /camp root` and asserts exit 0 plus the expected outside-campaign relative root. Validation: `go test ./internal/campaign` and `go test ./tests/integration -tags integration -run "^TestRoot_(AtCampaignRoot|FromNestedDirectory|JSON|NotInCampaign|RespectsCampRootOverrideOutsideCampaign)$"`.

obey-agent (approved):
Verdict: Approve

Overview:
I re-reviewed PR #207 at head `fd9bba7d9ef2668e7fb0543c1c687eb8da91e0be`, with fresh attention on the post-review commits and their interaction with the larger linked-project stack. The latest head fixes the real `CAMP_ROOT` regression without reopening the linked-project correctness issues we already worked through, and the recent error-package cleanup is structurally sound.

What I re-validated:
- `internal/campaign/detect.go` restores `CAMP_ROOT` to a hard override before cwd/symlink-based campaign discovery, which brings `camp root` back in line with `origin/main` for out-of-tree tooling.
- The new integration coverage in `tests/integration/root_test.go` exercises that exact external-tooling path.
- The linked non-git guard centralization in `internal/project/git_requirements.go` is a good refactor: it removes repeated branch-local checks without weakening behavior, and the existing linked project remote/worktree flows still reject correctly.
- The error-routing cleanup does not appear to introduce a functional regression in the linked-project or campaign-targeting flows.

Verification:
- `go test ./internal/campaign ./internal/project ./internal/config ./cmd/camp/project ./cmd/camp/project/remote ./cmd/camp/project/worktree ./cmd/camp ./cmd/camp/intent`
- `go test ./tests/integration -tags integration -run 'TestRoot_RespectsCampRootOverrideOutsideCampaign|TestProject_(Link_PicksCampaignOutsideCurrentContext|Link_NonGitDir|.*Add_UsesCurrentCampaignEvenIfUnregistered|AddLink_RejectsDuplicateTargetInSameCampaign|AddLink_AllowsRelinkAfterUnlink)|TestProjectRun_'`
- `go test ./...` (same pre-existing `cmd/camp/leverage` failures reproduce on `origin/main`; I did not treat those as PR regressions)
- direct binary repro comparing `origin/main`, the previously reviewed head, and the current head for the `CAMP_ROOT` override path

What’s done well:
- The latest commit is narrowly scoped and fixes the contract break at the right layer instead of papering over it elsewhere.
- The recent refactors reduce duplicated linked-project guard logic and bring the stack into the repo’s error-handling conventions.
- Integration coverage on the linked-project flows remains materially better than where this PR started.

Staff standard:
Yes. On the current head, I would be comfortable merging this.

obey-agent (approved):
Verdict: Approve

Overview:
I re-reviewed PR #207 at the new head `dd528608e13bdce83fe6169d049f774f61be0e73`. The delta since the previous approval is test-only, and I do not see a new correctness or maintainability regression in the current head.

What changed in this delta:
- linked-project filesystem checks that were previously using host `t.TempDir()` paths were moved toward the containerized integration harness
- leverage and skills tests were updated to rely on `CAMP_ROOT` instead of process-wide `chdir`, which is a better fit now that the override semantics are restored
- the linked delete-block behavior is still covered, now through integration rather than a host-filesystem unit test

What I validated:
- `go test ./cmd/camp/leverage -run 'TestLeverageReset_'`
- `go test ./internal/campaign -run 'TestDetect'`
- `go test ./internal/git/commit -run 'TestProjectAction_Values|TestProjectAction_Values_LinkedActions'`
- `go test ./internal/skills -run 'TestFindSkillsDir|TestFindSkillsDir_NoSkillsDir'`
- `go test ./internal/project`
- `go test ./tests/integration -tags integration -run 'TestProject_(Link_GitRepo|Link_PicksCampaignOutsideCurrentContext|Unlink_LinkedProject|Remove_LinkedProjectDeleteBlocked)|TestGo_FromLinkedProjectCwd|TestRun_ProjectDispatch_LinkedProject|TestProjectRun_AutoDetectFromLinkedCwd'`

Baseline note:
- `go test ./cmd/camp/leverage` still has the same broader campaign-detection failures that reproduce outside this delta; I did not treat that package-wide baseline noise as a regression from this commit.

Staff standard:
Yes. On the current head, this remains mergeable.

obey-agent:
Follow-up comment to consolidate the outstanding review state on this PR, because several earlier `obey-agent` inline threads are still left open in GitHub even though later commits and review passes superseded them.

Superseded / no longer blocking on the current head:
- linked marker overwrite across campaigns: fixed by rejecting links to a repo already linked to another live campaign
- `CAMP_ROOT` hard-override regression: fixed, with follow-up integration coverage
- repeated `SourceLinkedNonGit` guard drift: fixed via `ResolveResult.RequireGit(...)` and `NonGitOperationError`
- `.camp` marker atomic write gap: fixed
- the earlier "unregistered campaigns should still be discoverable from linked cwd" blocker was withdrawn after product intent was clarified that unregistered campaigns are intentionally not discoverable outside their own cwd

Still-open follow-up items on the current head `dd528608e13bdce83fe6169d049f774f61be0e73` that were not fixed in this PR:
- `internal/campaign/registry_lookup.go` still duplicates registry-loading logic that already exists in `internal/config`; it would be better to converge on one source of truth instead of carrying a parallel reader
- `internal/nav/exec.go` and `cmd/camp/cmdutil/command.go` still call uncached `campaign.Detect(...)` in command-exec hot paths; `DetectCached` or caller-threaded campaign root would be the cleaner design
- `internal/fsutil/atomic_write.go` is a good extraction for the new linked-project writes, but the older ad-hoc atomic-write sites elsewhere in the codebase still have not been migrated to it
- `internal/project/link.go` still mixes some raw `fmt.Errorf`-style error construction / stderr-only warning surfacing for ignore-file update failures; that would still benefit from a cleaner typed-error / explicit-warning pass

I’m leaving this as a top-level PR comment instead of another inline note because the inline follow-ups kept getting buried. Current overall review state remains `APPROVED`; the list above is intended to make the remaining follow-up debt explicit, not to reopen the blockers that were already fixed.

PR #208, opened 2026-04-10
fix: allow concept names as camp go navigation targets

lancekrogers:
## Summary
- `cgo docs`, `cgo ai_docs`, `cgo festivals`, etc. now work alongside the shortcut keys (`d`, `ai`, `f`)
- Added `CampaignPaths.AsMap()` to expose path concept names as a map
- Updated `BuildCategoryMappings` to accept optional `CampaignPaths` and merge concept-name mappings alongside shortcut-key mappings (shortcuts win on collision)

## Root cause
`BuildCategoryMappings` only iterated over the `shortcuts` section of `jumps.yaml`, which maps abbreviation keys (`d`, `ai`, `f`) to categories. The `paths` section defines full concept names (`docs`, `ai_docs`, `festivals`) but was never used for navigation resolution. So `cgo docs` fell through to fuzzy search with no category and failed.

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes (all packages)
- [x] New test `TestBuildCategoryMappings_WithPaths` covers concept name resolution
- [x] Updated completion test expectations for expanded mapping count

### DIFF
--- a/cmd/camp/navigation/go.go
+++ b/cmd/camp/navigation/go.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Obedience-Corp/camp/cmd/camp/cmdutil"
 	"github.com/Obedience-Corp/camp/internal/config"
 	"github.com/Obedience-Corp/camp/internal/nav"
+	"github.com/Obedience-Corp/camp/internal/nav/fuzzy"
 	"github.com/Obedience-Corp/camp/internal/nav/index"
 	"github.com/Obedience-Corp/camp/internal/pins"
 	"github.com/Obedience-Corp/camp/internal/state"
@@ -59,9 +60,9 @@ func runGo(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// Build category mappings from config shortcuts
-	// This allows config shortcuts to work with fuzzy search
-	configMappings := nav.BuildCategoryMappings(cfg.Shortcuts(), cfg.PathsMap())
+	// Resolve configured navigation from shortcuts, long-form directory aliases,
+	// and configured concepts.
+	resolved := nav.ResolveConfiguredTarget(cfg, args)
 
 	// Handle toggle keyword: "t" or "toggle"
 	if len(args) > 0 && (args[0] == "toggle" || args[0] == "t") {
@@ -80,8 +81,24 @@ func runGo(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	// Parse shortcuts using config mappings resolved from the campaign config.
-	result := nav.ParseShortcut(args, configMappings)
+	result := nav.ParseResult{
+		Category:   nav.CategoryAll,
+		Query:      strings.Join(args, " "),
+		IsShortcut: false,
+	}
+	if resolved.Matched {
+		if resolved.RelativePath != "" {
+			return handleRelativePathNavigation(ctx, campaignRoot, resolved.RelativePath, resolved.Query, printOnly, command)
+		}
+		if resolved.Drill {
+			return handleRelativePathNavigation(ctx, campaignRoot, resolved.Category.Dir(), resolved.Query, printOnly, command)
+		}
+		result = nav.ParseResult{
+			Category:   resolved.Category,
+			Query:      resolved.Query,
+			IsShortcut: true,
+		}
+	}
 
 	// Check for sub-shortcut in remaining args
 	// Example: "camp p fest cli" -> result.Query="fest", subShortcut="cli"
@@ -358,6 +375,96 @@ func handleCustomNavShortcut(ctx context.Context, sc config.ShortcutConfig, camp
 		fmt.Printf("cd %s\n", jumpResult.Path)
 	}
 	return nil
+}
+
+// handleRelativePathNavigation resolves a configured relative path plus optional
+// query and executes the standard camp go output flow for the resolved target.
+func handleRelativePathNavigation(ctx context.Context, campaignRoot, relativePath, query string, printOnly bool, command []string) error {
+	targetPath, err := resolveRelativePathNavigation(ctx, campaignRoot, relativePath, query)
+	if err != nil {
+		return err
+	}
+
+	if len(command) > 0 {
+		execResult, err := nav.ExecInDir(ctx, targetPath, command)
+		if err != nil {
+			return err
+		}
+		if execResult.ExitCode != 0 {
+			return camperrors.NewCommand("", execResult.ExitCode, "", nil)
+		}
+		return nil
+	}
+
+	cwd, _ := os.Getwd()
+	_ = state.SetLastLocation(ctx, campaignRoot, cwd)
+
+	if printOnly {
+		fmt.Println(targetPath)
+	} else {
+		fmt.Printf("cd %s\n", targetPath)
+	}
+	return nil
+}
+
+func resolveRelativePathNavigation(ctx context.Context, campaignRoot, relativePath, query string) (string, error) {
+	if query == "" {
+		jumpResult, err := nav.JumpToPathFromRoot(ctx, campaignRoot, relativePath)
+		if err != nil {
+			return "", err
+		}
+		return jumpResult.Path, nil
+	}
+
+	basePath := filepath.Join(campaignRoot, relativePath)
+	exactPath := filepath.Join(basePath, query)
+	if info, err := os.Stat(exactPath); err == nil && info.IsDir() {
+		return exactPath, nil
+	}
+
+	if strings.Contains(query, "/") {
+		parts := strings.SplitN(query, "/", 2)
+		prefixPath, err := fuzzyResolveDirectory(ctx, basePath, parts[0], relativePath)
+		if err != nil {
+			return "", err
+		}
+		nestedPath := filepath.Join(prefixPath, parts[1])
+		if info, err := os.Stat(nestedPath); err == nil && info.IsDir() {
+			return nestedPath, nil
+		}
+		return "", fmt.Errorf("path does not exist: %s/%s", strings.TrimRight(relativePath, "/"), query)
+	}
+
+	return fuzzyResolveDirectory(ctx, basePath, query, relativePath)
+}
+
+func fuzzyResolveDirectory(ctx context.Context, basePath, query, relativePath string) (string, error) {
+	if ctx.Err() != nil {
+		return "", ctx.Err()
+	}
+
+	entries, err := os.ReadDir(basePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "", fmt.Errorf("path does not exist: %s", relativePath)
+		}
+		return "", camperrors.Wrap(err, "failed to read navigation path")
+	}
+
+	var names []string
+	for _, entry := range entries {
+		if !entry.IsDir() || strings.HasPrefix(entry.Name(), ".") {
+			continue
+		}
+		names = append(names, entry.Name())
+	}
+
+	matches := fuzzy.FilterMulti(names, query)
+	if len(matches) == 0 {
+		return "", fmt.Errorf("no directories matching %q in %s", query, strings.TrimRight(relativePath, "/"))
+	}
+
+	return filepath.Join(basePath, matches[0].Target), nil
 }
 
 // evalSymlinks resolves symlinks in a path, returning the original path if resolution fails.
--- a/cmd/camp/navigation/go.go
+++ b/cmd/camp/navigation/go.go
@@ -61,7 +61,7 @@ func runGo(cmd *cobra.Command, args []string) error {
 
 	// Build category mappings from config shortcuts
 	// This allows config shortcuts to work with fuzzy search
-	configMappings := nav.BuildCategoryMappings(cfg.Shortcuts())
+	configMappings := nav.BuildCategoryMappings(cfg.Shortcuts(), cfg.PathsMap())
 
 	// Handle toggle keyword: "t" or "toggle"
 	if len(args) > 0 && (args[0] == "toggle" || args[0] == "t") {
--- a/cmd/camp/navigation/go.go
+++ b/cmd/camp/navigation/go.go
@@ -60,10 +60,6 @@ func runGo(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// Resolve configured navigation from shortcuts, long-form directory aliases,
-	// and configured concepts.
-	resolved := nav.ResolveConfiguredTarget(cfg, args)
-
 	// Handle toggle keyword: "t" or "toggle"
 	if len(args) > 0 && (args[0] == "toggle" || args[0] == "t") {
 		return handleToggle(ctx, campaignRoot, printOnly)
@@ -80,6 +76,11 @@ func runGo(cmd *cobra.Command, args []string) error {
 			}
 		}
 	}
+
+	// Resolve configured navigation from shortcuts, long-form directory aliases,
+	// and configured concepts. Deferred until after the short-circuits above so
+	// toggle and custom-path shortcuts don't pay the resolution cost.
+	resolved := nav.ResolveConfiguredTarget(cfg, args)
 
 	result := nav.ParseResult{
 		Category:   nav.CategoryAll,
@@ -408,6 +409,10 @@ func handleRelativePathNavigation(ctx context.Context, campaignRoot, relativePat
 }
 
 func resolveRelativePathNavigation(ctx context.Context, campaignRoot, relativePath, query string) (string, error) {
+	if ctx.Err() != nil {
+		return "", ctx.Err()
+	}
+
 	if query == "" {
 		jumpResult, err := nav.JumpToPathFromRoot(ctx, campaignRoot, relativePath)
 		if err != nil {
@@ -428,11 +433,14 @@ func resolveRelativePathNavigation(ctx context.Context, campaignRoot, relativePa
 		if err != nil {
 			return "", err
 		}
+		if ctx.Err() != nil {
+			return "", ctx.Err()
+		}
 		nestedPath := filepath.Join(prefixPath, parts[1])
 		if info, err := os.Stat(nestedPath); err == nil && info.IsDir() {
 			return nestedPath, nil
 		}
-		return "", fmt.Errorf("path does not exist: %s/%s", strings.TrimRight(relativePath, "/"), query)
+		return "", camperrors.Wrapf(errNavigationPathNotFound, "%s/%s", strings.TrimRight(relativePath, "/"), query)
 	}
 
 	return fuzzyResolveDirectory(ctx, basePath, query, relativePath)
@@ -446,7 +454,7 @@ func fuzzyResolveDirectory(ctx context.Context, basePath, query, relativePath st
 	entries, err := os.ReadDir(basePath)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return "", fmt.Errorf("path does not exist: %s", relativePath)
+			return "", camperrors.Wrap(errNavigationPathNotFound, relativePath)
 		}
 		return "", camperrors.Wrap(err, "failed to read navigation path")
 	}
@@ -461,11 +469,18 @@ func fuzzyResolveDirectory(ctx context.Context, basePath, query, relativePath st
 
 	matches := fuzzy.FilterMulti(names, query)
 	if len(matches) == 0 {
-		return "", fmt.Errorf("no directories matching %q in %s", query, strings.TrimRight(relativePath, "/"))
+		return "", camperrors.Wrapf(errNavigationNoMatch, "%q in %s", query, strings.TrimRight(relativePath, "/"))
 	}
 
 	return filepath.Join(basePath, matches[0].Target), nil
 }
+
+// errNavigationPathNotFound indicates the requested navigation target directory
+// could not be resolved to an existing path under the campaign root.
+var errNavigationPathNotFound = camperrors.New("navigation path does not exist")
+
+// errNavigationNoMatch indicates fuzzy resolution found no matching directory.
+var errNavigationNoMatch = camperrors.New("no directories match navigation query")
 
 // evalSymlinks resolves symlinks in a path, returning the original path if resolution fails.
 func evalSymlinks(path string) (string, error) {
--- a/internal/complete/complete.go
+++ b/internal/complete/complete.go
@@ -3,6 +3,8 @@ package complete
 
 import (
 	"context"
+	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
@@ -29,23 +31,32 @@ func Generate(ctx context.Context, args []string) ([]string, error) {
 	ctx, cancel := context.WithTimeout(ctx, Timeout)
 	defer cancel()
 
-	// Load shortcuts from campaign config
-	shortcuts := loadShortcutMappings(ctx)
+	cfg, campaignRoot, err := config.LoadCampaignConfigFromCwd(ctx)
+	if err != nil {
+		return nil, nil
+	}
+
+	topLevelNames := nav.TopLevelNavigationNames(cfg)
 
 	// No args - complete category shortcuts
 	if len(args) == 0 {
-		return shortcutKeys(shortcuts), nil
+		return topLevelNames, nil
 	}
 
-	// Check if first arg is a category shortcut
-	result := nav.ParseShortcut(args, shortcuts)
-	if result.IsShortcut {
+	resolved := nav.ResolveConfiguredTarget(cfg, args)
+	if resolved.Matched {
+		if resolved.RelativePath != "" {
+			return completeInRelativePath(ctx, campaignRoot, resolved.RelativePath, resolved.Query)
+		}
+		if resolved.Drill {
+			return completeDrillInCategory(ctx, campaignRoot, resolved.Category, resolved.Query)
+		}
 		// Complete within category
-		return completeInCategory(ctx, result.Category, result.Query)
+		return completeInCategory(ctx, resolved.Category, resolved.Query)
 	}
 
 	// Not a shortcut - complete from all targets and shortcuts
-	return completeAll(ctx, args[0], shortcuts)
+	return completeAll(ctx, args[0], topLevelNames)
 }
 
 // GenerateRich returns completion candidates with fuzzy matching and path descriptions.
@@ -54,30 +65,45 @@ func GenerateRich(ctx context.Context, args []string) ([]RichCategoryGroup, erro
 	ctx, cancel := context.WithTimeout(ctx, Timeout)
 	defer cancel()
 
-	// Determine query and category
-	var query string
-	var cat nav.Category
+	cfg, campaignRoot, err := config.LoadCampaignConfigFromCwd(ctx)
+	if err != nil {
+		return nil, nil
+	}
+
+	resolved := nav.ResolveConfiguredTarget(cfg, args)
+	query := ""
 	if len(args) > 0 {
 		query = args[0]
 	}
-
-	// Check if first arg is a category shortcut
-	shortcuts := loadShortcutMappings(ctx)
-	result := nav.ParseShortcut(args, shortcuts)
-	if result.IsShortcut {
-		cat = result.Category
-		query = result.Query
-	}
-
-	// Get campaign root
-	jumpResult, err := nav.DirectJump(ctx, nav.CategoryAll)
-	if err != nil {
-		return nil, err
+	cat := nav.CategoryAll
+	if resolved.Matched {
+		if resolved.RelativePath != "" {
+			candidates, err := completeInRelativePathRich(ctx, campaignRoot, resolved.RelativePath, resolved.Query)
+			if err != nil {
+				return nil, err
+			}
+			return []RichCategoryGroup{{
+				Category:   strings.TrimRight(resolved.RelativePath, "/"),
+				Candidates: candidates,
+			}}, nil
+		}
+		if resolved.Drill {
+			candidates, err := completeDrillInCategoryRich(ctx, campaignRoot, resolved.Category, resolved.Query)
+			if err != nil {
+				return nil, err
+			}
+			return []RichCategoryGroup{{
+				Category:   string(resolved.Category),
+				Candidates: candidates,
+			}}, nil
+		}
+		cat = resolved.Category
+		query = resolved.Query
 	}
 
 	// Worktrees use hierarchical completion: project dirs first, then project@branch
-	if result.IsShortcut && cat == nav.CategoryWorktrees {
-		candidates, err := completeWorktreeRich(ctx, jumpResult.Path, query)
+	if resolved.Matched && cat == nav.CategoryWorktrees {
+		candidates, err := completeWorktreeRich(ctx, campaignRoot, query)
 		if err != nil {
 			return nil, err
 		}
@@ -88,8 +114,8 @@ func GenerateRich(ctx context.Context, args []string) ([]RichCategoryGroup, erro
 	}
 
 	// Subdirectory completion for queries containing "/"
-	if result.IsShortcut && strings.Contains(query, "/") {
-		subdirCandidates, err := CompleteSubdirectoryRich(ctx, jumpResult.Path, cat, query)
+	if resolved.Matched && strings.Contains(query, "/") {
+		subdirCandidates, err := CompleteSubdirectoryRich(ctx, campaignRoot, cat, query)
 		if err == nil && len(subdirCandidates) > 0 {
 			return []RichCategoryGroup{{
 				Category:   string(cat),
@@ -99,7 +125,7 @@ func GenerateRich(ctx context.Context, args []string) ([]RichCategoryGroup, erro
 	}
 
 	// Get or build index
-	idx, err := index.GetOrBuild(ctx, jumpResult.Path, false)
+	idx, err := index.GetOrBuild(ctx, campaignRoot, false)
 	if err != nil {
 		return nil, err
 	}
@@ -110,9 +136,6 @@ func GenerateRich(ctx context.Context, args []string) ([]RichCategoryGroup, erro
 
 	// Get fuzzy completion candidates
 	q := index.NewQuery(idx)
-	if cat == "" {
-		cat = nav.CategoryAll
-	}
 	candidates := q.FuzzyComplete(query, cat)
 
 	// Group by category
@@ -136,16 +159,6 @@ func GenerateRich(ctx context.Context, args []string) ([]RichCategoryGroup, erro
 	return grouped, nil
 }
 
-// loadShortcutMappings loads shortcuts from campaign config.
-// Returns empty map if not in a campaign or on error.
-func loadShortcutMappings(ctx context.Context) map[string]nav.Category {
-	cfg, _, err := config.LoadCampaignConfigFromCwd(ctx)
-	if err != nil {
-		return nil
-	}
-	return nav.BuildCategoryMappings(cfg.Shortcuts(), cfg.PathsMap())
-}
-
 // shortcutKeys returns the keys from a shortcuts map.
 func shortcutKeys(shortcuts map[string]nav.Category) []string {
 	if len(shortcuts) == 0 {
@@ -162,8 +175,11 @@ func shortcutKeys(shortcuts map[string]nav.Category) []string {
 // Returns nil if not in a campaign.
 func CategoryShortcuts() []string {
 	ctx := context.Background()
-	shortcuts := loadShortcutMappings(ctx)
-	return shortcutKeys(shortcuts)
+	cfg, _, err := config.LoadCampaignConfigFromCwd(ctx)
+	if err != nil {
+		return nil
+	}
+	return nav.TopLevelNavigationNames(cfg)
 }
 
 // completeInCategory returns completion candidates within a specific category.
@@ -231,13 +247,13 @@ func completeInCategory(ctx context.Context, cat nav.Category, query string) ([]
 }
 
 // completeAll returns completion candidates from all categories plus shortcuts.
-func completeAll(ctx context.Context, query string, shortcuts map[string]nav.Category) ([]string, error) {
+func completeAll(ctx context.Context, query string, topLevelNames []string) ([]string, error) {
 	var candidates []string
 
-	// Add matching category shortcuts first
-	for shortcut := range shortcuts {
-		if strings.HasPrefix(shortcut, query) {
-			candidates = append(candidates, shortcut)
+	// Add matching top-level navigation names first.
+	for _, name := range topLevelNames {
+		if strings.HasPrefix(strings.ToLower(name), strings.ToLower(query)) {
+			candidates = append(candidates, name)
 		}
 	}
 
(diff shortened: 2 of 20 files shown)


internal/config/types.go
```diff
@@ -65,6 +65,15 @@ func (c *CampaignConfig) Paths() CampaignPaths {
 	return DefaultCampaignPaths()
 }
 
+// PathsMap returns the raw concept-name→path map from jumps.yaml.
+// Returns from Jumps if loaded, otherwise returns defaults.
+func (c *CampaignConfig) PathsMap() map[string]string {
```
obey-agent: `PathsMap()` is returning the raw YAML map here, not the effective campaign paths after `NormalizeIntentNavigation()` and `ApplyDefaults()` run. That means partial or legacy `jumps.yaml` files still miss concept-name aliases in practice: e.g. `cgo intents` won't resolve if the stored path is still `workflow/intents/`, and omitted keys never get aliases even though defaults are active elsewhere. Can we derive this map from the normalized `CampaignPaths` instead of the raw unmarshaled section?

obey-agent: `PathsMap()` is returning the raw YAML map here, not the effective campaign paths after `NormalizeIntentNavigation()` and `ApplyDefaults()` run. That means partial or legacy `jumps.yaml` files still miss concept-name aliases in practice: e.g. `cgo intents` won't resolve if the stored path is still `workflow/intents/`, and omitted keys never get aliases even though defaults are active elsewhere. Can we derive this map from the normalized `CampaignPaths` instead of the raw unmarshaled section?

obey-agent (changes requested):
Verdict: Request Changes

Overview:
The direction is right and it fixes the happy path, but the alias map still does not reflect the effective campaign paths after normalization and defaulting. That leaves partial and legacy `jumps.yaml` configs with missing concept-name resolution even though the rest of the settings pipeline normalizes them.

Key Findings:
- `internal/config/types.go`: `PathsMap()` returns raw YAML state instead of the normalized/defaulted config, so concept-name aliases can still be absent for legacy or partial settings.

What's Done Well:
- The call sites were updated consistently.
- The UX improvement is well scoped and the added tests are pointed at the intended behavior.

Staff Standard:
Not yet. I would want the effective config and the alias map to agree before merging.

obey-agent (approved):
Verdict: Approve

Follow-up:
The effective-path alias mismatch is fixed. `PathsMap` now stays aligned with the normalized/defaulted jumps config, so concept-name navigation no longer diverges for legacy or partial settings.

internal/config/settings.go
```diff
@@ -95,6 +108,7 @@ func SaveJumpsConfig(ctx context.Context, campaignRoot string, cfg *JumpsConfig)
 		}
 	}
 	normalized.NormalizeIntentNavigation()
+	normalized.refreshPathsMap()
```
obey-agent: **Aliased map mutation — defensive copy missing**

`normalized := *cfg` is a shallow copy, so `normalized.PathsMap` still points at `cfg.PathsMap` (maps are reference types). When `normalized.refreshPathsMap()` writes entries at line 111, it mutates the caller's `cfg.PathsMap` as a side effect of `SaveJumpsConfig`.

Contrast with the block immediately above (lines 104–108) which explicitly deep-copies `Shortcuts` before mutating, specifically to avoid this aliasing. `PathsMap` is handled inconsistently.

Two separate problems fall out of this:

1. `SaveJumpsConfig` now has an undocumented side effect on the caller's `PathsMap`. Behavior is also asymmetric: if `cfg.PathsMap` was nil, `normalized.refreshPathsMap()` creates a new map and assigns it only to `normalized`, leaving the caller's nil map untouched; if `cfg.PathsMap` was non-nil, the caller's map is mutated. That asymmetry is the kind of thing that causes hard-to-diagnose bugs six months from now.
2. `normalized` is discarded after `yaml.Marshal`, and `PathsMap` is tagged `yaml:"-"`, so for the *serialization* path the refresh is inert — it's strictly a side-channel on the caller.

Fix options, in order of preference:
- **Drop `normalized.refreshPathsMap()` from Save entirely.** Save's job is to persist the config; refreshing in-memory state is the caller's problem. This is the most honest fix and matches what the marshal needs.
- If you want Save to leave the caller's in-memory map consistent, deep-copy `PathsMap` the same way `Shortcuts` is copied above, and refresh the copy.

internal/config/types.go
```diff
@@ -147,6 +156,29 @@ type CampaignPaths struct {
 	Dungeon string `yaml:"dungeon,omitempty"`
 }
 
+// asMap returns a fallback concept-name→path map derived from the struct fields.
+// This is only used when no jumps.yaml is present and we need defaults.
```
obey-agent: **Doc comment contradicts actual usage**

> This is only used when no jumps.yaml is present and we need defaults.

That is not what the code does. `asMap()` is now called from `JumpsConfig.refreshPathsMap()` (settings.go:213), which runs on **every** `LoadJumpsConfig` and inside `ApplyDefaults()` — i.e. precisely when jumps.yaml **is** present. It's the primary production overlay source, not a fallback.

Please update the comment to describe the actual role: "Returns the canonical concept-name→path map derived from the current `CampaignPaths` struct values. Used by `JumpsConfig.refreshPathsMap` to keep `PathsMap` in sync with the typed `Paths` struct after loads, normalization, and defaults."

This matters because the next engineer reading this will assume they can rely on the fallback framing and skip considering it for the load-time path, which is exactly the misread I almost made reviewing this PR.

internal/config/settings.go
```diff
@@ -70,7 +74,16 @@ func LoadJumpsConfig(ctx context.Context, campaignRoot string) (*JumpsConfig, er
...
+	// Capture the raw paths map so callers can preserve custom aliases.
+	var raw struct {
+		Paths map[string]string `yaml:"paths"`
+	}
+	if err := yaml.Unmarshal(data, &raw); err == nil && raw.Paths != nil {
+		cfg.PathsMap = raw.Paths
+	}
+
 	cfg.NormalizeIntentNavigation()
+	cfg.refreshPathsMap()
```
obey-agent: **Incomplete coverage of partial jumps.yaml at load time**

`LoadJumpsConfig` calls `NormalizeIntentNavigation` + `refreshPathsMap`, but **not** `ApplyDefaults`. The consequence, traced carefully:

1. User's jumps.yaml only defines a subset of paths (very plausible for an older campaign or hand-edited config).
2. `yaml.Unmarshal(data, &cfg)` populates only the specified struct fields — rest stay `""`.
3. `raw.Paths` = the raw map keys from YAML (only the ones the user wrote).
4. `cfg.PathsMap = raw.Paths` → still only partial.
5. `refreshPathsMap()` overlays `j.Paths.asMap()`, but `asMap()` skips empty struct fields (see types.go:164–167). So paths the user didn't explicitly set remain absent from both `Paths` and `PathsMap`.

Result: for a partial jumps.yaml, `cgo docs` works if `docs:` is in YAML, but `cgo projects` silently falls back to fuzzy search if `projects:` is missing — which is the exact class of bug this PR is trying to fix.

`loadJumps` in `campaign.go` happens to call `jumps.ApplyDefaults()` immediately after `LoadJumpsConfig` returns, which masks this for the primary `camp go` path. But `LoadJumpsConfig` is also called directly from three sites in `cmd/camp/navigation/shortcuts.go` (518, 732, 855) without `ApplyDefaults`, and `SaveJumpsConfig` on the result will round-trip only whatever ended up in `cfg.Paths` — which silently prunes the intended defaults for those users.

Suggested fix: call `cfg.ApplyDefaults()` inside `LoadJumpsConfig` before `refreshPathsMap()`. That makes the contract uniform across call sites and puts the "make this struct usable" invariant in the loader where it belongs. It also removes the need for the duplicate call in `campaign.go:loadJumps`.

If you'd rather keep Load strict and make callers opt in, at minimum the new test below should be replaced with one that writes a partial jumps.yaml to a tempdir and round-trips it through `LoadJumpsConfig`, since that is the path real users hit.

obey-agent (commented):

**Verdict**: Request Changes

**Overview**: The intent is sound and the `BuildCategoryMappings(shortcuts, pathsMap)` signature change is the right shape. The implementation, however, is leaky at two important seams — map aliasing in `SaveJumpsConfig`, and incomplete coverage for partial jumps.yaml files at load time. The new unit test sidesteps the real production path, so the bug the PR is meant to prevent can still reach users through `LoadJumpsConfig` call sites that don't follow up with `ApplyDefaults`.

### Key Findings

1. **`SaveJumpsConfig` aliases the caller's `PathsMap`.** The shallow copy `normalized := *cfg` shares the map, and `normalized.refreshPathsMap()` mutates it as a side effect. The block immediately above explicitly deep-copies `Shortcuts` for exactly this reason — the handling is now inconsistent. The refresh call is also inert for serialization (`PathsMap` is `yaml:"-"` and `normalized` is discarded after `yaml.Marshal`), so the cleanest fix is to drop `normalized.refreshPathsMap()` from Save. See inline at `internal/config/settings.go:111`.

2. **`LoadJumpsConfig` doesn't apply defaults.** Partial jumps.yaml → partial `cfg.Paths` → `asMap()` skips empty fields → partial `PathsMap`. Users with an old/hand-edited config hit the same fuzzy-search fallthrough the PR is trying to eliminate. The main `LoadCampaignConfig` → `loadJumps` path compensates by calling `ApplyDefaults` after the loader returns, but the three direct call sites in `cmd/camp/navigation/shortcuts.go` (518, 732, 855) don't. Move `cfg.ApplyDefaults()` into `LoadJumpsConfig` before `refreshPathsMap()` so the invariant lives in one place. See inline at `internal/config/settings.go:86`.

3. **`asMap()` doc comment contradicts the code.** "Only used when no jumps.yaml is present" is wrong — `refreshPathsMap` calls it on every load. This matters because the comment sets the wrong mental model for the next maintainer. See inline at `internal/config/types.go:160`.

4. **Test coverage gap.** `TestJumpsConfigRefreshPathsMapUsesEffectivePaths` hand-constructs a `JumpsConfig` and explicitly calls `ApplyDefaults()`, which is precisely the step that's missing from the real load path. Please add a test that writes a **partial** jumps.yaml (e.g. only `paths: {docs: ...}`) to a tempdir, round-trips it through `LoadJumpsConfig`, and asserts that both `projects` and `docs` resolve — that's the class of regression the PR is meant to prevent.

### What's Done Well

- Layering of concept names beneath explicit shortcuts in `BuildCategoryMappings` is the right precedence and is documented in the function comment.
- Accepting `nil` for `pathsMap` keeps the existing call sites and test fixtures ergonomic.
- Preserving custom aliases via the raw-map second `yaml.Unmarshal` is a tasteful way to let users keep non-standard path keys without adding struct fields.
- Test and production call-site updates are consistent (`cfg.Shortcuts(), cfg.PathsMap()`), and the completion count assertions were updated correctly to match the new 11+11 mapping fan-out.

### Staff Standard

I would not sign off on this as my own work in its current state. The aliasing bug in `SaveJumpsConfig` is the kind of subtle defect that surfaces as a flaky test or a user-reported "why did my PathsMap grow unexpectedly?" months down the line, and the loader coverage gap undermines the PR's stated goal for a real class of users. Both are fixable in a small follow-up commit; happy to re-review once the three inline comments are addressed.

obey-agent (commented):

**Verdict**: Request Changes

**Overview**: The intent is sound and the `BuildCategoryMappings(shortcuts, pathsMap)` signature change is the right shape. The implementation, however, is leaky at two important seams — map aliasing in `SaveJumpsConfig`, and incomplete coverage for partial jumps.yaml files at load time. The new unit test sidesteps the real production path, so the bug the PR is meant to prevent can still reach users through `LoadJumpsConfig` call sites that don't follow up with `ApplyDefaults`.

### Key Findings

1. **`SaveJumpsConfig` aliases the caller's `PathsMap`.** The shallow copy `normalized := *cfg` shares the map, and `normalized.refreshPathsMap()` mutates it as a side effect. The block immediately above explicitly deep-copies `Shortcuts` for exactly this reason — the handling is now inconsistent. The refresh call is also inert for serialization (`PathsMap` is `yaml:"-"` and `normalized` is discarded after `yaml.Marshal`), so the cleanest fix is to drop `normalized.refreshPathsMap()` from Save. See inline at `internal/config/settings.go:111`.

2. **`LoadJumpsConfig` doesn't apply defaults.** Partial jumps.yaml → partial `cfg.Paths` → `asMap()` skips empty fields → partial `PathsMap`. Users with an old/hand-edited config hit the same fuzzy-search fallthrough the PR is trying to eliminate. The main `LoadCampaignConfig` → `loadJumps` path compensates by calling `ApplyDefaults` after the loader returns, but the three direct call sites in `cmd/camp/navigation/shortcuts.go` (518, 732, 855) don't. Move `cfg.ApplyDefaults()` into `LoadJumpsConfig` before `refreshPathsMap()` so the invariant lives in one place. See inline at `internal/config/settings.go:86`.

3. **`asMap()` doc comment contradicts the code.** "Only used when no jumps.yaml is present" is wrong — `refreshPathsMap` calls it on every load. This matters because the comment sets the wrong mental model for the next maintainer. See inline at `internal/config/types.go:160`.

4. **Test coverage gap.** `TestJumpsConfigRefreshPathsMapUsesEffectivePaths` hand-constructs a `JumpsConfig` and explicitly calls `ApplyDefaults()`, which is precisely the step that's missing from the real load path. Please add a test that writes a **partial** jumps.yaml (e.g. only `paths: {docs: ...}`) to a tempdir, round-trips it through `LoadJumpsConfig`, and asserts that both `projects` and `docs` resolve — that's the class of regression the PR is meant to prevent.

### What's Done Well

- Layering of concept names beneath explicit shortcuts in `BuildCategoryMappings` is the right precedence and is documented in the function comment.
- Accepting `nil` for `pathsMap` keeps the existing call sites and test fixtures ergonomic.
- Preserving custom aliases via the raw-map second `yaml.Unmarshal` is a tasteful way to let users keep non-standard path keys without adding struct fields.
- Test and production call-site updates are consistent (`cfg.Shortcuts(), cfg.PathsMap()`), and the completion count assertions were updated correctly to match the new 11+11 mapping fan-out.

### Staff Standard

I would not sign off on this as my own work in its current state. The aliasing bug in `SaveJumpsConfig` is the kind of subtle defect that surfaces as a flaky test or a user-reported "why did my PathsMap grow unexpectedly?" months down the line, and the loader coverage gap undermines the PR's stated goal for a real class of users. Both are fixable in a small follow-up commit; happy to re-review once the three inline comments are addressed.

obey-agent (commented):

**Verdict**: Request Changes

**Overview**: The intent is sound and the `BuildCategoryMappings(shortcuts, pathsMap)` signature change is the right shape. The implementation, however, is leaky at two important seams — map aliasing in `SaveJumpsConfig`, and incomplete coverage for partial jumps.yaml files at load time. The new unit test sidesteps the real production path, so the bug the PR is meant to prevent can still reach users through `LoadJumpsConfig` call sites that don't follow up with `ApplyDefaults`.

### Key Findings

1. **`SaveJumpsConfig` aliases the caller's `PathsMap`.** The shallow copy `normalized := *cfg` shares the map, and `normalized.refreshPathsMap()` mutates it as a side effect. The block immediately above explicitly deep-copies `Shortcuts` for exactly this reason — the handling is now inconsistent. The refresh call is also inert for serialization (`PathsMap` is `yaml:"-"` and `normalized` is discarded after `yaml.Marshal`), so the cleanest fix is to drop `normalized.refreshPathsMap()` from Save. See inline at `internal/config/settings.go:111`.

2. **`LoadJumpsConfig` doesn't apply defaults.** Partial jumps.yaml → partial `cfg.Paths` → `asMap()` skips empty fields → partial `PathsMap`. Users with an old/hand-edited config hit the same fuzzy-search fallthrough the PR is trying to eliminate. The main `LoadCampaignConfig` → `loadJumps` path compensates by calling `ApplyDefaults` after the loader returns, but the three direct call sites in `cmd/camp/navigation/shortcuts.go` (518, 732, 855) don't. Move `cfg.ApplyDefaults()` into `LoadJumpsConfig` before `refreshPathsMap()` so the invariant lives in one place. See inline at `internal/config/settings.go:86`.

3. **`asMap()` doc comment contradicts the code.** "Only used when no jumps.yaml is present" is wrong — `refreshPathsMap` calls it on every load. This matters because the comment sets the wrong mental model for the next maintainer. See inline at `internal/config/types.go:160`.

4. **Test coverage gap.** `TestJumpsConfigRefreshPathsMapUsesEffectivePaths` hand-constructs a `JumpsConfig` and explicitly calls `ApplyDefaults()`, which is precisely the step that's missing from the real load path. Please add a test that writes a **partial** jumps.yaml (e.g. only `paths: {docs: ...}`) to a tempdir, round-trips it through `LoadJumpsConfig`, and asserts that both `projects` and `docs` resolve — that's the class of regression the PR is meant to prevent.

### What's Done Well

- Layering of concept names beneath explicit shortcuts in `BuildCategoryMappings` is the right precedence and is documented in the function comment.
- Accepting `nil` for `pathsMap` keeps the existing call sites and test fixtures ergonomic.
- Preserving custom aliases via the raw-map second `yaml.Unmarshal` is a tasteful way to let users keep non-standard path keys without adding struct fields.
- Test and production call-site updates are consistent (`cfg.Shortcuts(), cfg.PathsMap()`), and the completion count assertions were updated correctly to match the new 11+11 mapping fan-out.

### Staff Standard

I would not sign off on this as my own work in its current state. The aliasing bug in `SaveJumpsConfig` is the kind of subtle defect that surfaces as a flaky test or a user-reported "why did my PathsMap grow unexpectedly?" months down the line, and the loader coverage gap undermines the PR's stated goal for a real class of users. Both are fixable in a small follow-up commit; happy to re-review once the three inline comments are addressed.

obey-agent (approved):
Verdict: Approve

Follow-up:
The current head fixes the remaining config-loader issues. `LoadJumpsConfig` now applies defaults before refreshing the effective alias map, direct callers no longer depend on ad hoc `ApplyDefaults()` calls, `SaveJumpsConfig` no longer mutates the caller's runtime-only `PathsMap`, and config coverage now exercises partial `jumps.yaml` plus caller-map non-mutation.

obey-agent (commented):
## Staff-Level Re-Review — Approve

**Verdict**: Approve

**Overview**: All three inline issues from my prior "Request Changes" review are addressed at `425cbaea`, and the fixes are exactly the right shape. The two new tests in `settings_test.go` (+66 lines) cover precisely the gaps I flagged — partial jumps.yaml round-tripped through the real `LoadJumpsConfig` path, and a defensive-copy assertion on `SaveJumpsConfig`'s non-mutation of caller state.

### How each prior finding landed

1. **`SaveJumpsConfig` aliasing the caller's `PathsMap`** — fixed at its cleanest: the `normalized.refreshPathsMap()` call is simply removed. Save now does exactly what serialization needs and nothing more. No defensive deep-copy workaround, no subtle aliasing behavior. `TestSaveJumpsConfig_DoesNotMutateCallerPathsMap` nails this directly by constructing a `cfg` with pre-populated `PathsMap`, calling Save, then asserting the caller's map was *not* mutated (specifically that it did NOT gain the `projects` default from Save's side). Good, targeted coverage.

2. **`LoadJumpsConfig` not applying defaults** — fixed by swapping `cfg.refreshPathsMap()` for `cfg.ApplyDefaults()`, which already calls `refreshPathsMap` internally as its last step. This collapses two invariants into one call and makes the loader self-sufficient. Consequence: the redundant `jumps.ApplyDefaults()` calls in `campaign.go:loadJumps` and `internal/project/worktrees_path.go` are removed (−1 line each), so the "apply defaults" invariant now lives in one place instead of three. That's architecturally cleaner than my original suggestion. `TestLoadJumpsConfig_AppliesDefaultsAndPreservesRawAliases` writes a real partial YAML to a tempdir with only `intents` and a custom `documentation` alias, round-trips it through `LoadJumpsConfig`, and asserts: (a) `intents` gets normalized to `.campaign/intents/` by `NormalizeIntentNavigation`, (b) `projects` gets populated from `ApplyDefaults`, and (c) the custom `documentation` alias survives untouched. That's exactly the "partial jumps.yaml through the real loader" path I asked for.

3. **`asMap()` doc comment contradicting actual usage** — fixed. New text: "Returns the canonical concept-name→path map derived from the current CampaignPaths struct values. JumpsConfig.refreshPathsMap uses it to keep the effective alias map aligned with normalized and defaulted runtime paths." That's accurate, names the caller, and sets the correct mental model for the next maintainer.

### What's Still Done Well

- The fix uses *less* code than the previous revision: one call-site removed from Save, two redundant `ApplyDefaults` calls removed from downstream callers, and the load path collapses to `NormalizeIntentNavigation → ApplyDefaults` with ApplyDefaults doing the rest. Subtractive.
- Test coverage is targeted at the exact regression classes I flagged, not speculative adjacent cases. Each test is small (~30 lines) and asserts one invariant.
- The `asMap` doc update names `JumpsConfig.refreshPathsMap` directly as the caller, so a reader following the code can jump straight to the consumer.

### Staff Standard

Approve. This is how review feedback should be addressed: the fix is smaller than the original code, the tests cover the exact gaps I flagged, and the architectural consequence (one-place ApplyDefaults invariant) is better than my original suggestion of "add ApplyDefaults inside Load but leave the downstream calls intact." Ship it.

internal/config/types.go
```diff
@@ -147,6 +156,30 @@ type CampaignPaths struct {
 	Dungeon string `yaml:"dungeon,omitempty"`
 }
 
+// asMap returns the canonical concept-name→path map derived from the current
+// CampaignPaths struct values. JumpsConfig.refreshPathsMap uses it to keep the
+// effective alias map aligned with normalized and defaulted runtime paths.
+func (p CampaignPaths) asMap() map[string]string {
```
obey-agent: This is not the intended design, the intended design is that these default concepts we include in camp are documented in the yaml file, users can edit that file and the camp go jumping should work with their edits, or whatever is in that file.  Jumping to concepts defined in that yaml should work just like with the shortcuts.

obey-agent (changes requested):

obey-agent (approved):
Addressed the navigation review gaps. camp go now resolves configured shortcut keys, long-form directory aliases, and configured concepts separately, and first-token drill syntax now works via design/... and de@... filesystem traversal with dynamic shell completion.

obey-agent (commented):
## Staff-Level Re-Review — Substantial Architectural Change (still Approve, with two follow-ups)

**Verdict**: Approve

Commit `3033311` "fix: resolve cgo long-form aliases from config" is a much larger change than the title suggests — 985 lines across 16 files, introducing `internal/nav/configured_navigation.go` with a new layered resolution architecture. I want to flag that scope growth explicitly since my prior approval was for a narrower fix, and note a couple of observations on the new surface.

### What changed

The `PathsMap`/`asMap` approach I previously approved has been **replaced** with a cleaner design:

1. **`ResolveConfiguredTarget`** — unified entry point with explicit resolution order: shortcuts → concepts → long-form path aliases.
2. **`BuildPathAliasMappings`** — derives aliases from navigation shortcut **target paths** (e.g. `workflow/design/` → `design`) instead of from a separate `PathsMap` field. This is a better single-source-of-truth design.
3. **`NormalizeNavigationName`** — case-insensitive, trailing-slash-tolerant matching across all three layers.
4. **Drill syntax** — `p/query`, `p@query` patterns handled via `splitSlashDrillArgs` / `splitShortcutDrillArgs`.
5. **`BuildCategoryMappings`** signature reverted to single-arg (no `pathsMap`), consistent with the new layering.

Architecturally this is cleaner than what I approved last round. The ambiguity handling in `BuildPathAliasMappings` (drop the alias rather than silently pick one) is the right default for user-safety.

### Observations (non-blocking)

1. **`JumpsConfig.PathsMap` is now dead state.** `CampaignConfig.PathsMap()` was removed in this commit (`types.go` −33 lines), and the `asMap()` method was moved to package-private `effectivePathsMap` in `settings.go`. But `JumpsConfig.PathsMap` is still a struct field and `refreshPathsMap` still populates it. Nothing outside the package reads it anymore. Either delete the field + `refreshPathsMap` (tidy), or leave a comment explaining why it's kept around for backward-compat (if that's the intent).

2. **`ResolveConfiguredTarget` builds shortcut mappings twice.** In the drill-check branch and again in the main branch:
   ```go
   shortcutMappings := BuildCategoryMappings(cfg.Shortcuts())  // build 1
   // ...
   shortcutMappings := BuildCategoryMappings(cfg.Shortcuts())  // build 2 (identical)
   ```
   Same applies to `BuildPathAliasMappings` being re-derived inside `resolveDrillTarget`. For a CLI the micro-cost is trivial, but hoisting both to the top of `ResolveConfiguredTarget` would be cleaner and future-proof if the shortcut map ever grows.

3. **PR scope** — the PR title is "fix: allow concept names as camp go navigation targets" but this commit adds drill syntax, fuzzy sub-directory matching, and long-form-alias resolution. That's a feature expansion. Consider splitting these into a follow-up PR if the base fix has already proven stable, or updating the PR title/description so reviewers know to expect the broader scope.

### What's Still Done Well

- The new code is well-tested: 136 lines of new tests in `configured_navigation_test.go` plus 30 lines of integration coverage.
- Explicit resolution order documented in the `ResolveConfiguredTarget` docstring.
- Ambiguity handling is conservative and safe.
- `NormalizeNavigationName` unifies case/slash handling so there's one canonical normalization across the codebase.
- Shell integration updated (`bash`, `fish`, `zsh`) — the three shells stay in sync.

### Staff Standard

Architecture is cleaner than what I approved last round — this is a net improvement. Ship it. The two observations above are follow-up cleanups, not blockers. The PR-scope comment is process feedback for next time, not a request to amend this PR.

obey-agent:
Design note for the current head (`3033311`):

This PR now treats `camp go` / `cgo` navigation as three distinct sources of names instead of collapsing everything into `CampaignPaths.asMap()`.

1. Shortcut keys from `.campaign/settings/jumps.yaml`
- Examples: `de`, `ai`, `p`, `i`
- These still work exactly as before.

2. Long-form directory aliases derived from configured navigation shortcut targets
- Examples: if a configured shortcut points at `workflow/design/`, then `design` should work.
- If a configured shortcut points at `ai_docs/`, then `ai_docs` should work.
- This is a directory-alias layer, not the concept layer.

3. Configured concepts from `.campaign/campaign.yaml`
- User-defined concept names should also resolve.
- This covers custom concepts in `campaign.yaml` instead of only built-in path names.

The intent is to support the UX you described:
- `cgo de` and `cgo design` should both work
- `cgo ai` and `cgo ai_docs` should both work
- configured concepts in `campaign.yaml` should work
- shortcuts are preserved; long-form names are added, not substituted

Drill behavior added on this head:
- `cgo design/` enters drill mode rooted at `workflow/design/`
- `cgo design/<tab>` cycles through actual filesystem entries under that directory
- each additional `/` drills one level deeper
- `cgo de@` is the shortcut form of the same behavior
- `cgo de@<tab>` cycles entries under the shortcut target
- shortcut drill and slash drill are filesystem-based traversal from the resolved target path, not index-only fuzzy lookup

What changed in code:
- Added a dedicated configured-navigation resolver in `internal/nav/configured_navigation.go`
- `cmd/camp/navigation/go.go` now resolves navigation through:
  - shortcut keys
  - long-form directory aliases
  - configured concepts
- `internal/complete/complete.go` now completes those same names and supports first-token drill completion for `design/...` and `de@...`
- shell init scripts were updated so first-token drill completion actually works in zsh/bash/fish
- tests were added/updated for long-form aliases, configured concepts, and drill behavior

Important `PathsMap` note:
- `JumpsConfig.PathsMap` is no longer part of runtime navigation resolution on this branch.
- Navigation no longer reads `CampaignConfig.PathsMap()` and no longer uses `CampaignPaths.asMap()` as a concept/navigation source of truth.
- `PathsMap` is currently only maintained inside `internal/config/settings.go` as an internal effective map for the jumps loader/tests.
- So yes: for navigation, it is effectively dead now.
- I did not fully remove the `JumpsConfig.PathsMap` field in this PR because that is a separate config cleanup decision, but the navigation design on this head no longer depends on it.

The design line here is:
- shortcuts stay in `jumps.yaml`
- concepts stay in `campaign.yaml`
- long-form directory aliases come from configured navigation targets
- those are related, but they are not the same layer anymore

internal/config/settings.go
```diff
@@ -28,6 +28,10 @@ type JumpsConfig struct {
 	Paths CampaignPaths `yaml:"paths,omitempty"`
 	// Shortcuts defines custom navigation and command shortcuts.
 	Shortcuts map[string]ShortcutConfig `yaml:"shortcuts,omitempty"`
+	// PathsMap is the effective name→path map for jumps paths.
+	// It starts with the raw YAML keys, then overlays the normalized/defaulted
+	// standard campaign paths so alias lookups stay aligned with runtime config.
+	PathsMap map[string]string `yaml:"-"`
```
obey-agent: **Dead code.** `PathsMap` is added to `JumpsConfig` with ~50 lines of implementation (raw YAML capture, `refreshPathsMap`, `effectivePathsMap`) plus ~100 lines of tests, but nothing in production code ever reads it. A repo-wide `rg PathsMap` only hits `settings.go` and `settings_test.go`. Long-form directory aliases are derived from `cfg.Shortcuts()` via `BuildPathAliasMappings`, not from `PathsMap`, so a user-supplied key like `paths.documentation: docs/` in jumps.yaml will not create a navigable alias — despite what `TestLoadJumpsConfig_AppliesDefaultsAndPreservesRawAliases` appears to demonstrate.

If this is intended for future use, land it with the consumer or drop it for now. Shipping a dead field inflates surface area that future maintainers will have to reason about and keep in sync with the rest of the config layer.

internal/complete/complete.go
```diff
@@ -136,16 +159,6 @@ func GenerateRich(ctx context.Context, args []string) ([]RichCategoryGroup, erro
...
-func loadShortcutMappings(ctx context.Context) map[string]nav.Category {
-	cfg, _, err := config.LoadCampaignConfigFromCwd(ctx)
-	if err != nil {
-		return nil
-	}
-	return nav.BuildCategoryMappings(cfg.Shortcuts())
-}
-
 // shortcutKeys returns the keys from a shortcuts map.
 func shortcutKeys(shortcuts map[string]nav.Category) []string {
```
obey-agent: **Dead function.** `shortcutKeys` is no longer called anywhere after the refactor (previously used by `Generate` and `CategoryShortcuts`, both of which now call `nav.TopLevelNavigationNames` directly). `rg shortcutKeys` shows only the declaration here. Remove it rather than leaving stale helpers — they confuse future readers trying to understand the data flow.

cmd/camp/navigation/go.go
```diff
@@ -360,6 +377,96 @@ func handleCustomNavShortcut(ctx context.Context, sc config.ShortcutConfig, camp
...
+	for _, entry := range entries {
+		if !entry.IsDir() || strings.HasPrefix(entry.Name(), ".") {
+			continue
+		}
+		names = append(names, entry.Name())
+	}
+
+	matches := fuzzy.FilterMulti(names, query)
+	if len(matches) == 0 {
+		return "", fmt.Errorf("no directories matching %q in %s", query, strings.TrimRight(relativePath, "/"))
```
obey-agent: **Use the project error framework.** These three `fmt.Errorf` call sites (L435, L449, L464) are new production code, and the project's guideline is to wrap via `camperrors` (you already import it and use `camperrors.Wrap` on L451). The trailing `/`-trim formatting is fine, but a plain `fmt.Errorf` string is not wrappable by callers and loses observability — no type, no sentinel, no `errors.Is` target.

Minimum: use `camperrors.Wrapf(..., "...")` or introduce a sentinel like `ErrNavigationPathNotFound` in `internal/errors` and wrap it here. This matters because shell completion / scripted callers rely on being able to distinguish 'not found' from 'I/O failed'.

internal/nav/configured_navigation.go
```diff
@@ -0,0 +1,288 @@
...
+	return names
+}
+
+// ResolveConfiguredTarget resolves the first navigation argument against the
+// current campaign configuration.
+//
+// Resolution order:
+//  1. Configured shortcut keys from jumps.yaml
+//  2. Configured concept names from campaign.yaml
+//  3. Long-form directory aliases derived from configured navigation shortcuts
```
obey-agent: **Docstring disagrees with implementation.** The doc claims the order is (1) shortcut keys → (2) concepts → (3) path aliases, but the actual flow is:

1. Shortcut drill (`de@foo`) via `splitShortcutDrillArgs`
2. Shortcut keys (plain `de`)
3. Slash drill (`design/foo`) via `splitSlashDrillArgs` → `resolveDrillTarget`
4. Concept names (`resolveDrillTarget` first branch)
5. Path aliases derived from shortcuts (`resolveDrillTarget` second branch)

Concept names shadow path aliases silently when both exist at the same normalized key. That precedence decision is reasonable but invisible — please encode it explicitly in the docstring so it doesn't get swapped by accident later. Also mention the `@` and `/` drill syntaxes; they're the user-facing surface and should be visible here.

internal/nav/configured_navigation.go
```diff
@@ -0,0 +1,288 @@
...
+			return ConfiguredTarget{
+				Category: parsed.Category,
+				Query:    parsed.Query,
+				Drill:    true,
+				Matched:  true,
+			}
+		}
+	}
+
+	shortcutMappings := BuildCategoryMappings(cfg.Shortcuts())
```
obey-agent: **Redundant work — compute once, branch twice.** `BuildCategoryMappings(cfg.Shortcuts())` is rebuilt twice on any input that contains `@` (once in the drill branch at L102, then again unconditionally at L114). `BuildPathAliasMappings` is similarly recomputed inside `resolveDrillTarget` at L159, and yet a third time inside `TopLevelNavigationNames`. Each rebuild iterates every shortcut, does `path.Base`/`strings.TrimRight`, and allocates a fresh map.

This runs on every tab completion, so it's in a shell-latency-sensitive hot path (shell tab completion is gated on the 200ms `complete.Timeout`). Compute both maps once at the top of `ResolveConfiguredTarget`, or memoize them on `*CampaignConfig`. The ergonomics are also better — you won't have ambiguity collisions diverging between sites.

internal/complete/complete.go
```diff
@@ -231,13 +247,13 @@ func completeInCategory(ctx context.Context, cat nav.Category, query string) ([]
...
+func completeAll(ctx context.Context, query string, topLevelNames []string) ([]string, error) {
 	var candidates []string
 
-	// Add matching category shortcuts first
-	for shortcut := range shortcuts {
-		if strings.HasPrefix(shortcut, query) {
-			candidates = append(candidates, shortcut)
+	// Add matching top-level navigation names first.
+	for _, name := range topLevelNames {
+		if strings.HasPrefix(strings.ToLower(name), strings.ToLower(query)) {
```
lancekrogers: **Normalization inconsistency.** Everywhere else in this PR, navigation tokens go through `NormalizeNavigationName` (lowercase + trim trailing `/`). Here in `completeAll`, you roll your own `strings.HasPrefix(strings.ToLower(name), strings.ToLower(query))` — which does NOT trim a trailing slash from the query. 

Consequence: if a user has typed `design/` and `ResolveConfiguredTarget` hasn't matched (because say concepts list is empty and there's no `design` path alias), this falls through to `completeAll` with query="design/". The top-level name stored in `topLevelNames` is "design" (no slash). Prefix match fails. No suggestions.

Use `NormalizeNavigationName` on both sides here, or better yet, do the normalization once up front and compare normalized-to-normalized consistently.

cmd/camp/navigation/go.go
```diff
@@ -59,9 +60,9 @@ func runGo(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// Build category mappings from config shortcuts
-	// This allows config shortcuts to work with fuzzy search
-	configMappings := nav.BuildCategoryMappings(cfg.Shortcuts())
+	// Resolve configured navigation from shortcuts, long-form directory aliases,
+	// and configured concepts.
+	resolved := nav.ResolveConfiguredTarget(cfg, args)
```
lancekrogers: **Wasted call on short-circuit paths.** `ResolveConfiguredTarget` runs unconditionally here, but the next ~15 lines short-circuit out for the toggle keyword and custom-path navigation shortcuts. For those cases (every `cgo t`, every `cgo foo` where `foo` is a non-standard-path shortcut) we pay the full resolution cost — sort shortcut keys, build both mappings, parse — and throw the result away.

Move the `resolved := nav.ResolveConfiguredTarget(cfg, args)` call down to just above line 84 where it's actually consumed, so the short-circuits stay cheap.

internal/complete/complete.go
```diff
@@ -271,3 +287,196 @@ func completeAll(ctx context.Context, query string, shortcuts map[string]nav.Cat
...
+	if strings.Contains(query, "/") {
+		return CompleteSubdirectoryRich(ctx, campaignRoot, cat, query)
+	}
+	return listPathCandidatesRich(ctx, basePath, relativePath, query)
+}
+
+func listPathCandidates(ctx context.Context, absPath, prefix string) ([]string, error) {
+	entries, err := os.ReadDir(absPath)
+	if err != nil {
+		return nil, nil
```
lancekrogers: **Errors are swallowed silently.** `listPathCandidates`, `listPathCandidatesRich`, `completeSubdirectoryInPath`, and `completeSubdirectoryInPathRich` all do `if err != nil { return nil, nil }` on `os.ReadDir`. That pattern is reasonable for 'missing dir in completion' but it also silently drops permission errors, I/O errors, and bad symlinks. When a user has a corrupted workflow dir or a permission issue, they'll see 'no completions' with zero diagnostic signal.

Differentiate: swallow `os.IsNotExist`, but bubble up (or at least log to a debug channel / campaign audit log) the real I/O failures. If we care enough to call `ctx.Err()` inside the loop, we should care enough to distinguish 'nothing to suggest' from 'I/O broke.'

internal/config/settings_test.go
```diff
@@ -181,3 +181,99 @@ func TestSaveJumpsConfig_NormalizesLegacyIntentNavigation(t *testing.T) {
...
+	}
+	if got := cfg.PathsMap["projects"]; got != "projects/" {
+		t.Fatalf("PathsMap[projects] = %q, want %q", got, "projects/")
+	}
+	if got := cfg.PathsMap["documentation"]; got != "docs/" {
+		t.Fatalf("PathsMap[documentation] = %q, want %q", got, "docs/")
+	}
+}
+
+func TestSaveJumpsConfig_DoesNotMutateCallerPathsMap(t *testing.T) {
```
lancekrogers: **Test name is misleading.** This asserts `SaveJumpsConfig` does not mutate the caller's `PathsMap`, but it only 'passes' because `SaveJumpsConfig` never calls `ApplyDefaults`. `ApplyDefaults` → `refreshPathsMap` writes directly into `j.PathsMap` (settings.go:211-219), so any caller that runs `LoadJumpsConfig` then `ApplyDefaults` on their own config *would* see their PathsMap mutated. 

Either (a) rename to `TestSaveJumpsConfig_DoesNotApplyDefaults` so the actual invariant is clear, or (b) make `refreshPathsMap` operate on a copy and assign atomically so the non-mutation property holds regardless of call order. Option (b) is friendlier to future callers.

cmd/camp/navigation/go.go
```diff
@@ -360,6 +377,96 @@ func handleCustomNavShortcut(ctx context.Context, sc config.ShortcutConfig, camp
...
+		jumpResult, err := nav.JumpToPathFromRoot(ctx, campaignRoot, relativePath)
+		if err != nil {
+			return "", err
+		}
+		return jumpResult.Path, nil
+	}
+
+	basePath := filepath.Join(campaignRoot, relativePath)
+	exactPath := filepath.Join(basePath, query)
+	if info, err := os.Stat(exactPath); err == nil && info.IsDir() {
```
lancekrogers, line 426: **Context cancellation gap.** `fuzzyResolveDirectory` starts with a `ctx.Err()` check, but the caller `resolveRelativePathNavigation` does not. So the `os.Stat(exactPath)` on line 421, the `fuzzyResolveDirectory` call on line 427, the `os.Stat(nestedPath)` on line 432 — all proceed even if ctx is already cancelled/timed-out. For the shell-integration path, `complete.Timeout` is 200ms; a stale call here after timeout hits the disk anyway.

Please add a `ctx.Err()` gate at the top of `resolveRelativePathNavigation` and before each `os.Stat` (at least the second one). Stats are cheap but not free, and the project CLAUDE guidance is 'always check ctx.Err() before starting long operations.'

obey-agent (commented):
## Verdict: Request Changes

## Overview
This PR introduces a real improvement — letting users type concept names, long-form directory aliases, and drill syntax (`design/foo`, `de@foo`) at the first navigation arg. The core design (dedicated `ResolveConfiguredTarget` in `internal/nav`, shared between `cgo` and the completion engine) is sound and the test coverage for the happy paths is solid. But there are several issues that need cleanup before this ships.

## Key Findings
- **`JumpsConfig.PathsMap` is dead code** — 50+ lines of implementation plus ~100 lines of test, no production consumer. `BuildPathAliasMappings` still derives aliases from `cfg.Shortcuts()`, not from the new `PathsMap`. See inline on settings.go:34.
- **`fmt.Errorf` in new production paths** — violates the project's stated error convention. See inline on go.go:464.
- **Dead function** `shortcutKeys` left in complete.go after the refactor. See inline on complete.go:163.
- **Docstring in `ResolveConfiguredTarget` disagrees with the implementation** — it describes 3 resolution steps, the code has 5 with non-trivial precedence choices (concept shadows alias, drill-form precedes plain). See inline on configured_navigation.go:94.
- **Redundant hot-path work** — `BuildCategoryMappings` and `BuildPathAliasMappings` recomputed 2–3× per invocation in a shell-completion path gated on a 200ms timeout. See inline on configured_navigation.go:114.
- **Normalization inconsistency** in `completeAll` (custom `strings.ToLower` vs `NormalizeNavigationName` everywhere else) — breaks completion when a user types a trailing `/`. See inline on complete.go:255.
- **Silent error swallowing** in the new `listPathCandidates*` / `completeSubdirectoryInPath*` helpers — permission and I/O errors become 'no completions' with no diagnostic signal. See inline on complete.go:357.
- **Context-cancellation gap** in `resolveRelativePathNavigation` — does `os.Stat` without checking `ctx.Err()`. See inline on go.go:421.
- **Misleading test name** — `TestSaveJumpsConfig_DoesNotMutateCallerPathsMap` passes only because `SaveJumpsConfig` doesn't call `ApplyDefaults`; any caller running `ApplyDefaults` directly would see mutation. See inline on settings_test.go:254.
- **`resolved := nav.ResolveConfiguredTarget(cfg, args)` runs before toggle / custom-shortcut short-circuits** — wasted work on common paths. See inline on go.go:65.

## What's Done Well
- The introduction of a single `ResolveConfiguredTarget` that both `cgo` and shell completion share is exactly the right abstraction — no parallel implementation of the resolution logic, which would have been easy to regress.
- `NormalizeNavigationName` is a clean primitive and it's applied consistently in the new nav package.
- `BuildPathAliasMappings` correctly drops ambiguous aliases rather than silently picking one — good defensive design.
- Integration tests in `tests/integration/navigation_test.go` cover the user-facing `cgo design`, `cgo ai_docs`, `cgo design/festival_app`, and `cgo de@festival_site` flows — those exercises the full stack end-to-end.
- Build passes, all existing tests pass.

## Staff Standard
No. Not in this state. The core design is good and I'd be proud of the overall architecture, but shipping a dead `PathsMap` field with ~150 lines of plumbing, `fmt.Errorf` in new code after the project explicitly forbids it, and stale helper functions is not staff-level delivery. Tighten the six 'must-fix' items (PathsMap, fmt.Errorf, dead shortcutKeys, redundant computation, completeAll normalization, error swallowing) and this will be ready.

PR #212, opened 2026-04-13
fix: restore --campaign value space-separated syntax for intent add

lancekrogers:
## Summary

The integration test `TestIntentAdd_TargetCampaignWritesToSelectedRoot` was failing on `lr/develop` with:

\`\`\`
camp exited with code 1: Error: accepts at most 1 arg(s), received 2
\`\`\`

## Root cause

`#206` added a bare-flag interactive picker for `--campaign` by setting `NoOptDefVal` on the flag. Cobra's pflag library treats a flag with `NoOptDefVal` as "optional-value": the value must be attached with `=`, and space-separated `--campaign value` **never consumes the next token**. So:

\`\`\`
camp intent add --campaign intent-target-dest "Targeted Intent" --no-commit
\`\`\`

parsed as: `--campaign` (uses sentinel), plus positionals `["intent-target-dest", "Targeted Intent"]` → two positionals violates `MaximumNArgs(1)`.

The test was written in `2c04f84` (Mar 16) before `NoOptDefVal` was added in `5f90da0` / `ff83a1a`. The regression wasn't caught when `#206` merged.

## Fix

Preprocess `os.Args` in `Execute()` before cobra parses, rewriting `--campaign value` → `--campaign=value` (and `-c value` → `-c=value`) when the next token is not a flag. Bare `--campaign` (end-of-args or followed by another flag) is left untouched so the `NoOptDefVal` sentinel still fires and the bare-flag picker UX from #206 is preserved.

Result — all three UX contracts now hold simultaneously:
- \`--campaign foo\` / \`-c foo\` (space-separated) — value "foo" ✓ **(was broken)**
- \`--campaign=foo\` / \`-c=foo\` (equals form) — value "foo" ✓
- \`--campaign\` / \`-c\` (bare) — interactive picker ✓

## Files

- \`cmd/camp/args_normalize.go\` — new pure function \`normalizeOptionalValueFlagArgs\`
- \`cmd/camp/args_normalize_test.go\` — 13 unit-test cases covering gluing, already-glued, bare-at-end, bare-before-flag, \`--\` separator, and unrelated flags
- \`cmd/camp/root.go\` — wire the normalizer into \`Execute()\`, documented with a comment explaining why it's needed

## Test plan

- [x] \`just test unit\` — 2116 / 2116 pass (63 packages)
- [x] \`just test integration\` — 137 / 137 pass (previously 136/137 with \`TestIntentAdd_TargetCampaignWritesToSelectedRoot\` failing)
- [x] Failing test \`TestIntentAdd_TargetCampaignWritesToSelectedRoot\` runs green in isolation
- [x] New \`TestNormalizeOptionalValueFlagArgs\` table-driven test covers all argument permutations

## Scope

The \`optionalValueFlags\` registry in \`args_normalize.go\` is a one-entry map today (\`--campaign\` / \`-c\`) because that's the only \`NoOptDefVal\` flag in the codebase (confirmed via grep). The comment on the registry notes it should be kept in sync with any future flag that sets \`NoOptDefVal\`.

### DIFF
--- a/cmd/camp/args_normalize.go
+++ b/cmd/camp/args_normalize.go
@@ -1,0 +1,79 @@
+package main
+
+import (
+	"strings"
+)
+
+// optionalValueFlags are long/short flag tokens whose flag definitions set
+// NoOptDefVal so that a bare occurrence (no value) falls back to a sentinel
+// default — see cmd/camp/intent/add.go. Cobra's pflag library refuses to
+// consume the next space-separated argument as the value for these flags,
+// which means users who run `--campaign foo "title"` end up with "foo"
+// treated as a positional argument instead of the flag's value.
+//
+// normalizeOptionalValueFlagArgs rewrites the space-separated form to the
+// `--flag=value` form before cobra ever sees os.Args, so the standard
+// Unix convention (`--flag value`) works as expected while the bare-flag
+// picker UX is preserved.
+//
+// Keep this list in sync with every flag that sets NoOptDefVal.
+var optionalValueFlags = map[string]struct{}{
+	"--campaign": {},
+	"-c":         {},
+}
+
+// normalizeOptionalValueFlagArgs rewrites `--flag value` to `--flag=value`
+// for flags listed in optionalValueFlags, returning a new slice. Bare
+// occurrences (last arg, or followed by another flag starting with `-`)
+// are left untouched so the NoOptDefVal sentinel still fires.
+//
+// Tokens after a standalone `--` separator are treated as positionals and
+// not rewritten.
+func normalizeOptionalValueFlagArgs(args []string) []string {
+	out := make([]string, 0, len(args))
+	sawDoubleDash := false
+	for i := 0; i < len(args); i++ {
+		arg := args[i]
+
+		if sawDoubleDash {
+			out = append(out, arg)
+			continue
+		}
+		if arg == "--" {
+			sawDoubleDash = true
+			out = append(out, arg)
+			continue
+		}
+
+		// Skip `--flag=value` forms — already glued.
+		if strings.Contains(arg, "=") {
+			out = append(out, arg)
+			continue
+		}
+
+		if _, ok := optionalValueFlags[arg]; !ok {
+			out = append(out, arg)
+			continue
+		}
+
+		// Flag is in our set. Look at the next token to decide whether to glue.
+		next := ""
+		hasNext := i+1 < len(args)
+		if hasNext {
+			next = args[i+1]
+		}
+
+		// Bare flag cases — keep the flag as-is and let NoOptDefVal apply.
+		//   - last argument
+		//   - next token is another flag (starts with `-`, or is the `--` separator)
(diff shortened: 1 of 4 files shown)


obey-agent (approved):
Verdict: Approve

Overview:
I re-read this change from scratch and didn't find a blocking regression. The argument normalizer is narrow, restores the documented `--campaign value` form, and preserves the bare-flag interactive picker path.

Key Findings:
- No blocking issues found in the current revision.

What's Done Well:
- Keeps the fix scoped to the one `NoOptDefVal` flag that actually regressed.
- Adds solid table coverage for glued, bare, separator, and end-to-end argument permutations.

Staff Standard:
Yes. I would be comfortable submitting this change as my own work.

lancekrogers:
Pushed \`f28e170\` closing the two follow-up gaps from the self-review above.

### 1. Integration matrix — \`TestIntentAdd_TargetCampaignFlagForms\`

Exercises all four valid invocation forms through cobra end-to-end (not just the original long-space form):

- \`--campaign <name>\` — long, space-separated (the original bug)
- \`--campaign=<name>\` — long, equals-attached
- \`-c <name>\` — short, space-separated
- \`-c=<name>\` — short, equals-attached

Each sub-test uses a dedicated target campaign so there is no shared state between forms, and a final assertion confirms no intent leaked into the source campaign across any of the four. If cobra/pflag ever changes optional-value semantics, the specific failing form will be named.

### 2. Drift guard — \`TestOptionalValueFlagsCoverNoOptDefValFlags\`

Walks the entire cobra command tree from \`rootCmd\`, finds every **non-bool** flag with \`NoOptDefVal != ""\`, and asserts each such flag token (both long and short forms) is present in the hand-maintained \`optionalValueFlags\` map in \`cmd/camp/args_normalize.go\`.

The bool exclusion is important: pflag auto-assigns \`NoOptDefVal="true"\` to every \`BoolP\` / \`Bool\`, so without the filter the test would flag every bool in the tree. The doc comment on the test explains this clearly for future readers.

**Sanity-checked:** temporarily added a second \`NoOptDefVal\` elsewhere, ran the test, got:
\`\`\`
non-bool flags with NoOptDefVal set but missing from optionalValueFlags in cmd/camp/args_normalize.go:
  [--type -t]

Add each token to the optionalValueFlags map so space-separated \`--flag value\` parses correctly.
\`\`\`

Actionable error message pointing at the file to edit. Reverted the temp change.

### Verification

- \`go test ./cmd/camp/ -run TestOptionalValueFlagsCoverNoOptDefValFlags\` — pass (raw)
- \`go test ./cmd/camp/ -run TestNormalizeOptionalValueFlagArgs\` — pass (14 cases, unchanged + new guard)
- \`go test -tags=integration ./tests/integration/ -run TestIntentAdd_TargetCampaignFlagForms\` — pass (4 sub-tests)
- \`just test all\` — 2117 unit + 141 integration (from raw \`go test -v\`), 0 failures

### No production code changes

This commit is test-only. The actual fix (\`f40b2a1\`) is unchanged.

PR #214, opened 2026-04-14
fix(project): reject duplicate --link targets up front

lancekrogers:
Targets [#207](https://github.com/Obedience-Corp/camp/pull/207). Closes the inline comment on \`internal/project/link.go:95\` ([r3074896429](https://github.com/Obedience-Corp/camp/pull/207#discussion_r3074896429)).

## Root cause (not what the original comment diagnosed)

The reviewer observed the right symptom but proposed the wrong lever. Two \`camp project add --link\` invocations with the same source path and different \`--name\` aliases would both succeed at the filesystem level — two distinct symlinks, both pointing at the same target — but \`List()\` applies URL-based dedup (\`internal/project/list.go:230\` \`deduplicateByRemoteURL\`) that keeps only one entry per repo URL. The other alias then silently disappears from \`camp project list\`, \`camp project run\`, and \`camp leverage\`.

The dedup itself is **intentional and valuable** — it suppresses the submodule-plus-linked-clone duplicate of the same repo, which is exactly what \`camp leverage\` needs to behave. So rather than weakening the dedup, we reject the duplicate at link time where a clear error is possible.

## Change

1. New \`ErrProjectAlreadyLinked\` error in \`internal/project/add.go\` that names the existing alias, the attempted alias, and the shared target, and points the user at \`camp project unlink <existing>\` to re-link.

2. New \`findExistingLinkToTarget\` helper in \`internal/project/link.go\` that walks \`campaignRoot/projects/\`, resolves every symlink via \`filepath.EvalSymlinks\`, and returns the name of any that resolves to the same target as the incoming link. Uses \`EvalSymlinks\` on both sides so path-shape differences (\`/private/var/...\` vs \`/var/...\` on macOS) don't cause false negatives.

3. \`AddLinked\` calls the helper before \`os.Symlink\` and before writing the \`.camp\` marker — nothing is created on rejection, no cleanup path needed.

4. The URL dedup in \`list.go\` is unchanged. It still handles submodule↔linked overlap.

## Tests

- \`TestAddLinked_RejectsDuplicateTarget\` — links a dir, attempts a second link under a different name, asserts:
  - \`ErrProjectAlreadyLinked\` with correct \`ExistingName\` / \`AttemptedName\`
  - Error unwraps to \`camperrors.ErrAlreadyExists\` (so generic \`errors.Is\` routing still works)
  - The second symlink was **not** created
  - The first symlink still resolves to the target
- \`TestAddLinked_AllowsRelinkAfterUnlink\` — guards that the rejection is scoped to active duplicates; unlinking the original frees the target for re-linking under any name.

## Verification

- \`just test all\`: 2134 unit tests + 145 integration tests pass (from 2132 / 145 on the base branch — +2 for the new cases).
- Existing \`ErrProjectExists\` (destination-path collision) behavior is untouched; the new rejection is additive.

## Scope notes

This PR doesn't touch the other open reviewer finding on \`--no-register\` + \`--link\` interaction — [separate thread](https://github.com/Obedience-Corp/camp/pull/207#discussion_r3075824015) is still live with a pushback argument.


internal/project/link.go
```diff
@@ -96,6 +96,23 @@ func AddLinked(ctx context.Context, campaignRoot, localPath string, opts LinkOpt
 		return nil, &ErrProjectExists{Name: name, Path: destPath}
 	}
 
+	// Reject if another symlink in this campaign already points to the same
+	// target. Two aliases of the same linked directory would survive the
+	// symlink creation but collide inside List()'s URL-based dedup, which
+	// would silently hide one alias from `project list`, `project run`, and
+	// `leverage`. Failing up front with a clear error beats a silent trap.
+	existing, err := findExistingLinkToTarget(campaignRoot, absLocal, fullPath)
```
obey-agent: On the current base this block is dead code: `codex/camp-optional-submodules` already added `ensureLinkedTargetUnique(...)` just above it, and that helper returns the generic string error before we ever reach `findExistingLinkToTarget`. I cherry-picked this commit onto the current base and `go test ./internal/project` fails in `TestAddLinked_RejectsDuplicateTarget` for exactly that reason. Please fold the structured `ErrProjectAlreadyLinked` path into the existing guard (or replace that earlier check) so the merge result actually exercises this code.

internal/project/add.go
```diff
@@ -49,6 +49,33 @@ func (e *ErrProjectExists) Unwrap() error {
...
+	ExistingName string
+	// Target is the absolute path of the linked directory (symlink target).
+	Target string
+	// AttemptedName is the alias the caller just tried to use.
+	AttemptedName string
+}
+
+func (e *ErrProjectAlreadyLinked) Error() string {
+	return fmt.Sprintf(
+		"project %q is already linked as %q → %s; unlink first with 'camp project unlink %s' to re-link under a different name",
```
obey-agent, line 69: This remediation points users to `camp project unlink`, but there is no such CLI command today; linked projects are unlinked via `camp project remove <name>`. As written, the nicer error type still sends the user to a nonexistent command. Please update the message to the actual command or add the command in the same change.

obey-agent (changes requested):
Verdict: Request Changes

The intent here is good: rejecting duplicate `--link` targets at creation time is the right place to solve the silent-alias-loss problem, and the added structured error plus unit coverage are moving in the right direction. But on the current `codex/camp-optional-submodules` base, the merge result does not actually deliver the new behavior.

Key findings:
- `internal/project/link.go`: on the current base, `ensureLinkedTargetUnique(...)` already rejects the duplicate a few lines earlier, so the new `findExistingLinkToTarget` / `ErrProjectAlreadyLinked` path is dead code in the merge result. I cherry-picked this commit onto the current base and `go test ./internal/project` fails in `TestAddLinked_RejectsDuplicateTarget` because the code still returns the older generic string error.
- `internal/project/add.go`: the new error text tells users to run `camp project unlink`, but that command does not exist. Linked projects are removed/unlinked via `camp project remove <name>` today.

What’s done well:
- The PR preserves URL-based dedup in `List()` instead of weakening it for a narrower symptom.
- The new targeted unit tests are the right level to lock down this behavior once the merge-result path is corrected.

Staff standard:
- Not yet. I would not want to merge this as-is because the actual merge result on the current base fails the new unit expectation and the user-facing remediation path is wrong.

lancekrogers:
Follow-up in 7dd6383: the tests I added in the first commit used \`t.TempDir()\`, which writes to the host filesystem. This repo's standard is that **any** filesystem-touching test runs inside the shared Docker container via testcontainers-go. That rule is explicit precisely so buggy filesystem code can't hit real paths the user cares about, and I should have followed it from the start.

Reverted the unit-level \`t.TempDir()\` tests from \`internal/project/linked_test.go\` and added equivalent coverage to \`tests/integration/project_linked_test.go\` using the established \`GetSharedContainer(t)\` + \`RunCampInDir\` pattern:

- \`TestProject_AddLink_RejectsDuplicateTargetInSameCampaign\` — CLI exits non-zero with an "already linked" message naming the existing alias, the second symlink is not created, and \`project list\` shows only the original alias.
- \`TestProject_AddLink_AllowsRelinkAfterUnlink\` — link, \`project remove\`, re-link under a different alias, assert the new symlink is created.

Both tests exercise the real \`camp\` binary end-to-end inside the alpine container; nothing touches the host filesystem.

\`just test all\` on the updated branch: 2132 unit + 146 integration, zero failures. No production code change — the fix in 5a01f9c stands.

internal/project/link.go
obey-agent: Follow-up: the proposed `camp project link` / `camp project unlink` UX does not change this specific merge-result problem. Even if we add those commands here, the duplicate-target guard still needs to be consolidated so the structured `ErrProjectAlreadyLinked` path is the one that actually executes on the current base.

internal/project/add.go
obey-agent: Follow-up: if this PR grows to add the dedicated `camp project link` / `camp project unlink` commands, this specific concern is resolved. In that shape, the remediation text is pointing at a real command and the UX becomes clearer than overloading `project remove` for linked workspaces.

obey-agent:
Implemented follow-up changes on this branch after rebasing it onto the current `codex/camp-optional-submodules` base.

What changed:
- added canonical `camp project link <path>` and `camp project unlink <name>` commands
- kept linked-project behavior routed through the existing internal link/unlink logic instead of creating a second implementation
- updated `camp project add --link` help text to point users at `project link` as the preferred UX
- fixed the duplicate-target rejection so the structured `ErrProjectAlreadyLinked` path is now the one that actually executes on the current base
- kept `project remove` for submodule removal, while `project unlink` is now the dedicated linked-project inverse

The dedicated `project unlink` command also resolves the earlier user-facing error-path concern, since remediation text now points at a real command.

Verification run on this branch:
- `go test ./cmd/camp/... ./internal/project`
- `go test ./tests/integration -tags integration -run 'TestProject_(Link|Unlink|Help)|TestProjectRun_AutoDetectFromLinkedCwd|TestGo_FromLinkedProjectCwd|TestRun_ProjectDispatch_LinkedProject'`

obey-agent:
Follow-up patch on this branch after clarifying the product model for linked projects.

Reason for the patch:
- linked projects are still the usability-first alternative to submodules
- but the campaign repo needs to track the linked project entry in git so campaign history, visualization, and reporting reflect the real project list a user sees locally
- the portability drawback of machine-specific symlink targets is accepted here because campaigns are currently a single-user workspace model

What changed in this patch:
- stopped excluding `projects/<name>` from the campaign repo
- `camp project link` now auto-commits the campaign-side symlink add by default
- `camp project unlink` now auto-commits the campaign-side symlink removal by default
- kept the linked repo's `.camp` marker local/ignored; that metadata is still machine-local reverse-lookup state, not shared campaign structure
- kept `camp project commit` support for linked git repos, and added integration coverage proving that path

Verification run on this patch:
- `go test ./cmd/camp/project ./cmd/camp/project/linked ./internal/project`
- `go test ./tests/integration -tags integration -run 'TestProject_(Link_GitRepo|Unlink_LinkedProject|Commit_LinkedGitRepo|Link_TargetCampaignOutsideCurrentContext)'`

internal/project/link.go
```diff
@@ -134,6 +153,49 @@ func AddLinked(ctx context.Context, campaignRoot, localPath string, opts LinkOpt
...
+func Unlink(ctx context.Context, campaignRoot, name string, opts UnlinkOptions) (*UnlinkResult, error) {
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	if err := ValidateProjectName(name); err != nil {
+		return nil, err
+	}
+
+	projectPath := filepath.Join(campaignRoot, "projects", name)
```
obey-agent, line 163: @/tmp/camp214-inline.md

obey-agent (changes requested):
Verdict: Request Changes
Risk: medium.

Overview:
This PR fixes the duplicate-linked-target invisibility bug and introduces dedicated `camp project link` / `camp project unlink` commands. The duplicate-target guard is a real improvement, and the normal link/unlink flows test cleanly, but the new unlink implementation is not correct for custom-path links.

Key Findings:
- [P1] `internal/project/link.go:167` still assumes every linked entry lives at `projects/<name>`. I reproduced `camp project link <repo> --path custom/external` followed by `camp project unlink external` failing with `project not found: external`, and the returned `UnlinkResult.Path` would also stage the wrong path for the auto-commit.

What's Done Well:
- The new duplicate-target error is better than the previous silent alias collision because it fails before the campaign enters a partially hidden state.
- The CLI split between submodule add/remove and link/unlink is clearer for operators.
- The added tests around duplicate rejection and relink-after-unlink cover the intended happy paths well.

Staff Standard:
Not yet. I would want custom-path links to round-trip through the new link/unlink command pair before merging this UX surface.

obey-agent:
Updated this branch on top of the current `codex/camp-optional-submodules` base and pushed `46650cd` for re-review.

The linked-project surface is narrower now:
- removed `camp project add --link`; `camp project add` is back to submodule-only behavior
- `camp project link [path]` is now the canonical entrypoint, and when `path` is omitted it links the current working directory
- `camp project unlink [name]` is now the canonical inverse, and when `name` is omitted it infers the current linked project from the working directory
- dropped the linked-project destination override so linked projects always live at `projects/<name>`

This keeps the feature aligned with the product goal here: make it easier to attach an existing workspace to a campaign without teaching users a second layer of `add` flags or custom path rules.

Also regenerated CLI docs so `camp project link` / `camp project unlink` now show up in the reference output.

Verification run on this branch:
- `go test ./cmd/camp/project/... ./internal/project`
- `go test ./tests/integration -tags integration -run 'TestProject_(Link|Unlink|Help|Add_UsesCurrentCampaignEvenIfUnregistered)|TestProjectRun_AutoDetectFromLinkedCwd|TestGo_FromLinkedProjectCwd|TestRun_ProjectDispatch_LinkedProject'`

obey-agent:
Follow-up pushed in `f6b3c7d` to address the linked test harness problem.

I removed `internal/project/linked_test.go` entirely instead of keeping any machine-local filesystem tests there. That file was using `t.TempDir()`, direct symlink creation, `os.Chdir`, and git repo setup on the developer machine.

The missing behavior coverage was moved into the container-backed integration suite in `tests/integration/project_linked_test.go`, including:
- legacy `.camp` marker rejection
- `project unlink` rejecting non-linked projects

Verification:
- `go test ./internal/project`
- `go test ./tests/integration -tags integration -run 'TestProject_(Link|Unlink|Help|Add_UsesCurrentCampaignEvenIfUnregistered)|TestProjectRun_AutoDetectFromLinkedCwd|TestGo_FromLinkedProjectCwd|TestRun_ProjectDispatch_LinkedProject'`

obey-agent:
Added a dedicated containerized integration test for the outside-campaign picker path in Manage git submodules and project repositories in the campaign.

A project is a git repository tracked as a submodule under the projects/ directory.
Projects can be added from remote URLs or existing local repositories.

Examples:
  camp project list                    List all projects
  camp project add git@github.com:org/repo.git  Add a new project
  camp project remove api-service      Remove a project

Usage:
  camp project [flags]
  camp project [command]

Available Commands:
  add         Add a project to campaign
  commit      Commit changes in a project submodule
  list        List projects in campaign
  new         Create a new project in campaign
  prune       Delete merged branches in a project
  remote      Manage remotes for a project
  remove      Remove a project from campaign
  run         Run a command inside a project directory
  worktree    Manage worktrees for a project

Flags:
  -h, --help   help for project

Global Flags:
      --config string   config file (default: ~/.obey/campaign/config.json)
      --no-color        disable colored output
      --verbose         enable verbose output

Use "camp project [command] --help" for more information about a command.. The new coverage runs the command from an unassociated project directory with no  flag, drives the real fuzzy picker through a PTY-backed  harness, selects a specific campaign interactively, and asserts the symlink plus  marker land in the chosen campaign.

obey-agent:
Added a dedicated containerized integration test for the outside-campaign picker path in camp project link. The new coverage runs from an unassociated project directory with no --campaign flag, drives the real fuzzy picker through a PTY-backed docker exec -it harness, selects a specific campaign interactively, and asserts the symlink plus .camp marker land in the chosen campaign.

PR #215, opened 2026-04-14
refactor: shell-init generation into go:embed templates with behavior tests

### LINKED ISSUE #210: refactor shell-init generation into templates with behavior-level tests
## Problem

`camp shell-init` currently generates bash/zsh/fish integration by embedding large shell programs as Go raw string literals in `internal/shell/*.go`.

That works, but it has real maintenance costs:
- reviewability is poor because the actual shell programs are buried inside Go source
- behavioral changes are hard to reason about because shell structure and Go string assembly are mixed together
- tests are mostly string-shape and syntax checks rather than shell-behavior checks
- cross-shell duplication is high, which makes future completion/wrapper fixes riskier than they should be

This became more obvious while extending `cgo` completion/navigation behavior in #208.

## Goal

Keep `camp shell-init` as the product surface, but refactor the implementation so shell code is easier to review, test, and evolve.

## Proposed Direction

- Move shell init programs into real template/assets files instead of large inline Go string literals.
- Use `go:embed` or equivalent so Go is only responsible for small substitutions such as shortcut/completion entries.
- Preserve the current user-facing behavior for `camp`, `cgo`, `cr`, `csw`, `cint`, and `cie`.
- Reduce duplication across bash/zsh/fish where practical without forcing a false abstraction.

## Testing Expectations

Add behavioral coverage, not just string checks:
- generate each shell init script
- load/source it in the target shell
- stub `camp` where needed
- verify wrapper behavior for directory-changing commands and completion entrypoints

Where feasible, these shell-behavior tests should run in a controlled/containerized environment rather than depending on the host shell/filesystem.

## Acceptance Criteria

- shell init sources are no longer authored as large inline raw string literals in `internal/shell/*.go`
- generated output still supports the current wrapper/completion behavior across bash, zsh, and fish
- syntax validation exists for each supported shell
- at least one behavior-level test exists per shell family for wrapper/completion integration
- follow-up changes do not regress the UX added in #208

## Context

This is intentionally a follow-up to #208 rather than part of that PR.

obey-agent:
## Summary

Closes #210.

- **Moved** bash/zsh/fish shell init programs from inline Go raw string literals (`const bashInitTemplate = ...`) into standalone `internal/shell/templates/*.sh.tmpl` files
- **Introduced** `go:embed` + `text/template` to load and render templates at runtime, with dynamic shortcut data injected via a `templateData` struct
- **Reduced** `bash.go`, `zsh.go`, and `fish.go` from ~150-240 lines each to ~12 lines (thin render calls)
- **Added** behavior-level tests (`behavior_test.go`) that:
  - Generate the init script for each shell family
  - Source it in a real bash/zsh/fish process
  - Stub the `camp` binary to simulate `--print` output
  - Verify `camp go` / `cgo` perform `cd` to the correct directory
  - Verify `camp switch` / `csw` perform `cd`
  - Verify `cr`, `cint`, `cie` pass through to the real binary
  - Verify completion functions are registered
- **Added** per-shell syntax validation tests (`bash -n` / `zsh -n` / `fish -n`)
- All 57 tests pass (4 fish tests skip gracefully when fish is not installed)
- Generated output is byte-identical to the previous inline implementation

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/shell/ -race` passes (57 tests, 0 failures)
- [x] `go test ./...` passes (full project)
- [x] `go vet ./internal/shell/` passes
- [x] Behavior tests verify `camp go`, `cgo`, `csw`, `cr`, `cint`, `cie` wrapper behavior in real bash and zsh
- [x] Syntax validation tests run `bash -n` and `zsh -n` on generated output
- [x] Fish tests skip gracefully when fish is not available
- [ ] Verify no regression on `camp shell-init bash|zsh|fish` output in a real shell session


obey-agent (commented):
Verdict: Approve

Overview:
This refactor materially improves maintainability without changing behavior. Risk is low: I verified the generated `shell-init` output is byte-identical to `main` for bash, zsh, and fish, and the new behavior tests exercise the shell wrappers in real interpreters instead of just snapshotting strings.

Key Findings:
- None.

What's Done Well:
- The template extraction keeps the shell-specific entry points very small and pushes the shell bodies into the right asset layer.
- The new behavior and syntax tests are meaningful; they validate directory-changing wrapper behavior, passthrough aliases, completion registration, and shell syntax in actual bash/zsh/fish processes.
- `go test -race ./internal/shell`, `go vet ./internal/shell`, and the byte-for-byte output comparison against `main` all came back clean.

Staff Standard:
Yes. I would be comfortable shipping this refactor as my own work.

PR #216, opened 2026-04-14
feat: programmatic CLI parity for intent add and edit

### LINKED ISSUE #209: Intent CLI needs programmatic parity with TUI/editor for create and edit
## Summary

Agents need the CLI and humans need the TUI / editor. Both are first-class paths and both should support **create**, **view**, and **edit** of intents at full fidelity.

Today, the CLI has gaps that force agents out of the programmatic path for anything beyond a title-only stub:

- **Create (`intent add`):** positional title is the only non-interactive input. No way to set body, concept, or author from the command line.
- **Edit (`intent edit`):** always opens `$EDITOR`. No programmatic field updates.
- **View (`intent show`):** already supports `--format json|yaml|text`. No changes needed here — this confirms the pattern is already established for other commands to match.

This issue is about closing the create and edit gaps without touching the TUI/editor flows that humans use.

## Principle

> Every intent operation a human can do through the TUI or editor, an agent can do through CLI flags, stdin, or files — and vice versa.

Neither path is a second-class citizen. The TUI stays because humans prefer it. The CLI grows because agents cannot use a TUI.

### `camp intent add` (`cmd/camp/intent/add.go`)

- `camp intent add "Title"` → ultra-fast non-interactive path. Title-only.
- `camp intent add` (no args) → BubbleTea TUI. Interactive.
- `camp intent add --full` → BubbleTea TUI including body textarea. Interactive.
- `camp intent add --edit "Title"` → opens `$EDITOR`. Interactive.

`intent.CreateOptions` already has `Title`, `Type`, `Concept`, `Body`, `Author`. The CLI path only populates `Title`, `Type`, and hard-codes `Author = "agent"` (`add.go:99-113`).

### `camp intent edit` (`cmd/camp/intent/edit.go`)

- `camp intent edit [id]` → resolves the intent and always opens `$EDITOR`.
- No flags exist to set field values directly.

### `camp intent show` (`cmd/camp/intent/show.go`)

- `camp intent show <id> [-f text|json|yaml]` — already programmatic. Nothing to change here.

### `intent add` — programmatic creation with full field support

Add flags that let a non-interactive caller set every `CreateOptions` field:

| Flag | Type | Purpose |
| --- | --- | --- |
| `--body <string>` | string | Body as a literal string argument |
| `--body-file <path>` | string | Read body from a file. `-` means stdin |
| `--concept <name>` | string | Set the concept field |
| `--author <string>` | string | Override the hard-coded `"agent"` attribution |

### `intent edit` — programmatic field updates

Add flags that let a non-interactive caller update any editable field without launching `$EDITOR`:

| Flag | Type | Purpose |
| --- | --- | --- |
| `--title <string>` | string | Update the title |
| `--body <string>` | string | Replace the body with a literal string |
| `--body-file <path>` | string | Replace the body from a file. `-` means stdin |
| `--append-body <string>` \| `--append-body-file <path>` | string | Append to the existing body rather than replace |
| `--type <string>` | string | Update the type (idea, feature, bug, research, chore) |
| `--concept <name>` | string | Update the concept |
| `--status <string>` | string | Update the status (only transitions that the service already allows) |
| `--priority <string>` | string | Update priority |
| `--horizon <string>` | string | Update horizon |
| `--author <string>` | string | Record the editing author in the audit log |

### Mutual exclusivity (both commands)

- `--body` and `--body-file` are mutually exclusive.
- `--append-body` and `--append-body-file` are mutually exclusive (edit only).
- `--body` / `--body-file` cannot combine with `--append-body` / `--append-body-file` (edit only).

### Precedence with existing interactive flags

- `add --edit` combined with `--body` / `--body-file` pre-fills the editor template with the provided body. The editor still opens because the human asked for it.
- `add --full` combined with any programmatic body flag exits with a usage error — mixing TUI with CLI inputs is ambiguous.
- `edit` with any programmatic flag (e.g. `--title`, `--body`, `--status`) SKIPS `$EDITOR` and performs a direct update via the service. Absent any programmatic flag, behavior is unchanged (opens editor).
- `edit --edit` with programmatic flags: use the flags to pre-fill a temporary file, then open `$EDITOR` on it, preserving the "human asked for the editor" intent.

### Non-interactive enforcement

If stdin is not a TTY AND no title positional is provided for `add`, exit with a clear usage error instead of entering the TUI. Same principle for `edit` when no ID and no programmatic flags are provided.

### Create

```bash
# Existing ultra-fast path — unchanged
camp intent add "Add dark mode"

# New: title + inline body, purely programmatic
camp intent add "Add dark mode" \
  --body "Toggle under Settings > Appearance. Persists across sessions via localStorage."

# New: body from a markdown file an agent wrote earlier
camp intent add "Migrate billing to Stripe v3" --body-file ./artifacts/billing-migration.md

# New: body piped from stdin
agent-output | camp intent add "Extracted plan from session" --body-file -

# New: every field set in one call
camp intent add "Fix auth race condition" \
  -t bug \
  --concept auth \
  --body-file ./findings.md \
  --author festival-executor-sonnet

# Combined with --edit: body pre-populates the editor template
camp intent add "Long-form proposal" --body-file ./draft.md --edit
```

### Edit

```bash
# Today: always interactive
camp intent edit auth-race

# New: flip status programmatically
camp intent edit auth-race --status active --author release-operator

# New: append an agent's new finding to an existing intent without an editor
camp intent edit auth-race --append-body-file ./new-evidence.md --author agent

# New: bulk pipelines become one-liners
for id in $(camp intent list --status proposed --format json | jq -r '.[].id'); do
  camp intent edit "$id" --status active --author scheduler
done

# Still supported: human path, editor opens as today
camp intent edit auth-race
```

### Shared body-resolution helper

A single helper serves both `add` and `edit`:

```go
// resolveBody returns (body, set, err).
//   set=false means no --body or --body-file was provided (edit treats as "do not change").
//   set=true  means body was explicitly provided; empty string is a valid value.
func resolveBody(cmd *cobra.Command) (string, bool, error) { ... }
```

Implementation:

- If `--body` is set, return its value, `true`.
- If `--body-file` is set: `-` → read `cmd.InOrStdin()`; otherwise read the file. Enforce a 10 MiB cap with a clear error on overflow.
- Otherwise return `""`, `false`.

Trim a single trailing newline so pipe-based bodies are tidy; preserve internal whitespace.

### `intent add` extension

In `runIntentAdd` (`add.go:99-113`), after the existing `len(args) > 0` check:

```go
body, bodySet, err := resolveBody(cmd)
if err != nil {
    return err
}
concept, _ := cmd.Flags().GetString("concept")
author, _ := cmd.Flags().GetString("author")
if author == "" {
    author = "agent"
}

opts := intent.CreateOptions{
    Title:   args[0],
    Type:    intent.Type(intentType),
    Concept: concept,
    Author:  author,
}
if bodySet {
    opts.Body = body
}
```

Guard: if `--body` / `--body-file` / `--concept` / `--author` are set WITHOUT a title positional AND not in a TTY, exit with usage error.

Guard: `--full` + any new body flag → usage error.

### `intent edit` extension

Add an `UpdateOptions` struct (or equivalent) and an `IntentService.Update(ctx, id, opts)` method if not already present. The CLI layer maps flags → `UpdateOptions`:

```go
type UpdateOptions struct {
    Title       *string   // pointer to distinguish "unset" from "empty"
    Type        *Type
    Concept     *string
    Status      *Status
    Priority    *string
    Horizon     *string
    Body        *string   // full replacement
    AppendBody  *string   // append instead of replace
    Author      string    // for audit event
}
```

`runIntentEdit` becomes:

1. Resolve the intent as it does today (ID or fuzzy picker if no ID and TTY).
2. If any programmatic flag is set, call `svc.Update(ctx, id, opts)` and emit an audit event. Skip `$EDITOR`.
3. If no programmatic flags AND no `$EDITOR` override: open the editor exactly as today.
4. If `--edit` AND programmatic flags: pre-fill a temp file with the merged result, then open `$EDITOR` (human explicitly asked).

### Audit events

Every update MUST produce an audit event with:

- `Type`: `EventEdit` (or per-field granularity if already supported)
- `Author`: from `--author` flag or from `git config user.name` for TUI/editor paths
- `Changes`: map of field → (old, new) for the fields actually updated

This keeps the audit trail usable for agents and humans.

### Flag registration

```go
// add.go
flags.String("body", "", "Body as a literal string (non-interactive)")
flags.String("body-file", "", "Read body from file; use '-' for stdin")
flags.String("concept", "", "Set the concept name")
flags.String("author", "", "Override the author attribution (default: 'agent' for non-TUI paths)")
intentAddCmd.MarkFlagsMutuallyExclusive("body", "body-file")

// edit.go
flags.String("title", "", "Update the title")
flags.String("body", "", "Replace the body with a literal string")
flags.String("body-file", "", "Replace the body from file; use '-' for stdin")
flags.String("append-body", "", "Append a literal string to the existing body")
flags.String("append-body-file", "", "Append contents of file to the existing body; '-' for stdin")
flags.String("concept", "", "Update the concept")
flags.String("priority", "", "Update priority")
flags.String("horizon", "", "Update horizon")
flags.String("author", "", "Record the editing author in the audit log")
// type and status are kept as -t / -s but re-purposed from 'filter' to 'update' when combined with an ID.
// Prefer adding new distinct flags like --set-type and --set-status to avoid overloading.
intentEditCmd.MarkFlagsMutuallyExclusive("body", "body-file", "append-body", "append-body-file")
intentEditCmd.MarkFlagsMutuallyExclusive("body", "append-body")
intentEditCmd.MarkFlagsMutuallyExclusive("body-file", "append-body-file")
```

Note the existing `edit` flags `-s/--status`, `-t/--type`, `-p/--project` are **picker filters**, not update targets. Introduce `--set-status`, `--set-type`, and `--set-concept` (aliased from `--concept`) for updates to avoid overloading the same flag name with two different semantics. Document this clearly.

### Backward compatibility

All new flags are additive. Every current invocation works unchanged. Picker-filter flags keep their existing meaning.

### `intent add`

- [ ] `--body`, `--body-file`, `--concept`, `--author` implemented with the shared resolver.
- [ ] `--body` and `--body-file` mutually exclusive.
- [ ] `--body-file -` reads stdin.
- [ ] 10 MiB cap on `--body-file` with clear error.
- [ ] `--full` + body flags = usage error.
- [ ] `--edit` + body flags pre-fills the editor template.
- [ ] Non-TTY + body flags + no title positional = usage error, no TUI launch.
- [ ] Ultra-fast path populates `CreateOptions.Body`, `Concept`, `Author`.
- [ ] `--author` unset → defaults to `"agent"` (unchanged).

### `intent edit`

- [ ] `UpdateOptions` exists on the service (or equivalent pointer-semantics struct).
- [ ] `--title`, `--body`, `--body-file`, `--append-body`, `--append-body-file`, `--set-type`, `--set-status`, `--set-concept`, `--priority`, `--horizon`, `--author` implemented.
- [ ] Any programmatic flag present → skip `$EDITOR`, update via service, emit audit event.
- [ ] No programmatic flags → open `$EDITOR` exactly as today.
- [ ] `--edit` + programmatic flags → pre-fill temp file, then editor.
- [ ] Status transitions respect service-level validation.
- [ ] Picker filter flags (`-s/--status`, `-t/--type`, `-p/--project`) preserve today's "filter picker" meaning.
- [ ] Audit event records old/new per changed field and the author.

### Docs and tests

- [ ] `docs/cli-reference/camp_intent_add.md` and `camp_intent_edit.md` updated with the new flags and examples from this issue.
- [ ] `--help` output for both commands lists the new flags with one-line descriptions.
- [ ] Unit tests in `add_test.go` and `edit_test.go` cover every new flag combination plus the mutual-exclusivity errors.
- [ ] Integration tests cover the end-to-end flow of agent pipelines creating and mutating intents without ever invoking a TUI or editor.

## Motivation Recap

Without this change, automation callers (fest, camp-sync, CI, agent sessions, scheduled pipelines) cannot:

- Create intents with a meaningful body.
- Update an existing intent's status or body.
- Record agent attribution in the audit log.

Their only options today are title-only stubs plus a manual follow-up in the TUI/editor, or writing the intent file directly on disk — which bypasses the service, audit trail, validation, and auto-commit. That is fragile.

Both problems disappear once `add` and `edit` have programmatic parity with the TUI and editor paths. The TUI / editor paths are unchanged; humans keep what they like. Agents stop being second-class.

## Related

- `internal/intent/service.go` — `CreateOptions.Body/Concept/Author` already exist; may need `UpdateOptions` for edit.
- `cmd/camp/intent/add.go:99-113` — the ultra-fast path.
- `cmd/camp/intent/edit.go` — currently editor-only.
- `cmd/camp/intent/show.go` — reference for the CLI-parity pattern; already supports `--format json|yaml|text`.

lancekrogers:
## Summary

Closes #209. Gives agents full programmatic parity with the TUI/editor paths for `intent add` and `intent edit`.

- **`intent add`**: new `--body`, `--body-file`, `--concept`, `--author` flags with mutual exclusivity enforcement, non-TTY guard, and `--full`/`--edit` interaction rules
- **`intent edit`**: new `--title`, `--body`, `--body-file`, `--append-body`, `--append-body-file`, `--set-type`, `--set-status`, `--set-concept`, `--priority`, `--horizon`, `--author` flags that skip $EDITOR and apply updates via `UpdateDirect`
- **Service layer**: `UpdateOptions` struct with pointer fields (nil = unset), `UpdateDirect` method with validation, status-aware file moves, and `FieldChange` tracking
- **Audit**: `EventEdit` type with per-field old/new change records in the JSONL audit log
- **Shared**: `resolveBody(cmd)` helper for body flag resolution with 10 MiB stdin cap

## Test plan

- [x] Unit tests for `resolveBody` (no flags, body, body-file, mutual exclusivity, missing file, empty body, size limit at boundary)
- [x] Unit tests for `validateEditBodyFlags` (all 4 body flags in isolation, body+body-file, append+append-file, all replace-vs-append combos)
- [x] Unit tests for `hasProgrammaticEditFlags` (none set, each of 11 flags individually)
- [x] Unit tests for `buildUpdateOptions` (title only, multiple fields, append body, no flags)
- [x] Unit tests for `UpdateDirect` (title, body, append, multi-field, status change with file move, no-op identical values, no fields specified, not found, invalid type/priority/horizon validation, context cancellation, append to empty body, persistence round-trip)
- [x] Integration tests for full agent pipeline (create with body/concept/author -> programmatic edit -> verify audit trail, create then append, create then status change with file move, multiple successive edits)
- [x] `go vet ./...` clean
- [x] `go test ./...` all pass (no regressions)
- [x] `go build ./...` clean

### DIFF
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -44,6 +44,16 @@ Fast capture is optimized for speed - ideas are saved immediately.
 Use --full when you want to add a body description in the form.
 Use --edit when you need the complete template in your editor.
 
+PROGRAMMATIC (agent) FLAGS:
+  --body              Set intent body from a literal string
+  --body-file         Read intent body from a file (- for stdin)
+  --concept           Set the concept field (e.g., "projects/camp")
+  --author            Override the default author attribution
+
+  --body and --body-file are mutually exclusive.
+  --full + body flags is a usage error.
+  --edit + body flags pre-fills the editor template.
+
 Examples:
   camp intent add "Add dark mode"        Ultra-fast capture
   camp intent add -c obey-campaign "Add dark mode"
@@ -51,7 +61,10 @@ Examples:
   camp intent add --campaign             Pick a target campaign interactively
   camp intent add --full                 Full TUI (includes body)
   camp intent add -e "Complex feature"   Deep capture with editor
-  camp intent add -t feature "New API"   Set type explicitly`,
+  camp intent add -t feature "New API"   Set type explicitly
+  camp intent add "Fix login" --body "The login page returns 500"
+  camp intent add "Migrate DB" --body-file spec.md --concept projects/camp
+  echo "body" | camp intent add "Idea" --body-file -`,
 	Args: cobra.MaximumNArgs(1),
 	RunE: runIntentAdd,
 }
@@ -66,6 +79,12 @@ func init() {
 	flags.StringP("campaign", "c", "", "Target campaign by name or ID; omit value to pick interactively")
 	flags.Bool("no-commit", false, "Don't create a git commit")
 	flags.Lookup("campaign").NoOptDefVal = noOptCampaign
+
+	// Programmatic (agent) flags
+	flags.String("body", "", "Set intent body as a literal string")
+	flags.String("body-file", "", "Read intent body from file (- for stdin, 10 MiB cap)")
+	flags.String("concept", "", "Set the concept field (e.g., projects/camp)")
+	flags.String("author", "", "Override the default author attribution")
 }
 
 func runIntentAdd(cmd *cobra.Command, args []string) error {
@@ -77,6 +96,19 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 	fullMode, _ := cmd.Flags().GetBool("full")
 	targetCampaign, _ := cmd.Flags().GetString("campaign")
 	noCommit, _ := cmd.Flags().GetBool("no-commit")
+	conceptFlag, _ := cmd.Flags().GetString("concept")
+	authorFlag, _ := cmd.Flags().GetString("author")
+
+	// Resolve body from --body / --body-file (mutual exclusivity checked inside)
+	body, bodySet, err := resolveBody(cmd)
+	if err != nil {
+		return err
+	}
+
+	// --full + body flags = usage error (body in TUI conflicts with programmatic body)
+	if fullMode && bodySet {
+		return fmt.Errorf("--full and --body/--body-file are mutually exclusive")
+	}
 
 	campaignResolver := newIntentAddCampaignResolver(cmd.ErrOrStderr())
 	cfg, campaignRoot, err := campaignResolver.resolve(ctx, targetCampaign, cmd.Flags().Changed("campaign"))
@@ -96,15 +128,26 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 		return camperrors.Wrap(err, "ensuring intent directories")
 	}
 
-	// Ultra-fast path: title provided as argument (non-TUI = agent author)
+	// Determine if any programmatic flags were used (signals agent/CLI path)
+	hasProgrammaticFlags := bodySet || cmd.Flags().Changed("concept") || cmd.Flags().Changed("author")
+
+	// Ultra-fast path: title provided as argument
 	if len(args) > 0 {
-		opts := intent.CreateOptions{
-			Title:  args[0],
-			Type:   intent.Type(intentType),
-			Author: "agent",
+		// Default author for CLI-with-title is "agent", but --author overrides
+		author := "agent"
+		if cmd.Flags().Changed("author") {
+			author = authorFlag
 		}
 
-		// Deep capture overrides ultra-fast
+		opts := intent.CreateOptions{
+			Title:   args[0],
+			Type:    intent.Type(intentType),
+			Author:  author,
+			Body:    body,
+			Concept: conceptFlag,
+		}
+
+		// Deep capture overrides ultra-fast; body flags pre-fill the template
 		if useEditor {
 			return runDeepCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
 		}
@@ -112,10 +155,18 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 		return runFastCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
 	}
 
-	// TUI path: use git config author (human)
-	author := git.GetUserName(ctx)
+	// No title argument: non-TTY requires a title (can't launch TUI)
+	if !navtui.IsTerminal() && !hasProgrammaticFlags {
+		return fmt.Errorf("title argument required in non-interactive mode\n       Usage: camp intent add <title> [flags]")
+	}
 
-	// Build navigation shortcuts map (key → path) for @ completion
+	// TUI path: use git config author (human), unless --author overrides
+	author := git.GetUserName(ctx)
+	if cmd.Flags().Changed("author") {
+		author = authorFlag
+	}
+
+	// Build navigation shortcuts map (key -> path) for @ completion
 	shortcuts := make(map[string]string)
 	for key, sc := range cfg.Shortcuts() {
 		if sc.HasPath() {
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -128,9 +128,6 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 		return camperrors.Wrap(err, "ensuring intent directories")
 	}
 
-	// Determine if any programmatic flags were used (signals agent/CLI path)
-	hasProgrammaticFlags := bodySet || cmd.Flags().Changed("concept") || cmd.Flags().Changed("author")
-
 	// Ultra-fast path: title provided as argument
 	if len(args) > 0 {
 		// Default author for CLI-with-title is "agent", but --author overrides
@@ -155,8 +152,10 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 		return runFastCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
 	}
 
-	// No title argument: non-TTY requires a title (can't launch TUI)
-	if !navtui.IsTerminal() && !hasProgrammaticFlags {
+	// No title argument: non-TTY always requires a title (can't launch TUI)
+	// Programmatic flags like --body/--concept/--author supplement a title,
+	// they don't replace it.
+	if !navtui.IsTerminal() {
 		return fmt.Errorf("title argument required in non-interactive mode\n       Usage: camp intent add <title> [flags]")
 	}
 
--- a/cmd/camp/intent/edit.go
+++ b/cmd/camp/intent/edit.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Obedience-Corp/camp/internal/git/commit"
 	"github.com/Obedience-Corp/camp/internal/intent"
 	"github.com/Obedience-Corp/camp/internal/intent/audit"
+	navtui "github.com/Obedience-Corp/camp/internal/nav/tui"
 	"github.com/Obedience-Corp/camp/internal/paths"
 )
 
@@ -132,6 +133,10 @@ func runIntentEdit(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	} else {
+		// No ID + programmatic flags + no TTY = deterministic error
+		if programmatic && !navtui.IsTerminal() {
+			return fmt.Errorf("intent ID required in non-interactive mode\n       Usage: camp intent edit <id> [flags]")
+		}
 		// No ID - show fuzzy picker
 		selectedIntent, err = pickIntent(ctx, svc, statusFilter, typeFilter, projectFilter)
 		if err != nil {
@@ -219,7 +224,12 @@ func runProgrammaticEdit(
 
 	// Auto-commit (unless --no-commit)
 	if !noCommit {
-		files := commit.NormalizeFiles(campaignRoot, updated.Path, audit.FilePath(intentsDir))
+		filesToCommit := []string{updated.Path, audit.FilePath(intentsDir)}
+		// Status changes move the file — stage the old path deletion too
+		if target.Path != updated.Path {
+			filesToCommit = append(filesToCommit, target.Path)
+		}
+		files := commit.NormalizeFiles(campaignRoot, filesToCommit...)
 		commitResult := commit.Intent(ctx, commit.IntentOptions{
 			Options: commit.Options{
 				CampaignRoot:  campaignRoot,
--- a/cmd/camp/intent/edit.go
+++ b/cmd/camp/intent/edit.go
@@ -12,28 +12,57 @@ import (
 	"github.com/Obedience-Corp/camp/internal/config"
 	"github.com/Obedience-Corp/camp/internal/editor"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/git/commit"
 	"github.com/Obedience-Corp/camp/internal/intent"
+	"github.com/Obedience-Corp/camp/internal/intent/audit"
 	"github.com/Obedience-Corp/camp/internal/paths"
 )
 
 var intentEditCmd = &cobra.Command{
 	Use:   "edit [id]",
 	Short: "Edit an existing intent",
-	Long: `Edit an intent in your preferred editor.
+	Long: `Edit an intent in your preferred editor or programmatically via flags.
 
-If ID is provided, opens the intent directly (supports partial matching).
-If no ID is provided, shows a fuzzy picker to select an intent.
+If no programmatic flags are given, opens the intent in $EDITOR.
+If any programmatic flag is present, applies the update directly and
+emits an audit event — no editor is launched.
 
-PARTIAL ID MATCHING:
-  Full ID:       20260119-153412-add-retry-logic
-  Time suffix:   153412-add-retry
-  Slug portion:  add-retry
+PICKER / EDITOR PATH:
+  If ID is provided, opens the intent directly (supports partial matching).
+  If no ID is provided, shows a fuzzy picker to select an intent.
+
+PROGRAMMATIC FLAGS (skip $EDITOR):
+  --title            Set a new title
+  --body             Replace the body with a literal string
+  --body-file        Replace the body from a file (- for stdin)
+  --append-body      Append text to the existing body
+  --append-body-file Append text from a file (- for stdin)
+  --set-type         Change the intent type
+  --set-status       Change the intent status
+  --set-concept      Change the concept field
+  --priority         Change priority (low, medium, high)
+  --horizon          Change horizon (now, next, later, someday)
+  --author           Override the author attribution
+
+MUTUAL EXCLUSIVITY:
+  --body vs --body-file
+  --append-body vs --append-body-file
+  --body/--body-file vs --append-body/--append-body-file (replace vs append)
+
+FILTER FLAGS (for picker only, not update targets):
+  -s/--status        Filter picker by status
+  -t/--type          Filter picker by type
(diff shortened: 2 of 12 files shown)


cmd/camp/intent/edit.go
```diff
@@ -88,17 +142,205 @@ func runIntentEdit(cmd *cobra.Command, args []string) error {
...
+	}
+
+	fmt.Printf("Intent updated: %s\n", updated.Path)
+	for _, c := range changes {
+		fmt.Printf("  %s: %q -> %q\n", c.Field, c.Old, c.New)
+	}
+
+	// Auto-commit (unless --no-commit)
+	if !noCommit {
+		files := commit.NormalizeFiles(campaignRoot, updated.Path, audit.FilePath(intentsDir))
```
obey-agent: Status-changing programmatic edits need the source path in the selective commit scope too. Right now `camp intent edit <id> --set-status ready` removes the inbox file on disk, but the auto-commit stages only `updated.Path` and `.intents.jsonl`. I reproduced this leaving the old inbox file tracked in Git and deleted-but-unstaged after the command printed `Committed changes to git`. `move` and `archive` already include both source and destination paths; this path should do the same (for example by including `target.Path` or returning the original path from `UpdateDirect`).

cmd/camp/intent/add.go
```diff
@@ -96,26 +128,45 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
...
 		if useEditor {
 			return runDeepCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
 		}
 
 		return runFastCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
 	}
 
-	// TUI path: use git config author (human)
+	// No title argument: non-TTY requires a title (can't launch TUI)
+	if !navtui.IsTerminal() && !hasProgrammaticFlags {
```
obey-agent: This non-TTY guard now exempts `--body`/`--body-file`/`--concept`/`--author`, but those flags still do not provide a title. In a non-interactive session, `camp intent add --body "..."` now falls through into BubbleTea and dies with `TUI error: could not open a new TTY` instead of returning the explicit `title argument required` usage error. Since there is still no programmatic title flag, missing-title adds should be rejected here regardless of those other flags.

cmd/camp/intent/edit.go
```diff
@@ -88,17 +142,205 @@ func runIntentEdit(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	// Edit the intent
+	// Programmatic path: apply updates directly, skip $EDITOR
+	if programmatic {
```
obey-agent, line 150: The control flow reaches the fuzzy picker before we get here. In a non-interactive agent session, `camp intent edit --title "new title"` with no ID now fails with `failed to initialize the fuzzy finder: terminal not cursor addressable` instead of a deterministic `id required in non-interactive mode` error. If update flags are set and no TTY is available, we should reject the missing ID before calling `pickIntent` rather than relying on the interactive picker path.

obey-agent (commented):
Verdict: Request Changes

Overview:
This PR moves intent add/edit much closer to real agent parity and the new service-layer coverage is directionally strong. Risk is medium-high right now, though, because the direct edit path still has two non-interactive control-flow holes and one status-move commit bug that leaves the repo dirty after claiming success.

Key Findings:
- `cmd/camp/intent/edit.go`: status-changing programmatic edits stage only the destination file plus `.intents.jsonl`, so the old path deletion is left unstaged. I reproduced `camp intent edit <id> --set-status ready` printing `Committed changes to git` while Git still showed the old inbox file as deleted-but-unstaged.
- `cmd/camp/intent/add.go`: missing-title adds in non-interactive mode can now bypass the explicit usage error if `--body`, `--body-file`, `--concept`, or `--author` is set, and they fall into BubbleTea with `could not open a new TTY` instead.
- `cmd/camp/intent/edit.go`: programmatic edits without an ID still enter the fuzzy picker path, so `camp intent edit --title ...` in a non-interactive session fails with `terminal not cursor addressable` instead of a deterministic `id required` error.

What's Done Well:
- The `UpdateOptions` pointer model is the right shape for direct edits and keeps unset vs empty-string semantics explicit.
- The new tests cover a lot of useful positive-path behavior across add/edit/update/audit layers.
- The audit-log extension for per-field edit changes is a good direction for operator traceability.

Staff Standard:
Not yet. I would want the non-interactive agent paths to fail deterministically and the status-change auto-commit path to preserve repository correctness before I would be comfortable submitting this as my own work.

obey-agent (commented):
Verdict: Approve

Overview:
The latest head addresses the three correctness issues from my prior review. Risk is now low to medium: the non-interactive add/edit paths fail deterministically again, and the status-changing edit path now carries the source deletion into the selective commit scope, so the repo no longer ends the command in a falsely-clean state.

Key Findings:
- No remaining blocking findings on the updated head.

What's Done Well:
- The fixes are minimal and land in the right places in the control flow instead of adding parallel handling.
- I re-ran the prior repros and confirmed: `camp intent add --body ...` now returns the expected title-required usage error, `camp intent edit --title ...` without an ID now returns the expected ID-required usage error, and `camp intent edit <id> --set-status ready` now commits the move as a rename rather than leaving the old path deleted-but-unstaged.
- `go test ./cmd/camp/intent ./internal/intent/... ./internal/git/commit` and `go build ./...` passed on the updated head.

Staff Standard:
Yes. I would be comfortable submitting this updated version as my own work.

Residual Risk:
- The fix commit did not add explicit regression tests for the three reviewed failures, so those exact edge cases still rely on manual verification today.

cmd/camp/intent/add.go
```diff
@@ -96,26 +128,44 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
...
 		}
 
 		return runFastCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
 	}
 
-	// TUI path: use git config author (human)
+	// No title argument: non-TTY always requires a title (can't launch TUI)
+	// Programmatic flags like --body/--concept/--author supplement a title,
+	// they don't replace it.
+	if !navtui.IsTerminal() {
```
obey-agent, line 158: The new comment says programmatic flags like `--body` and `--concept` "supplement a title" rather than replace it, but the guard here only runs for non-TTY sessions. In an interactive shell, `camp intent add --body "prefilled body" --no-commit` still drops into the TUI title prompt instead of rejecting the missing title. Because `runIntentAddTUI` only receives `DefaultType`, `FullMode`, `Author`, `CampaignRoot`, and `Shortcuts`, the supplied `--body` / `--concept` values are never wired through and are effectively ignored. If a title is required for the programmatic flags, enforce that before the TUI branch regardless of TTY; otherwise plumb those flags into the TUI path.

obey-agent (changes requested):
Verdict: Request Changes

Overview:
This is still strong work overall. The direct add/edit path, the service-layer update model, and the audit extensions are all materially better than the base branch, and the updated head fixed the three non-interactive issues from the prior review. Re-reviewing with fresh eyes turned up one remaining control-flow gap in `intent add`: interactive runs with programmatic flags but no title still fall into the TUI path, where the supplied values are not honored.

Key Findings:
- `cmd/camp/intent/add.go`: the missing-title guard only fires for non-TTY sessions. In an interactive terminal, `camp intent add --body 'prefilled body' --no-commit` opens the title TUI instead of rejecting the missing title, and the provided `--body` / `--concept` values are not wired into `tui.AddOptions`, so they are effectively ignored.

What's Done Well:
- The updated head correctly fixed the earlier non-interactive regressions: missing-title add now errors deterministically, programmatic edit without an ID errors deterministically, and status-changing edits now stage the old path deletion in the selective commit.
- `UpdateOptions` with pointer fields remains the right model for direct edits and keeps unset vs empty-string semantics explicit.
- The per-field audit records and the new test coverage across add/edit/update layers are solid improvements.

Staff Standard:
Not yet. I would want the interactive programmatic-add path to either reject a missing title consistently or actually honor the supplied flags before I’d be comfortable submitting this as my own work.

Risk:
Medium. The feature is close, but the current add control flow still has one user-facing mismatch between documented semantics and runtime behavior.

obey-agent (changes requested):
Verdict: Request Changes

Follow-up clarification on the remaining blocker:

The outstanding `intent add` issue is not just that interactive runs with programmatic flags but no title fall into the TUI path. The concrete bug is that the supplied programmatic values are then dropped.

Evidence chain:
- `cmd/camp/intent/add.go` only rejects missing titles in non-TTY mode.
- In an interactive shell, `camp intent add --body "prefilled body" --no-commit` therefore reaches `runIntentAddTUI(...)` instead of failing fast.
- `runIntentAddTUI(...)` only receives `tui.AddOptions{DefaultType, FullMode, Author, CampaignRoot, Shortcuts}`.
- `internal/intent/tui.AddOptions` has no fields for prefilled title, body, or concept.

So the current head has a real semantic mismatch: `--body` / `--concept` are accepted, but in this control-flow branch they are not honored at all.

What I want changed:
- If a title is required for programmatic add flags, reject the missing title before the TUI branch regardless of TTY.
- Otherwise, plumb the programmatic fields through the TUI path so the documented behavior is actually true.

Everything else from my prior re-review still stands: the earlier non-interactive regressions appear fixed, targeted tests pass, and this add-path control-flow gap is the remaining blocker I found.

obey-agent:
Fresh visibility note on the remaining blocker for the current PR head `8a57938ec228bff8e5439b5757b42cdec9f198de`:

The outstanding `intent add` issue is not just that interactive runs with programmatic flags but no title fall into the TUI path. The concrete bug is that the supplied programmatic values are then dropped.

Evidence chain:
- `cmd/camp/intent/add.go` only rejects missing titles in non-TTY mode.
- In an interactive shell, `camp intent add --body "prefilled body" --no-commit` therefore reaches `runIntentAddTUI(...)` instead of failing fast.
- `runIntentAddTUI(...)` only receives `tui.AddOptions{DefaultType, FullMode, Author, CampaignRoot, Shortcuts}`.
- `internal/intent/tui.AddOptions` has no fields for prefilled title, body, or concept.

So the current head has a real semantic mismatch: `--body` / `--concept` are accepted, but in this control-flow branch they are not honored at all.

What I want changed:
- If a title is required for programmatic add flags, reject the missing title before the TUI branch regardless of TTY.
- Otherwise, plumb the programmatic fields through the TUI path so the documented behavior is actually true.

For clarity: I checked the PR head before posting this comment, and it is still `8a57938ec228bff8e5439b5757b42cdec9f198de`, so there are no newer code commits for me to re-review beyond the head that already has the current `CHANGES_REQUESTED` review.

PR #219, opened 2026-04-14
test: cover submodule path handling in project remote set-url / rename

lancekrogers:
Targets [#207](https://github.com/Obedience-Corp/camp/pull/207). Pure test addition — **no production code change**.

## Why

Commit \`75948bf\` on this branch replaced

\`\`\`go
submodulePath := strings.TrimPrefix(resolved.Path, campRoot+\"/\")
\`\`\`

with

\`\`\`go
submodulePath := resolved.LogicalPath
\`\`\`

in all four files under \`cmd/camp/project/remote/\` (list, remove, rename, set_url). Both expressions must produce a campaign-relative key (\`projects/<name>\`) because that key ends up as the \`.gitmodules\` section header and as the \`submodule.<path>.url\` config name. If either ever drifted toward an absolute path, \`.gitmodules\` would become non-portable across machines — the whole point of the submodule manifest.

There was **zero integration coverage** for this surface on the base branch. \`grep set-url tests/integration/\` returned no hits. These tests close that gap.

## Tests added

Both in \`tests/integration/project_remote_submodule_test.go\` using the containerized pattern (\`GetSharedContainer\` + \`RunCampInDir\` + \`.gitmodules\` read-back).

### \`TestProject_RemoteSetURL_SubmoduleKeyIsCampaignRelative\`
- Adds a submodule via \`project add --local\`.
- Runs \`project remote set-url <new-url> --project <name> --no-verify --no-stage\` to isolate the \`.gitmodules\` write.
- Asserts:
  - \`.gitmodules\` header is \`[submodule \"projects/<name>\"]\` both before and after the update.
  - The new URL is written at that campaign-relative key.
  - No \`[submodule \"/...]\` (absolute) or \`[submodule \"\"]\` (empty) sections were introduced.
  - \`git config -f .gitmodules submodule.projects/<name>.url\` returns the new URL directly.

### \`TestProject_RemoteRename_ToOriginUpdatesCampaignRelativeKey\`
- Adds a submodule, attaches an \`upstream\` remote, removes \`origin\`.
- Runs \`project remote rename upstream origin\` — exercises the \`newName == origin\` branch that re-declares \`.gitmodules\` with the new canonical URL.
- Asserts the campaign-relative key is preserved and the new URL lands under it.

## Mutation-tested

Temporarily patched set_url.go to \`submodulePath := resolved.Path\` (absolute) to confirm the guard fires. The first test failed with exactly the expected diff:

\`\`\`
Expected: git@github.com:obedience-corp/renamed-submodule.git
Actual:   /test/remote-seturl-origin
\`\`\`

URL got written under the absolute-path key and git-config lookup under the campaign-relative key returned the old value — exactly the regression the test is guarding against. Reverted.

## Verification

- \`just test all\`: 2140 unit + 147 integration, zero failures (from 2138 + 145 on the base — +2 for the new cases).
- Runs entirely inside the shared alpine container. Nothing touches the host filesystem.

## Scope

This PR only adds tests. The production code under test was already correct on #207; this simply makes the invariant enforceable going forward. If someone touches the remote subcommands and breaks the path semantics, CI will say so.

### DIFF
--- a/tests/integration/project_remote_submodule_test.go
+++ b/tests/integration/project_remote_submodule_test.go
@@ -1,0 +1,143 @@
+//go:build integration
+// +build integration
+
+package integration
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// These tests guard the submodule path handling in the `camp project remote`
+// subcommands. The underlying concern: commits 75948bf and successors
+// switched `submodulePath := strings.TrimPrefix(resolved.Path, campRoot+"/")`
+// to `submodulePath := resolved.LogicalPath` across list/remove/rename/set_url
+// in cmd/camp/project/remote/. Both forms should produce a campaign-relative
+// key (e.g. `projects/my-sub`) — if either path ever drifted toward an
+// absolute path or an empty string, `.gitmodules` would get a broken
+// `[submodule ".../projects/name"]` section and the submodule would become
+// unusable across machines (the whole point of `.gitmodules` is portability).
+//
+// Before this file there was zero integration coverage for `remote set-url`
+// or `remote rename` touching `.gitmodules`. These tests close that gap.
+
+// TestProject_RemoteSetURL_SubmoduleKeyIsCampaignRelative exercises
+// `camp project remote set-url` on a real submodule and asserts the key
+// written into `.gitmodules` is the campaign-relative path (`projects/<name>`)
+// not an absolute path or anything else.
+func TestProject_RemoteSetURL_SubmoduleKeyIsCampaignRelative(t *testing.T) {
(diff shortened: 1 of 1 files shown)


obey-agent (approved):
Verdict: Approve
Risk: low.

Overview:
This PR adds the missing integration coverage for `.gitmodules` key handling in `camp project remote set-url` and `camp project remote rename`. The tests exercise real submodules and verify the canonical `git config -f .gitmodules submodule.projects/<name>.url` path, which is the right regression guard for this area.

Key Findings:
- None.

What's Done Well:
- The tests protect the actual portability risk here: accidentally writing absolute or empty submodule section keys into `.gitmodules`.
- The assertions stay behavior-focused instead of duplicating implementation details.
- Coverage is narrow, useful, and low-noise.

Staff Standard:
Yes. I would be comfortable submitting this test coverage as-is.

PR #227, opened 2026-04-17
follow up linked-project cleanup for #226

lancekrogers:
## What changed
- move campaign registry file access behind a shared `internal/config/registryfile` helper so linked-project detection no longer keeps a parallel raw registry reader
- thread known campaign roots through command execution paths so `cmdutil` and `nav` stop doing redundant uncached detection in hot exec paths
- migrate additional ad-hoc atomic-write sites onto `internal/fsutil.WriteFileAtomically`
- surface linked-project cleanup problems as explicit warnings in CLI results instead of stderr-only noise
- add focused tests for atomic writes and registry-backed linked-project detection

## Why
`camp#207` was mergeable, but it left a small cluster of non-blocking cleanup debt tracked in `#226`: duplicated registry-file access, repeated detection in exec paths, older atomic-write implementations, and linked-project warning surfacing that was easy to miss.

## Impact
- reduces drift risk between config-owned registry persistence and campaign detection
- avoids unnecessary detection work in command-exec paths that already know the active campaign root
- consolidates more persistence behavior on the shared atomic-write helper
- makes linked-project follow-up cleanup failures visible to users and reviewers

## Validation
- `go test ./internal/campaign ./internal/config ./internal/fsutil ./internal/nav ./internal/nav/index ./internal/pins ./internal/project ./internal/workitem/priority ./cmd/camp/cmdutil ./cmd/camp/project ./cmd/camp/project/linked ./cmd/camp`
- `go test ./tests/integration -tags integration -run 'TestProject_(Link_GitRepo|Link_RejectsDuplicateTargetInSameCampaign|Unlink_LinkedProject|Remove_LinkedProjectDeleteBlocked)|TestGo_FromLinkedProjectCwd|TestRun_ProjectDispatch_LinkedProject|TestProjectRun_AutoDetectFromLinkedCwd'`

## Tracking
- Closes #226
- Related: #221

PR #229, opened 2026-04-17
fix: use campaign.EnvCampaignRoot constant in cmdutil

lancekrogers:
## Summary

Stacked on #227. Restores the use of `campaign.EnvCampaignRoot` constant in `cmd/camp/cmdutil/command.go` instead of the hardcoded `"CAMP_ROOT"` string introduced in #227.

## Why

#227 dropped the `internal/campaign` import from `cmdutil` and inlined the env var name as a magic string. Meanwhile `internal/nav/exec.go` still uses `campaign.EnvCampaignRoot+"="+campaignRoot`. Two places setting the same env var, two different ways — they will drift if the constant ever changes.

There is no import cycle: `internal/campaign` imports only `config/registryfile` and `fsutil`. The original code imported `campaign` cleanly.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [ ] Run unit tests in CI


obey-agent (approved):

**Verdict**: Approve

**Overview**: One-line fix restoring the use of `campaign.EnvCampaignRoot` constant in `cmd/camp/cmdutil/command.go`. Closes a real drift hazard I should have caught in the #227 review.

### The Bug This Fixes

After #227:
- `cmd/camp/cmdutil/command.go` hardcodes the env var name as the magic string `"CAMP_ROOT"`
- `internal/nav/exec.go` uses `campaign.EnvCampaignRoot`

Two places set the same env var, two different ways. If the constant ever changes (for example to namespace it under `CAMP_` prefix or version it), the hardcoded string would silently drift. This is exactly the kind of subtle drift that produces "works in one command, silently broken in another" bugs later.

### Verification

- Confirmed via clone+build: `internal/campaign` package dependencies are `internal/config/registryfile` and `internal/fsutil` only. No import cycle when `cmdutil` re-adds the `campaign` import.
- Build, vet, and tests pass on the stack tip.

### Staff Standard

Yes. This is a correct, minimal fix that closes a real drift hazard. Ready to merge.

PR #230, opened 2026-04-17
refactor: collapse nav.ExecInDir to single signature

lancekrogers:
## Summary

Stacked on #229. Eliminates the dead `ExecInDir` wrapper from #227 and consolidates to one canonical signature.

## Why

#227 added `ExecInDirFromRoot` to thread caller-supplied campaign roots through, but kept the original `ExecInDir` as a wrapper that:

1. Still did its own redundant `campaign.Detect` (defeating the stated purpose of the PR)
2. Silently swallowed detect errors with bare `campaignRoot = \"\"` assignment

After #227 all production callers were already migrated to `ExecInDirFromRoot`. Only tests still used the old wrapper.

## What changed

- Deleted the `ExecInDir` wrapper that did redundant `campaign.Detect`
- Renamed `ExecInDirFromRoot` → `ExecInDir` (single canonical name)
- Migrated production callers (`cmd/camp/navigation/go.go`, `internal/nav/exec.go::ExecInCategory`)
- Migrated test callers (`internal/nav/exec_test.go`)
- Callers that don't care about `CAMP_ROOT` pass empty string

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./internal/nav/...` passes

## External callers check (obey-agent flag from #227)

`grep -rn ExecuteCommand` confirmed no external wrappers exist for the new `ExecuteCommand` signature — all 4 production callers already pass `campaignRoot`.


obey-agent (approved):

**Verdict**: Approve

**Overview**: Eliminates dead code left over from the #207→#227 migration and collapses `nav.ExecInDir`/`ExecInDirFromRoot` to a single canonical signature. Catches a subtle bug I missed in the #227 review.

### The Bug This Fixes

After #227, `internal/nav/exec.go` had two functions:

```go
func ExecInDir(ctx, dir, command)           // old — did redundant campaign.Detect
func ExecInDirFromRoot(ctx, dir, root, cmd) // new — threaded caller-supplied root
```

Two problems with the old wrapper:

1. **It still did `campaign.Detect`** — defeating the stated purpose of #227's exec-path cleanup. The whole point of #227 was "don't re-detect in hot paths," and this wrapper quietly undermined that.
2. **It swallowed detect errors silently** with bare `campaignRoot = ""` — no context, no wrap, no logging. Any error from detect would just result in no `CAMP_ROOT` being set, which is indistinguishable from "we're not in a campaign."

I should have flagged this in the #227 review. Good catch by the author.

### Verification

Cloned the stack tip and verified:
- All 3 production `nav.ExecInDir` callers in `cmd/camp/navigation/go.go` pass the new 4-arg signature with `campaignRoot` threaded from `runGo`
- `internal/nav/exec.go::ExecInCategory` migrated correctly
- All 7 test callers in `internal/nav/exec_test.go` pass `""` as empty campaignRoot (intentional — tests don't care about CAMP_ROOT)
- No stragglers with old signature: `grep "ExecInDir\|ExecInDirFromRoot"` found only the new signature and `ExecInDirFromRoot` is completely gone
- `go build ./... && go vet ./... && go test ./internal/nav/...` all pass

### External Callers

PR body mentions the `ExecuteCommand` grep from #227 — I also verified `grep ExecuteCommand` returns only the 4 production callers in `cmd/camp/run.go` (3) and `cmd/camp/project/run.go` (1), all passing the new signature. Clean.

### What's Done Well

- Removing dead code that shipped as "migration tail" from #227 — catching this within a day of merge is the right cadence.
- Single canonical name (`ExecInDir`) is the correct end state; the `FromRoot` suffix existed only as a transition name.
- Test migration is complete — no old-signature callers left behind to rot.

### Staff Standard

Yes. Clean refactor, catches a real bug (silent error swallow), verified end-to-end. Ready to merge.

PR #231, opened 2026-04-17
test: add registry round-trip safety tests

lancekrogers:
## Summary

Stacked on #230. Adds drift-detection tests for the asymmetric registry persistence introduced in #227, plus removes dead code from `LoadRegistry`.

## Why

#227 split the on-disk schema (`registryfile.Campaign`) from the in-memory type (`RegisteredCampaign`). `SaveRegistry` marshals the full `Registry` struct, but `LoadRegistry` reads through `registryfile.File` and manually copies fields. Today the JSON tags happen to match, so nothing breaks. But if anyone adds a new JSON-tagged field to `RegisteredCampaign` without updating `registryfile.Campaign`, that field will be silently dropped on every load.

## What changed

Three new tests in `internal/config/registry_test.go`:

- **`TestRegistry_RoundTrip`** — Save → Load preserves all values across multiple campaigns
- **`TestRegisteredCampaign_AllJSONFieldsPersist`** — reflective enumeration of every JSON-tagged field on `RegisteredCampaign`, asserts each non-zero value is non-zero after round-trip
- **`TestRegistryFile_Schema_StaysAlignedWithRegisteredCampaign`** — extracts the actual JSON keys written to disk by `SaveRegistry`, asserts every one can be read back via `registryfile.File`

Also cleans up `LoadRegistry`:

- Removed the unreachable \"Initialize map if nil\" check (`NewRegistry()` already initializes)
- Removed the duplicate \"Populate IDs from map keys\" loop (the field-copy loop above already sets `ID`)

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/config/... -run Registry` passes


obey-agent (approved):

**Verdict**: Approve

**Overview**: Adds three drift-detection tests for the asymmetric registry persistence introduced in #227, plus removes two genuinely dead code blocks from `LoadRegistry`. This is exactly the kind of defensive instrumentation the #227 architecture needed.

### The Drift Hazard This Guards Against

After #227:
- `SaveRegistry` marshals the full `Registry` struct directly to JSON
- `LoadRegistry` reads through `registryfile.File` and manually copies fields into `RegisteredCampaign`

Today the JSON tags happen to match, so the round-trip works. But if anyone adds a new JSON-tagged field to `RegisteredCampaign` without also adding it to `registryfile.Campaign`, the field will be:
1. Written to disk correctly (SaveRegistry marshals the full struct)
2. **Silently dropped on every subsequent load**

This is the worst kind of data-persistence bug — writes succeed, reads return stale/missing data, and no error is raised. The three new tests are the right way to lock this down.

### Test Analysis

**`TestRegistry_RoundTrip`** — Basic Save→Load round-trip with 2 campaigns covering both campaign types (Product, Research). Handles `time.Time` precision correctly via `LastAccess.Equal()` rather than `==`. Covers the happy path end-to-end.

**`TestRegisteredCampaign_AllJSONFieldsPersist`** — The critical test. Uses reflection to enumerate every JSON-tagged field on `RegisteredCampaign` and asserts each non-zero input value remains non-zero after round-trip. This is the test that fires when someone adds a new field without updating the load path. The `jsonTag == "-"` check correctly excludes `ID` (which is populated from the map key, not serialized). The comma-stripping for `omitempty` handles tagged variants.

**`TestRegistryFile_Schema_StaysAlignedWithRegisteredCampaign`** — The most robust check. Marshals a `RegisteredCampaign` directly to see what JSON keys are written, then performs a full Save→Load cycle and inspects what `registryfile.File` actually reads back. If the write side and the read side ever diverge at the key level, this fails with a pointer to the exact file to update (`internal/config/registryfile/registryfile.go`).

All three tests combine to provide strong coverage:
- Round-trip correctness (test 1)
- Field-level persistence (test 2)
- Schema-level alignment (test 3)

### Dead Code Removal Verified

The two removed blocks:

**"Initialize map if nil"** — I verified this is unreachable. `NewRegistry()` at line 37 initializes `Campaigns` as a non-nil map, and the field-copy loop at lines 39-47 writes to that same map. The old unmarshal path could set the field to nil on empty input; the new field-copy path cannot. Removal is safe.

**"Populate IDs from map keys"** — Verified dead. The field-copy loop at line 41 explicitly sets `ID: id` from the map key already. The trailing loop was a leftover from the pre-#227 direct-unmarshal path where `ID: json:"-"` meant the field needed post-hoc population. After the field-copy refactor, it's duplicate work.

### Validation

- Cloned stack tip, ran `go test ./internal/config/... ./internal/nav/... ./cmd/camp/cmdutil/...` → all pass
- `go build ./...` and `go vet ./...` clean
- Traced the reflection logic manually through `RegisteredCampaign` struct fields to confirm `ID` is correctly excluded and `LastAccess` time round-trip edge case is handled

### What's Done Well

- The reflection-based test is the right abstraction — it guards against *future* field additions without requiring test updates each time a field lands.
- The schema-alignment test points directly at the file to update when it fires (`internal/config/registryfile/registryfile.go`) — future debuggers get a clear signal, not a mysterious test failure.
- Dead code removal is targeted and verified, not speculative.

### Staff Standard

Yes. This is textbook defensive instrumentation for a legitimate drift hazard. Ready to merge.

---

## Stack Summary (#227 → #229 → #230 → #231)

I've now reviewed all four linked-project cleanup PRs. The stack is well-constructed:

| PR | Purpose | Status |
|---|---|---|
| #227 | Four structural follow-ups from #207 consolidation | Approved |
| #229 | Restore `campaign.EnvCampaignRoot` constant (drift fix) | Approved |
| #230 | Collapse `ExecInDir`/`ExecInDirFromRoot` + remove silent-error wrapper | Approved |
| #231 | Drift-detection tests for registry persistence + dead code removal | Approved |

The author responded to my #227 observations with three focused, independently-mergeable PRs. Each is small, verifiable, and addresses a concrete concern. #229 and #230 catch real bugs I missed in #227. Good stack-PR discipline.

obey-agent (commented):
## Re-Review After Commit `47ba0a4c`

The rewrite is a strict improvement. The original tests (which I approved) used `t.TempDir()` + `t.Setenv("XDG_CONFIG_HOME")` + real `SaveRegistry`/`LoadRegistry` calls — which is the exact pattern the project convention prohibits for unit tests. I should have flagged that in my prior review.

### What Changed

- New `loadRegistryFromBytes` helper does what `LoadRegistry` does (registryfile unmarshal + manual field copy) but operates on bytes, not files.
- All three tests (`TestRegistry_RoundTrip`, `TestRegisteredCampaign_AllJSONFieldsPersist`, `TestRegistryFile_Schema_StaysAlignedWithRegisteredCampaign`) now use `json.Marshal` directly.
- Zero filesystem touches in the host unit suite.

### Drift-Detection Still Holds

The critical property of these tests — "if someone adds a new JSON-tagged field to `RegisteredCampaign` without updating `registryfile.Campaign`, the tests fail" — is preserved. The filesystem was incidental to that guarantee; the JSON encode/decode path is what the tests actually care about.

The end-to-end filesystem behavior is covered by existing container-based integration tests (`TestRegister_Campaign`, `TestList_RegisteredCampaigns`) that exercise the real `SaveRegistry`/`LoadRegistry` path.

Prior approval stands, now with cleaner test architecture.

PR #233, opened 2026-04-17
Tag CLI-created shortcuts as user-sourced (fixes #224)

### LINKED ISSUE #224: camp shortcuts add should set source: user
## Context

Discovered while fixing PR #222. `camp shortcuts add <name> <path>` creates shortcuts without setting `Source: "user"` on the `ShortcutConfig`.

**Current behavior** (`cmd/camp/navigation/shortcuts.go:556-560`):
```go
sc := config.ShortcutConfig{
    Path:        shortcutPath,
    Description: description,
    Concept:     concept,
    // Source not set — defaults to empty string
}
```

**Impact**: The repair system (`internal/scaffold/repair.go:441-452`) uses `Source` to distinguish user vs auto shortcuts:
- `source: auto` → auto-generated, can be updated/replaced by repair
- `source: user` → user-created, always preserved
- empty source → "legacy" fallback heuristic (check against defaults)

Shortcuts created via `camp shortcuts add` get the legacy treatment instead of being explicitly tagged as user-defined. This works correctly today because the heuristic falls through to "unknown key = user-defined", but it's semantically wrong and could break if the heuristic logic changes.

## Fix

Set `Source: config.ShortcutSourceUser` in `runShortcutsAddJump` since shortcuts added by the user via CLI are by definition user-sourced.

## Related

- PR #222 (repair integration test fix)

lancekrogers:
## Summary
- `camp shortcuts add` was writing `ShortcutConfig` entries without `Source`, so the repair system (`internal/scaffold/repair.go`) fell through to the legacy default-match heuristic. This works today but is semantically wrong and would break if the heuristic changes.
- Extract `newUserShortcut(path, description, concept)` and use it in `runShortcutsAddJump` so CLI-added shortcuts are explicitly tagged `Source: ShortcutSourceUser`.
- Add unit tests covering the field-set behavior and the downstream `isAutoShortcut()` classification, so a regression fails at the unit level instead of silently drifting through the repair path.

## Context
Discovered while fixing PR #222. Full analysis in issue #224.

## Test plan
- [x] `go test -short ./...` — all packages pass
- [x] `go test -v -run TestNewUserShortcut ./cmd/camp/navigation/...` — new tests (`TestNewUserShortcut_SetsSourceUser` table-driven + `TestNewUserShortcut_NotTreatedAsAuto`) pass
- [x] `just lint` — no new warnings introduced
- [ ] Manual: `camp shortcuts add test-shortcut projects/` and confirm `.campaign/jumps.yaml` shows `source: user` for the new entry

Fixes #224

### DIFF
--- a/cmd/camp/navigation/shortcuts.go
+++ b/cmd/camp/navigation/shortcuts.go
@@ -552,12 +552,9 @@ func runShortcutsAddJump(cmd *cobra.Command, args []string) error {
 		fmt.Printf("%s Adding shortcut '%s'\n", ui.SuccessIcon(), shortcutName)
 	}
 
-	// Create shortcut config
-	sc := config.ShortcutConfig{
-		Path:        shortcutPath,
-		Description: description,
-		Concept:     concept,
-	}
+	// Create shortcut config. Shortcuts added via the CLI are always
+	// user-sourced so the repair system preserves them on reset/diff.
+	sc := newUserShortcut(shortcutPath, description, concept)
 
 	// Add/update the shortcut
 	jumps.Shortcuts[shortcutName] = sc
@@ -577,6 +574,18 @@ func runShortcutsAddJump(cmd *cobra.Command, args []string) error {
 	}
 
 	return nil
+}
+
+// newUserShortcut builds a ShortcutConfig for a shortcut added via the CLI.
+// Source is always ShortcutSourceUser so the repair system preserves it on
+// reset and diff operations (see internal/scaffold/repair.go).
+func newUserShortcut(path, description, concept string) config.ShortcutConfig {
+	return config.ShortcutConfig{
+		Path:        path,
+		Description: description,
+		Concept:     concept,
+		Source:      config.ShortcutSourceUser,
+	}
 }
 
 // isAutoShortcut returns true if the shortcut was auto-generated (not user-defined).
--- a/cmd/camp/navigation/shortcuts_test.go
+++ b/cmd/camp/navigation/shortcuts_test.go
@@ -281,3 +281,70 @@ func TestResetHasFlags(t *testing.T) {
 		}
 	}
 }
+
+// TestNewUserShortcut_SetsSourceUser guards against regression of issue #224.
+// Shortcuts added via the CLI must be tagged ShortcutSourceUser so the repair
+// system preserves them instead of treating them as legacy entries subject to
+// the default-match heuristic.
+func TestNewUserShortcut_SetsSourceUser(t *testing.T) {
+	tests := []struct {
+		name        string
+		path        string
+		description string
+		concept     string
+	}{
+		{
+			name:        "path only",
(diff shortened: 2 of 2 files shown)


obey-agent (approved):
Verdict: Approve

Overview: This is the right-sized fix for #224. It makes CLI-created shortcuts explicit about ownership instead of relying on the legacy fallback heuristic, and the added unit coverage exercises both the field population and the downstream auto-vs-user classification.

Key Findings:
- No staff-level correctness or design issues found.

What's Done Well:
- The behavior change stays tightly scoped to the CLI creation path.
- The helper keeps the intent obvious and avoids repeating the `Source` contract at call sites.
- The tests check the semantic behavior the repair flow actually depends on.

Staff Standard: Yes. I would be comfortable submitting this as-is.

obey-agent (approved):

**Verdict**: Approve

Clean, targeted fix for a real semantic bug. 82 lines, 2 files, precisely scoped.

### The Bug It Fixes

`camp shortcuts add` was constructing `ShortcutConfig` without setting `Source`. The repair system's source-based classification then fell through to the legacy default-match heuristic — which happened to work today because unknown keys coincidentally resolved to "user-defined," but was semantically wrong. A future refactor of the heuristic could start misclassifying user shortcuts as auto-generated and overwriting them on reset/diff.

### The Fix

Extract `newUserShortcut(path, description, concept)` that sets `Source: config.ShortcutSourceUser` explicitly, and use it in `runShortcutsAddJump`. The helper's doc comment names the invariant and the downstream system it protects (`internal/scaffold/repair.go`).

### Test Analysis

Two tests, each covering a different layer of the contract:

**`TestNewUserShortcut_SetsSourceUser`** (table-driven) — verifies `Source = ShortcutSourceUser` AND that the other fields (`Path`, `Description`, `Concept`) round-trip correctly. Three cases: path only, concept only, all fields. Catches regressions in both the source tagging and the field plumbing.

**`TestNewUserShortcut_NotTreatedAsAuto`** — this is the critical test. It verifies the *downstream semantic property*: `isAutoShortcut()` returns false for user-sourced shortcuts across multiple key scenarios including keys that collide with default keys. The test comment explicitly names the property being guarded: "Even if the key happens to collide with a default key, a user-sourced shortcut must not be classified as auto." This is what the repair system actually cares about.

Testing both the data shape and the downstream classification is the right pattern — it catches regressions at both the structural level and the semantic level.

### What's Done Well

- Traces the bug to root cause (repair system's classification heuristic) rather than patching the symptom.
- Names the invariant in a helper function so future callers get the correct shape by default.
- Doc comments on both the helper and the call site explain *why*, not just what.
- Tests verify downstream classification, not just field-set behavior — the test failure mode would be "user shortcut got treated as auto" which is the actual user-facing regression.
- References #224 in context and traces back to PR #222 discovery.

### Staff Standard

Yes. Small, correct, well-tested. Ready to merge.

PR #234, opened 2026-04-17
fix: prune merged detached worktrees during fresh

lancekrogers:
## Summary
- teach prune/fresh to remove detached worktrees whose HEAD is already merged into the fresh sync base
- keep unmerged detached worktrees intact and skip the current checkout/git dir paths
- add unit and integration coverage for merge reachability and detached worktree cleanup behavior

## Root Cause
`camp fresh` only pruned merged local branches and branch-backed worktrees. Detached review worktrees stayed registered even when their commits were already merged into `main`, so repeated post-merge cleanup left stale worktrees behind.

## Impact
After this change, `camp fresh` cleans up the merged detached review worktrees it can prove are already contained in the target sync ref, while preserving detached worktrees that still contain unmerged work.

## Validation
- `go test ./internal/git`
- `go test ./internal/commands/fresh -tags integration`
- `go test ./cmd/camp/project ./cmd/camp`

### DIFF
--- a/internal/git/branches.go
+++ b/internal/git/branches.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os/exec"
 	"strings"
@@ -251,4 +252,31 @@ func DeleteRemoteBranch(ctx context.Context, repoPath, branch string) error {
 		return camperrors.Wrapf(err, "delete remote branch %s: %s", branch, strings.TrimSpace(string(output)))
 	}
 	return nil
+}
+
+// IsAncestor reports whether ancestor is reachable from descendant.
+func IsAncestor(ctx context.Context, repoPath, ancestor, descendant string) (bool, error) {
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+	if strings.TrimSpace(ancestor) == "" {
+		return false, fmt.Errorf("ancestor is required")
+	}
+	if strings.TrimSpace(descendant) == "" {
+		return false, fmt.Errorf("descendant is required")
+	}
+
+	cmd := exec.CommandContext(ctx, "git", "-C", repoPath,
+		"merge-base", "--is-ancestor", ancestor, descendant)
+	err := cmd.Run()
+	if err == nil {
+		return true, nil
+	}
+
+	var exitErr *exec.ExitError
+	if errors.As(err, &exitErr) && exitErr.ExitCode() == 1 {
+		return false, nil
+	}
+
+	return false, camperrors.Wrapf(err, "check %s reachable from %s", ancestor, descendant)
 }
--- a/internal/git/branches.go
+++ b/internal/git/branches.go
@@ -260,10 +260,10 @@ func IsAncestor(ctx context.Context, repoPath, ancestor, descendant string) (boo
 		return false, ctx.Err()
 	}
 	if strings.TrimSpace(ancestor) == "" {
-		return false, fmt.Errorf("ancestor is required")
+		return false, camperrors.Wrap(camperrors.ErrInvalidInput, "ancestor is required")
 	}
 	if strings.TrimSpace(descendant) == "" {
-		return false, fmt.Errorf("descendant is required")
+		return false, camperrors.Wrap(camperrors.ErrInvalidInput, "descendant is required")
 	}
 
 	cmd := exec.CommandContext(ctx, "git", "-C", repoPath,
--- a/internal/prune/prune.go
+++ b/internal/prune/prune.go
@@ -27,6 +27,7 @@ const (
 	StatusWorktreeWouldRemove Status = "wt would remove"
 
 	SkipReasonActiveWorktree SkipReason = "active_worktree"
+	SkipReasonDirtyWorktree  SkipReason = "dirty_worktree"
 )
 
 // Options holds configuration for a prune operation.
@@ -64,26 +65,26 @@ func Execute(ctx context.Context, name, path string, opts Options) ProjectResult
 	pr := ProjectResult{Name: name, Path: path}
 
 	baseRef := strings.TrimSpace(opts.BaseRef)
-	var (
-		merged []string
-		err    error
-	)
 	if baseRef == "" {
-		merged, err = git.MergedBranches(ctx, path)
-	} else {
-		merged, err = git.MergedBranchesFromRef(ctx, path, baseRef)
+		baseRef = git.DefaultBranch(ctx, path)
+		if baseRef == "" {
+			pr.Error = "could not determine default branch"
+			return pr
+		}
 	}
+
+	merged, err := git.MergedBranchesFromRef(ctx, path, baseRef)
 	if err != nil {
 		pr.Error = err.Error()
 		return pr
 	}
 
+	deleteDetachedWorktrees(ctx, path, baseRef, opts, &pr)
 	if len(merged) == 0 && !opts.Remote {
 		return pr
 	}
 
 	deleteLocalBranches(ctx, path, merged, opts, &pr)
-	deleteDetachedWorktrees(ctx, path, baseRef, opts, &pr)
 
 	// Remote branch deletion uses the original merged list intentionally —
 	// remote branches should be cleaned up regardless of local deletion outcome.
@@ -119,7 +120,38 @@ func deleteDetachedWorktrees(ctx context.Context, path, baseRef string, opts Opt
 		}
 
 		merged, err := git.IsAncestor(ctx, path, entry.Commit, baseRef)
-		if err != nil || !merged {
+		if err != nil {
+			pr.Results = append(pr.Results, Result{
+				Branch: formatDetachedWorktreeLabel(entry.Commit),
+				Status: StatusError,
+				Detail: fmt.Sprintf("merge-base check for %s: %s", entry.Path, err),
+			})
+			continue
+		}
+		if !merged {
+			continue
+		}
+
+		clean, err := detachedWorktreeClean(ctx, entry.Path)
+		if err != nil {
+			pr.Results = append(pr.Results, Result{
+				Branch: formatDetachedWorktreeLabel(entry.Commit),
+				Status: StatusError,
+				Detail: fmt.Sprintf("dirty check for %s: %s", entry.Path, err),
+			})
+			continue
+		}
+		if !clean {
+			detail := fmt.Sprintf("dirty detached worktree: %s", entry.Path)
+			if opts.DryRun {
+				detail = fmt.Sprintf("would keep dirty detached worktree: %s", entry.Path)
+			}
+			pr.Results = append(pr.Results, Result{
+				Branch:     formatDetachedWorktreeLabel(entry.Commit),
+				Status:     StatusSkipped,
+				Detail:     detail,
+				SkipReason: SkipReasonDirtyWorktree,
+			})
 			continue
 		}
 
@@ -157,6 +189,14 @@ func formatDetachedWorktreeLabel(commit string) string {
 		commit = commit[:shortSHA]
 	}
 	return "detached@" + commit
+}
+
+func detachedWorktreeClean(ctx context.Context, path string) (bool, error) {
+	hasChanges, err := git.HasChanges(ctx, path)
+	if err != nil {
+		return false, err
+	}
+	return !hasChanges, nil
 }
 
 // detectWorktreesForBranches lists worktrees and returns a map of branch name → worktree entry
--- a/internal/prune/prune.go
+++ b/internal/prune/prune.go
@@ -3,6 +3,7 @@ package prune
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/Obedience-Corp/camp/internal/git"
@@ -82,6 +83,7 @@ func Execute(ctx context.Context, name, path string, opts Options) ProjectResult
 	}
 
 	deleteLocalBranches(ctx, path, merged, opts, &pr)
+	deleteDetachedWorktrees(ctx, path, baseRef, opts, &pr)
 
 	// Remote branch deletion uses the original merged list intentionally —
 	// remote branches should be cleaned up regardless of local deletion outcome.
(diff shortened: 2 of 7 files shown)


internal/prune/prune.go
```diff
@@ -92,6 +94,71 @@ func Execute(ctx context.Context, name, path string, opts Options) ProjectResult
...
+			Detail: entry.Path,
+		}
+
+		if opts.DryRun {
+			result.Status = StatusWorktreeWouldRemove
+			pr.Results = append(pr.Results, result)
+			continue
+		}
+
+		if err := wt.Remove(ctx, entry.Path, true); err != nil {
```
obey-agent, line 174: `git worktree remove --force` will delete a dirty detached worktree, including untracked files. That means a review worktree whose HEAD is merged but still has local WIP gets silently destroyed by `camp fresh`. We need a cleanliness check in `entry.Path` and should only auto-remove clean detached worktrees; otherwise emit a skipped result.

internal/prune/prune.go
```diff
@@ -82,6 +83,7 @@ func Execute(ctx context.Context, name, path string, opts Options) ProjectResult
 	}
 
 	deleteLocalBranches(ctx, path, merged, opts, &pr)
+	deleteDetachedWorktrees(ctx, path, baseRef, opts, &pr)
```
obey-agent: This call never runs in the branch-already-deleted case because `Execute` still returns early on `len(merged) == 0 && !opts.Remote`. That is exactly the stale detached-review-worktree scenario we see after a feature branch has already been pruned locally. Please move detached-worktree cleanup ahead of the early return, or make the return depend on both merged branches and removable detached worktrees.

obey-agent (changes requested):
Verdict: Request Changes

Overview: The direction is good and the ancestry-based cleanup is the right primitive, but the current implementation still has one data-loss path and one control-flow hole. Those both need to be closed before I would trust `fresh` to clean up detached worktrees automatically.

Key Findings:
- `internal/prune/prune.go:137`: force-removing detached worktrees without checking for dirty/untracked state can destroy user work.
- `internal/prune/prune.go:86`: detached cleanup is still skipped whenever there are no merged local branches, so stale detached review worktrees survive once their source branch has already been deleted.

What's Done Well:
- Using reachability instead of branch names is the right way to reason about detached worktrees.
- The new integration coverage for merged vs unmerged detached worktrees is valuable.
- The fresh output change makes the cleanup observable when it does happen.

Staff Standard: Not yet. I would not be comfortable shipping this until the dirty-worktree safety check and the branchless detached-worktree case are both covered.

obey-agent (approved):

**Verdict**: Approve (with one small observation)

Targeted fix that closes a real gap in `camp fresh` — detached review worktrees (typical of PR-review checkouts) weren't being cleaned up even after the underlying branch was merged. 212 lines across 5 files, well-scoped.

### The Gap This Fixes

`camp fresh` pruned merged local branches and branch-backed worktrees, but detached worktrees (`git worktree add --detach`) at commits already merged into the sync base were left registered. Over time that leaves stale review checkouts on disk.

### Architecture

Introduces `git.IsAncestor(ctx, repoPath, ancestor, descendant)` as a primitive in `internal/git/branches.go`. Uses `git merge-base --is-ancestor` with correct exit-code discrimination:
- exit 0 → ancestor reachable from descendant (true)
- exit 1 → not reachable (false)
- any other error → wrapped as actual error

This is the right primitive — it's the fundamental "is X merged into Y" question that prune needs to ask. Reusable for future cleanup work.

The caller (`deleteDetachedWorktrees` in `internal/prune/prune.go`) uses this to identify detached worktrees whose `HEAD` commit is already reachable from the sync base, and removes them via `worktree.Remove(ctx, path, true)`. Skips the current checkout and the git-dir path. Skips `IsLocked` worktrees (respects user intent).

### Preservation Semantics Verified

The integration test `TestIntegration_ExecuteFresh_RemovesMergedDetachedWorktreesOnly` is especially good — it creates BOTH:
- a merged detached worktree (feature-merged → merged into main)
- an unmerged detached worktree (detached with additional commits never merged)

and asserts that only the merged one is removed. This is the critical safety property: `camp fresh` must never delete detached worktrees that still contain unmerged work. The test verifies the directory and the worktree registration both stay.

### Observation (not blocking)

**`entry.Branch != "HEAD (detached)"` is a string-format dependency.** Git's exact output for `git worktree list --porcelain` is stable in practice, but it's a brittle coupling. If `worktree.GitWorktreeEntry` has (or could gain) a typed `Detached bool` field, the check becomes semantic rather than stringly-typed. Not blocking — just noting as future cleanup.

**Silent-continue on `IsAncestor` errors.** The code does `if err != nil || !merged { continue }` — combining "error" and "not merged" into the same branch. If `git merge-base` fails for some unusual reason (e.g. corrupted ref, missing commit), that worktree is silently skipped with no logging. Arguably should `slog.Warn` the error case so operators can debug stuck worktrees. Conservative default is fine though.

### What's Done Well

- Targeted helper in the right package (`internal/git` exposes a generic primitive; `internal/prune` owns the policy).
- `IsLocked` worktrees explicitly preserved.
- Current worktree and git-dir explicitly skipped (prevents deleting the active checkout).
- `git worktree prune` invoked after removals to clean admin directory state — catches cases where `remove` doesn't fully update internals.
- Dry-run status (`StatusWorktreeWouldRemove`) properly separated from live removal (`StatusWorktreeRemoved`).
- Fresh UI detail line shows both branch count and worktree count (`worktree(s)`) so users know what was actually cleaned.
- Integration test exercises real git commands against real worktrees — doesn't mock the primitive.

### Staff Standard

Yes. Well-scoped, well-tested, preserves the right invariant. Ready to merge.

internal/git/branches.go
```diff
@@ -252,3 +253,30 @@ func DeleteRemoteBranch(ctx context.Context, repoPath, branch string) error {
...
 	return nil
 }
+
+// IsAncestor reports whether ancestor is reachable from descendant.
+func IsAncestor(ctx context.Context, repoPath, ancestor, descendant string) (bool, error) {
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+	if strings.TrimSpace(ancestor) == "" {
+		return false, fmt.Errorf("ancestor is required")
```
obey-agent: **Blocking — project error package.**

`fmt.Errorf("ancestor is required")` and `fmt.Errorf("descendant is required")` violate the repo's standing rule that the project error package must be used for all error construction. The rest of this function already uses `camperrors.Wrapf` at line 281, so the divergence is localized.

Change to something like:

```go
if strings.TrimSpace(ancestor) == "" {
    return false, camperrors.New("ancestor is required")
}
if strings.TrimSpace(descendant) == "" {
    return false, camperrors.New("descendant is required")
}
```

(or whichever constructor the rest of the package uses for validation — e.g. `camperrors.Validation(...)`). This is the same rule that's driven the repo-wide `fmt.Errorf` cleanup work you've already landed elsewhere.

internal/prune/prune.go
```diff
@@ -92,6 +95,110 @@ func Execute(ctx context.Context, name, path string, opts Options) ProjectResult
 	return pr
 }
 
+func deleteDetachedWorktrees(ctx context.Context, path, baseRef string, opts Options, pr *ProjectResult) {
+	entries, err := worktree.NewGitWorktree(path).List(ctx)
+	if err != nil {
```
obey-agent, line 101: **Non-blocking — silently-swallowed listing error.**

`worktree.NewGitWorktree(path).List(ctx)` can fail (e.g. git lock held, corrupt `.git/worktrees/`). Today this function just `return`s with no trace, so the fresh output will claim "nothing to prune" even though it couldn't inspect the worktree state.

The rest of this function already uses the `Result{Status: StatusError, Detail: ...}` pattern for IsAncestor / HasChanges failures — same treatment here would keep the observability story consistent:

```go
entries, err := worktree.NewGitWorktree(path).List(ctx)
if err != nil {
    pr.Results = append(pr.Results, Result{
        Status: StatusError,
        Detail: fmt.Sprintf("worktree list: %s", err),
    })
    return
}
```

Not blocking since it preserves the prior behavior for the non-failing case, but worth doing while you're in here.

internal/prune/prune.go
```diff
@@ -92,6 +95,110 @@ func Execute(ctx context.Context, name, path string, opts Options) ProjectResult
...
+		return
+	}
+
+	skipPaths := map[string]struct{}{
+		filepath.Clean(path): {},
+	}
+	if gitDir, err := git.Output(ctx, path, "rev-parse", "--absolute-git-dir"); err == nil {
+		skipPaths[filepath.Clean(gitDir)] = struct{}{}
+	}
+
```
obey-agent, line 116: **Nit — redundant construction.**

`worktree.NewGitWorktree(path)` is called twice in this function: once inline at line 98 for `.List(ctx)` and again at line 110 stored in `wt`. They're the same object; either reuse the second instance for the List call, or call `.List(ctx)` through `wt` — avoids the impression that they're two different views.

internal/prune/prune.go
```diff
@@ -92,6 +95,110 @@ func Execute(ctx context.Context, name, path string, opts Options) ProjectResult
...
+		result.Status = StatusWorktreeRemoved
+		pr.Results = append(pr.Results, result)
+		removedAny = true
+	}
+
+	if !opts.DryRun && removedAny {
+		wt.Prune(ctx, false)
+	}
+}
+
```
obey-agent, line 190: **Non-blocking — discarded Prune error.**

`wt.Prune(ctx, false)` drops its return value. If the gc-style prune fails (concurrent git operation, filesystem perms), the failure silently vanishes. Either:

- Surface it as a `Result{Status: StatusError, Detail: ...}`, or
- Add a one-line comment explaining it's best-effort cleanup so the next reader doesn't mistake it for oversight.

The first is preferable for consistency with the rest of the function's error-surfacing pattern.

obey-agent (commented):

### Overview

The core change is sound: correctly teach `fresh` to prune detached worktrees whose HEAD is an ancestor of the sync base ref, preserve dirty ones, and avoid touching the current checkout or `--absolute-git-dir`. The new `git.IsAncestor` helper properly distinguishes the "not an ancestor" exit-1 case from real command errors, and the test matrix covers the three interesting states (merged+clean → removed, dirty → kept, unmerged → kept). The change does what it says on the tin.

That said, there's one blocking rule violation and several observability/consistency issues worth cleaning up before merge.

### Blocking

1. **`internal/git/branches.go:263,266` — `fmt.Errorf` in a non-test file.** Repo rule is project error package only. See inline comment for the exact change. This is the same convention that drove 05d644b's repo-wide cleanup.

### Non-blocking (strongly recommended while in the area)

2. **Silently-swallowed `List(ctx)` error** at `internal/prune/prune.go:100`. Function returns with no trace; fresh's output will look like a no-op when the real story is a worktree-list failure. Inline comment with suggested shape.
3. **Silently-discarded `wt.Prune(ctx, false)`** at `internal/prune/prune.go:185`. Either surface as a `StatusError` Result or add a one-line "best-effort" comment.
4. **Redundant `worktree.NewGitWorktree(path)` construction** at lines 98 and 110. Reuse one instance.

### Non-blocking (systemic, separate from this PR)

5. **`fresh_integration_test.go` `t.TempDir()` + real git harness.** The new tests (~133 lines) match the file's pre-existing convention (258 lines before this PR) but violate the campaign rule that filesystem-mutating tests run in containerized isolation, not host TempDir. This is a systemic issue in `fresh_integration_test.go`, not introduced by #234 — worth a dedicated cleanup PR that migrates the whole file to the container harness used elsewhere in the repo, not a gate on this one.
6. **Magic string `"HEAD (detached)"` at `prune.go:113`.** Matches `parseWorktreeList`'s convention at `internal/worktree/git.go:257-258`, so it's correct — but an `IsDetached bool` field on `GitWorktreeEntry` would make this coupling explicit and survive future parser refactors. Follow-up.

### What's Done Well

- `git.IsAncestor` correctly distinguishes exit-1 ("not ancestor") from real errors using `errors.As` on `*exec.ExitError`. Many implementations get this wrong.
- The `HasChanges`-based clean check catches untracked files (verified against `internal/git/commit.go:412-434`), so the dirty-protection test isn't vacuous.
- `skipPaths` covers both the current checkout and `--absolute-git-dir`, which prevents the classic foot-gun of pruning the repo's own worktree.
- Error results carry the worktree path in `Detail`, so operators can actually locate what went wrong — good observability hygiene.
- Test triples (merged+clean → removed, merged+dirty → kept, unmerged → kept) give the reviewer confidence the intent is captured.

### Staff Standard

Close. The implementation reasoning is staff-level — correct exit-code handling, good invariant coverage, tests that encode intent rather than implementation. One blocking rule violation to fix (`fmt.Errorf`) and two small observability holes. Address those and this is ready to merge.

internal/prune/prune.go
obey-agent: Already addressed on `d28dc5f`. Detached worktrees are now checked for local changes before any forced removal, and dirty ones are reported as skipped instead of being deleted.

internal/prune/prune.go
obey-agent: Already addressed on `d28dc5f`. Detached-worktree cleanup now runs before the early return, so branchless stale review worktrees are still considered even when there are no merged local branches left.

internal/git/branches.go
obey-agent: Addressed in `a3d2abd`. `IsAncestor` now uses the project error package for empty-ref validation, and the regression test covers both missing-ancestor and missing-descendant inputs.

internal/prune/prune.go
obey-agent: Addressed in `a3d2abd`. `deleteDetachedWorktrees` now reuses the same `GitWorktree` instance for listing and mutation, and a `worktree list` failure is surfaced as a `StatusError` result instead of being silently swallowed.

internal/prune/prune.go
obey-agent: Addressed in `a3d2abd`. The extra `NewGitWorktree` construction is gone; the function now creates one instance up front and uses it for both `List` and `Remove`/`Prune`.

internal/prune/prune.go
obey-agent: Addressed in `a3d2abd`. `worktree prune` failures are now surfaced through `StatusError` results after both detached-worktree cleanup and branch-backed worktree removal. The host-filesystem test cleanup is tracked separately in issue #238.

obey-agent:
Follow-up on the latest review body: `09f49df` adds a typed `GitWorktreeEntry.IsDetached` flag and switches detached-worktree pruning over to that semantic field instead of the raw `"HEAD (detached)"` string check.

The earlier observability concern about silently continuing on `IsAncestor` errors is already handled on the current branch head via the `StatusError` reporting added in `a3d2abd`. The remaining host-filesystem integration-test cleanup stays tracked separately in #238.

PR #235, opened 2026-04-17
Tolerate duplicate scaffold files during intent migration

lancekrogers:
## Problem

`camp init --repair` fails on campaigns scaffolded before `f476a19` (Mar 16, 2026 — the commit that moved the canonical intent root from `workflow/intents/` to `.campaign/intents/`):

```
Error: failed to compute repair plan: destination already exists for
  .../.campaign/intents/dungeon/.crawl.yaml: intent migration conflict: conflict
```

Hit this on CC_GO_PLANNING (originally scaffolded Mar 5).

## Root Cause

Repair's phase ordering is self-conflicting:

- **Phase 2** writes canonical intent scaffold (including `.campaign/intents/dungeon/.crawl.yaml`)
- **Phase 8** plans the legacy `workflow/intents/` → `.campaign/intents/` migration and refuses any conflicts

Migration's conflict check at `migration_filesystem.go:148` already recognized `.gitkeep` as safe-to-skip when the destination exists. But `.crawl.yaml` — also scaffold-generated, also listed in `legacyIntentScaffoldFiles` for post-migration cleanup — had no such exception. Migration thus rejects its own idempotent output.

## Fix

Replace the hard-coded `.gitkeep` check with `isIntentScaffoldBasename()`, derived from the existing `legacyIntentScaffoldFiles` list. Any filename already declared as scaffold (today: `.gitkeep`, `.crawl.yaml`) is safe-to-skip when the destination exists.

Applied in both:
- `collectIntentTreeMoves` — planning path (`PlanLegacyIntentRootMigration`)
- `moveIntentTree` — execution path (`migrateLegacyIntentRoot` → `EnsureDirectories`)

So plan and apply stay consistent, and adding a new scaffold file to the list in the future extends the tolerance automatically.

## Blast Radius

Affects only campaigns scaffolded before `f476a19` that run `camp init --repair` after it. New campaigns never hit this path.

## Test plan
- [x] `go test -short ./internal/intent/... ./internal/scaffold/...` — all pass
- [x] `go test -short ./...` — no regressions
- [x] New tests exercise both the planning path (`TestIntentService_PlanLegacyIntentRootMigration_SkipsDuplicateCrawlConfig`), the execution path (`TestIntentService_EnsureDirectories_DuplicateCrawlConfigTolerated`), and the helper (`TestIsIntentScaffoldBasename`)
- [x] Manually verified on CC_GO_PLANNING: `camp init --repair` now completes after removing the legacy duplicate — this PR makes that manual step unnecessary for the next affected campaign

### DIFF
--- a/internal/intent/migration_filesystem.go
+++ b/internal/intent/migration_filesystem.go
@@ -146,7 +146,7 @@ func collectIntentTreeMoves(srcDir, dstDir string, moves *[]PlannedPathMove) err
 		}
 
 		if _, err := os.Stat(dstPath); err == nil {
-			if entry.Name() == ".gitkeep" {
+			if isIntentScaffoldBasename(entry.Name()) {
 				continue
 			}
 			return camperrors.Wrapf(ErrIntentMigrationConflict, "destination already exists for %s", dstPath)
@@ -263,7 +263,7 @@ func moveIntentTree(srcDir, dstDir string) error {
 		}
 
 		if _, err := os.Stat(dstPath); err == nil {
-			if entry.Name() == ".gitkeep" {
+			if isIntentScaffoldBasename(entry.Name()) {
 				if err := os.Remove(srcPath); err != nil {
 					return camperrors.Wrapf(err, "removing %s", srcPath)
 				}
--- a/internal/intent/migration_filesystem.go
+++ b/internal/intent/migration_filesystem.go
@@ -147,7 +147,13 @@ func collectIntentTreeMoves(srcDir, dstDir string, moves *[]PlannedPathMove) err
 
 		if _, err := os.Stat(dstPath); err == nil {
 			if isIntentScaffoldBasename(entry.Name()) {
-				continue
+				matches, matchErr := scaffoldFilesMatch(srcPath, dstPath)
+				if matchErr != nil {
+					return matchErr
+				}
+				if matches {
+					continue
+				}
 			}
 			return camperrors.Wrapf(ErrIntentMigrationConflict, "destination already exists for %s", dstPath)
 		} else if !os.IsNotExist(err) {
@@ -264,10 +270,16 @@ func moveIntentTree(srcDir, dstDir string) error {
 
 		if _, err := os.Stat(dstPath); err == nil {
 			if isIntentScaffoldBasename(entry.Name()) {
-				if err := os.Remove(srcPath); err != nil {
-					return camperrors.Wrapf(err, "removing %s", srcPath)
+				matches, matchErr := scaffoldFilesMatch(srcPath, dstPath)
+				if matchErr != nil {
+					return matchErr
 				}
-				continue
+				if matches {
+					if err := os.Remove(srcPath); err != nil {
+						return camperrors.Wrapf(err, "removing %s", srcPath)
+					}
+					continue
+				}
 			}
 			return camperrors.Wrapf(ErrIntentMigrationConflict, "destination already exists for %s", dstPath)
 		} else if !os.IsNotExist(err) {
@@ -284,4 +296,18 @@ func moveIntentTree(srcDir, dstDir string) error {
 	}
 
 	return nil
+}
+
+func scaffoldFilesMatch(srcPath, dstPath string) (bool, error) {
+	srcData, err := os.ReadFile(srcPath)
+	if err != nil {
+		return false, camperrors.Wrapf(err, "reading %s", srcPath)
+	}
+
+	dstData, err := os.ReadFile(dstPath)
+	if err != nil {
+		return false, camperrors.Wrapf(err, "reading %s", dstPath)
+	}
+
+	return bytes.Equal(srcData, dstData), nil
 }
--- a/internal/intent/service_test.go
+++ b/internal/intent/service_test.go
@@ -1522,6 +1522,69 @@ func TestIntentService_PlanLegacyIntentRootMigration_SkipsDuplicateCrawlConfig(t
 	}
 }
 
+// Divergent scaffold content is treated as a real conflict rather than being
+// silently discarded. This preserves any user edits to legacy .crawl.yaml.
+func TestIntentService_EnsureDirectories_DivergentCrawlConfigConflicts(t *testing.T) {
+	campaignRoot := t.TempDir()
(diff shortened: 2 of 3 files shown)


internal/intent/migration_filesystem.go
```diff
@@ -146,7 +146,7 @@ func collectIntentTreeMoves(srcDir, dstDir string, moves *[]PlannedPathMove) err
 		}
 
 		if _, err := os.Stat(dstPath); err == nil {
-			if entry.Name() == ".gitkeep" {
+			if isIntentScaffoldBasename(entry.Name()) {
```
lancekrogers, line 149: **Observation — divergent-content edge case.**

This change silently skips the legacy scaffold file when the destination exists. Works perfectly for the documented scenario (byte-identical scaffold content). But consider: a user edited `.crawl.yaml` in the legacy root (e.g. added a custom `excludes:` entry before the f476a19 canonical switch). Today's migration would error out, preserving the divergent legacy copy. After this change, the legacy `.crawl.yaml` is silently discarded via the `moveIntentTree` removal branch.

For `.gitkeep` this is harmless. For `.crawl.yaml`, there's a narrow data-loss window. Options:

1. **Accept it** — scaffold files are campaign-managed, not user-managed. Document the behavior. This is my lean.
2. **Compare content before skipping** — skip only if `bytes.Equal`, error otherwise. Safer, small cost. See `moveIntentMarkerFile` at `migration_filesystem.go:77-86` which already does this for OBEY.md.
3. **Keep legacy on divergence** — warn and move on, like a 3-way merge fallback.

Don't think this blocks merge given the scaffold's semantics, but worth calling out so it's an explicit decision rather than accidental. A test with divergent content would pin the chosen behavior in place for the next reader.

internal/intent/service_test.go
```diff
@@ -1431,6 +1431,117 @@ func TestIntentService_PlanLegacyIntentRootCleanup(t *testing.T) {
...
 }
 
+// TestIntentService_EnsureDirectories_DuplicateCrawlConfigTolerated guards the
+// regression where `camp init --repair` would fail on campaigns scaffolded
+// before the canonical-intent-root migration: Phase 2 of repair writes a
+// fresh .crawl.yaml at the canonical path, then Phase 8 migration refused to
+// reconcile the identical legacy copy. Scaffold files listed in
+// legacyIntentScaffoldFiles (currently .gitkeep and .crawl.yaml) must be
+// treated as safe-to-skip when the destination already exists, not just
+// .gitkeep.
```
lancekrogers, line 1441: **Non-blocking — systemic `t.TempDir()` + filesystem-mutation pattern.**

The three new tests add to the existing `service_test.go` convention of real-FS migration tests (`TestIntentService_EnsureDirectories_MigratesLegacyRootAndAudit` and siblings). That pattern violates campaign rule §11 (filesystem-mutating tests must run in containerized isolation). Same situation as PR #234's `fresh_integration_test.go`.

Not introduced by this PR, and containerizing the whole file is out of scope here. Worth a separate cleanup PR that migrates `service_test.go`'s migration tests to the container harness — not a gate on merging #235.

obey-agent (commented):

### Overview

Targeted bug fix for the `camp init --repair` self-conflict on legacy campaigns. The root-cause write-up in the PR description correctly identifies repair's phase-ordering hazard (Phase 2 scaffold writes the canonical `.crawl.yaml`, Phase 8 migration then treats it as a conflict) and applies the right layer of fix: migration becomes tolerant of its own scaffold output rather than forcing a refactor of repair's phase ordering.

### Implementation Review

- `intentScaffoldBasenames` is derived from the existing `legacyIntentScaffoldFiles` list via `filepath.Base`, so the source of truth stays single. Future scaffold additions propagate automatically — this is the right level of indirection.
- Fix applied in both `collectIntentTreeMoves` (planning) and `moveIntentTree` (execution), so plan and apply stay consistent. The execution path correctly removes the legacy copy after skipping; the planning path correctly omits the move (cleanup is handled by the existing `PlanLegacyIntentRootCleanup` path).
- Helper is a pure function with no hidden state, clean to test.

### Observations (non-blocking, see inline)

1. **Divergent-content edge case** (`migration_filesystem.go:149`). For byte-identical scaffold files this is obviously correct. For user-edited `.crawl.yaml` (e.g. custom `excludes:` entries written before f476a19), the legacy copy is silently discarded via the `moveIntentTree` removal branch. Worth an explicit decision + a test pinning the behavior. Three options in the inline comment. Lean: accept as-is, document, add divergent-content test.
2. **Test-harness pattern violation** (`service_test.go:1441`). New tests match the file's existing `t.TempDir()` convention, which already violates the campaign rule that filesystem-mutating tests run in containerized isolation. Not introduced here — follow-up cleanup PR.

### What's Done Well

- The commit message traces the self-conflict to a specific prior commit (`f476a19`) and explains who's affected (campaigns scaffolded before Mar 16, 2026). Future readers debugging this area will thank you.
- Three tests cover the three distinct responsibilities — helper (unit), planning path, execution path — and each has assertions that would actually catch a regression (e.g. the "don't find .crawl.yaml in the planned moves" check is direct, not a side-effect assertion).
- The `isIntentScaffoldBasename` name is boring and descriptive in the best possible way — anyone scanning this code knows exactly what it does.
- Fix preserves the existing conflict-error for true user-data conflicts (intent files, OBEY.md divergence is still gated by `moveIntentMarkerFile`'s content-equality check). Only the scaffold list gets the relaxed treatment.

### Staff Standard

Yes. Fix is minimal, rule-following, targeted at a real user hazard with clear root-cause attribution. The divergent-content question is worth answering explicitly before merge but doesn't gate approval. Ready to ship after that decision is captured in a comment or test.

internal/intent/migration_filesystem.go
obey-agent: Addressed in `7865df6`. Scaffold-file duplicates are now skipped only when source and destination contents are byte-identical; divergent `.crawl.yaml` content preserves the prior conflict path instead of being discarded. Added regression coverage for both the planning and execution paths.

internal/intent/service_test.go
obey-agent: Tracked separately in #237. PR #235 stays scoped to the duplicate scaffold migration fix; the broader containerization cleanup is follow-up work.

PR #239, opened 2026-04-17
test: containerize intent migration filesystem coverage

obey-agent:
## Summary
- move the legacy intent migration coverage into the containerized integration harness
- cover full legacy-root migration, marker-only migration, and legacy scaffold cleanup through command paths
- drop the overlapping host filesystem-mutation tests from `internal/intent/service_test.go`

## Why
Issue #237 tracks the remaining intent migration tests that still mutate the host filesystem directly with `t.TempDir()`. This keeps the migration behavior covered while shifting the real filesystem mutation into the shared container harness.

## Validation
- `go test ./internal/intent`
- `go test -tags=integration ./tests/integration -run 'TestIntentList_MigratesLegacyIntentRootAndAudit|TestIntentList_MigratesLegacyMarkerWithoutIntentState|TestInitRepair_RemovesLegacyIntentScaffoldResidue' -count=1`

Closes #237

PR #240, opened 2026-04-17
test: containerize fresh integration filesystem coverage

### LINKED ISSUE #238: Containerize fresh integration filesystem tests
## Summary
`internal/commands/fresh/fresh_integration_test.go` still uses host `t.TempDir()` with real git/worktree filesystem mutation. That predates the detached-worktree cleanup fix in PR #234, but the file still violates the campaign rule that filesystem-mutating tests should run in containerized isolation.

## Follow-up
- move the `fresh` integration filesystem cases into the containerized integration harness
- preserve coverage for merged, dirty, and unmerged detached worktrees
- narrow or remove the host-side real-filesystem tests once the containerized replacements exist

Tracked separately so PR #234 can stay scoped to the detached-worktree cleanup behavior.

obey-agent:
## Summary
- move the `fresh` filesystem-mutation coverage into the containerized integration harness
- cover branch creation, existing-branch handling, detached-worktree cleanup, nested submodule drift, and remote-branch retention through `camp fresh` command paths
- remove the host-side `internal/commands/fresh/fresh_integration_test.go` file

## Why
Issue #238 tracks the remaining `fresh` integration tests that still use host `t.TempDir()` with real git/worktree mutation. This keeps the real `camp fresh` behavior covered while moving the filesystem mutation into the shared container harness.

## Validation
- `go test ./internal/commands/fresh`
- `go test -tags=integration ./tests/integration -run 'TestFresh_' -count=1`

Closes #238


tests/integration/fresh_test.go
```diff
@@ -0,0 +1,348 @@
+//go:build integration
```
obey-agent, line 1: **Floating image tag — non-deterministic container environment**

The shared harness in `helpers.go` pulls `alpine:latest`, which is already present here rather than introduced by this PR. However this PR adds 7 new test scenarios that all depend on that tag. `alpine:latest` resolves to a different layer on every runner cache miss, meaning the git version inside the container can silently change between CI runs and locally. Pin to a digest (e.g. `alpine:3.20@sha256:...`) so the environment is reproducible. This is especially important for worktree and submodule tests where git behaviour differs across minor versions.

This is an existing harness issue surfaced by the expanded coverage — worth fixing before this series lands.

tests/integration/fresh_test.go
```diff
@@ -0,0 +1,348 @@
...
+set -e
+git -C %[1]s checkout -b feature-merged
+printf 'feature\n' > %[1]s/feature.txt
+git -C %[1]s add .
+git -C %[1]s commit -m 'Feature work'
+`, projectDir))
+
+	mainWorktree := "/test/fresh-default-elsewhere-main"
+	stableWorktree := "/test/fresh-default-elsewhere-stable"
+	mergedSiblingWorktree := "/test/fresh-default-elsewhere-sidecar"
```
obey-agent: **Hardcoded worktree paths create cross-test contamination risk**

Tests in `TestFresh_HandlesDefaultBranchInAnotherWorktree`, `TestFresh_RemovesOnlyMergedDetachedWorktrees`, etc. use fixed container paths (`/test/fresh-default-elsewhere-main`, `/test/fresh-detached-main`, etc.). As long as tests run sequentially with `Reset()` between them this is safe, but:

1. Two tests sharing a fixture name (now or in a future test) will collide silently.
2. The defensive `rm -rf` guards inside setup shell scripts paper over the issue rather than preventing it structurally.

Recommend deriving worktree paths from the `name` parameter already threaded through setup helpers (e.g. `/test/` + name + `-` + role) so uniqueness is guaranteed mechanically and the manual cleanup guards can be removed.

tests/integration/fresh_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
```
obey-agent, line 13: **`runContainerShell` helper duplicates existing `ExecCommand` pattern — promote to `helpers.go`**

`runContainerShell` and `gitOutput` are defined at the top of this file. `ExecCommand` already exists on `TestContainer` in `helpers.go`. The new helpers are free functions that take `*TestContainer` as a parameter, bypassing the method set. If another test file needs a similar shell-runner helper (PR #239's intent migration tests run similar shell blocks inline), this pattern won't be reused.

These two helpers belong in `helpers.go` as methods on `TestContainer` (`tc.Shell(t, script)` and `tc.GitOutput(t, dir, args...)`) so they're available across the entire integration suite without copy-paste. The fresh tests would then call `tc.Shell` and `tc.GitOutput` consistently with the rest of the harness API.

obey-agent (commented):
## Review: test: containerize fresh integration filesystem coverage

### Verdict: Request Changes

The migration goal is correct and well-scoped. The old host-side `fresh_integration_test.go` (457 lines using `t.TempDir()` and real git) is fully removed and replaced with 348 lines running inside the shared container harness. All 7 original test scenarios are ported with equivalent coverage. The PR does exactly what it says.

**Top 3 findings:**

1. **`runContainerShell` / `gitOutput` are free functions instead of `TestContainer` methods** (line 14–28). These helpers are file-local and not available to other test files in the package. PR #239 (intent migration) and any future test files will hand-roll their own shell invocation patterns. Both belong in `helpers.go` as `tc.Shell(t, script)` and `tc.GitOutput(t, dir, args...)` — the existing `ExecCommand` method is the right base.

2. **Hardcoded worktree paths inside tests** (e.g. `/test/fresh-default-elsewhere-main`, `/test/fresh-detached-main`). The `Reset()` between tests prevents contamination today, but these fixed strings create silent future collision risk and force manual `rm -rf` guards inside setup scripts. Derive worktree paths from the `name` parameter already threaded through setup helpers (`/test/` + name + `-` + role) so uniqueness is structural, not coincidental.

3. **Shared harness uses `alpine:latest` (floating tag)** — not introduced by this PR but all 7 new worktree/submodule tests amplify the risk. A silent git version bump between CI runs could cause intermittent failures in these scenarios. Pin to a digest before this series lands.

**Overlap with #239:** No duplication of harness infrastructure — both PRs extend the same shared `helpers.go` / `main_test.go` harness, which is correct. The issue is that fresh_test.go introduces local helpers (`runContainerShell`, `gitOutput`) instead of contributing them back to the shared harness where #239 and other callers would benefit.

**Missing coverage:** No test for `camp fresh` when the container has no git binary (startup failure path), and no `testing.Short()` skip guard. Given that each test calls `Reset()` (full container wipe + reinit), the suite will be slow on CI — a `t.Skip` on short mode is advisable.

**Mechanics are solid:** containerization actually isolates from host FS, cleanup is handled by `AutoRemove: true` + `Cleanup()` in `TestMain`, context is propagated throughout the harness, and tests check both positive and negative exit codes properly.

obey-agent (commented):
## Review: test: containerize fresh integration filesystem coverage

### Verdict: Request Changes

The migration goal is correct and well-scoped. The old host-side `fresh_integration_test.go` (457 lines using `t.TempDir()` and real git) is fully removed and replaced with 348 lines running inside the shared container harness. All 7 original test scenarios are ported with equivalent coverage. The PR does exactly what it says.

**Top 3 findings:**

1. **`runContainerShell` / `gitOutput` are free functions instead of `TestContainer` methods** (line 14–28). These helpers are file-local and not available to other test files in the package. PR #239 (intent migration) and any future test files will hand-roll their own shell invocation patterns. Both belong in `helpers.go` as `tc.Shell(t, script)` and `tc.GitOutput(t, dir, args...)` — the existing `ExecCommand` method is the right base.

2. **Hardcoded worktree paths inside tests** (e.g. `/test/fresh-default-elsewhere-main`, `/test/fresh-detached-main`). The `Reset()` between tests prevents contamination today, but these fixed strings create silent future collision risk and force manual `rm -rf` guards inside setup scripts. Derive worktree paths from the `name` parameter already threaded through setup helpers (`/test/` + name + `-` + role) so uniqueness is structural, not coincidental.

3. **Shared harness uses `alpine:latest` (floating tag)** — not introduced by this PR but all 7 new worktree/submodule tests amplify the risk. A silent git version bump between CI runs could cause intermittent failures in these scenarios. Pin to a digest before this series lands.

**Overlap with #239:** No duplication of harness infrastructure — both PRs extend the same shared `helpers.go` / `main_test.go` harness, which is correct. The issue is that fresh_test.go introduces local helpers (`runContainerShell`, `gitOutput`) instead of contributing them back to the shared harness where #239 and other callers would benefit.

**Missing coverage:** No test for `camp fresh` when the container has no git binary (startup failure path), and no `testing.Short()` skip guard. Given that each test calls `Reset()` (full container wipe + reinit), the suite will be slow on CI — a `t.Skip` on short mode is advisable.

**Mechanics are solid:** containerization actually isolates from host FS, cleanup is handled by `AutoRemove: true` + `Cleanup()` in `TestMain`, context is propagated throughout the harness, and tests check both positive and negative exit codes properly.

obey-agent:
## Review: test: containerize fresh integration filesystem coverage

### Verdict: Request Changes

The migration goal is correct and well-scoped. The old host-side `fresh_integration_test.go` (457 lines using `t.TempDir()` and real git) is fully removed and replaced with 348 lines running inside the shared container harness. All 7 original test scenarios are ported with equivalent coverage. The PR does exactly what it says.

**Top 3 findings:**

1. **`runContainerShell` / `gitOutput` are free functions instead of `TestContainer` methods** (lines 14–28). These helpers are file-local and unavailable to other test files in the package. PR #239 (intent migration) and any future test files will hand-roll their own shell invocation patterns. Both belong in `helpers.go` as `tc.Shell(t, script)` and `tc.GitOutput(t, dir, args...)` — the existing `ExecCommand` method is the right base.

2. **Hardcoded worktree paths inside tests** (e.g. `/test/fresh-default-elsewhere-main`). The `Reset()` between tests prevents contamination today, but fixed strings create silent future collision risk and force manual `rm -rf` guards in setup scripts. Derive worktree paths from the `name` parameter already threaded through setup helpers so uniqueness is structural.

3. **Shared harness uses `alpine:latest` (floating tag)** — not introduced here but all 7 new worktree/submodule tests amplify the risk. A silent git version bump between CI runs could cause intermittent failures in worktree/submodule tests. Pin to a digest before this series lands.

**Overlap with #239:** No harness duplication — both PRs extend the same shared `helpers.go` / `main_test.go`, which is correct. The issue is that `fresh_test.go` introduces local helpers that should be promoted to the shared harness for reuse.

**Missing coverage:** No `testing.Short()` skip guard. Each test calls `Reset()` (full container wipe), so the suite will be slow on CI without it.

**Mechanics are solid:** FS isolation is real (no host bind mounts), cleanup is handled by `AutoRemove: true` + `Cleanup()` in `TestMain`, and tests check both positive and negative exit codes correctly.

tests/integration/fresh_test.go
```diff
@@ -0,0 +1,348 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, 0, exitCode, "stable worktree branch should remain")
+
+	worktrees := gitOutput(t, tc, projectDir, "worktree", "list", "--porcelain")
+	assert.Contains(t, worktrees, mainWorktree)
+	assert.Contains(t, worktrees, stableWorktree)
+	assert.NotContains(t, worktrees, mergedSiblingWorktree)
+}
+
+func TestFresh_RemovesOnlyMergedDetachedWorktrees(t *testing.T) {
```
obey-agent, line 217: Coverage regression: the old host-side suite had a distinct `TestIntegration_ExecuteFresh_RemovesMergedDetachedWorktreesAfterBranchDeleted` case that asserted pruning still removes a merged detached worktree after the source branch (e.g. `feature-merged`) has been deleted with `git branch -d`. This exercises a different code path in the pruner (branch-lookup must not be required to classify the worktree as merged). The containerized suite does not cover this scenario — `TestFresh_RemovesOnlyMergedDetachedWorktrees` keeps `feature-merged` alive throughout. Please port that case so the pruner's merged-detection-after-branch-deletion path stays covered.

obey-agent (commented):
## Verdict: Request Changes

## Overview
The PR cleanly moves `fresh`'s filesystem-mutating integration coverage out of host `t.TempDir()` and into the shared container harness — directly satisfying the project rule that workflows exercising real git/worktree mutation must run in containerized isolation. The CLI-level invocation (`RunCampInDir` with real `camp fresh ...` flags) is actually stronger coverage than the old direct `executeFresh()` calls. However, one distinct scenario from the old suite was not ported, leaving a real regression in pruner coverage.

## Key Findings
- Missing coverage: `TestIntegration_ExecuteFresh_RemovesMergedDetachedWorktreesAfterBranchDeleted` was not carried over. It asserted that a merged detached worktree is still pruned after the source branch has been deleted (`git branch -d feature-merged`) — a distinct code path in the worktree pruner from `TestFresh_RemovesOnlyMergedDetachedWorktrees`, which keeps the branch alive. Please port it.
- Old suite had 8 integration tests; new suite has 7. The delta is exactly the branch-deleted case above.

## What's Done Well
- Complete removal of host `t.TempDir()` + real git mutation for `fresh` (no leftovers in `internal/commands/fresh`; `settings_fresh_test.go` is pure config-parse and correctly stays host-side).
- Shared container + `Reset()` between tests gives clean isolation; unique per-test names (`fresh-create-push`, `fresh-existing-branch`, ...) avoid cross-test collisions.
- Setup helpers (`setupFreshCampaignWithSubmodule`, `setupFreshCampaignWithNestedSubmoduleProject`) are tight and reusable; `runContainerShell` and `gitOutput` keep the tests readable.
- Exercises the real CLI surface (flags, wiring) rather than the internal `executeFresh()` function — catches regressions the old suite would miss.
- `GIT_ALLOW_PROTOCOL=file` used correctly for file:// submodule operations under modern git defaults.
- File is 348 lines; structure is clear; assertions carry diagnostic context (output strings on failure).

## Staff Standard
Meets the containerization bar and improves coverage fidelity in most dimensions, but ships with a net loss of one real test scenario versus the file it replaces. Fix the missing detached-after-branch-deleted case and this is a clean merge.

tests/integration/fresh_test.go
lancekrogers: Fixed in 3592a1ae3f74b8a61a63d5e4fd73978edec81ab6. Pinned the shared harness image in `helpers.go` to `alpine:3.21@sha256:f27cad91...` (multi-arch manifest list digest so it works across runner architectures). Left a comment explaining the why and the bump procedure. Verified the suite still passes against the pinned image.

tests/integration/fresh_test.go
lancekrogers: Fixed in 3592a1ae3f74b8a61a63d5e4fd73978edec81ab6. Added `worktreePath(name, role)` helper that derives paths from the per-test `name` constant, and removed all the defensive `rm -rf` guards from setup shell blocks — uniqueness is now structural. Each test declares its name once (`const name = "fresh-..."`) and every worktree in that test reads from it. A future test picking a colliding role string would fail cleanly on worktree-add rather than silently reusing a stale directory.

tests/integration/fresh_test.go
lancekrogers: Fixed in 3592a1ae3f74b8a61a63d5e4fd73978edec81ab6. Moved `runContainerShell` and `gitOutput` to `helpers.go` as `tc.Shell(t, script)` and `tc.GitOutput(t, dir, args...)` methods on `TestContainer`. They now use `tc.ExecCommand` under the hood exactly as you suggested. All of `fresh_test.go` switched over; the free functions are gone. PR #239 and future test files get the shared API without copy-paste.

tests/integration/fresh_test.go
lancekrogers: Fixed in 3592a1ae3f74b8a61a63d5e4fd73978edec81ab6. Ported the scenario as `TestFresh_RemovesMergedDetachedWorktreeAfterBranchDeleted`. The test creates the `feature-merged` branch, forks a detached worktree at it, checks out a scratch branch so `git branch -d feature-merged` succeeds, merges + pushes on main, deletes the source branch, then runs `camp fresh`. Asserts the detached worktree is pruned — exercising the commit-ancestry classification path rather than any branch-name lookup. Verified passing in the container suite (2.50s runtime). Also added a comment in the test explaining why this case is distinct from `TestFresh_RemovesOnlyMergedDetachedWorktrees` so the next reader knows what would regress without it.

obey-agent (commented):
## Verdict: Comment

## Overview
Re-reviewed at commit `3318c2a`. The prior blocking findings are all addressed: the shared shell/git helpers now live in `helpers.go`, worktree paths are derived structurally from each test's name, the shared container image is pinned to a digest, and the missing detached-worktree-after-branch-deletion scenario is back under test. I did not find any remaining correctness, coverage, or maintainability issues in the current diff.

## Key Findings
- None.
- GitHub does not allow `obey-agent` to approve this PR because `obey-agent` is the PR author, so this re-review is recorded as a comment-only review.

## Validation
- `go test ./internal/commands/fresh` (expected `[no test files]` now that the host-side suite is removed)
- `go test -tags=integration ./tests/integration -run 'TestFresh_' -count=1`

## What's Done Well
- The migration now preserves the meaningful `fresh` filesystem-mutation coverage while enforcing the project's containerized-isolation rule.
- `skipIfShort`, the pinned harness image, and shared `tc.Shell` / `tc.GitOutput` helpers improve the broader integration harness rather than solving this file in isolation.
- `TestFresh_RemovesMergedDetachedWorktreeAfterBranchDeleted` restores a real pruner edge case that would otherwise regress silently.

## Staff Standard

PR #243, opened 2026-04-18
[codex] fix leverage recent history snapshot persistence

lancekrogers:
## What changed
- persist current leverage snapshots during normal `camp leverage` runs
- reuse one HEAD commit lookup per git repo when saving those snapshots
- show an explicit note when recent 7/30 day leverage is unavailable because historical snapshots have not been seeded yet
- add tests covering snapshot persistence and the missing-history output path

## Why
`camp leverage` computed the recent 7-day and 30-day values only from stored snapshots. If `.campaign/leverage/snapshots/` was empty or missing, the recent section disappeared even though the main leverage table still rendered.

## Root cause
The command relied on snapshot history for recent leverage, but normal `camp leverage` runs did not write a current snapshot. That meant the feature only worked after a manual `camp leverage backfill` or `camp leverage snapshot`, and could appear to "break again" whenever snapshot history was absent.

## Impact
- normal `camp leverage` usage now keeps current snapshot data fresh automatically
- campaigns with no seeded history get a clear instruction to run `camp leverage backfill` once instead of silently losing the recent section
- the change preserves the existing backfill-based historical model for true 7-day and 30-day comparisons

## Validation
- `go test ./cmd/camp/leverage ./internal/leverage`
- `just build-camp`
- `just test short`
- manual verification with `./projects/camp/bin/camp leverage` showing `Last 7 days` and `Last 30 days` again after the run populated snapshots

### DIFF
--- a/cmd/camp/leverage/main_command.go
+++ b/cmd/camp/leverage/main_command.go
@@ -81,6 +81,7 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	elapsed := intleverage.ElapsedMonths(cfg.ProjectStart, now)
 
 	var scores []*intleverage.LeverageScore
+	var snapshotInputs []currentSnapshotInput
 	for _, proj := range resolved {
 		if ctx.Err() != nil {
 			return ctx.Err()
@@ -102,6 +103,11 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 			FallbackElapsed: elapsed,
 		})
 		scores = append(scores, score)
+		snapshotInputs = append(snapshotInputs, currentSnapshotInput{
+			project: proj,
+			result:  result,
+			score:   score,
+		})
 	}
 
 	if projectFilter != "" && len(scores) == 0 {
@@ -132,6 +138,13 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	}
 
 	store := intleverage.NewFileSnapshotStore(intleverage.DefaultSnapshotDir(setup.Root))
+	existingSnapshots, listErr := store.ListProjects(ctx)
+	hadSnapshots := listErr == nil && len(existingSnapshots) > 0
+	if authorFilter == "" && peopleOverride == 0 {
+		if err := persistCurrentSnapshots(ctx, store, snapshotInputs, now, nil); err != nil {
+			fmt.Fprintf(cmd.ErrOrStderr(), "Warning: failed to save leverage snapshots: %v\n", err)
+		}
+	}
 	week7, has7 := intleverage.RecentLeverage(ctx, store, scores, effectivePeople, now.AddDate(0, 0, -7))
 	month30, has30 := intleverage.RecentLeverage(ctx, store, scores, effectivePeople, now.AddDate(0, 0, -30))
 
@@ -139,7 +152,13 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 		return leverageOutputJSON(cmd, agg, scores)
 	}
 
-	recent := recentLeverage{week7: week7, has7: has7, month30: month30, has30: has30}
+	recent := recentLeverage{
+		week7:         week7,
+		has7:          has7,
+		month30:       month30,
+		has30:         has30,
+		needsBackfill: authorFilter == "" && peopleOverride == 0 && len(scores) > 0 && !hadSnapshots && !has7 && !has30,
+	}
 	opts := leverageOutputOpts{
 		authorFilter:   authorFilter,
 		authorExcluded: authorExcluded,
--- a/cmd/camp/leverage/main_command_test.go
+++ b/cmd/camp/leverage/main_command_test.go
@@ -258,6 +258,160 @@ func TestLeverageConfigCommand_ValidationDate(t *testing.T) {
 	}
 }
 
+type snapshotStoreMock struct {
+	saved map[string]*intleverage.Snapshot
+}
+
+func newSnapshotStoreMock() *snapshotStoreMock {
+	return &snapshotStoreMock{saved: make(map[string]*intleverage.Snapshot)}
+}
+
+func (m *snapshotStoreMock) Save(_ context.Context, snapshot *intleverage.Snapshot) error {
+	key := snapshot.Project + ":" + snapshot.Date
+	m.saved[key] = snapshot
+	return nil
+}
+
+func (m *snapshotStoreMock) Load(_ context.Context, project, date string) (*intleverage.Snapshot, error) {
+	return m.saved[project+":"+date], nil
+}
+
+func (m *snapshotStoreMock) List(_ context.Context, project string) ([]string, error) {
+	var dates []string
+	prefix := project + ":"
+	for key := range m.saved {
+		if strings.HasPrefix(key, prefix) {
+			dates = append(dates, strings.TrimPrefix(key, prefix))
+		}
+	}
+	return dates, nil
+}
+
+func (m *snapshotStoreMock) LoadAll(ctx context.Context, project string) ([]*intleverage.Snapshot, error) {
+	dates, _ := m.List(ctx, project)
+	var snapshots []*intleverage.Snapshot
+	for _, date := range dates {
+		snapshots = append(snapshots, m.saved[project+":"+date])
+	}
+	return snapshots, nil
+}
+
+func (m *snapshotStoreMock) ListProjects(_ context.Context) ([]string, error) {
+	projects := make(map[string]struct{})
+	for key := range m.saved {
+		parts := strings.SplitN(key, ":", 2)
+		projects[parts[0]] = struct{}{}
+	}
+	var out []string
+	for project := range projects {
+		out = append(out, project)
+	}
+	return out, nil
+}
+
+func TestPersistCurrentSnapshots_SavesEachProjectAndReusesHeadLookup(t *testing.T) {
+	store := newSnapshotStoreMock()
+	commitDate := time.Date(2026, 4, 18, 10, 0, 0, 0, time.UTC)
+	sampledAt := commitDate.Add(2 * time.Hour)
+
+	inputs := []currentSnapshotInput{
+		{
+			project: intleverage.ResolvedProject{
+				Name:   "obey-platform-monorepo",
+				GitDir: "/tmp/obey-platform-monorepo",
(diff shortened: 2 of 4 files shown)


obey-agent (commented):
## Verdict: Request Changes

## Overview

The intent is right — `camp leverage` shouldn't require a manual `camp leverage snapshot`/`backfill` dance before the Last 7/30 days rows populate, and threading the current snapshot through every unfiltered run is the correct long-term fix. The implementation is mostly clean (context propagation, `camperrors.Wrapf` on error paths, injected `headCommitResolver` for tests, HEAD reuse across subprojects of a monorepo). There is one non-obvious correctness bug that makes the new path strictly worse than pre-PR behavior for a specific class of repos, plus a few maintainability and test-coverage gaps that are easy to close.

## Key Findings

1. **Auto-persisted snapshot becomes its own "prior" when HEAD is older than the window.** `NewSnapshot` derives `Date` from `commitDate`, so on a repo whose HEAD is 10 days old the current run writes a snapshot dated 10 days ago with `Leverage` pointing at the current score. `findMostRecent` then returns that snapshot for `since=now-7d`, and `RecentLeverage` computes `priorPM == currentPM` → displays `Last 7 days: 0 (0%)`. The `!has7 && !has30` gate on `needsBackfill` suppresses the backfill hint in that same state, so the user never learns why they're seeing zeroes. Active-repo tests don't exercise this path because fresh HEADs produce snapshots dated today, which `findMostRecent` correctly skips. See inline comment on `main_command.go` L144. Fix either by making `RecentLeverage` skip snapshots whose `SampledAt` is within the current run, or by deferring the first auto-persist until a subsequent run.

2. **`hadSnapshots` is derived from `ListProjects` (project names only) and swallows `listErr`.** An empty-but-present project directory or an unreadable snapshot root both register as "history exists", permanently suppressing the backfill hint. See inline on L141. Consider `HasAnySnapshots(ctx)` on `SnapshotStorer` or list-then-check; surface listErr as a stderr warning.

3. **The `authorFilter == "" && peopleOverride == 0` predicate is duplicated three times.** Name the concept once (`unfiltered := …`). Prevents drift when the next filter is added. See inline on L160.

4. **Save-failure path is warned-only but untested.** `snapshotStoreMock.Save` always returns nil; no test asserts the `camperrors.Wrapf` messages from `persistCurrentSnapshots` actually propagate, and nothing exercises the ctx.Err() branch. See inline on `main_command_test.go` L272.

## What's Done Well

- `headCommitResolver` plus the `byGitDir` memoization is clean: one git call per repo even when multiple projects (e.g. `obey-platform-monorepo` and `obey-platform-monorepo@festui`) share a worktree.
- Dependency-injected resolver lets the test assert the memoization contract directly (`headCalls != 1`) instead of stubbing a global.
- Context.Context threaded through the new code path, with explicit `ctx.Err()` checks at loop boundaries.
- Error wrapping uses the project's `camperrors.Wrapf` rather than `fmt.Errorf`, matching internal convention.
- Keeping the auto-persist behind `authorFilter == ""` and `peopleOverride == 0` avoids polluting the snapshot store with runs whose Leverage numbers are filter-scoped and not comparable across time.
- Table and JSON output paths are both considered; the new UI hint only renders on the table path, where it belongs.

## Staff Standard

Not yet. The inventory-quality bar (context propagation, error wrapping, interface-backed testability) is fine. The problem is that the new behavior reports a misleading `Last 7 days: 0` for any campaign whose HEAD commit is older than the 7/30-day window, and in that same state it silences the only hint telling the user how to recover. That's the specific failure mode the PR advertises as "fixed", so shipping it as-is trades one confusing UX (no recent row at all) for a worse one (a zero that looks authoritative). Close the window-interaction bug, add the regression test around it, and this is mergeable.

cmd/camp/leverage/main_command.go
```diff
@@ -132,14 +138,27 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	}
 
 	store := intleverage.NewFileSnapshotStore(intleverage.DefaultSnapshotDir(setup.Root))
+	existingSnapshots, listErr := store.ListProjects(ctx)
+	hadSnapshots := listErr == nil && len(existingSnapshots) > 0
+	if authorFilter == "" && peopleOverride == 0 {
+		if err := persistCurrentSnapshots(ctx, store, snapshotInputs, now, nil); err != nil {
```
obey-agent, line 144: **Auto-persisting the current run's snapshot can make `RecentLeverage` report a misleading `0` for projects whose HEAD commit is older than the 7/30-day window.**

`NewSnapshot` (internal/leverage/snapshot.go L74) sets `Date = commitDate.Format("2006-01-02")`, not `sampledAt`. If HEAD is 10 days old, this run writes a snapshot dated 10 days ago with `Leverage` pointing at the current `*LeverageScore`. On the same run, `RecentLeverage(since=now-7d)` calls `findMostRecent(snapshots, since)` (history.go L317-331), which picks that snapshot because its Date is `<= since`. `priorPM = snap.Leverage.EstimatedPeople * snap.Leverage.EstimatedMonths` equals the current run's `currentPM`, so `deltaPM = 0` and `has7 = true`.

That triggers two problems:
1. The output shows `Last 7 days: 0 (0%)` when the campaign has no real historical baseline — strictly worse than the pre-PR "no recent section" behavior.
2. The `needsBackfill` gate on L160 short-circuits on `!has7 && !has30`, so the "run `camp leverage backfill`" hint never appears for these campaigns. The user never learns they need to seed history.

Fresh-HEAD / active-repo cases (most tests, probably most manual testing) don't hit this because `findMostRecent(snapshots, since)` returns nil when every snapshot Date > since. Sparse or long-inactive repos are where this bites.

Two ways to resolve without reverting the feature:
- Teach `RecentLeverage` to skip any snapshot whose `SampledAt` is within the current run (or within the same clock minute), so the current-run snapshot is never its own prior.
- Or stop auto-persisting when `hadSnapshots == false` and only lay down the current snapshot on the NEXT run, after the user has had a chance to backfill.

Either way, a unit test in `internal/leverage/history_test.go` covering "single snapshot whose Date == since" is worth adding; that's the regression bar this PR creates.

cmd/camp/leverage/main_command.go
```diff
@@ -132,14 +138,27 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	}
 
 	store := intleverage.NewFileSnapshotStore(intleverage.DefaultSnapshotDir(setup.Root))
+	existingSnapshots, listErr := store.ListProjects(ctx)
```
obey-agent, line 141: **`hadSnapshots` is computed from `ListProjects`, which returns project *names* rather than actual snapshot files.**

On L141 the code probes with `store.ListProjects(ctx)` and treats any non-empty slice as "history exists". A project directory under `.campaign/leverage/snapshots/<project>/` that's present but empty (e.g. a prior failed write that created the dir without contents, or someone manually cleared the JSON files) will register as "had snapshots" and suppress the backfill hint forever.

Consider gating on whether any project actually has at least one snapshot file — for example call `store.List(ctx, name)` for each name in `ListProjects` and short-circuit on the first non-empty result. Or introduce `HasAnySnapshots(ctx) (bool, error)` on `SnapshotStorer` so callers don't have to reason about partial directory state.

Relatedly: `listErr` is swallowed into `hadSnapshots = false`, with no warning. If the snapshot root exists but is unreadable (permissions, corrupted JSON), the user sees "no history, try backfill" when the real problem is filesystem health. At least emit a one-line warning on stderr so the failure isn't silent.

cmd/camp/leverage/main_command.go
```diff
@@ -132,14 +138,27 @@ func runLeverage(cmd *cobra.Command, args []string) error {
...
 		return leverageOutputJSON(cmd, agg, scores)
 	}
 
-	recent := recentLeverage{week7: week7, has7: has7, month30: month30, has30: has30}
+	recent := recentLeverage{
+		week7:         week7,
+		has7:          has7,
+		month30:       month30,
+		has30:         has30,
+		needsBackfill: authorFilter == "" && peopleOverride == 0 && len(scores) > 0 && !hadSnapshots && !has7 && !has30,
```
obey-agent, line 160: **The same `authorFilter == "" && peopleOverride == 0` gate is repeated three times in `runLeverage` (L131, L143, L160).**

Each is semantically "is this an unfiltered aggregate run?". That concept is worth naming once:

```go
unfiltered := authorFilter == "" && peopleOverride == 0
```

...and reused everywhere. Besides readability, it prevents the three sites from drifting if a new filter (e.g. `--project`, `--since`) ends up in the same bucket. Staff-level maintainability nudge rather than a correctness issue.

cmd/camp/leverage/main_command_test.go
```diff
@@ -258,6 +258,160 @@ func TestLeverageConfigCommand_ValidationDate(t *testing.T) {
...
+}
+
+func newSnapshotStoreMock() *snapshotStoreMock {
+	return &snapshotStoreMock{saved: make(map[string]*intleverage.Snapshot)}
+}
+
+func (m *snapshotStoreMock) Save(_ context.Context, snapshot *intleverage.Snapshot) error {
+	key := snapshot.Project + ":" + snapshot.Date
+	m.saved[key] = snapshot
+	return nil
```
obey-agent, line 272: **`persistCurrentSnapshots` reports save failures as a single stderr warning and keeps going; no test covers the failure path.**

On main_command.go L143-147 a Save error becomes `Warning: failed to save leverage snapshots: %v` and execution continues. That's reasonable for a CLI — we don't want to kill the leverage table over a snapshot write — but two things are missing:

1. `snapshotStoreMock.Save` in the new test (L272) always returns `nil`. No test verifies the `camperrors.Wrapf` propagation or the "reading HEAD commit for X" wrap in `persistCurrentSnapshots`. At minimum add one mock that returns a synthetic error on Save and asserts the wrapped message reaches the caller; another that returns a synthetic error from `resolveHead` and asserts the HEAD-lookup wrap is applied.

2. No test covers context cancellation for `persistCurrentSnapshots`, even though the function explicitly checks `ctx.Err()` at the top of the loop. A table-driven test that cancels after the first Save would lock in that behavior and prevent regressions.

Both are small additions; they close the "works on my machine" gap for the error path the PR adds.

lancekrogers:
Superseded by #244 after renaming the branch and PR title to match repository naming preferences.

PR #244, opened 2026-04-18
fix: persist leverage snapshots for recent history

lancekrogers:
## What changed
- persist current leverage snapshots during normal `camp leverage` runs
- reuse one HEAD commit lookup per git repo when saving those snapshots
- show an explicit note when recent 7/30 day leverage is unavailable because historical snapshots have not been seeded yet
- add tests covering snapshot persistence and the missing-history output path

## Why
`camp leverage` computed the recent 7-day and 30-day values only from stored snapshots. If `.campaign/leverage/snapshots/` was empty or missing, the recent section disappeared even though the main leverage table still rendered.

## Root cause
The command relied on snapshot history for recent leverage, but normal `camp leverage` runs did not write a current snapshot. That meant the feature only worked after a manual `camp leverage backfill` or `camp leverage snapshot`, and could appear to "break again" whenever snapshot history was absent.

## Impact
- normal `camp leverage` usage now keeps current snapshot data fresh automatically
- campaigns with no seeded history get a clear instruction to run `camp leverage backfill` once instead of silently losing the recent section
- the change preserves the existing backfill-based historical model for true 7-day and 30-day comparisons

## Validation
- `go test ./cmd/camp/leverage ./internal/leverage`
- `just build-camp`
- `just test short`
- manual verification with `./projects/camp/bin/camp leverage` showing `Last 7 days` and `Last 30 days` again after the run populated snapshots

### DIFF
--- a/cmd/camp/leverage/helpers.go
+++ b/cmd/camp/leverage/helpers.go
@@ -324,6 +324,8 @@ func persistCurrentSnapshots(ctx context.Context, store intleverage.SnapshotStor
 		resolveHead = getHeadCommit
 	}
 
+	// TODO: Add snapshot retention trimming. Daily automatic snapshots solve
+	// recent-history freshness but will grow unbounded without an age-based cap.
 	type commitMeta struct {
 		hash string
 		date time.Time
--- a/cmd/camp/leverage/helpers.go
+++ b/cmd/camp/leverage/helpers.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/Obedience-Corp/camp/internal/campaign"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
@@ -244,6 +245,14 @@ type scoreParams struct {
 	FallbackElapsed float64
 }
 
+type currentSnapshotInput struct {
+	project intleverage.ResolvedProject
+	result  *intleverage.SCCResult
+	score   *intleverage.LeverageScore
+}
+
+type headCommitResolver func(context.Context, string) (string, time.Time, error)
+
 func computeProjectScore(ctx context.Context, proj intleverage.ResolvedProject, result *intleverage.SCCResult, params scoreParams) *intleverage.LeverageScore {
 	var projActualPM float64
 	var projPeople int
@@ -308,6 +317,43 @@ func computeProjectScore(ctx context.Context, proj intleverage.ResolvedProject, 
 	}
 
 	return score
+}
+
+func persistCurrentSnapshots(ctx context.Context, store intleverage.SnapshotStorer, inputs []currentSnapshotInput, sampledAt time.Time, resolveHead headCommitResolver) error {
+	if resolveHead == nil {
+		resolveHead = getHeadCommit
+	}
+
+	type commitMeta struct {
+		hash string
+		date time.Time
+	}
+
+	byGitDir := make(map[string]commitMeta)
+
+	for _, input := range inputs {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
+		meta, ok := byGitDir[input.project.GitDir]
+		if !ok {
+			hash, date, err := resolveHead(ctx, input.project.GitDir)
+			if err != nil {
+				return camperrors.Wrapf(err, "reading HEAD commit for %s", input.project.Name)
+			}
+			meta = commitMeta{hash: hash, date: date}
+			byGitDir[input.project.GitDir] = meta
+		}
+
+		scc := intleverage.SCCResultToSnapshotSCC(input.result)
+		snapshot := intleverage.NewSnapshot(input.project.Name, meta.hash, meta.date, sampledAt, scc, input.score, input.project.Authors)
+		if err := store.Save(ctx, snapshot); err != nil {
+			return camperrors.Wrapf(err, "saving current snapshot for %s", input.project.Name)
+		}
+	}
+
+	return nil
 }
 
 func buildScoreRows(scores []*intleverage.LeverageScore) [][]string {
--- a/cmd/camp/leverage/main_command.go
+++ b/cmd/camp/leverage/main_command.go
@@ -81,6 +81,7 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	elapsed := intleverage.ElapsedMonths(cfg.ProjectStart, now)
 
 	var scores []*intleverage.LeverageScore
+	var snapshotInputs []currentSnapshotInput
 	for _, proj := range resolved {
 		if ctx.Err() != nil {
 			return ctx.Err()
@@ -102,6 +103,11 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 			FallbackElapsed: elapsed,
 		})
 		scores = append(scores, score)
+		snapshotInputs = append(snapshotInputs, currentSnapshotInput{
+			project: proj,
+			result:  result,
+			score:   score,
+		})
 	}
 
 	if projectFilter != "" && len(scores) == 0 {
@@ -132,6 +138,13 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	}
 
 	store := intleverage.NewFileSnapshotStore(intleverage.DefaultSnapshotDir(setup.Root))
+	existingSnapshots, listErr := store.ListProjects(ctx)
+	hadSnapshots := listErr == nil && len(existingSnapshots) > 0
+	if authorFilter == "" && peopleOverride == 0 {
+		if err := persistCurrentSnapshots(ctx, store, snapshotInputs, now, nil); err != nil {
+			fmt.Fprintf(cmd.ErrOrStderr(), "Warning: failed to save leverage snapshots: %v\n", err)
+		}
+	}
 	week7, has7 := intleverage.RecentLeverage(ctx, store, scores, effectivePeople, now.AddDate(0, 0, -7))
 	month30, has30 := intleverage.RecentLeverage(ctx, store, scores, effectivePeople, now.AddDate(0, 0, -30))
 
@@ -139,7 +152,13 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 		return leverageOutputJSON(cmd, agg, scores)
 	}
 
-	recent := recentLeverage{week7: week7, has7: has7, month30: month30, has30: has30}
+	recent := recentLeverage{
+		week7:         week7,
+		has7:          has7,
+		month30:       month30,
+		has30:         has30,
+		needsBackfill: authorFilter == "" && peopleOverride == 0 && len(scores) > 0 && !hadSnapshots && !has7 && !has30,
+	}
 	opts := leverageOutputOpts{
 		authorFilter:   authorFilter,
 		authorExcluded: authorExcluded,
--- a/cmd/camp/leverage/main_command.go
+++ b/cmd/camp/leverage/main_command.go
@@ -138,8 +138,6 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	}
 
 	store := intleverage.NewFileSnapshotStore(intleverage.DefaultSnapshotDir(setup.Root))
-	existingSnapshots, listErr := store.ListProjects(ctx)
-	hadSnapshots := listErr == nil && len(existingSnapshots) > 0
 	if authorFilter == "" && peopleOverride == 0 {
 		if err := persistCurrentSnapshots(ctx, store, snapshotInputs, now, nil); err != nil {
 			fmt.Fprintf(cmd.ErrOrStderr(), "Warning: failed to save leverage snapshots: %v\n", err)
@@ -157,7 +155,7 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 		has7:          has7,
 		month30:       month30,
 		has30:         has30,
-		needsBackfill: authorFilter == "" && peopleOverride == 0 && len(scores) > 0 && !hadSnapshots && !has7 && !has30,
+		needsBackfill: authorFilter == "" && peopleOverride == 0 && len(scores) > 0 && !has7 && !has30,
 	}
 	opts := leverageOutputOpts{
 		authorFilter:   authorFilter,
--- a/cmd/camp/leverage/main_command_test.go
+++ b/cmd/camp/leverage/main_command_test.go
@@ -357,7 +357,10 @@ func TestPersistCurrentSnapshots_SavesEachProjectAndReusesHeadLookup(t *testing.
 		t.Fatalf("saved %d snapshots, want 2", len(store.saved))
 	}
 
-	snap := store.saved["obey-platform-monorepo:"+commitDate.Format("2006-01-02")]
+	snap, err := store.Load(context.Background(), "obey-platform-monorepo", commitDate.Format("2006-01-02"))
+	if err != nil {
+		t.Fatalf("Load: %v", err)
+	}
 	if snap == nil {
 		t.Fatal("expected root project snapshot to be saved")
 	}
--- a/cmd/camp/leverage/main_command_test.go
+++ b/cmd/camp/leverage/main_command_test.go
@@ -258,6 +258,160 @@ func TestLeverageConfigCommand_ValidationDate(t *testing.T) {
 	}
 }
 
+type snapshotStoreMock struct {
+	saved map[string]*intleverage.Snapshot
+}
+
+func newSnapshotStoreMock() *snapshotStoreMock {
+	return &snapshotStoreMock{saved: make(map[string]*intleverage.Snapshot)}
+}
+
+func (m *snapshotStoreMock) Save(_ context.Context, snapshot *intleverage.Snapshot) error {
+	key := snapshot.Project + ":" + snapshot.Date
+	m.saved[key] = snapshot
+	return nil
+}
+
+func (m *snapshotStoreMock) Load(_ context.Context, project, date string) (*intleverage.Snapshot, error) {
+	return m.saved[project+":"+date], nil
+}
+
+func (m *snapshotStoreMock) List(_ context.Context, project string) ([]string, error) {
+	var dates []string
+	prefix := project + ":"
+	for key := range m.saved {
+		if strings.HasPrefix(key, prefix) {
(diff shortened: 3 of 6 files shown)


cmd/camp/leverage/main_command.go
```diff
@@ -132,14 +138,27 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	}
```
obey-agent, line 139: **Ordering hazard: `hadSnapshots` is sampled before the write, but the write occurs *after* this read.**

```go
existingSnapshots, listErr := store.ListProjects(ctx)
hadSnapshots := listErr == nil && len(existingSnapshots) > 0
if authorFilter == "" && peopleOverride == 0 {
    if err := persistCurrentSnapshots(...); err != nil { ... }
}
week7, has7 := intleverage.RecentLeverage(ctx, store, ...)
```

The intent is to show the backfill hint only on the *very first* run — but the logic is: sample → write → read-for-recent. On the first run `hadSnapshots` is `false`, the write seeds the store, then `RecentLeverage` reads the freshly-written snapshot and may return `has7 = true`. That means `needsBackfill` uses a stale `hadSnapshots` (false) but `has7` is now true, so the hint is correctly suppressed... by accident.

More critically, the condition is ORed together:

```go
needsBackfill: authorFilter == "" && peopleOverride == 0 && len(scores) > 0 && !hadSnapshots && !has7 && !has30
```

On the **second run** (snapshots now exist), `hadSnapshots` is `true`, so `needsBackfill` is false — correct. But if `RecentLeverage` fails (e.g. corrupt file) and returns `has7 = false`, `hadSnapshots` is the only guard preventing a spurious backfill hint. If `listErr != nil` (directory unreadable), `hadSnapshots` is silently `false` and the hint fires again.

Suggest separating the concerns:
1. Check and snapshot in one step before reading recent leverage.
2. Derive `needsBackfill` purely from the post-write state: `!has7 && !has30 && len(scores) > 0` — no need for `hadSnapshots` at all. If the write just happened and recent values are still unavailable, backfill is genuinely needed.

cmd/camp/leverage/main_command.go
```diff
@@ -132,14 +138,27 @@ func runLeverage(cmd *cobra.Command, args []string) error {
...
 
 	store := intleverage.NewFileSnapshotStore(intleverage.DefaultSnapshotDir(setup.Root))
+	existingSnapshots, listErr := store.ListProjects(ctx)
+	hadSnapshots := listErr == nil && len(existingSnapshots) > 0
+	if authorFilter == "" && peopleOverride == 0 {
```
obey-agent, line 141: **Non-atomic write: a crash mid-write leaves a corrupt JSON file that silently blocks future reads.**

`FileSnapshotStore.Save` does:
```go
path := filepath.Join(dir, snapshot.Date+".json")
if err := os.WriteFile(path, data, 0o644); err != nil { ... }
```

`os.WriteFile` is a direct in-place overwrite. If the process is killed or loses power mid-write, the file is left partially written. On the next `camp leverage` run, `Load` will call `json.Unmarshal` on a truncated file and return a wrapped error, causing `RecentLeverage` to silently skip that project — effectively making recent-history disappear again until someone manually removes the corrupt file.

This PR writes a snapshot on every normal run (the missing piece before), so the surface area is now larger. Standard mitigation is an atomic rename:
```go
tmp := path + ".tmp"
if err := os.WriteFile(tmp, data, 0o644); err != nil { return ... }
return os.Rename(tmp, path)
```

`Rename` is atomic on POSIX when both paths are on the same filesystem (always true here since `tmp` lives in the same directory). This is the fix at the storage layer; the PR description correctly identifies the root cause but the store itself still has this latent durability gap.

cmd/camp/leverage/helpers.go
```diff
@@ -310,6 +319,43 @@ func computeProjectScore(ctx context.Context, proj intleverage.ResolvedProject,
...
+			byGitDir[input.project.GitDir] = meta
+		}
+
+		scc := intleverage.SCCResultToSnapshotSCC(input.result)
+		snapshot := intleverage.NewSnapshot(input.project.Name, meta.hash, meta.date, sampledAt, scc, input.score, input.project.Authors)
+		if err := store.Save(ctx, snapshot); err != nil {
+			return camperrors.Wrapf(err, "saving current snapshot for %s", input.project.Name)
+		}
+	}
+
```
obey-agent, line 357: **Retention is unbounded — snapshots grow forever.**

Every normal `camp leverage` run now calls `persistCurrentSnapshots`, which calls `store.Save` once per project per day (the key is `YYYY-MM-DD`). Running daily for a year on a 20-project campaign produces 20 × 365 = 7,300 JSON files. Running multiple times a day on the same date simply overwrites (fine), but there is no pruning path.

The PR description says "recent history" needs a bound, but no retention/trim policy is implemented or even documented as a known gap. Suggest one of:
- A constant `maxSnapshotAgeDays = 90` plus a post-save trim pass that deletes files older than the threshold.
- Or at minimum a follow-up issue/TODO comment so the gap is tracked.

Without a bound this is a maintenance debt that grows silently and will only be noticed when someone wonders why `.campaign/leverage/snapshots/` is large.

cmd/camp/leverage/main_command_test.go
```diff
@@ -258,6 +258,160 @@ func TestLeverageConfigCommand_ValidationDate(t *testing.T) {
...
+		t.Fatalf("persistCurrentSnapshots failed: %v", err)
+	}
+
+	if headCalls != 1 {
+		t.Fatalf("resolveHead called %d times, want 1", headCalls)
+	}
+
+	if len(store.saved) != 2 {
+		t.Fatalf("saved %d snapshots, want 2", len(store.saved))
+	}
```
obey-agent, line 358: **Test mock implements `SnapshotStorer` but the key in the mock diverges from the store contract.**

`FileSnapshotStore` keys snapshots by `YYYY-MM-DD` derived from `CommitDate`. The mock keys by `snapshot.Date`, which is also `CommitDate.Format("2006-01-02")` — consistent. However, the assertion at the bottom of `TestPersistCurrentSnapshots_SavesEachProjectAndReusesHeadLookup` is:

```go
snap := store.saved["obey-platform-monorepo:"+commitDate.Format("2006-01-02")]
```

This embeds the key construction directly in the test body — if `NewSnapshot` ever changes how `Date` is derived (e.g. uses `SampledAt` instead of `CommitDate`), this assertion will silently pass against a nil snapshot and the `snap == nil` check will catch it, but only for that one specific lookup.

Prefer looking up through the mock's own `Load` method with the project name and date string, so the test exercises the same retrieval path as production:
```go
snap, err := store.Load(ctx, "obey-platform-monorepo", commitDate.Format("2006-01-02"))
```
Minor but it keeps the test honest about the interface contract.

obey-agent (commented):
## Review: fix: persist leverage snapshots for recent history

**Verdict: REQUEST_CHANGES** — 2 blocking issues, 1 tracked debt, 1 minor.

---

### What the PR does well

Root cause identification is correct and the fix is the right architectural move. `persistCurrentSnapshots` is clean, single-responsibility, testable via the injected `headCommitResolver` func, and correctly deduplicates HEAD lookups per git repo. Error handling uses `camperrors.Wrapf` throughout. Tests cover the two new paths (snapshot persistence, backfill hint). No parallel persistence layer was introduced — this extends the existing `SnapshotStorer`.

---

### Blocking

**1. hadSnapshots ordering hazard (main_command.go:138-143)**
`ListProjects` is called *before* the write, then `RecentLeverage` is called *after* it. On the first run, snapshots are written and `has7` can come back true, so `needsBackfill` is false — correct by accident. The real problem is the error path: if `ListProjects` fails (`listErr != nil`), `hadSnapshots` is silently false, so a corrupt or unreadable store directory will keep showing the backfill hint on every run. Simplify: drop `hadSnapshots` entirely and derive `needsBackfill` from `!has7 && !has30 && len(scores) > 0` after the write. That is always accurate regardless of store state.

**2. Non-atomic write in FileSnapshotStore.Save (snapshot.go — pre-existing, now more exposed)**
`os.WriteFile` is an in-place overwrite. A mid-write crash leaves a truncated JSON file that will cause silent project omission on the next run — reproducing the bug this PR fixes. The write path is now exercised on every normal run, making the window wider. Fix: write to `<path>.tmp` then `os.Rename` into place. Rename is atomic on the same filesystem. This is in `internal/leverage/snapshot.go`, not a file changed by this PR, but the PR is the right moment to fix it.

---

### Should-track

**3. Snapshot retention is unbounded (helpers.go:persistCurrentSnapshots)**
Daily runs on a 20-project campaign = ~7,300 JSON files/year. No pruning path exists and none is documented as a known gap. Add a TODO or a follow-up issue; a 90-day rolling window would be sufficient for 7/30-day recent history.

---

### Minor

**4. Test assertion bypasses the Load interface (main_command_test.go:358)**
The test reads directly from `store.saved[key]` instead of `store.Load()`. Use `Load` so the test is honest about the retrieval contract.

obey-agent (commented):
## Review: fix: persist leverage snapshots for recent history

**Verdict: REQUEST_CHANGES** — 2 blocking issues, 1 tracked debt, 1 minor.

---

### What the PR does well

Root cause identification is correct and the fix is the right architectural move. `persistCurrentSnapshots` is clean, single-responsibility, testable via the injected `headCommitResolver` func, and correctly deduplicates HEAD lookups per git repo. Error handling uses `camperrors.Wrapf` throughout. Tests cover the two new paths (snapshot persistence, backfill hint). No parallel persistence layer was introduced — this extends the existing `SnapshotStorer`.

---

### Blocking

**1. hadSnapshots ordering hazard (main_command.go:138-143)**
`ListProjects` is called *before* the write, then `RecentLeverage` is called *after* it. On the first run, snapshots are written and `has7` can come back true, so `needsBackfill` is false — correct by accident. The real problem is the error path: if `ListProjects` fails (`listErr != nil`), `hadSnapshots` is silently false, so a corrupt or unreadable store directory will keep showing the backfill hint on every run. Simplify: drop `hadSnapshots` entirely and derive `needsBackfill` from `!has7 && !has30 && len(scores) > 0` after the write. That is always accurate regardless of store state.

**2. Non-atomic write in FileSnapshotStore.Save (snapshot.go — pre-existing, now more exposed)**
`os.WriteFile` is an in-place overwrite. A mid-write crash leaves a truncated JSON file that will cause silent project omission on the next run — reproducing the bug this PR fixes. The write path is now exercised on every normal run, making the window wider. Fix: write to `<path>.tmp` then `os.Rename` into place. Rename is atomic on the same filesystem. This is in `internal/leverage/snapshot.go`, not a file changed by this PR, but the PR is the right moment to fix it.

---

### Should-track

**3. Snapshot retention is unbounded (helpers.go:persistCurrentSnapshots)**
Daily runs on a 20-project campaign = ~7,300 JSON files/year. No pruning path exists and none is documented as a known gap. Add a TODO or a follow-up issue; a 90-day rolling window would be sufficient for 7/30-day recent history.

---

### Minor

**4. Test assertion bypasses the Load interface (main_command_test.go:358)**
The test reads directly from `store.saved[key]` instead of `store.Load()`. Use `Load` so the test is honest about the retrieval contract.

obey-agent (commented):
## Review: fix: persist leverage snapshots for recent history

**Verdict: REQUEST_CHANGES** — 2 blocking issues, 1 tracked debt, 1 minor.

---

### What the PR does well

Root cause identification is correct and the fix is the right architectural move. persistCurrentSnapshots is clean, single-responsibility, testable via the injected headCommitResolver func, and correctly deduplicates HEAD lookups per git repo. Error handling uses camperrors.Wrapf throughout. Tests cover the two new paths (snapshot persistence, backfill hint). No parallel persistence layer was introduced — this extends the existing SnapshotStorer.

---

### Blocking

**1. hadSnapshots ordering hazard (main_command.go:138-143)**
ListProjects is called before the write, then RecentLeverage is called after it. On the first run, snapshots are written and has7 can come back true, so needsBackfill is false — correct by accident. The real problem is the error path: if ListProjects fails (listErr != nil), hadSnapshots is silently false, so a corrupt or unreadable store directory will keep showing the backfill hint on every run. Simplify: drop hadSnapshots entirely and derive needsBackfill from !has7 && !has30 && len(scores) > 0 after the write. That is always accurate regardless of store state.

**2. Non-atomic write in FileSnapshotStore.Save (snapshot.go — pre-existing, now more exposed)**
os.WriteFile is an in-place overwrite. A mid-write crash leaves a truncated JSON file that will cause silent project omission on the next run — reproducing the bug this PR fixes. The write path is now exercised on every normal run, making the window wider. Fix: write to <path>.tmp then os.Rename into place. Rename is atomic on the same filesystem. This is in internal/leverage/snapshot.go, not a file changed by this PR, but the PR is the right moment to fix it.

---

### Should-track

**3. Snapshot retention is unbounded (helpers.go:persistCurrentSnapshots)**
Daily runs on a 20-project campaign = ~7,300 JSON files/year. No pruning path exists and none is documented as a known gap. Add a TODO or a follow-up issue; a 90-day rolling window is sufficient for 7/30-day recent history.

---

### Minor

**4. Test assertion bypasses the Load interface (main_command_test.go:358)**
The test reads directly from store.saved[key] instead of store.Load(). Use Load so the test is honest about the retrieval contract.

obey-agent (commented):
## Review: fix: persist leverage snapshots for recent history

**Verdict: REQUEST_CHANGES** — 2 blocking issues, 1 tracked debt, 1 minor.

Root cause identification is correct and the fix is the right architectural move. `persistCurrentSnapshots` is clean, single-responsibility, testable via the injected `headCommitResolver` func, and correctly deduplicates HEAD lookups per git repo. Error handling uses `camperrors.Wrapf` throughout. Tests cover the two new paths. No parallel persistence layer was introduced.

---

### Blocking

**1. hadSnapshots ordering hazard (main_command.go:138-143)**
`ListProjects` is called before the write, then `RecentLeverage` after. If `ListProjects` fails (`listErr != nil`), `hadSnapshots` is silently false, and a corrupt/unreadable store directory triggers the backfill hint on every run. Drop `hadSnapshots` entirely — derive `needsBackfill` purely from `!has7 && !has30 && len(scores) > 0` after the write. That is always accurate.

**2. Non-atomic write in FileSnapshotStore.Save (snapshot.go)**
`os.WriteFile` is an in-place overwrite. A mid-write crash leaves truncated JSON that silently drops a project from `RecentLeverage` — reproducing the bug this PR fixes. The write path is now exercised on every normal run, so the exposure is larger. Fix: write to `<path>.tmp` then `os.Rename` into place (POSIX-atomic on same fs). This is pre-existing in `snapshot.go` but this PR is the right moment to close it.

---

### Should-track

**3. Snapshot retention is unbounded**
Daily runs on a 20-project campaign produce ~7,300 JSON files/year with no pruning path and no documented gap. Add a TODO or follow-up issue; a 90-day rolling window is sufficient for 7/30-day recent history.

---

### Minor

**4. Test assertion bypasses the Load interface (main_command_test.go:358)**
Assertion reads `store.saved[key]` directly. Use `store.Load()` so the test exercises the retrieval contract.

obey-agent (changes requested):
## Review: fix: persist leverage snapshots for recent history

**Verdict: REQUEST_CHANGES** — 2 blocking issues, 1 tracked debt, 1 minor.

Root cause identification is correct and the fix is the right architectural move. `persistCurrentSnapshots` is clean, single-responsibility, testable via the injected `headCommitResolver` func, and correctly deduplicates HEAD lookups per git repo. Error handling uses `camperrors.Wrapf` throughout. Tests cover the two new paths. No parallel persistence layer was introduced.

---

### Blocking

**1. hadSnapshots ordering hazard (main_command.go:138-143)**
`ListProjects` is called before the write, then `RecentLeverage` after. If `ListProjects` fails (`listErr != nil`), `hadSnapshots` is silently `false`, and a corrupt/unreadable store directory triggers the backfill hint on every run. Drop `hadSnapshots` entirely — derive `needsBackfill` purely from `!has7 && !has30 && len(scores) > 0` after the write. That is always accurate regardless of store errors.

**2. Non-atomic write in FileSnapshotStore.Save (snapshot.go)**
`os.WriteFile` is an in-place overwrite. A mid-write crash leaves truncated JSON that silently drops a project from `RecentLeverage` — reproducing the bug this PR fixes. The write path is now exercised on every normal run, so the exposure is larger. Fix: write to `<path>.tmp` then `os.Rename` into place (POSIX-atomic on same filesystem). Pre-existing in `snapshot.go` but this PR is the right moment to close it.

---

### Should-track

**3. Snapshot retention is unbounded**
Daily runs on a 20-project campaign produce ~7,300 JSON files/year with no pruning path and no documented gap. Add a TODO or follow-up issue; a 90-day rolling window is sufficient for 7/30-day recent history.

---

### Minor

**4. Test assertion bypasses the Load interface (main_command_test.go:358)**
Assertion reads `store.saved[key]` directly. Use `store.Load()` so the test exercises the retrieval contract.

cmd/camp/leverage/main_command.go
obey-agent: Addressed in e05afaf. I removed the pre-write `hadSnapshots` check entirely, and `needsBackfill` now derives only from the post-write recent-history result: `!has7 && !has30 && len(scores) > 0`. That makes the hint accurate even if store listing fails.

cmd/camp/leverage/main_command.go
obey-agent: Addressed in e05afaf. `FileSnapshotStore.Save` now uses the existing repo helper `fsutil.WriteFileAtomically`, so snapshot updates go through same-directory temp file + fsync + rename instead of in-place overwrite.

cmd/camp/leverage/helpers.go
obey-agent: Tracked in e05afaf. I added a TODO at the persistence point calling out retention trimming as the next follow-up, so the unbounded-growth gap is now explicitly recorded next to the automatic snapshot write path.

cmd/camp/leverage/main_command_test.go
obey-agent: Addressed in e05afaf. The test now retrieves the saved snapshot through `store.Load(...)` instead of reaching into the mock's backing map directly, so it exercises the store contract rather than the mock internals.

PR #249, opened 2026-04-20
docs(skills): teach camp-projects skill to choose link vs submodule

lancekrogers:
## Summary

Updates the embedded \`camp-projects\` skill template so AI agents understand when to reach for \`camp project add\` vs \`camp project link\`, and actively **ask the user** when the choice isn't obvious from context.

## Background

Most users don't want the complexity of git submodules. Linking is typically the better default when a project is already cloned on the user's machine. But submodules are still the right call when the campaign needs to travel across devices (linked projects only work where the linked path exists on the filesystem).

The prior skill template predated \`camp project link\` and described projects only as submodules, so agents would silently default to \`camp project add --local\` even when the user had no intent to version the project with the campaign.

## Changes

**\`internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl\`**
- Broadened the front-matter description so the skill triggers on \"add / link / remove\" flows, not just submodule commits.
- Added a decision framework with three cases:
  - URL given → \`camp project add <url>\`
  - Already cloned locally → **ask the user**; link is the default recommendation, submodule only when the user explicitly wants multi-device use.
  - User wants multi-device → submodules are required; explain the tradeoff.
- Added a Remove section distinguishing safe \`unlink\` from history-modifying \`remove\`.
- Common Mistakes now includes the new pitfalls: silently adding as a submodule without asking, and telling users a submodule is required when a link would work fine.

**\`internal/scaffold/campaign/templates/.campaign/skills/references/camp-command-contracts.md.tmpl\`**
- New \"Adding and Linking Projects\" section with the exact command forms and flags, so the skill can cite the reference.
- Fixed the Pull/Push Scope section: \`camp pull\` / \`camp push\` without flags touch the campaign root repo (not projects). \`all\` works across every project regardless of whether it's a submodule or a linked local directory.
- Fixed the Status Scope wording: \`--sub\` acts on the current project (submodule or link), not exclusively submodules.

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go test ./internal/scaffold/... ./internal/skills/...\` green
- [x] Rebuilt via \`just install stable\` and ran \`camp init\` in a fresh temp campaign — verified the rendered \`.campaign/skills/camp-projects/SKILL.md\` and \`.campaign/skills/references/camp-command-contracts.md\` contain the new content end to end.

### DIFF
--- a/internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl
+++ b/internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl
@@ -62,9 +62,11 @@ camp refs-sync
 camp refs-sync projects/camp
 ```
 
-## Worktrees
+## Other Project Operations
 
-See shared reference.
+Other project subcommands — `list`, `new`, `run`, `prune`, `remote`,
+`worktree` — plus the project-scoped `camp fresh` (post-merge reset) are
+all documented in the shared reference.
 
 ## Common Mistakes
 
--- a/internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl
+++ b/internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl
@@ -8,16 +8,18 @@ description: Manage campaign projects with `camp project`. Use when the user wan
 Read shared contracts first: `../references/camp-command-contracts.md`.
 
 A project in a campaign is either a git submodule under `projects/` or a
-symlink (a "linked project") to a directory on the same machine. Most
-users prefer linking. Default toward link unless the user has stated a
-reason for a submodule.
+linked project — a symlink placed at `projects/<name>` that points to a
+directory elsewhere on the same machine. Most users prefer linking.
+Default toward link unless the user has stated a reason for a submodule.
 
 ## Decide: `add` or `link` — ASK the User
 
 - **URL given, not cloned locally** → `camp project add <url>`. Only
   reasonable choice.
 - **Already cloned on the user's machine** → ask: "Link it, or add as a
-  submodule?" Default recommendation is `camp project link <path>`. Use
+  submodule?" Default recommendation is `camp project link <path>`, which
+  creates a symlink at `projects/<name>` pointing back to the user's
+  existing directory (the directory stays where it is on disk). Use
   `camp project add --local <path>` only when the user has explicitly
   asked for multi-device use.
 - **User wants multi-device campaign** → submodules are required. Linked
--- a/internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl
+++ b/internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl
@@ -1,73 +1,54 @@
 ---
 name: camp-projects
-description: Manage campaign projects with `camp project`. Use when the user wants to add, link, or remove a project; when committing inside `projects/*`; when deciding status/pull/push scope (root vs submodule vs all); or when managing project worktrees.
+description: Manage campaign projects with `camp project`. Use when the user wants to add, link, or remove a project; when deciding between `camp project add` and `camp project link`; or when committing inside `projects/*`.
 ---
 
 # Campaign Projects
 
 Read shared contracts first: `../references/camp-command-contracts.md`.
 
-A project in a campaign is either a **git submodule** under `projects/` or a
-**linked local directory** (a symlink under `projects/` pointing to a
-directory on the same machine). Most users prefer linking — submodules add
-complexity they don't always want — so default toward linking unless the
-user has stated a reason for a submodule.
+A project in a campaign is either a git submodule under `projects/` or a
+symlink (a "linked project") to a directory on the same machine. Most
+users prefer linking. Default toward link unless the user has stated a
+reason for a submodule.
 
-## Decide: `add` or `link` — and ASK the User First
+## Decide: `add` or `link` — ASK the User
 
-When the user wants to bring a project into the campaign, identify the case
-and ask before choosing. **Do not assume.**
+- **URL given, not cloned locally** → `camp project add <url>`. Only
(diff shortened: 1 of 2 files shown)


internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl
```diff
@@ -1,12 +1,74 @@
...
 
 Read shared contracts first: `../references/camp-command-contracts.md`.
 
+A project in a campaign is either a **git submodule** under `projects/` or a
+**linked local directory** (a symlink under `projects/` pointing to a
+directory on the same machine). Most users prefer linking — submodules add
+complexity they don't always want — so default toward linking unless the
+user has stated a reason for a submodule.
+
+## Decide: `add` or `link` — and ASK the User First
```
obey-agent: This section turns the skill into a mini tutorial and duplicates detailed command semantics that the file already delegates to `../references/camp-command-contracts.md`. The neighboring skill templates stay terse and keep exact command contracts centralized in the shared reference. Here we now have two sources of truth for the same behavior, and they already drifted later in the file. I would keep the skill focused on the decision rule (`link` vs `add --local`) and let the reference carry the low-level mechanics.

internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl
```diff
@@ -1,12 +1,74 @@
...
+
+### When in doubt
+
+Ask. Never convert an already-local directory into a submodule without
+explicit user confirmation — that modifies the campaign's git history and
+can surprise users who expected a lightweight link.
+
+## Remove
+
+- `camp project unlink <name>` — remove a linked project. Safe: deletes the
```
obey-agent: `camp project remove` is narrower in this template than it is in the CLI. Today `internal/project/remove.go` safely unlinks linked projects by default and only rejects the destructive `--delete` path; `cmd/camp/project/remove.go` even documents this as “prefer unlink”, not “remove is wrong for links”. Framing `remove` as submodule-only will teach agents an inaccurate contract and make them avoid a supported path.

internal/scaffold/campaign/templates/.campaign/skills/camp-projects/SKILL.md.tmpl
```diff
@@ -20,15 +82,26 @@ camp refs-sync
 camp refs-sync projects/camp
 ```
 
+Linked projects commit to their own repositories; the campaign-root git does
+not track their commits and does not need `camp refs-sync` for them.
+
 ## Scope-Safe Status / Sync
 
+`camp status`, `camp pull`, and `camp push` without arguments act on the
```
obey-agent: This overstates current linked-project support. The new text says `camp status/pull/push --sub` works for the current project whether it is a submodule or a link, and that `all` spans every project. The implementation is narrower today: `internal/git/resolve.go` still errors when `--sub` resolves to a linked repo (`current directory is in a git repository but not a submodule of the campaign`), and the `all` paths in `cmd/camp/status_all.go`, `cmd/camp/pull.go`, and `cmd/camp/push.go` only enumerate submodules. If we merge this wording as-is, agents will confidently recommend commands that fail for linked projects.

obey-agent (changes requested):
Verdict: Request Changes

Overview:
This is solving a real gap in the `camp-projects` skill, but the current draft reads more like end-user documentation than an agent skill and it also teaches command contracts that the CLI does not actually honor today. I think the direction is right, but it needs to be tightened back to skill-file form and corrected to match the implementation.

Key Findings:
- The new `--sub` / `all` guidance overstates linked-project support. The implementation still treats `--sub` as submodule-only and the bulk `all` paths enumerate submodules, not linked projects.
- The Remove guidance narrows `camp project remove` more than the CLI does today. Linked projects are safely unlinked by `remove` unless the destructive `--delete` path is requested.
- The skill now duplicates a lot of low-level command-contract prose that already belongs in `references/camp-command-contracts.md`, which creates two sources of truth and already led to drift in this PR.

What's Done Well:
- The core intent is good: agents do need better guidance about `link` versus `add --local`, and the tradeoff explanation around machine-local links versus versioned submodules is directionally useful.

Staff Standard:
I would not be comfortable shipping this as-is because it will train agents on incorrect behavior and the current format is heavier than the rest of the skill set.

internal/scaffold/campaign/templates/.campaign/skills/references/camp-command-contracts.md.tmpl
```diff
@@ -11,6 +11,27 @@ camp go <shortcut> --print
...
+camp project add <url> [--name <n>] [--campaign <name-or-id>]
+
+# Add an existing local directory as a git submodule (modifies campaign history)
+camp project add --local <path> [--name <n>] [--campaign <name-or-id>]
+
+# Link an existing local directory as a machine-local symlink under projects/
+camp project link <path> [--name <n>] [--campaign <name-or-id>]
+camp project link                                        # link current directory
+
+# Remove
```
obey-agent: The top-level skill now matches the implementation much better, but the shared reference still has two incorrect contracts in this section: `camp project remove <name>` is documented here as submodule-only even though `internal/project/remove.go` safely unlinks linked projects by default, and the line below says linked projects are not versioned with the campaign even though `camp project link` records the symlink entry in campaign git history (`cmd/camp/project/linked/common.go`). Since the skill explicitly tells agents to read this reference first, these inaccuracies are still blocking.

obey-agent (changes requested):
Re-review:

The `camp-projects` skill template itself is in much better shape now. The linked-project wording is clearer, the `remove` behavior in the skill matches the implementation, and the scope-limit warning for linked projects is a good correction.

One blocking issue remains in `references/camp-command-contracts.md`: it still says `camp project remove <name>` is submodule-only and that linked projects are not versioned with the campaign. Both statements conflict with the current CLI behavior, and this file is the shared contract the skill tells agents to read first. I would land this once that reference section is brought back in sync with the implementation.

PR #251, opened 2026-04-20
fix(prune): detect squash-merged branches via gone upstream

lancekrogers:
## Summary

\`camp fresh\` and \`camp project prune\` only detect merged branches via git's ancestry check (\`git branch --merged\`), which misses **squash-merges**: squash creates a brand-new commit on the base branch, and the PR branch tip is never an ancestor of base.

GitHub's default merge strategy is squash + delete source branch. After \`git fetch --prune\`, the deleted remote source shows up as \`[origin/<branch>: gone]\` in tracking metadata. That signal is pure-git and forge-agnostic — anything that deletes merged branches upstream produces it — so **no \`gh\` CLI integration is required**.

## Motivation — concretely

Running \`camp fresh --dry-run\` inside \`projects/camp\` with three merged (squash) PR branches reported \`nothing to prune\`. Confirmed the PRs were squash-merged:
- Each main commit has exactly one parent (not two — not a merge commit)
- Commit messages use the \`(#NNN)\` squash format
- \`git merge-base --is-ancestor <branch-tip> origin/main\` returns false for each

Meanwhile \`git fetch --prune\` + \`git branch -vv\` shows all three as \`: gone]\`.

## Changes

- **\`internal/git/branches.go\`** — adds \`GoneBranches\` (lists local branches whose upstream tracking ref is gone, excluding the current branch) and \`DeleteBranchForce\` (\`git branch -D\`; \`-d\` refuses squash-merged branches).
- **\`internal/git/remote.go\`** — adds \`FetchRemotePrune\`.
- **\`internal/prune/prune.go\`** — unions ancestry-merged and gone-upstream branches into a single candidate list via \`unionBranches\`. Gone-upstream entries are marked for \`-D\` deletion; ancestry-merged entries continue to use \`-d\`. Remote branch deletion uses the ancestry-merged list only (a gone-upstream branch is already deleted upstream). The confirmation prompt and \`Result.Detail\` annotate which branches were detected via gone upstream.
- **\`internal/commands/fresh/fresh.go\`** — runs \`fetch --prune\` right before the prune step so gone markers reflect the current remote.

## Safety

- Active worktrees still block branch deletion (existing \`SkipWorktreeBranches\` path applies equally to gone-upstream entries).
- The existing confirmation prompt now calls out gone-upstream branches explicitly (\`(gone upstream — force delete)\`), so users see the distinction before deleting.
- \`--dry-run\` continues to work end-to-end.
- No \`gh\` CLI, no GitHub API, no network calls beyond \`git fetch --prune\`.

## Test plan

- [x] \`go test ./internal/git/... ./internal/prune/... ./internal/commands/fresh/...\` green.
- [x] New unit tests pass:
  - \`TestGoneBranches_DetectsDeletedUpstream\`
  - \`TestGoneBranches_NoGoneUpstreams\`
  - \`TestGoneBranches_ExcludesCurrentBranch\`
  - \`TestGoneBranches_CancelledContext\`
  - \`TestDeleteBranchForce_RemovesSquashMergedBranch\`
- [x] Fresh integration tests still green: \`go test ./tests/integration/ -run TestFresh -tags integration\`.

### DIFF
--- a/internal/commands/fresh/fresh.go
+++ b/internal/commands/fresh/fresh.go
@@ -209,8 +209,15 @@ func executeFresh(ctx context.Context, name, path string, opts freshOptions) err
 		}
 	}
 
-	// Step 3: Prune merged branches
+	// Step 3: Prune merged and gone-upstream branches.
+	// Refresh remote tracking before checking gone markers so squash-merged
+	// PRs (whose source branch was deleted upstream) show up as prunable.
 	if opts.prune {
+		if !opts.dryRun {
+			if err := git.FetchRemotePrune(ctx, path, "origin"); err != nil {
+				return camperrors.Wrap(err, "fetch --prune origin")
+			}
+		}
 		pruneOpts := prune.Options{
 			DryRun:  opts.dryRun,
 			Force:   true, // Skip confirmation — fresh is deliberate
--- a/internal/commands/fresh/fresh.go
+++ b/internal/commands/fresh/fresh.go
@@ -215,11 +215,15 @@ func executeFresh(ctx context.Context, name, path string, opts freshOptions) err
 	// 'git fetch --prune' first.
 	if opts.prune {
 		pruneOpts := prune.Options{
-			DryRun:        opts.dryRun,
-			Force:         true, // Skip confirmation — fresh is deliberate
-			Remote:        opts.pruneRemote,
+			DryRun: opts.dryRun,
+			Force:  true, // Skip confirmation — fresh is deliberate
+			Remote: opts.pruneRemote,
+			// Refresh even on dry-run: 'git fetch --prune' only updates
+			// remote-tracking refs, not the worktree, so the dry-run
+			// preview must include it or squash-merged branches stay
+			// invisible until the user fetches manually.
 			BaseRef:       syncState.baseRef,
-			RefreshRemote: !opts.dryRun,
+			RefreshRemote: true,
 		}
 		pr := prune.Execute(ctx, name, path, pruneOpts)
 		if pr.Error != "" {
--- a/internal/commands/fresh/fresh.go
+++ b/internal/commands/fresh/fresh.go
@@ -210,19 +210,16 @@ func executeFresh(ctx context.Context, name, path string, opts freshOptions) err
 	}
 
 	// Step 3: Prune merged and gone-upstream branches.
-	// Refresh remote tracking before checking gone markers so squash-merged
-	// PRs (whose source branch was deleted upstream) show up as prunable.
+	// Prune flow itself refreshes remote tracking (RefreshRemote below),
+	// so squash-merged PRs show up here without requiring the user to run
+	// 'git fetch --prune' first.
 	if opts.prune {
-		if !opts.dryRun {
-			if err := git.FetchRemotePrune(ctx, path, "origin"); err != nil {
-				return camperrors.Wrap(err, "fetch --prune origin")
-			}
-		}
 		pruneOpts := prune.Options{
-			DryRun:  opts.dryRun,
-			Force:   true, // Skip confirmation — fresh is deliberate
-			Remote:  opts.pruneRemote,
-			BaseRef: syncState.baseRef,
+			DryRun:        opts.dryRun,
+			Force:         true, // Skip confirmation — fresh is deliberate
+			Remote:        opts.pruneRemote,
+			BaseRef:       syncState.baseRef,
+			RefreshRemote: !opts.dryRun,
 		}
 		pr := prune.Execute(ctx, name, path, pruneOpts)
 		if pr.Error != "" {
--- a/internal/prune/prune.go
+++ b/internal/prune/prune.go
@@ -41,6 +41,16 @@ type Options struct {
 	// SkipWorktreeBranches preserves merged branches that still have active
 	// worktrees instead of removing the worktree and deleting the branch.
 	SkipWorktreeBranches bool
+
+	// RefreshRemote causes Execute to run 'git fetch --prune <remote>'
+	// before checking upstream:track state. Required for gone-upstream
+	// detection to reflect the current remote. Set true for interactive
+	// entrypoints (camp fresh, camp project prune) and false for library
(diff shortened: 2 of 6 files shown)


internal/prune/prune.go
```diff
@@ -79,22 +79,66 @@ func Execute(ctx context.Context, name, path string, opts Options) ProjectResult
...
 	}
 
+	// Gone-upstream branches catch squash-merged PRs: git's ancestry-based
+	// --merged check misses them because squash creates a brand-new commit
+	// on base that shares no history with the branch. When the remote
+	// source branch is deleted (GitHub's "Automatically delete head
+	// branches" default), fetch --prune marks the upstream as gone, which
+	// is the same "done" signal across any forge that deletes merged
+	// branches.
+	gone, err := git.GoneBranches(ctx, path)
```
obey-agent, line 116: Treating any `%(upstream:track)` value containing `gone` as safe to force-delete is not actually equivalent to "merged." A branch also becomes `[origin/foo: gone]` when someone deletes the remote branch without merging it. In that case `git merge-base --is-ancestor <branch> origin/main` is still false, but this path adds it to `candidates` and `fresh` will delete it with `-D` non-interactively. We need an additional merge-equivalence check here, or we need to keep `gone upstream` as an opt-in/prompt-only hint instead of auto-deleting local work.

internal/commands/fresh/fresh.go
```diff
@@ -209,8 +209,15 @@ func executeFresh(ctx context.Context, name, path string, opts freshOptions) err
 		}
 	}
 
-	// Step 3: Prune merged branches
+	// Step 3: Prune merged and gone-upstream branches.
+	// Refresh remote tracking before checking gone markers so squash-merged
+	// PRs (whose source branch was deleted upstream) show up as prunable.
 	if opts.prune {
+		if !opts.dryRun {
+			if err := git.FetchRemotePrune(ctx, path, "origin"); err != nil {
```
obey-agent: This refresh only happens in `fresh`. `camp project prune` and `camp project prune all` still call `prune.Execute` without a preceding `fetch --prune`, so the new `GoneBranches` path stays stale there until the user manually refreshes tracking metadata. Since the PR description says `camp project prune` is fixed too, I think the fetch/prune step needs to live in the prune flow itself or the scope/claims need to be narrowed.

obey-agent (changes requested):
Verdict: Request Changes

This fixes a real usability gap around squash-merged branches, and the targeted test command from the PR body is green locally: `go test ./internal/git/... ./internal/prune/... ./internal/commands/fresh/...`.

Key Findings:
- `internal/prune/prune.go`: `[origin/<branch>: gone]` is not reliable evidence that the branch was merged. Deleting a remote branch without merging produces the same signal, and the new path force-deletes those local branches during `camp fresh`. I reproduced that exact case locally: after deleting an unmerged remote branch, `git branch -vv` showed `[gone]` while `git merge-base --is-ancestor feature origin/main` still returned exit code `1`.
- `internal/commands/fresh/fresh.go`: only `fresh` refreshes tracking metadata before calling into prune. The direct `camp project prune` paths still depend on the user having manually run `git fetch --prune`, so the squash-merge detection this PR claims to add is incomplete there.

What's Done Well:
- The ancestry-vs-gone split is easy to follow.
- The new git tests around deleted upstreams and forced local deletion are solid.

Staff Standard:
- Not yet. I would not ship a non-interactive `-D` path that can remove unmerged local work based only on upstream deletion.

internal/commands/fresh/fresh.go
```diff
@@ -209,13 +209,17 @@ func executeFresh(ctx context.Context, name, path string, opts freshOptions) err
...
 		pruneOpts := prune.Options{
-			DryRun:  opts.dryRun,
-			Force:   true, // Skip confirmation — fresh is deliberate
-			Remote:  opts.pruneRemote,
-			BaseRef: syncState.baseRef,
+			DryRun:        opts.dryRun,
+			Force:         true, // Skip confirmation — fresh is deliberate
+			Remote:        opts.pruneRemote,
+			BaseRef:       syncState.baseRef,
+			RefreshRemote: !opts.dryRun,
```
obey-agent: The safety fix is much better now, but this still disables remote refresh in dry-run mode. That means the original motivating case in the PR body (`camp fresh --dry-run` missing squash-merged branches until the user manually fetches) still happens whenever local tracking metadata is stale. `git fetch --prune` does not touch the worktree, only remote-tracking refs, so if we want dry-run to preview current prune behavior it still needs the refresh step or the scope/docs need to say dry-run relies on pre-refreshed tracking state.

obey-agent (changes requested):
Verdict: Request Changes

The updated head addressed the earlier unsafe `gone-upstream => force delete` issue by adding a merge-equivalence check and by moving remote refresh into the prune flow. The targeted test command from the PR body is still green locally: `go test ./internal/git/... ./internal/prune/... ./internal/commands/fresh/...`.

Key Findings:
- `internal/commands/fresh/fresh.go` / `cmd/camp/project/prune.go`: remote refresh is now disabled when `--dry-run` is set. That leaves the original stale-tracking dry-run failure mode in place, so `camp fresh --dry-run` and `camp project prune --dry-run` can still miss squash-merged branches unless the user fetched manually first.

What's Done Well:
- The merge-equivalence filter is the right direction for distinguishing squash-merges from remote-deleted unmerged branches.
- The new git coverage is materially stronger than the previous revision.

Staff Standard:
- Close, but not yet. I’d want dry-run to preview the current remote state as well, or the command contract to narrow explicitly.

obey-agent (approved):
Verdict: Approve

Overview:
This addresses the dry-run stale-tracking gap from the previous review. `fresh` and `project prune` now refresh remote-tracking refs before evaluating gone-upstream branches, so the preview matches the real prune behavior the PR is trying to add.

What's Done Well:
- The fix stays inside the existing prune abstraction instead of creating a second path.
- The comments make the dry-run semantics explicit.
- `fresh` and `project prune` now behave consistently.

Staff Standard:
Yes. This now matches the PR's stated behavior.

PR #252, opened 2026-04-21
fix(project): resolve nested submodules correctly in ResolveFromCwd

lancekrogers:
## Summary

When running `camp fresh` (or any command relying on `ResolveFromCwd`) from inside a nested submodule checkout — e.g. `projects/obey-platform-monorepo/obey/` — the resolver silently targeted the *outer* monorepo instead of the nested submodule. The parent was on `main`, so every invocation reported

```
── Checkout main                     already on it
── Pull (ff-only)                  up-to-date
── Prune merged branches           nothing to prune
Fresh! Synced to main.
```

…while the user's actual working repo sat on a feature branch with stale merged branches and orphan worktrees that never got cleaned up.

## Root causes

1. **First-match resolver.** `ResolveFromCwd` in `internal/project/resolve.go` iterated projects and returned the first whose path contained cwd. Iteration order put the outer monorepo ahead of its submodule entries, so the outer always won.
2. **Base-name dedup dropping nested submodules.** `deduplicateByRemoteURL` in `internal/project/list.go` removed any monorepo subproject whose base name matched a standalone project. That meant `obey-platform-monorepo@obey` was dropped whenever top-level `obey` existed, even though the two are separate checkouts on potentially different branches/commits.

## Fix

- `ResolveFromCwd` now picks the project with the **longest (deepest) matching path**, so nested submodules win over their parent monorepos.
- Nested submodules are always kept in the project list. Only standalone projects are subject to URL-based dedup.

## Behavior change

`camp project list` now includes nested submodule entries that collide by base name with a standalone project (e.g. `obey-platform-monorepo@obey`, `@fest`, `@camp`, `@obey-shared`). Callers that care about dedup can filter on `Project.MonorepoRoot`. This is intentional — the two checkouts are operationally independent and `camp fresh` / `camp <cmd> --project mono@foo` need to address them directly.

## Verification

Repro against obey-campaign after the fix:

```
$ cd projects/obey-platform-monorepo/obey
$ camp fresh --dry-run
  obey-platform-monorepo@obey (dry-run)
  ── Checkout main                     already on it
  ── Would pull (ff-only)
  ── Prune merged branches           would delete: codex/issue-56-system-prompt-templates,
      codex/provider-list-command, codex/review-obey-pr58, lr/camputil-migration,
      lr/codex-output-schema, lr/obey-smoke-fixes; would remove 8 worktree(s)
```

From the monorepo root, resolution still correctly targets the parent monorepo — no regression.

## Test plan

- [x] `just test unit` — 2188/2188 passing
- [x] New `TestResolveFromCwd_NestedSubmoduleWins` covers the resolver fix.
- [x] Flipped `TestList_GitmodulesSubmoduleDedupAgainstStandalone` → `TestList_GitmodulesSubmoduleKeptAlongsideStandalone` to assert the new (correct) behavior.
- [x] Smoke-tested end-to-end with dev build against `projects/obey-platform-monorepo/obey`.

### DIFF
--- a/internal/project/list.go
+++ b/internal/project/list.go
@@ -223,10 +223,11 @@ func latestCommitDate(ctx context.Context, absPath string) string {
 	return strings.TrimSpace(string(output))
 }
 
-// deduplicateByRemoteURL groups projects by git remote URL and keeps only the
-// copy with the most recent commit. Projects with empty URL are always kept.
-// Monorepo subprojects skip URL-based dedup (they share the parent's URL) but
-// are dropped if a standalone project with the same base name exists.
+// deduplicateByRemoteURL groups standalone projects by git remote URL and keeps
+// only the copy with the most recent commit. Projects with empty URL are always
+// kept. Monorepo subprojects are always kept — they are separate checkouts that
+// can sit on different branches/commits than any standalone project with the
+// same base name, and callers (e.g. `camp fresh`) need to target them directly.
 func deduplicateByRemoteURL(ctx context.Context, campaignRoot string, projects []Project) []Project {
 	if ctx.Err() != nil {
 		return projects
@@ -236,17 +237,9 @@ func deduplicateByRemoteURL(ctx context.Context, campaignRoot string, projects [
 	bestIdx := make(map[string]int)
 	bestDate := make(map[string]string)
 
-	// Collect standalone project names for monorepo dedup.
-	standaloneNames := make(map[string]bool)
-	for _, p := range projects {
-		if p.MonorepoRoot == "" {
-			standaloneNames[p.Name] = true
-		}
-	}
-
 	for i, p := range projects {
-		// Skip monorepo subprojects — they share the parent's URL and are
-		// deduped by name against standalone repos instead.
+		// Monorepo subprojects share the parent's URL. They are not subject to
+		// URL-based dedup — each nested submodule is an independent checkout.
 		if p.URL == "" || p.MonorepoRoot != "" {
 			continue
 		}
@@ -278,11 +271,7 @@ func deduplicateByRemoteURL(ctx context.Context, campaignRoot string, projects [
 		case p.URL == "":
 			result = append(result, p)
 		case p.MonorepoRoot != "":
-			// Keep monorepo subproject only if no standalone repo has the same base name.
-			baseName := filepath.Base(p.Path)
-			if !standaloneNames[baseName] {
-				result = append(result, p)
-			}
+			result = append(result, p)
 		case keep[i]:
 			result = append(result, p)
 		}
--- a/internal/project/list_test.go
+++ b/internal/project/list_test.go
@@ -398,9 +398,12 @@ func TestList_StandaloneNotExpanded(t *testing.T) {
 	}
 }
 
-func TestList_GitmodulesSubmoduleDedupAgainstStandalone(t *testing.T) {
-	// When a .gitmodules repo has a submodule named "foo" and there's also
-	// a standalone project named "foo", the submodule entry should be deduped.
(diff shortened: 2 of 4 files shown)


obey-agent (commented):
Verdict: Approve.

Reviewed head `3b331f88`. `go build ./...`, `go vet ./...`, and `go test -race ./internal/project/...` all clean locally (`ok .../internal/project 5.990s`).

### Key Findings

The root-cause analysis in the PR body is correct. Both bugs had the same shape: two iteration-order-dependent policies where a nested submodule happened to lose. The fix rewrites each policy to be iteration-order-independent, and uses `MonorepoRoot` as the unambiguous signal for "this is a monorepo subproject" — much cleaner than any heuristic on paths or base names.

1. `internal/project/resolve.go:104-129` — longest-match replaces first-match. Because `projectPath` and `logicalPath` of a nested submodule share a prefix with the parent monorepo's path plus the submodule's own directory name, the nested entry always has the longer match string by construction. The `len(match) > bestLen` strict inequality is fine; equal-length ties aren't possible across two projects matching the same cwd.

2. `internal/project/list.go:243, 273-274` — `deduplicateByRemoteURL` now early-returns for `MonorepoRoot != ""` on both the scan pass and the emit pass. The dedup policy (keep most-recent standalone per URL) is preserved intact; only the carve-out is added. Symmetrically applied in both passes, no drift risk.

### Edge Cases I Checked

- Two standalone projects that share a URL still dedup correctly against the most-recent clone. Confirmed via `TestList_GitmodulesSubmoduleKeptAlongsideStandalone` flow and by reading the emit loop.
- The second fallback in `ResolveFromCwd` (`git.FindProjectRootWithType` + exact-path match against projects) was not touched and does not need to be. It matches projects by exact path equality, not `isPathWithin`, so it never had the first-match bug. It runs only when the longest-match pass finds no hit, which is now rare for real submodule layouts.
- `cwd` vs `resolvedCwd` / `projectPath` vs `logicalPath` asymmetry is pre-existing; the fix does not introduce new cases. Not a regression.
- Naming collision between `obey-platform-monorepo@obey` and standalone `obey` is avoided by the `@` separator in monorepo subproject names.

### What's Done Well

- Fix surgically addresses the two actual policies instead of bolting on a "prefer nested" flag.
- `MonorepoRoot` is the right discriminator — it already existed, so the dedup carve-out does not introduce a new invariant.
- `TestResolveFromCwd_NestedSubmoduleWins` sets up the minimal repro (monorepo with one nested submodule, cwd inside the submodule) and asserts the `mono@obey` resolution, so the regression can't silently return.
- `TestList_GitmodulesSubmoduleDedupAgainstStandalone` → `TestList_GitmodulesSubmoduleKeptAlongsideStandalone` rename is the right move — asserting the new correct behavior with a new intent, rather than just deleting a test that used to lock in a bug.
- PR body documents the observable behavior change (`camp project list` emits more rows) and points callers at `Project.MonorepoRoot` for filtering. Future readers will need that.

### Non-blocking Notes

- The 8 stale worktrees and 6 merged branches the smoke test surfaced are operational cleanup this fix newly exposes. Not a code concern; worth a separate `camp fresh` pass on `projects/obey-platform-monorepo/obey` after this lands.
- Consumers of `camp project list` output in shell scripts or TUI filters may need a spot-check once this merges: anything that used `--project obey` while cd'd inside the monorepo will now disambiguate to the standalone clone rather than the monorepo subproject. Users cd'd into the monorepo subproject get the right answer automatically via `ResolveFromCwd`. Worth a release-note line.

### Staff Standard

Yes. The fix is scoped exactly to the two iteration-order bugs, the tests regression-lock both of them, and the behavior change is called out explicitly in the PR. Ready to merge.

PR #256, opened 2026-04-29
Remove camp init --skip-fest flag (#254)

### LINKED ISSUE #254: Remove camp init --skip-fest
## Summary

`camp init --skip-fest` appears to be an old CLI escape hatch that is likely inconsistent with the launch onboarding direction. A new campaign should normally receive Festival Methodology scaffolding by default, and skipping that setup risks producing workspaces that fail the expected first-run `fest next` path.

## Why this should be separate

PR #253 surfaced the flag while wiring `camp create`, but removing the existing `camp init` flag is a CLI behavior/API cleanup and should not be bundled into that PR. The PR should avoid expanding or reworking `--skip-fest`; this issue tracks deciding and removing the existing command surface intentionally.

## Proposed work

- Remove the `--skip-fest` flag from `camp init` if there is no confirmed valid use case.
- Remove related unit/integration tests and generated docs.
- Verify init/create onboarding still creates a usable `festivals/` workspace when `fest` is available.
- If an escape hatch is still needed for tests or unusual environments, replace the public flag with a narrower internal/test-only mechanism.

## Acceptance criteria

- `camp init --help` no longer lists `--skip-fest`.
- CLI reference docs no longer document `--skip-fest`.
- Init/create tests cover the default Festival initialization path without relying on the public flag.
- Any retained bypass is explicitly private or clearly documented as an unsupported/test-only path.

obey-agent:
Closes #254.

## Summary

- Remove the `--skip-fest` flag from `camp init` (and `Params.SkipFest`).
- New campaigns now always receive Festival Methodology scaffolding, so the documented first-run `fest next` path works out of the box.
- Replace the public flag with a private, undocumented `CAMP_INIT_SKIP_FEST` env var as a narrower test-only escape hatch.
- Update the workitem editor TTY integration test to set `CAMP_INIT_SKIP_FEST=1` in its base env instead of passing `--skip-fest`.
- Update `camp create`'s `absentFlags` assertion (skip-fest is no longer init-only — it doesn't exist anywhere on the public CLI).
- Regenerate `docs/cli-reference/camp_init.md` and `camp-reference.md`; both no longer mention the flag.

## Acceptance criteria (from #254)

- `camp init --help` no longer lists `--skip-fest`. ✓
- CLI reference docs no longer document `--skip-fest`. ✓
- Init/create tests cover the default Festival initialization path without relying on the public flag. ✓ (the workitem test still uses an escape hatch, but via env var, not a public flag)
- Any retained bypass is explicitly private or clearly documented as test-only. ✓ (`CAMP_INIT_SKIP_FEST` is documented in code as an unsupported test-only escape hatch and is not exposed in help or generated docs)

## Test plan

- [x] `just build` passes
- [x] `go test ./cmd/camp/... ./internal/scaffold/...` passes
- [x] `just docs` regenerated cleanly with no `skip-fest` references remaining
- [ ] CI integration suite green

### DIFF
--- a/cmd/camp/init/command.go
+++ b/cmd/camp/init/command.go
@@ -77,12 +77,6 @@ Use --no-git to skip git initialization.`,
 	return cmd
 }
 
-// envSkipFest is an unsupported, test-only escape hatch that bypasses
-// Festival Methodology initialization. It is intentionally not exposed
-// as a CLI flag — production users should always receive Festival
-// scaffolding so that the documented `fest next` first-run path works.
-const envSkipFest = "CAMP_INIT_SKIP_FEST"
-
 // Params is the full set of inputs the init flow needs, already
 // parsed from flags or constructed by another command (e.g., create).
 type Params struct {
@@ -265,10 +259,9 @@ func RunFlow(ctx context.Context, p Params, w Writers, isInteractive bool) error
 		commitRepairChanges(ctx, result, opts.RepairPlan, migrationCount, w)
 	}
 
-	// Initialize Festival Methodology (unless dry-run or test-only env override).
-	skipFest := os.Getenv(envSkipFest) != ""
+	// Initialize Festival Methodology (unless dry-run).
 	var festInitialized bool
-	if !p.DryRun && !skipFest {
+	if !p.DryRun {
 		festInitialized, _ = initializeFestivals(ctx, result.CampaignRoot, w)
 	}
 
--- a/cmd/camp/init/command.go
+++ b/cmd/camp/init/command.go
@@ -73,10 +73,15 @@ Use --no-git to skip git initialization.`,
 	cmd.Flags().Bool("dry-run", false, "Show what would be done without creating anything")
 	cmd.Flags().Bool("repair", false, "Add missing files to existing campaign")
 	cmd.Flags().Bool("yes", false, "Skip repair confirmation prompt (for scripting)")
-	cmd.Flags().Bool("skip-fest", false, "Skip automatic Festival Methodology initialization")
 
 	return cmd
 }
+
+// envSkipFest is an unsupported, test-only escape hatch that bypasses
+// Festival Methodology initialization. It is intentionally not exposed
+// as a CLI flag — production users should always receive Festival
+// scaffolding so that the documented `fest next` first-run path works.
+const envSkipFest = "CAMP_INIT_SKIP_FEST"
 
 // Params is the full set of inputs the init flow needs, already
 // parsed from flags or constructed by another command (e.g., create).
@@ -92,7 +97,6 @@ type Params struct {
 	DryRun        bool
 	Repair        bool
 	Yes           bool
-	SkipFest      bool
 	VerboseOutput bool
 }
 
@@ -140,7 +144,6 @@ func runInit(cmd *cobra.Command, args []string) error {
 		DryRun:        cmdutil.GetFlagBool(cmd, "dry-run"),
 		Repair:        cmdutil.GetFlagBool(cmd, "repair"),
 		Yes:           cmdutil.GetFlagBool(cmd, "yes"),
-		SkipFest:      cmdutil.GetFlagBool(cmd, "skip-fest"),
 		VerboseOutput: verboseOutput,
 	}
 	w := ChooseWriters()
@@ -262,12 +265,11 @@ func RunFlow(ctx context.Context, p Params, w Writers, isInteractive bool) error
 		commitRepairChanges(ctx, result, opts.RepairPlan, migrationCount, w)
 	}
 
-	// Initialize Festival Methodology (unless skipped or dry-run).
+	// Initialize Festival Methodology (unless dry-run or test-only env override).
+	skipFest := os.Getenv(envSkipFest) != ""
 	var festInitialized bool
-	if !p.DryRun && !p.SkipFest {
+	if !p.DryRun && !skipFest {
 		festInitialized, _ = initializeFestivals(ctx, result.CampaignRoot, w)
-	} else if p.SkipFest && !p.DryRun {
-		writeLine(w.HumanOut, ui.Info("Skipping Festival Methodology (--skip-fest)"))
 	}
 
 	// Print results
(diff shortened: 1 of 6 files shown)


cmd/camp/init/command.go
```diff
@@ -262,12 +265,11 @@ func RunFlow(ctx context.Context, p Params, w Writers, isInteractive bool) error
 		commitRepairChanges(ctx, result, opts.RepairPlan, migrationCount, w)
 	}
 
-	// Initialize Festival Methodology (unless skipped or dry-run).
+	// Initialize Festival Methodology (unless dry-run or test-only env override).
+	skipFest := os.Getenv(envSkipFest) != ""
```
obey-agent: Clarification: I am not arguing to keep `--skip-fest` or any user-facing optional path. Removing that flag is the right product behavior: `camp init` should initialize Festival scaffolding for users.

The issue I meant to flag is narrower and test-harness-only. This PR appears to introduce `CAMP_INIT_SKIP_FEST` so a non-container test can avoid Festival initialization, but that env var only guards the command-layer `initializeFestivals` call. `scaffold.Init(ctx, dir, opts)` still calls `internal/scaffold.initFestivalsIfNeeded` first, and that path invokes `fest init --path <dir>` whenever a `fest` binary is on `PATH`. I reproduced that with `CAMP_INIT_SKIP_FEST=1` and a fake `fest` shim; the shim was still called before this guard.

So the request is not “make skipping festivals optional.” It is: do not leave a partial test-only escape hatch that gives tests a false sense of isolation. Either remove the test-only env var and move the filesystem-mutating coverage into the containerized integration harness with `fest`/`PATH` controlled there, or make the env var cover every internal Festival init path strictly for that test harness. The user-facing command should not expose a skip flag.

obey-agent (commented):
**Verdict: Request Changes**

**Clarification:** I agree with the goal of this PR. `camp init --skip-fest` should be removed, and Festival initialization should not be optional for normal users.

The blocking concern I intended to raise is test-harness isolation, not product behavior. The PR appears to add `CAMP_INIT_SKIP_FEST` for a test path, but that env var only guards the command-layer Festival init call. `scaffold.Init` still has its own Festival init path, so a non-container test can still invoke a local `fest` binary even when the env var is set.

**Key Finding:** Do not leave a partial test-only skip mechanism that only covers one of two Festival init paths. Either remove that env var and move the filesystem-mutating coverage into the established containerized integration harness with `fest`/`PATH` controlled there, or make the env var cover every internal Festival init path strictly for tests. The public CLI should not expose a skip flag.

**What’s Done Well:** The public CLI surface and generated docs are aligned with the right product direction: new campaigns should get Festival scaffolding by default, and users should not be asked to decide whether the methodology exists.

**Staff Standard:** I would not merge until the test isolation story is explicit and end-to-end. With that resolved, the public flag removal is the correct direction.

Verification run:
- `just build default-build`
- `go test ./cmd/camp/... ./internal/scaffold/...`
- `just docs` followed by clean `git status --short`
- Manual fake-`fest` reproduction showing `CAMP_INIT_SKIP_FEST=1` still calls `fest init --path <campaign>` from `scaffold.Init`

internal/commands/workitem/workitem_integration_test.go
```diff
@@ -24,16 +24,21 @@ import (
 func TestIntegration_WorkitemEditorHandsOffTTY(t *testing.T) {
 	projectRoot := testProjectRoot(t)
 	campBinary := buildCampBinary(t, projectRoot)
+	festBinDir := buildFestBinaryDir(t, projectRoot)
 
 	tempRoot := t.TempDir()
```
obey-agent, line 29: Blocking: this still exercises a real filesystem-mutating `camp init` workflow on the host. The test builds `camp` and `fest` locally, creates a host `t.TempDir()`, runs `camp init` into it, adds an intent, writes an editor script, and drives a PTY against that host campaign. Building `fest` instead of bypassing it is the right product contract, but this kind of campaign creation/mutation belongs in the established containerized integration harness, not a host temp directory.

Please move this coverage into the gocontainer/testcontainers pattern already used under `tests/integration`, with `fest` installed/copied into that container PATH, and assert the real `camp init` -> `fest init` behavior there. The user-facing `--skip-fest` removal is correct; the test isolation story still needs to be containerized.

docs/cli-reference/camp_create.md
```diff
@@ -43,3 +43,4 @@ camp create <name> [flags]
 ### SEE ALSO
 
 * [camp](camp.md)	 - Campaign management CLI for multi-project AI workspaces
+
```
obey-agent, line 46: `git diff --check origin/main...HEAD` fails here with `new blank line at EOF`. Please regenerate/fix the doc so the whitespace gate is clean before merge.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The product direction is now clear and correct: remove public `camp init --skip-fest`, remove the hidden `CAMP_INIT_SKIP_FEST` bypass, and ensure `camp init` exercises the real Festival setup path. The new commit gets the behavior direction right by building/providing `fest` instead of skipping it.

**Key Findings:**
- Blocking: the updated workitem integration test still runs a real filesystem-mutating `camp init` workflow on the host using `t.TempDir()`. Per the project testing standard, campaign creation/repair/init workflows belong in the established gocontainer/testcontainers integration harness with `fest` installed in the container PATH. Inline comment posted on `internal/commands/workitem/workitem_integration_test.go`.
- Hygiene: `git diff --check origin/main...HEAD` fails on `docs/cli-reference/camp_create.md:46` with a new blank line at EOF. Inline comment posted.
- PR metadata: the PR body still describes `CAMP_INIT_SKIP_FEST` as retained, but the latest commit removes it. Please update the description so reviewers and future readers do not chase stale acceptance criteria.

**What’s Done Well:** The public CLI flag is gone, docs no longer advertise it, and the latest code no longer keeps a hidden env-var bypass. Requiring a real `fest` binary for the workflow under test is the right contract; it just needs to happen inside the container harness rather than on the developer machine.

**Staff Standard:** I would not merge this head yet because the test still violates the filesystem isolation rule. Once that test is moved into the containerized integration pattern and the whitespace/PR-body cleanup is done, the PR should be on the right track.

Verification:
- `rg -n "skip-fest|CAMP_INIT_SKIP_FEST|SkipFest" .` returns no matches.
- `git diff --check origin/main...HEAD` fails on `docs/cli-reference/camp_create.md:46`.
- I did not run the host-side workitem integration test because the issue is that it mutates host filesystem state.

PR #258, opened 2026-04-29
Always auto-commit dungeon moves

lancekrogers:
## Summary

- Remove the `--no-commit` escape hatch from `camp dungeon move`.
- Always run the dungeon crawl commit path after successful dungeon moves and docs routing.
- Stage moved destinations separately from tracked source deletions so untracked triage items can still auto-commit cleanly.
- Regenerate CLI reference docs and update the dungeon behavior rollout note.

## Why

Dungeon triage history is designed around automatic commits. Allowing `camp dungeon move --no-commit` makes it easy to bypass that audit trail, especially when routing workflow/explore work into docs.

## Validation

- `go test ./cmd/camp/dungeon ./internal/git/commit`
- `just docs`
- `go test -tags=integration ./tests/integration -run 'TestDungeonMove|TestDungeonCommands_NoDungeonContextError'`
- `git diff --check`

### DIFF
--- a/cmd/camp/dungeon/move.go
+++ b/cmd/camp/dungeon/move.go
@@ -141,6 +141,8 @@ func runDungeonMove(cmd *cobra.Command, args []string) error {
 	files := commit.NormalizeFiles(cmdCtx.CampaignRoot, destinationPaths...)
 	preStaged, err := stageTrackedMoveSourceDeletions(ctx, cmdCtx.CampaignRoot, sourcePaths)
 	if err != nil {
+		fmt.Printf("%s Move was applied on disk, but staging the source deletion failed.\n", ui.WarningIcon())
+		fmt.Printf("%s %v\n", ui.WarningIcon(), err)
 		return camperrors.Wrap(err, "staging move source deletions")
 	}
 	result := commit.Crawl(ctx, commit.CrawlOptions{
--- a/cmd/camp/dungeon/move.go
+++ b/cmd/camp/dungeon/move.go
@@ -1,6 +1,7 @@
 package dungeon
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"path/filepath"
@@ -8,6 +9,7 @@ import (
 
 	intdungeon "github.com/Obedience-Corp/camp/internal/dungeon"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/git"
 	"github.com/Obedience-Corp/camp/internal/git/commit"
 	"github.com/Obedience-Corp/camp/internal/ui"
 	"github.com/spf13/cobra"
@@ -21,6 +23,7 @@ var dungeonMoveCmd = &cobra.Command{
 Without --triage, moves an item already in the dungeon root to a status directory.
 With --triage, moves an item from the parent directory into the dungeon.
 With --triage and --to-docs, routes an item to an existing campaign-root docs/<subdirectory>.
+Moves are always auto-committed so dungeon history remains auditable.
 
 Statuses: completed, archived, someday
 
@@ -44,7 +47,6 @@ func init() {
 	flags := dungeonMoveCmd.Flags()
 	flags.Bool("triage", false, "Move from parent directory (not from dungeon root)")
 	flags.String("to-docs", "", "Route triage item into an existing campaign-root docs/<subdir> (requires --triage)")
-	flags.Bool("no-commit", false, "Don't create a git commit")
 }
 
 func runDungeonMove(cmd *cobra.Command, args []string) error {
@@ -52,7 +54,6 @@ func runDungeonMove(cmd *cobra.Command, args []string) error {
 
 	triageMode, _ := cmd.Flags().GetBool("triage")
 	toDocs, _ := cmd.Flags().GetString("to-docs")
-	noCommit, _ := cmd.Flags().GetBool("no-commit")
 
 	itemName := args[0]
 	status := ""
@@ -76,7 +77,8 @@ func runDungeonMove(cmd *cobra.Command, args []string) error {
 	svc := intdungeon.NewService(cmdCtx.CampaignRoot, cmdCtx.Dungeon.DungeonPath)
 
 	var description string
-	var movedPaths []string
+	var sourcePaths []string
+	var destinationPaths []string
 
 	if triageMode {
 		if toDocs != "" {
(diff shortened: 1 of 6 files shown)


obey-agent (commented):
## Staff Review — Always auto-commit dungeon moves

Verdict: **Needs changes.** The intent is sound and the implementation correctly threads `PreStaged` through `commit.Crawl`, but there is one user-visible failure-mode bug and a test-coverage gap that should be addressed before merge.

### Blocking

**`stageTrackedMoveSourceDeletions` failure does not warn that the move was already applied** (`cmd/camp/dungeon/move.go:142-145`).

By the time we reach line 142, `svc.MoveToDocs` / `svc.MoveToDungeonStatus` / `svc.MoveToDungeon` / `svc.MoveToStatus` has already mutated the working tree. If `git.FilterTracked` or `git.StageTrackedChanges` returns an error (transient git failure, missing git binary, broken index, lock contention), we return `camperrors.Wrap(err, "staging move source deletions")` without telling the user the disk move succeeded. They will see only:

```
staging move source deletions: ...
```

Compare to the `result.Err` branch at lines 159-162, which correctly emits:

```
Move was applied on disk, but auto-commit failed.
```

The pre-staging path needs the same warning. Otherwise users will assume the operation rolled back, retry, and double-move (or get an `ErrAlreadyExists` they cannot diagnose).

Suggested fix: emit the same warning before returning, e.g.:

```go
preStaged, err := stageTrackedMoveSourceDeletions(ctx, cmdCtx.CampaignRoot, sourcePaths)
if err != nil {
    fmt.Printf("%s Move was applied on disk, but staging the source deletion failed.\n", ui.WarningIcon())
    fmt.Printf("%s %v\n", ui.WarningIcon(), err)
    return camperrors.Wrap(err, "staging move source deletions")
}
```

### Significant

**Test coverage gap on the new behavior.** `TestDungeonMove_NoCommitFlagRemoved` only asserts the flag was unregistered. None of the integration tests in `tests/integration/dungeon_test.go` (which had `--no-commit` removed) actually assert that a commit was created with the expected message, that the source deletion was staged, or that the destination addition was committed. Without an end-to-end assertion the auto-commit path is structurally exercised but functionally unverified.

Recommended additions:
1. After each successful move in the integration tests, run `git log -1 --name-status` against the campaign repo and assert both the deletion of the source path and the addition of the destination path are in the same commit.
2. Add a unit-or-integration case for the failure path: simulate a broken git state and assert the user-facing message contains both the "Move was applied on disk" warning and the wrapped error.

### Notes

- `stageTrackedMoveSourceDeletions` is fine but cyclomatic complexity in `runDungeonMove` is creeping up. Consider extracting the `triageMode` branch into its own helper in a follow-up. Not blocking.
- Pre-existing `fmt.Errorf` calls remain (lines 66, 69, 122). The repo standard prefers `camperrors.Wrap`. Out of scope for this PR but worth a follow-up cleanup.
- The migration doc update (`docs/migrations/dungeon-behavior-upgrade-2026-03.md`) correctly removes `--no-commit` from the validation steps and adds the auto-commit confirmation. Good.

### Test Coverage Assessment

| Layer | Status |
|-------|--------|
| Unit (flag removal) | ✅ Present |
| Unit (auto-commit happy path) | ❌ Missing |
| Unit (staging failure → warning) | ❌ Missing |
| Integration (commit content assertion) | ⚠️ `--no-commit` removed but no assertion on commit existence |

Once the failure-path warning is added and at least one end-to-end test asserts the commit was actually created with the expected source deletion + destination addition, this is ready to ship.

obey-agent:
Validation update: `just test all` passed locally. Package phase: 65 packages, 2202/2202 tests passed. Integration phase: 1 suite, 209/209 tests passed.

obey-agent:
Addressed the review feedback in `1650c61`.

Changes made:
- Added the missing warning when the filesystem move succeeds but source-deletion pre-staging fails.
- Added containerized integration coverage for exact auto-commit contents: source deletion + destination addition in the same crawl commit.
- Added a containerized failure-path test that runs `/camp` with a `PATH` that cannot resolve `git`, proving the warning is emitted and the move remains applied on disk.

Validation:
- `just test integration-run 'TestDungeonMove_(ToStatus|TriageWithCommit_IncludesSourceDeletion|TriageStagingFailureWarnsMoveApplied|TriageDirectToStatusWithCommit|TriageToDocsDestination)$'`
- `go test ./cmd/camp/dungeon ./internal/git/commit`
- `git diff --check`
- `just test all` passed: package phase `65 packages`, `2202/2202`; integration phase `1 suite`, `210/210`.

obey-agent (commented):
## Staff Review (follow-up @ 1650c61) — Always auto-commit dungeon moves

Verdict: **Approve.** Both items from the prior review are addressed cleanly.

### What's resolved

1. **Failure-path warning when staging fails** — `cmd/camp/dungeon/move.go:144-145` now emits the same "Move was applied on disk, but staging the source deletion failed." warning before returning the wrapped error, matching the pattern at lines 159-162. Users will no longer mistake a staging failure for a no-op.

2. **End-to-end commit-content assertions** — `tests/integration/dungeon_test.go` now has an `assertLastDungeonMoveCommit` helper (lines 47-63) that walks the actual commit subject (`Crawl: dungeon crawl completed`), body, and `diff-tree --name-status` output. It is wired into the existing happy-path tests for status moves, triage-direct-to-status, and docs routing. The dated-status callers correctly use `assert.Regexp` against the returned diff for the `YYYY-MM-DD/<file>` pattern.

3. **New negative test for the failure path** — `TestDungeonMove_TriageStagingFailureWarnsMoveApplied` (lines 410-444) reproduces the exact scenario I flagged: it commits a tracked file, then runs the move with `PATH=/tmp/no-git` so `git.FilterTracked` cannot find the git binary. It asserts:
   - non-zero exit code
   - `"Moved tracked-broken.md"` (move applied)
   - `"Move was applied on disk, but staging the source deletion failed."` (the new warning)
   - `"staging move source deletions"` (the wrap)
   - destination path exists, source does not

   Solid coverage of the regression. The `PATH` trick is a clean way to force a deterministic git failure without monkey-patching the camp binary.

### Spot checks

- `tc.GitOutput` is the existing helper (used at `tests/integration/fresh_test.go:126`); no new infrastructure needed.
- Hardcoded subject `"Crawl: dungeon crawl completed"` matches `internal/git/commit/crawl.go:20`.
- Helper return value is the diff string, allowing additional `assert.Regexp` callers without refactoring it again later.

### Minor

Not blocking; pick up in a follow-up if convenient:

- The new `TestDungeonMove_TriageToDocsDestination` change adds `assert.Contains(t, output, "Committed")` — fine, but the literal string is duplicated across multiple tests. Worth a small `assertCommitted(output)` helper alongside `assertLastDungeonMoveCommit` if you anticipate adding more.
- Pre-existing `fmt.Errorf` calls in `move.go` (lines 66, 69, 122) remain. Out of scope here, but the project standard prefers `camperrors.Wrap`. Logging it as a follow-up.

LGTM, ready to merge once CI reports green (no checks visible on the branch right now).

PR #259, opened 2026-04-29
Fix workitem shell jumps

obey-agent:
Summary:
- make workitem selection output campaign-relative jump directories instead of absolute paths
- add shell-init handling for workitem/wi/workitems so interactive selection changes the parent shell directory
- cover relative output, shell wrapper generation, and shell-integrated TUI jumping

Tests:
- just fmt
- go test ./internal/commands/workitem/... ./internal/workitem/... ./internal/shell/...
- go test -tags=integration ./tests/integration -run Workitem
- go test ./cmd/camp/... ./internal/commands/workitem/... ./internal/workitem/... ./internal/shell/...
- just build-camp
- go test ./...


internal/shell/templates/bash.sh.tmpl
```diff
@@ -39,6 +39,43 @@ camp() {
...
+        esac
+      done
+      if [ "$passthrough" -eq 1 ]; then
+        command camp workitem "$@"
+        return
+      fi
+
+      local tmp root cmd_status
+      tmp=$(mktemp "${TMPDIR:-/tmp}/camp-workitem.XXXXXX") || return 1
+      command camp workitem "$@" --path-output "$tmp"
```
obey-agent, line 65: Correctness issue: this wrapper now appends `--path-output` even when the shell function itself is running non-interactively. With `shell-init bash` loaded, `camp workitem >/tmp/out` or `camp workitem | cat` silently selects the first item and `cd`s the caller into that work item with no output; before this change the binary preserved the non-interactive contract and returned `non-interactive use requires --json or --print flag`.

Please gate the shell-jump path on an interactive wrapper (`[ -t 0 ] && [ -t 1 ]` for bash/zsh, fish equivalent) and delegate to `command camp workitem "$@"` otherwise, so scripts and pipes keep the CLI semantics. The same guard should be applied to the zsh/fish templates, with a regression test for redirected or piped `camp workitem`.

internal/shell/templates/bash.sh.tmpl
```diff
@@ -39,6 +39,43 @@ camp() {
         fi
       fi
       ;;
+    workitem|wi|workitems)
+      shift
+      local arg passthrough
+      passthrough=0
+      for arg in "$@"; do
+        case "$arg" in
+          --help|-h|--json|--print)
```
obey-agent: Correctness issue: this pass-through check only recognizes exact `--json` and `--print` tokens, but Cobra also accepts boolean flag forms such as `--json=true` and `--print=true`. After loading shell-init, `camp workitem --json=true` gets rewritten with the hidden `--path-output` flag and fails with `--json and --path-output are mutually exclusive`; the direct binary emits JSON correctly.

Please handle the supported boolean flag forms before appending the shell-integration flag, and mirror the fix in the zsh/fish templates. A small shell-init regression test for `--json=true` would cover this path.

obey-agent (commented):
Verdict: Request Changes

Overview: This PR is pointed at the right user problem: workitem selection should be able to move the parent shell, and the command-side split between JSON, print, TUI, and hidden shell output is mostly clean. The current shell wrapper regresses non-interactive semantics, though, and also breaks supported Cobra boolean flag forms once shell-init is loaded.

Key Findings:
- Inline comment on `internal/shell/templates/bash.sh.tmpl`: redirected or piped `camp workitem` now silently selects the first item and changes the caller's directory because the wrapper always appends `--path-output`; it should only take the shell-jump path when the wrapper itself is interactive, and zsh/fish need the same guard.
- Inline comment on `internal/shell/templates/bash.sh.tmpl`: pass-through detection misses `--json=true` and `--print=true`, so valid direct binary invocations fail through shell-init with a hidden `--path-output` conflict; zsh/fish need the same flag handling.

What's Done Well: The command-level `selectedJumpPath` behavior is small and easy to reason about, and the integration coverage for the intended bash jump flow is a good start. The path-output file approach avoids command-substitution TTY problems for the interactive TUI.

Verification: `go test ./internal/commands/workitem ./internal/shell ./internal/workitem/...` passed in a temporary PR worktree. `go test ./cmd/camp/...` failed in existing leverage tests with `not in a campaign`, which I did not attribute to this PR. I also built the PR binary and reproduced both shell-wrapper regressions locally.

Staff Standard: I would not submit this as-is because the shell integration changes behavior in scripts/pipes and breaks valid CLI flag syntax. Once those wrapper edge cases are guarded and covered, the core approach looks reasonable.

obey-agent:
Addressed both review threads.

Changes:
- Shell workitem jump mode now only runs when stdin and stdout are TTYs; redirected/piped `camp workitem` delegates to the binary and preserves non-interactive CLI behavior.
- Workitem pass-through now recognizes Cobra boolean forms: `--json=*` and `--print=*` across bash, zsh, and fish.
- Added bash/zsh/fish shell-init regression coverage for non-interactive delegation and boolean flag pass-through.

Verification:
- `just fmt`
- `go test ./internal/shell/... ./internal/commands/workitem/...`
- `go test -tags=integration ./tests/integration -run 'ShellInit_.*Workitem|Workitem'`
- `go test ./...`
- `just build-camp`

PR #264, opened 2026-04-30
test: stop leverage unit tests from writing to host campaign

lancekrogers:
## Summary

`just test unit` was silently dirtying the host campaign's `.campaign/leverage/snapshots/` on every run. Five tests in `cmd/camp/leverage/main_command_test.go` drove the live cobra `RunE` through `executeLeverage(t)`, which calls `campaign.DetectCached` to walk up to the real campaign root and then persists today's snapshots there. Only the `sampled_at` timestamp changed on each run, which is why this hid for so long — the leverage scores looked stable but the files kept showing up dirty in `git status`.

The fix moves the five offenders into `tests/integration/leverage_test.go` where they run against a fresh ephemeral campaign inside the shared alpine container. The remaining unit tests either drive `runLeverageDir` (which never touches the snapshot store) or are pure function-level tests with a mocked `SnapshotStorer`, so they stay put.

### What moved

| Removed from unit tests | Now in integration tests |
| --- | --- |
| `TestLeverageCommand_TableOutput` | `TestLeverage_TableOutput` |
| `TestLeverageCommand_JSONOutput` | `TestLeverage_JSONOutput` |
| `TestLeverageCommand_ProjectFilter` | `TestLeverage_ProjectFilter` |
| `TestLeverageCommand_RunnerError` | `TestLeverage_RunnerError` |
| `TestLeverageConfigCommand_Display` | `TestLeverage_ConfigDisplay` |

### Harness changes

- `tests/integration/helpers.go`: added `buildSCCBinaryShared()`, which `go build`s `github.com/boyter/scc/v3@latest` for `GOOS=linux` and copies it to `/usr/local/bin/scc` in the shared container (same pattern as the existing camp/fest binary build).
- `tests/integration/main_test.go`: added a package-level `sccAvailable` flag mirroring `festAvailable`. The new leverage tests skip if scc didn't build.
- `TestLeverage_RunnerError` swaps the real scc binary for a failing stub so `exec.LookPath("scc")` still resolves but every `Run()` exits non-zero, exercising the same per-project `Warning: skipping` branch the original mock test covered.

### Verification

```
$ stat -f %Sm .campaign/leverage/snapshots/camp/2026-04-29.json   # before
Apr 30 04:36:04 2026
$ just test unit                                                  # ALL 2201 PASS, 80s
$ stat -f %Sm .campaign/leverage/snapshots/camp/2026-04-29.json   # after — unchanged
Apr 30 04:36:04 2026
```

Integration suite (`just test integration-run TestLeverage`):

```
--- PASS: TestLeverage_TableOutput (3.88s)
--- PASS: TestLeverage_JSONOutput  (1.55s)
--- PASS: TestLeverage_ProjectFilter (1.76s)
--- PASS: TestLeverage_RunnerError (1.65s)
--- PASS: TestLeverage_ConfigDisplay (1.15s)
ok  	github.com/Obedience-Corp/camp/tests/integration	27.708s
```

## Test plan

- [x] `just build default-build` — clean
- [x] `just test unit` — all 2201 tests pass
- [x] `just test integration-run TestLeverage` — all 5 tests pass
- [x] Host `.campaign/leverage/snapshots/` mtimes unchanged after `just test unit`
- [x] `gofmt -l` clean on touched files

### DIFF
--- a/tests/integration/helpers.go
+++ b/tests/integration/helpers.go
@@ -118,6 +118,23 @@ func NewSharedContainer() (*TestContainer, error) {
 		}
 	}
 
+	// Build and copy scc binary (best-effort — scc is required by leverage tests
+	// only). scc is a third-party Go binary at github.com/boyter/scc/v3 that the
+	// `camp leverage` command shells out to via PATH lookup.
+	sccBinary, err := buildSCCBinaryShared()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "WARN: scc binary not available: %v\n", err)
+		sccAvailable = false
+	} else {
+		defer os.RemoveAll(filepath.Dir(sccBinary))
+		if err := container.CopyFileToContainer(ctx, sccBinary, "/usr/local/bin/scc", 0o755); err != nil {
+			fmt.Fprintf(os.Stderr, "WARN: failed to copy scc binary into container: %v\n", err)
+			sccAvailable = false
+		} else {
+			sccAvailable = true
+		}
+	}
+
 	// Install git (required for project operations)
 	exitCode, output, err := container.Exec(ctx, []string{"apk", "add", "--no-cache", "git"})
 	if err != nil {
@@ -244,6 +261,42 @@ func buildFestBinaryShared() (string, error) {
 	cmd := fmt.Sprintf("cd %s && GOOS=linux GOARCH=%s go build -o %s ./cmd/fest", festRoot, runtime.GOARCH, binaryPath)
 	if err := runCommand(cmd); err != nil {
 		return "", fmt.Errorf("failed to build fest binary: %w", err)
+	}
+
+	return binaryPath, nil
+}
+
+// buildSCCBinaryShared builds the third-party scc binary into a temp directory
+// and returns its path. scc is required by the `camp leverage` command (it
+// shells out to it for source-code counting). We build it on the host with
+// GOOS=linux so the resulting binary runs in the alpine container, matching
+// the camp/fest binary build pattern above.
+//
+// Returns ("", error) if the build fails — callers should treat this as
+// non-fatal since scc is only required by leverage tests.
+func buildSCCBinaryShared() (string, error) {
+	binDir, err := os.MkdirTemp("", "scc-integration-bin-*")
+	if err != nil {
+		return "", fmt.Errorf("failed to create temporary scc binary directory: %w", err)
+	}
+
+	// Build scc from a throwaway Go module so we don't pollute the camp module.
+	// `go install` with GOOS != host OS ignores GOBIN, so we use a tmp module
+	// + `go build` pointing at the package import path.
+	modDir := filepath.Join(binDir, "buildmod")
+	if err := os.MkdirAll(modDir, 0o755); err != nil {
+		return "", fmt.Errorf("failed to create scc build module dir: %w", err)
+	}
+
+	binaryPath := filepath.Join(binDir, "scc")
+	cmd := fmt.Sprintf(
+		"cd %s && go mod init sccbuild >/dev/null 2>&1 && "+
+			"go get github.com/boyter/scc/v3@latest && "+
+			"GOOS=linux GOARCH=%s go build -o %s github.com/boyter/scc/v3",
+		modDir, runtime.GOARCH, binaryPath,
+	)
+	if err := runCommand(cmd); err != nil {
+		return "", fmt.Errorf("failed to build scc binary: %w", err)
 	}
 
 	return binaryPath, nil
--- a/tests/integration/helpers.go
+++ b/tests/integration/helpers.go
@@ -118,7 +118,7 @@ func NewSharedContainer() (*TestContainer, error) {
 		}
 	}
 
-	// Build and copy scc binary (best-effort — scc is required by leverage tests
+	// Build and copy scc binary (best-effort; scc is required by leverage tests
 	// only). scc is a third-party Go binary at github.com/boyter/scc/v3 that the
 	// `camp leverage` command shells out to via PATH lookup.
 	sccBinary, err := buildSCCBinaryShared()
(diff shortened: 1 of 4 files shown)


tests/integration/helpers.go
```diff
@@ -249,6 +266,42 @@ func buildFestBinaryShared() (string, error) {
...
+	// + `go build` pointing at the package import path.
+	modDir := filepath.Join(binDir, "buildmod")
+	if err := os.MkdirAll(modDir, 0o755); err != nil {
+		return "", fmt.Errorf("failed to create scc build module dir: %w", err)
+	}
+
+	binaryPath := filepath.Join(binDir, "scc")
+	cmd := fmt.Sprintf(
+		"cd %s && go mod init sccbuild >/dev/null 2>&1 && "+
+			"go get github.com/boyter/scc/v3@latest && "+
```
obey-agent, line 301: Please pin this to a known `scc` version instead of resolving `@latest` at test time. The latest commit now makes `scc` build failures fail loudly instead of skipping the migrated leverage tests, which fixes the silent-coverage-loss part. The remaining issue is that this regular PR-gate coverage still depends on live upstream `scc` state: a future release can change behavior/schema or break unrelated Camp PRs without any repo change. Pin this harness to an explicit supported version, and track `@latest` separately if we want an upstream compatibility canary.

obey-agent (commented):
**Verdict: Request Changes**

**Overview**
This PR correctly removes the host-mutating leverage command tests from the unit suite and rehomes the behavior coverage in the integration harness. The remaining unit coverage is scoped to pure validation/snapshot/output behavior, and the migrated container tests exercise the table, JSON, project filter, runner warning, and config-display paths against an ephemeral campaign.

**Key Findings**
- `tests/integration/helpers.go:294`: please pin the `scc` version instead of resolving `github.com/boyter/scc/v3@latest` during the test run. The new integration coverage should not change or skip based on upstream state on the day CI runs.

**What's Done Well**
- The host campaign mutation was identified at the actual Cobra execution boundary and removed from the unit suite instead of papering over the symptom.
- The migrated tests use the existing shared container harness and validate both successful output paths and the `scc` failure warning path.
- `just test unit` passed from the real `projects/camp` path, and the checked leverage snapshot checksum/mtime stayed unchanged. `just test integration-run TestLeverage` also passed.

**Staff Standard**
I would be proud to submit this once the third-party tool version is pinned. The shape of the fix is right; the remaining issue is test determinism for the newly important integration dependency.

tests/integration/helpers.go
```diff
@@ -249,6 +266,49 @@ func buildFestBinaryShared() (string, error) {
 	return binaryPath, nil
 }
 
+// buildSCCBinaryShared builds the third-party scc binary into a temp directory
+// and returns its path. scc is required by the `camp leverage` command (it
+// shells out to it for source-code counting). We build it on the host with
+// GOOS=linux so the resulting binary runs in the alpine container, matching
+// the camp/fest binary build pattern above.
+//
+// We deliberately use @latest rather than pinning a version. Real users
```
obey-agent, line 275: Update after `08e529b`: the silent-skip path is fixed because leverage integration tests now fail when `scc` cannot be built. My remaining concern is the `@latest` policy itself. This is regular PR-gate coverage for moved command tests, not a quarantined upstream-compatibility canary; a future upstream release can still change the tested behavior or fail unrelated Camp PRs with no repo change. Please pin the harness to an explicit supported `scc` version, and add a separate latest/canary job if we want drift detection.

obey-agent (commented):
**Verdict: Request Changes**

**Overview**
I re-reviewed the latest head `ab45a68dac288a4ec7e8d5db4a3704119bddc4fd`. The core fix remains sound: the host-mutating leverage command cases have been moved out of the unit suite and now run against an ephemeral integration campaign. The latest push only documents the choice to build `scc` from `@latest`; it does not resolve the prior determinism concern.

**Key Findings**
- `tests/integration/helpers.go:275`: regular PR-gate integration coverage should pin the `scc` version. Using `@latest` makes unrelated Camp PRs depend on upstream `scc` state, while build failures skip the moved leverage coverage entirely via `sccAvailable=false`. If we want an upstream compatibility canary, that should be separate from the deterministic harness for this PR.

**What's Done Well**
- The unit suite now avoids the previous host `.campaign/leverage` snapshot mutation path.
- The migrated integration tests cover table output, JSON output, project filtering, per-project `scc` runner warnings, and config display in a containerized campaign.
- Verification passed on the latest head: `just build default-build`, `just test unit` (2204/2204), and `just test integration-run TestLeverage`. The checked host snapshot checksum/mtime stayed unchanged across the unit run.

**Staff Standard**
I would be proud to submit the test relocation and coverage shape. I would not merge this PR with a floating third-party test dependency in the regular gate; pin that dependency first.

lancekrogers:
Going to push back hard here, because I think the framing of this review is wrong on the merits.

The premise of the review is "@latest is non-deterministic and a future upstream change can break unrelated Camp PRs." That is exactly what I want. scc is not a vendored library, it is a tool a real user installs themselves via `brew install scc` or `go install ...@latest`. If a future scc release breaks the CLI contract that `camp leverage` depends on, every Camp PR SHOULD see that the same day, not three months later when someone reaches for leverage and is surprised.

There is no such thing as a Camp PR that is "unrelated" to scc once leverage is shipped. Either leverage's contract with scc holds, or it does not. Pinning the test dep does not make leverage in production any more deterministic. It just hides the drift from the only people positioned to fix it.

The "track @latest separately as a canary" suggestion is more bureaucracy for the same signal. We have one integration surface that exercises the real CLI contract. That surface should reflect what users hit, not a museum copy.

The one substantive concern in the review is the silent skip: if scc-latest ever fails to build, `sccAvailable=false` skips the coverage instead of failing loudly. That is a real regression risk and I am addressing it separately by making the scc build failure fatal for the leverage tests rather than skip. That fix preserves @latest as policy and closes the only legitimate half of this review.

Not pinning. Please re-review with that lens.

obey-agent (commented):
**Verdict: Request Changes**

**Overview**
I re-reviewed PR #264 at current head `08e529b643d24cae66a9600ab4188ccde25cf490`. The main test relocation is still the right fix: the host-mutating leverage command scenarios are out of the unit suite and now run against an ephemeral integration campaign. The newest commit fixes the stale part of my previous review by making leverage integration tests fail loudly when `scc` cannot be built instead of silently skipping.

**Key Findings**
- `tests/integration/helpers.go`: the remaining issue is still the floating `github.com/boyter/scc/v3@latest` dependency in regular PR-gate coverage. Build failures now fail instead of skip, which is better, but unrelated Camp PRs can still break or change behavior due to upstream `scc` releases with no repo change. Pin this harness to an explicit supported `scc` version, and add a separate latest/canary job if we want drift detection.

**What's Done Well**
- The unit tests no longer exercise the live Cobra path that writes host leverage snapshots.
- The migrated integration tests cover table output, JSON output, project filtering, per-project `scc` runner warnings, and config display.
- `08e529b` improved the failure mode by preventing silent loss of leverage integration coverage when `scc` is unavailable.

**Verification**
- `just test unit` passed: 2204/2204.
- `just test integration-run TestLeverage` passed.

**Staff Standard**
I would be proud of the test relocation and the explicit failure mode. I would still pin the third-party tool version before merging this as regular gate coverage.

PR #265, opened 2026-04-30
fix(leverage): drop submodule entries shadowed by standalone clones

### LINKED ISSUE #263: Leverage score regression: double-counting projects that appear as both standalone projects and submodules
## Summary

`camp leverage` is double-counting work for repos that appear in the campaign both as standalone projects and as submodules of another project. For example, `obey-platform-monorepo` contains `camp`, `fest`, and other repos as submodules, and those same repos are also registered as top-level projects in the campaign. The leverage system used to deduplicate these but a regression has been introduced.

## Evidence

`.campaign/leverage/snapshots/` shows overlapping snapshot directories for the same underlying repos:

```
camp                              <- standalone project
obey-platform-monorepo            <- standalone project
obey-platform-monorepo@camp       <- camp as submodule of obey-platform-monorepo
obey-platform-monorepo@fest       <- fest as submodule
obey-platform-monorepo@festui
obey-platform-monorepo@guild-scaffold
obey-platform-monorepo@obey
obey-platform-monorepo@obey-hub
obey-platform-monorepo@obey-shared
obey-platform-monorepo@protos
fest                              <- standalone project
festival                          <- standalone project
festival@camp                     <- camp as submodule of festival
festival@fest                     <- fest as submodule of festival
```

The repos `camp` and `fest` (and others) end up scored more than once in the campaign aggregate.

## Expected Behavior

When a repo is registered as a standalone campaign project, it should not also be counted as a submodule contribution to another project's leverage score for the campaign-wide totals. The leverage system used to handle this dedup automatically.

## History

There is precedent for fixes in this area:
- `2c910e6 Fix campaign leverage double-counting authors across repos`
- `8b0e74b Fix author count deduplication using union-find on name and email`

`internal/leverage/campaign_effort.go` already dedupes authors by canonical `GitDir`, so the regression is likely upstream of that — either in project resolution (`internal/leverage/projects.go`) producing distinct `ResolvedProject` entries with different `GitDir` values for what is the same physical repo, or in the snapshot/aggregation layer combining per-project scores additively without dedup.

## Suggested Investigation

1. Confirm that `ResolvedProject.GitDir` resolves to the same canonical path for `projects/camp/.git` whether it's reached as a standalone project or via `projects/obey-platform-monorepo/<...>/camp` as a submodule. Symlink/canonicalisation differences would defeat the existing dedup.
2. Check whether the snapshot accumulator (under `.campaign/leverage/snapshots/`) sums per-project scores naively or whether it uses the same `CampaignActualPersonMonths` dedup path.
3. Decide whether the canonical answer should be: \"submodule snapshots are diagnostic only and never roll up into campaign totals when the same repo is also a standalone project\".

## Repro

In a campaign workspace where `obey-platform-monorepo` is registered alongside its submodule repos as standalone projects, run the standard leverage scoring command and inspect the campaign-wide actual person-months / author counts vs the per-project sums. Doubled values for shared submodule repos should be visible.

## Severity

Medium-high. Campaign-level effort metrics are user-facing and inform planning decisions; silent overcounting erodes trust in the score.

lancekrogers:
Fixes #263

## Summary

Campaign-wide leverage scoring was double-counting work for repos that appear both as standalone projects and as submodules of another project. For example, `obey-platform-monorepo` lists `camp`, `fest`, and others under its `.gitmodules`, and those repos are also registered as standalone projects under `projects/`. The same authors/LOC were being scored twice (once via the standalone clone and once via the submodule reference).

## Root cause

Commit `5d0b582` added a "skip dedup for monorepo subprojects" exception to `internal/project/list.go`'s `deduplicateByRemoteURL`. The motivation was correct — multiple submodules of the same monorepo would otherwise collapse onto each other — but the exception was over-broad. It worked because submodule entries were tagged with the **parent's** URL rather than their own, so the URL-based dedup could not actually distinguish them.

The result: when a submodule and a standalone clone of the same repo coexisted in the campaign (the common case for `camp`, `fest`, etc. alongside `obey-platform-monorepo`), both entries were preserved and both contributed to leverage totals.

## Fix

Two-layer change:

**1. `internal/project/list.go`** — each submodule entry now carries its **own** remote URL via `getGitRemoteURL(ctx, subFullPath)` instead of inheriting the parent monorepo's URL. This is the honest representation. `project.List` itself still keeps both `foo` and `mono@foo` so `camp fresh` and shell completion can target each checkout independently (`TestList_GitmodulesSubmoduleKeptAlongsideStandalone` preserved).

**2. `internal/leverage/projects.go`** — `resolveFromProjectList` runs a new `dropSubmodulesShadowedByStandalone` pass that removes submodule entries whose URL matches a standalone entry. The standalone wins because it is a direct clone with full history; submodule references may be pinned to older SHAs.

Submodule entries with no URL (uninitialised submodules, no remote configured) are kept unchanged — they cannot be shadowed.

## Tests

- `TestList_GitmodulesSubmoduleCarriesOwnURL` — new submodule entries carry their own remote URL.
- `TestDropSubmodulesShadowedByStandalone` (table) — unit coverage of the dedup helper across the shadow / no-shadow / empty-URL / no-op cases.
- `TestResolveProjects_FallbackDropsSubmoduleShadowedByStandalone` — end-to-end regression test for issue #263 using a real on-disk monorepo + `.gitmodules` + standalone child.

Existing tests preserved:
- `TestList_GitmodulesSubmoduleKeptAlongsideStandalone` — `project.List` still returns both forms.
- `TestList_DeduplicatesByRemoteURL` — standalone-vs-standalone dedup unchanged.
- `TestResolveProjects_FallbackMonorepoExpansion` — no-shadow monorepo still produces root + submodule entries.

## Test plan

- [x] `just test unit` — 2209/2209 pass
- [x] `just build default-build` — clean
- [ ] Manual verification on the obey campaign: `camp leverage` campaign totals should drop now that overlapping repos are not double-counted.

## Out of scope

The existing `.campaign/leverage/snapshots/<name>@<sub>/` directories on disk (e.g. `obey-platform-monorepo@camp`) will no longer be written to going forward, but historical snapshot files remain. Cleanup of those stale directories is a separate concern; happy to follow up if desired.

### DIFF
--- a/internal/leverage/projects.go
+++ b/internal/leverage/projects.go
@@ -61,11 +61,16 @@ func ResolveProjects(ctx context.Context, campaignRoot string, cfg *LeverageConf
 // resolveFromProjectList falls back to project.List() discovery.
 // Monorepo subprojects get SCCDir pointing to the subproject and GitDir
 // pointing to the monorepo root (where .git lives).
+//
+// Leverage scoring further dedups submodule entries against standalone
+// clones of the same git remote URL — see dropSubmodulesShadowedByStandalone.
 func resolveFromProjectList(ctx context.Context, campaignRoot string) ([]ResolvedProject, error) {
 	projects, err := project.List(ctx, campaignRoot)
 	if err != nil {
 		return nil, camperrors.Wrap(err, "list projects")
 	}
+
+	projects = dropSubmodulesShadowedByStandalone(projects)
 
 	resolved := make([]ResolvedProject, 0, len(projects))
 	for _, p := range projects {
@@ -94,6 +99,43 @@ func resolveFromProjectList(ctx context.Context, campaignRoot string) ([]Resolve
 	})
 
 	return resolved, nil
+}
+
+// dropSubmodulesShadowedByStandalone removes monorepo subproject
+// entries whose git remote URL matches a standalone (non-submodule)
+// project in the same list. This prevents leverage scoring from
+// double-counting a repo that appears both as `projects/foo` and as
+// `projects/<monorepo>/foo` via .gitmodules.
+//
+// `project.List` deliberately keeps both entries so that callers
+// like `camp fresh` can target each checkout by name. Leverage
+// scoring needs the opposite: a single canonical entry per repo.
+//
+// The standalone entry wins because it is a direct clone with full
+// history; submodule references may be pinned to older SHAs.
+//
+// Submodule entries with no URL (uninitialised submodule, no
+// remote configured) are kept unchanged — there is no standalone
+// to shadow them.
+func dropSubmodulesShadowedByStandalone(projects []project.Project) []project.Project {
+	standaloneURLs := make(map[string]struct{})
+	for _, p := range projects {
+		if p.MonorepoRoot != "" || p.URL == "" {
+			continue
+		}
+		standaloneURLs[p.URL] = struct{}{}
+	}
+
+	out := make([]project.Project, 0, len(projects))
+	for _, p := range projects {
+		if p.MonorepoRoot != "" && p.URL != "" {
+			if _, shadowed := standaloneURLs[p.URL]; shadowed {
+				continue
+			}
+		}
+		out = append(out, p)
+	}
+	return out
 }
 
 // resolveFromConfig resolves explicitly configured project entries.
--- a/internal/leverage/projects_test.go
+++ b/internal/leverage/projects_test.go
@@ -7,6 +7,8 @@ import (
 	"os/exec"
 	"path/filepath"
 	"testing"
+
+	"github.com/Obedience-Corp/camp/internal/project"
 )
 
 func TestResolveProjects_ConfigDriven(t *testing.T) {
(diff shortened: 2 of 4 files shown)


obey-agent (commented):
**Verdict: Approve**

**Overview**
I reviewed PR #265 at head `6e0d47cf6d3e6804f298dc51249d6ea8e3292d32`. The fix is well-scoped: `project.List` now records each submodule entry's own remote URL, while leverage resolution applies the narrower scoring-specific dedup that drops submodule entries shadowed by standalone clones. That keeps project operations able to target both checkouts while preventing campaign leverage totals from double-counting the same repo.

**Key Findings**
- No blocking findings.

**What's Done Well**
- The ownership boundary is right: generic project listing preserves both entries, and leverage applies its own canonicalization for scoring.
- The tests cover the helper behavior and the end-to-end fallback path with a standalone clone plus a monorepo submodule sharing the same remote.
- Verification passed for the changed surface: `gofmt -l` clean, `git diff --check` clean, `go test ./internal/project ./internal/leverage`, and `just build default-build`.

**Verification Caveat**
- `just test unit` from the detached `/tmp` review worktree failed in `cmd/camp/leverage` because this branch still contains the existing campaign-root-sensitive leverage command tests that PR #264 is addressing. I treated that as pre-existing/out-of-scope for this PR because the changed packages pass and the failure is not in this diff.

**Staff Standard**
Yes. I would be comfortable submitting this change as-is, with the caveat that PR #264 should land to make the full unit gate portable outside the campaign workspace.

PR #267, opened 2026-05-01
superseded: fix leverage repo URL dedup

lancekrogers:
## Summary

- Deduplicate leverage scoring projects by non-empty git remote URL, not only by standalone-vs-submodule shadowing.
- Apply the same leverage-specific dedup during config population and auto-detection so stale duplicate entries are pruned before scoring.
- Add regression coverage for duplicate submodule URLs across multiple monorepos and config pruning.

## Root Cause

The previous fix only removed a monorepo submodule entry when a standalone clone with the same remote URL existed. If the same repository appeared as submodules in multiple monorepos, leverage still resolved and scored each checkout separately, which double-counted COCOMO estimates.

## Validation

- `go test ./internal/leverage ./internal/project ./cmd/camp/leverage`
- `just test unit` (68 packages, 2283 tests)
- `just build-camp`

### DIFF
--- a/internal/leverage/config.go
+++ b/internal/leverage/config.go
@@ -75,6 +75,7 @@ func AutoDetectConfig(ctx context.Context, campaignRoot string) (*LeverageConfig
 	if err != nil {
 		return nil, camperrors.Wrap(err, "listing projects")
 	}
+	projects = deduplicateProjectsForLeverage(projects)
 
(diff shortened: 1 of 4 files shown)


lancekrogers:
Closing in favor of a replacement PR with a neutral branch and title.

PR #270, opened 2026-05-04
Add camp attach/detach for non-project symlinks

lancekrogers:
## Summary

Closes a bug where `camp pin` and other context-aware commands fail from arbitrary directories the user has symlinked into the campaign tree. Adds a small command surface (`camp attach` / `camp detach`) for binding non-project directories to a campaign, mirroring the linked-project marker pattern but without project ceremony.

- New `camp attach <path>` writes a `Kind="attachment"` `.camp` marker at the symlink's resolved target. Refuses paths inside a campaign, refuses to overwrite project markers (even with `--force`), follows symlinks once.
- New `camp detach <path>` removes attachment markers. Refuses on project markers (those use `camp project unlink`).
- `LinkMarker` schema bumped to V3 with a `Kind` field. Pre-V3 markers default to `KindProject` on read for backward compatibility.
- `Pin` gains an `AbsPath` field. `runPin` now accepts paths outside the campaign root when they resolve through an attachment marker pointing at the active campaign; otherwise the existing rejection stands.
- `camp pins` listing gains a `KIND` column. `cgo` / pin lookup resolves attachment pins via `AbsPath`.
- Linked-project writes now stamp `Kind: KindProject` explicitly.
- README updated with the attach workflow.

Design: [`workflow/design/non-project-symlinks/`](https://github.com/Obedience-Corp/obey-campaign/tree/main/workflow/design/non-project-symlinks) (in the campaign repo).

## Test plan

- [x] Unit tests pass (2298/2298, +9 new in `internal/attach`, +marker/pin coverage)
- [x] `golangci-lint run --new-from-rev=origin/main` is clean
- [x] End-to-end smoke test: created symlink into a tmp dir, `camp attach` it, `cd` into target, `camp pin` succeeds, `camp pins` shows `attachment` row, `camp unpin` removes it
- [x] Linked-project flow unchanged (existing `internal/project` tests pass)
- [x] Backward compat: v1/v2 markers on disk read as `Kind: project` (covered by `TestLinkMarker_LegacyDefaultsToProject`)

## What's not in this PR

- `camp doctor` warning for orphaned attachments (deferred per design Step 6)
- `camp attachments` listing command (deferred per `05-open-questions.md` Q2)
- Pre-existing lint issues elsewhere in the codebase

### DIFF
--- a/cmd/camp/attach/resolver.go
+++ b/cmd/camp/attach/resolver.go
@@ -6,8 +6,8 @@ import (
 
 	"github.com/Obedience-Corp/camp/cmd/camp/cmdutil"
 	"github.com/Obedience-Corp/camp/internal/config"
-	navtui "github.com/Obedience-Corp/camp/internal/nav/tui"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	navtui "github.com/Obedience-Corp/camp/internal/nav/tui"
 )
 
 // attachResolver picks a target campaign for camp attach. It mirrors the
--- a/cmd/camp/attach/resolver.go
+++ b/cmd/camp/attach/resolver.go
@@ -1,0 +1,91 @@
+package attach
+
+import (
+	"context"
+	"io"
+
+	"github.com/Obedience-Corp/camp/cmd/camp/cmdutil"
+	"github.com/Obedience-Corp/camp/internal/config"
+	navtui "github.com/Obedience-Corp/camp/internal/nav/tui"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+)
+
+// attachResolver picks a target campaign for camp attach. It mirrors the
+// linked-project resolver but skips project-specific registry checks because
+// attachments are not registered as projects.
+type attachResolver struct {
+	stderr        io.Writer
+	usageLine     string
+	isInteractive func() bool
+	loadCurrent   func(context.Context) (*config.CampaignConfig, string, error)
+	loadRegistry  func(context.Context) (*config.Registry, error)
+	loadCampaign  func(context.Context, string) (*config.CampaignConfig, error)
+	saveRegistry  func(context.Context, *config.Registry) error
+	pickCampaign  func(context.Context, *config.Registry) (config.RegisteredCampaign, error)
+}
+
+// NewResolver returns a CampaignResolver suitable for camp attach.
+func NewResolver(stderr io.Writer, usageLine string) CampaignResolver {
+	return attachResolver{
+		stderr:        stderr,
+		usageLine:     usageLine,
+		isInteractive: navtui.IsTerminal,
+		loadCurrent:   config.LoadCampaignConfigFromCwd,
+		loadRegistry:  config.LoadRegistry,
+		loadCampaign:  config.LoadCampaignConfig,
+		saveRegistry:  config.SaveRegistry,
+		pickCampaign:  cmdutil.PickCampaign,
+	}
+}
+
+func (r attachResolver) Resolve(ctx context.Context, targetCampaign string, targetChanged bool) (*config.CampaignConfig, string, error) {
+	if targetCampaign == NoOptCampaign {
+		targetCampaign = ""
+	}
+
+	if !targetChanged {
+		cfg, root, err := r.loadCurrent(ctx)
+		if err == nil {
+			return cfg, root, nil
+		}
+	}
+
+	reg, err := r.loadRegistry(ctx)
+	if err != nil {
+		return nil, "", camperrors.Wrap(err, "load registry")
+	}
+	if reg.Len() == 0 {
+		return nil, "", camperrors.Wrap(camperrors.ErrNotInitialized,
+			"no campaigns registered (use 'camp init' to create one)")
+	}
+
+	var selected config.RegisteredCampaign
+	switch targetCampaign {
+	case "":
+		if !r.isInteractive() {
+			return nil, "", camperrors.Wrapf(camperrors.ErrInvalidInput,
+				"campaign name required in non-interactive mode\n       Usage: %s", r.usageLine)
+		}
+		selected, err = r.pickCampaign(ctx, reg)
+		if err != nil {
+			return nil, "", err
+		}
+	default:
+		selected, err = cmdutil.ResolveCampaignSelection(targetCampaign, reg, r.stderr)
+		if err != nil {
+			return nil, "", err
+		}
+	}
+
+	cfg, err := r.loadCampaign(ctx, selected.Path)
+	if err != nil {
+		return nil, "", camperrors.Wrapf(err, "load target campaign %s", selected.Path)
+	}
+
+	reg.UpdateLastAccess(selected.ID)
+	if r.saveRegistry != nil {
+		_ = r.saveRegistry(ctx, reg)
+	}
+
+	return cfg, selected.Path, nil
+}
--- a/docs/cli-reference/camp_attach.md
+++ b/docs/cli-reference/camp_attach.md
@@ -41,4 +41,3 @@ camp attach <path> [flags]
 ### SEE ALSO
 
 * [camp](camp.md)	 - Campaign management CLI for multi-project AI workspaces
-
--- a/docs/cli-reference/camp_attach.md
+++ b/docs/cli-reference/camp_attach.md
@@ -1,0 +1,44 @@
+## camp attach
+
+Attach an external directory to a campaign
+
+### Synopsis
+
+Attach a non-project directory to a campaign by writing a .camp marker.
+
+The user manages the symlink (if any). camp attach only writes the marker at
+the resolved target so commands run from inside that directory know which
+campaign owns it.
+
+If the target is reached through a symlink, camp follows it once and writes
+the marker at the final directory.
+
+Examples:
+  camp attach ai_docs/examples/external-repo
+  camp attach ~/scratch/notes-link
+  camp attach /abs/path/to/dir --campaign platform
+
+```
+camp attach <path> [flags]
+```
+
+### Options
+
+```
+  -c, --campaign string   Target campaign by name or ID; defaults to current campaign or interactive picker
+      --force             Overwrite an existing attachment marker
+  -h, --help              help for attach
+```
+
+### Options inherited from parent commands
+
+```
+      --config string   config file (default: ~/.obey/campaign/config.json)
+      --no-color        disable colored output
+      --verbose         enable verbose output
+```
+
+### SEE ALSO
+
+* [camp](camp.md)	 - Campaign management CLI for multi-project AI workspaces
+
--- a/internal/attach/attach.go
+++ b/internal/attach/attach.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/Obedience-Corp/camp/internal/campaign"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/git"
 )
 
 // Options controls Attach behavior.
@@ -23,7 +24,7 @@ type Options struct {
 	Force bool
 }
 
-// Result describes the outcome of a successful attach.
+// Result describes the outcome of a successful attach or detach.
 type Result struct {
 	// Input is the path the user passed (for display purposes).
 	Input string
@@ -33,6 +34,12 @@ type Result struct {
 	CampaignID string
 	// FollowedSymlink is true when Input != Target.
 	FollowedSymlink bool
+	// GitExcludeUpdated is true when the operation also updated the
+	// target repo's .git/info/exclude file.
+	GitExcludeUpdated bool
+	// GitExcludeWarning is non-empty when a best-effort exclude update
+	// failed; the marker write/remove still succeeded.
+	GitExcludeWarning string
 }
 
 // Attach writes a Kind="attachment" marker at the resolved target of input,
@@ -40,8 +47,14 @@ type Result struct {
 //
 // Refuses if:
 //   - input does not resolve.
-//   - target is already inside a campaign root.
+//   - target is already inside any campaign tree (the selected one or
+//     another). A nested marker would shadow the parent .campaign/ during
+//     detection.
 //   - target already has a marker (unless opts.Force).
+//
+// When the target is a Git working tree, .camp is added to that repo's
+// .git/info/exclude on a best-effort basis so campaign-local state is not
+// accidentally committed to the attached repo.
 func Attach(ctx context.Context, campaignRoot, campaignID, input string, opts Options) (*Result, error) {
 	if ctx.Err() != nil {
 		return nil, ctx.Err()
@@ -66,13 +79,13 @@ func Attach(ctx context.Context, campaignRoot, campaignID, input string, opts Op
 		return nil, camperrors.Wrapf(camperrors.ErrInvalidInput, "attach target %q is not a directory", target)
 	}
 
-	if isUnderCampaignRoot(target, campaignRoot) {
-		return nil, camperrors.Wrapf(camperrors.ErrInvalidInput,
-			"target %q is already inside a campaign tree; attach is for external directories", target)
-	}
-
+	// Check for an existing marker at target first. A self-attached
+	// directory will detect its own marker via campaign.Detect, so we
+	// must distinguish that case before the any-campaign check.
 	markerPath := campaign.MarkerPath(target)
+	hasExistingMarker := false
 	if existing, readErr := campaign.ReadMarker(target); readErr == nil {
+		hasExistingMarker = true
 		if !opts.Force {
 			return nil, camperrors.Wrapf(camperrors.ErrAlreadyExists,
 				"marker already exists at %q (kind=%q); use --force to overwrite", markerPath, existing.Kind)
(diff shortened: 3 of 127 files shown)


internal/attach/attach.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	info, err := os.Stat(target)
+	if err != nil {
+		return nil, camperrors.Wrapf(err, "stat target %q", target)
+	}
+	if !info.IsDir() {
+		return nil, camperrors.Wrapf(camperrors.ErrInvalidInput, "attach target %q is not a directory", target)
+	}
+
+	if isUnderCampaignRoot(target, campaignRoot) {
```
obey-agent: This only checks the target against the selected campaign root. A path inside any other campaign tree will pass, then the new `.camp` file in that directory wins before the parent `.campaign` during detection, effectively hijacking context for part of another campaign. Since `attach` is documented as refusing paths inside a campaign, please detect/walk for any existing campaign context on the resolved target and reject it, not just the selected root.

internal/attach/attach.go
```diff
@@ -0,0 +1,170 @@
...
+	} else if !errors.Is(readErr, os.ErrNotExist) && !os.IsNotExist(readErr) {
+		return nil, camperrors.Wrapf(readErr, "read existing marker at %q", markerPath)
+	}
+
+	marker := campaign.LinkMarker{
+		Version:          campaign.LinkMarkerVersion,
+		Kind:             campaign.KindAttachment,
+		ActiveCampaignID: campaignID,
+	}
+	if err := campaign.WriteMarker(target, marker); err != nil {
```
obey-agent, line 120: `camp attach` writes campaign-local state into the external target, and the README explicitly calls out reference repos. For Git targets this leaves `.camp` as an untracked file and creates a real risk that campaign metadata gets committed to the attached repo. Please mirror the linked-project behavior by adding `.camp` to the target repo `.git/info/exclude` when the target is a Git repo, and remove that exclude entry on detach when this attachment marker is removed.

cmd/camp/attach/resolver.go
```diff
@@ -0,0 +1,91 @@
+package attach
+
+import (
+	"context"
+	"io"
+
+	"github.com/Obedience-Corp/camp/cmd/camp/cmdutil"
+	"github.com/Obedience-Corp/camp/internal/config"
+	navtui "github.com/Obedience-Corp/camp/internal/nav/tui"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
```
obey-agent, line 9: This file is not gofmt-clean. `gofmt -l cmd/camp/attach/resolver.go` reports this path; the diff is just the import ordering around `internal/errors` and `internal/nav/tui`. Please run gofmt before merge so the basic formatting gate is clean.

docs/cli-reference/camp_attach.md
```diff
@@ -0,0 +1,44 @@
...
+
+```
+      --config string   config file (default: ~/.obey/campaign/config.json)
+      --no-color        disable colored output
+      --verbose         enable verbose output
+```
+
+### SEE ALSO
+
+* [camp](camp.md)	 - Campaign management CLI for multi-project AI workspaces
```
obey-agent, line 43: `git diff --check origin/main...HEAD` reports a new blank line at EOF here, and the same issue in `docs/cli-reference/camp_detach.md`. Please regenerate or trim the generated CLI docs so diff-check passes cleanly.

obey-agent (commented):
Verdict: Request Changes

Overview: This PR is well-scoped and the marker/pin changes are generally integrated with the existing campaign detection model. The test coverage is solid, including legacy marker compatibility and attachment pin persistence. I am requesting changes because two filesystem-safety issues can leak or override campaign context in real user workflows.

Key Findings:
- `internal/attach/attach.go:69` only rejects paths inside the selected campaign root. Attaching a directory inside another campaign would write a `.camp` marker that wins before the parent `.campaign` during detection.
- `internal/attach/attach.go:94` writes `.camp` into external Git repos without adding it to `.git/info/exclude`, unlike linked projects. That leaves campaign-local state untracked and easy to accidentally commit.
- Formatting gates need cleanup: `gofmt -l` reports `cmd/camp/attach/resolver.go`, and `git diff --check origin/main...HEAD` reports blank lines at EOF in the generated attach/detach CLI docs.

What is Done Well:
- The V3 marker schema keeps legacy markers defaulting to `project`, which protects existing linked-project behavior.
- The attachment pin storage is explicit with `AbsPath`, and the targeted unit coverage for marker reads, detection, and pin round-tripping is useful.
- `golangci-lint run --new-from-rev=origin/main` is clean.

Staff Standard: Not yet. I would be comfortable with the direction, but I would not merge this until attach refuses any existing campaign context on the resolved target and handles Git exclude cleanup for `.camp` markers.

Verification:
- `go test ./internal/attach ./internal/campaign ./internal/pins ./cmd/camp/navigation` passed.
- `just test all` passed: 2,298 package tests and 229 integration tests.
- `golangci-lint run --new-from-rev=origin/main` passed.
- `gofmt -l ...` failed for `cmd/camp/attach/resolver.go`.
- `git diff --check origin/main...HEAD` failed for blank EOF lines in `docs/cli-reference/camp_attach.md` and `docs/cli-reference/camp_detach.md`.
- Full `just lint` still fails on pre-existing baseline lint debt outside the PR scope.

obey-agent:
Thanks for the review. All four findings addressed in commit 09eda68:

**1. `internal/attach/attach.go:69` — any-campaign rejection.**
`Attach` now calls `campaign.Detect` from the parent of the resolved target and refuses if it lands inside any campaign root, not just the selected one. The check runs after the marker-already-exists check so a re-attach (which previously had its own `.camp` and would self-detect) still gets the correct `ErrAlreadyExists`. New test: `TestAttach_RefusesPathInsideOtherCampaign` builds two campaigns and confirms a marker cannot be written inside the non-selected one.

**2. `internal/attach/attach.go:94` — Git exclude integration.**
When the target is a Git working tree, `Attach` now adds `.camp` to `.git/info/exclude`. `Detach` removes it symmetrically. Implemented via a new shared `internal/git` package (`EnsureInfoExclude`, `RemoveInfoExclude`, `IsRepo`) extracted from the linked-project helpers in `internal/project/link.go` so both surfaces share one implementation. Failures are surfaced via `Result.GitExcludeWarning` (best-effort; the marker write itself still succeeds). New tests: `TestAttach_AddsGitInfoExclude`, `TestDetach_RemovesGitInfoExclude`.

**3. `cmd/camp/attach/resolver.go` — gofmt.**
Ran gofmt; import block is now canonical.

**4. CLI doc trailing blank lines.**
Root-caused: `cobra/doc.GenMarkdownTree` emits files ending in `\n\n`. Added a `trimTrailingBlankLines` post-pass in `gendocs.go` that collapses trailing blank lines to a single `\n` for every `.md` in the output dir. Applied across all existing CLI reference docs in this commit so the codebase is consistent. `git diff --check origin/main...HEAD` is now clean.

**Verification:**
- 2301 unit tests pass (was 2298; +3 for the new behaviors).
- `golangci-lint run --new-from-rev=origin/main ./...` clean.
- `git diff --check origin/main...HEAD` clean.
- PR head SHA verified against local HEAD before this reply (09eda68).

obey-agent (commented):
Verdict: Approve with minor follow-ups.

Re-reviewed against PR head 09eda68. All four findings from the previous review are addressed cleanly, and the fixes hold up under inspection.

What I verified
- `internal/attach/attach.go:97-115` — any-campaign rejection now calls `campaign.Detect` from `filepath.Dir(target)`, which both reuses the rest of the CLI's detection path and avoids self-detecting an attachment marker at the target itself. The `selected` vs. `other` campaign error split is helpful for users.
- `internal/git/info_exclude.go` — `EnsureInfoExclude`/`RemoveInfoExclude`/`IsRepo` were extracted from `internal/project/link.go` and now serve both surfaces. `internal/project/link.go:381-388` is a thin wrapper over the new package, so the linked-project path is unchanged. Best-effort failure surfaces via `Result.GitExcludeWarning`, with the marker write still succeeding — matches the linked-project contract.
- `cmd/camp/gendocs.go` — `trimTrailingBlankLines` is a clean post-pass over the cobra/doc output dir; applied to all existing `*.md` so the codebase is consistent.
- `cmd/camp/attach/resolver.go` import block is now canonical.

Verification I ran
- `go test ./internal/attach/... ./internal/git/... ./internal/project/... ./internal/campaign/... ./internal/pins/...` — all pass.
- `gofmt -l ./internal/attach ./internal/git ./internal/project ./cmd/camp/attach ./cmd/camp/gendocs.go` — clean.
- `git diff --check origin/main...HEAD` — clean.
- `golangci-lint run --new-from-rev=origin/main ./internal/attach/... ./internal/git/... ./internal/project/... ./cmd/camp/...` — 0 issues.
- New tests `TestAttach_RefusesPathInsideOtherCampaign`, `TestAttach_AddsGitInfoExclude`, `TestDetach_RemovesGitInfoExclude` cover the regressions the prior review flagged.

Follow-ups (non-blocking)
1. `internal/attach/attach.go:104` — the `!hasExistingMarker` guard on the any-campaign check is wider than the comment justifies. Since `Detect` is called from `filepath.Dir(target)`, target's own marker is already invisible to the check; the guard isn't needed for self-skip. Its current effect is that `attach --force` on an existing attachment marker that lives inside a different campaign skips the shadowing check. Edge-casey but worth tightening — either drop the guard or limit it to the self-attach case (kind=attachment AND marker.ActiveCampaignID == campaignID).
2. `internal/attach/attach.go:185` / `git.RemoveInfoExclude` — the helper rewrites the exclude file unconditionally and the caller sets `GitExcludeUpdated = true` on any non-error return, even when `.camp` wasn't present. The CLI then prints "removed .camp from .git/info/exclude" misleadingly. Easy fix: have `RemoveInfoExclude` return a bool (or distinguish "removed" vs. "not present") and only set `GitExcludeUpdated` on actual removal.
3. `internal/project/link.go:381-388` — `ensureGitInfoExclude`/`removeGitInfoExclude` are now one-line wrappers over the new package. Could be removed and call sites updated to call `git.EnsureInfoExclude`/`git.RemoveInfoExclude` directly. Tidiness only.
4. `internal/git/info_exclude.go` has no direct unit tests; coverage is via `internal/attach` and `internal/project`. Acceptable since the helpers' previous home didn't have direct tests either, but worth a small `internal/git` test once more callers depend on it.

Staff Standard: Yes. The two security-relevant findings are properly closed; remaining items are cleanup, not correctness. Good to merge once the author has a chance to consider the follow-ups (or land them as a separate small PR).

internal/attach/attach.go
```diff
@@ -0,0 +1,218 @@
...
+		Target:          target,
+		CampaignID:      marker.EffectiveCampaignID(),
+		FollowedSymlink: target != abs,
+	}
+	if git.IsRepo(target) {
+		updated, err := git.RemoveInfoExclude(ctx, target, campaign.LinkMarkerFile)
+		switch {
+		case err != nil:
+			res.GitExcludeWarning = err.Error()
+		case updated:
```
obey-agent, line 189: The `(bool, error)` plumbing here is the user-visible fix of this commit (no more spurious "removed .camp from .git/info/exclude" lines when nothing was actually present). The contract is easy for a future refactor to break without anyone noticing because nothing tests the false case directly. Worth adding two small regression tests:

- `TestDetach_GitExcludeUpdatedFalseWhenAbsent` — git target, but `.git/info/exclude` either does not exist or does not contain `.camp`; call `Detach`; assert `res.GitExcludeUpdated == false` and no warning.
- `TestAttach_GitExcludeUpdatedFalseWhenAlreadyPresent` — pre-populate `.git/info/exclude` with `.camp`; call `Attach`; assert `res.GitExcludeUpdated == false`.

PR description already calls out direct `internal/git/info_exclude_test.go` coverage as deferred (follow-up #4), and that's fine, but at minimum the `Result` boolean should be pinned at the attach layer where the user-visible regression lived.

obey-agent (commented):
Verdict: Approve.

Re-reviewed against PR head `aa2f46f`. Two new commits since the previous approval at `09eda68` (`bcb54f4` and the squash merge `aa2f46f`) directly address follow-ups #1 and #2 from that review. Both fixes are correct and the previously-approved surface is unchanged.

What I verified

- `internal/attach/attach.go:97-115` — the `!hasExistingMarker` guard on the any-campaign rejection check has been removed. Because `Detect` is called from `filepath.Dir(target)`, the target's own marker is already invisible to detection, so the guard was only widening the trust boundary on `--force` re-attach onto a path that now lives inside a different campaign. The updated comment correctly reflects the new invariant.
- `internal/git/info_exclude.go` — `EnsureInfoExclude` / `RemoveInfoExclude` now return `(bool, error)`. The internal helpers (`ensurePatternInFile`, `removePatternFromFile`) only set the bool to `true` on an actual write, including the previously-buggy "pattern absent" path in remove (`if !removed { return false, nil }`). Atomic write semantics via `fsutil.WriteFileAtomically` are preserved.
- `internal/attach/attach.go:131-141` and `:185-195` — call sites now use a `switch` block so `GitExcludeWarning` is only set on real error and `GitExcludeUpdated` is only set when the file was actually modified. This is the user-visible fix.
- `internal/project/link.go:381-388` — wrappers discard the new bool. Acceptable as a tidiness deferral (PR description calls it out as follow-up #3 deferred); linked-project callers do not need the signal.

Verification I ran (at head `aa2f46f`)

- `go build ./...` — clean
- `go test ./internal/attach/... ./internal/git/... ./internal/project/... -count=1` — pass (attach 0.319s, git 17.475s, git/commit 1.952s, project 6.098s)
- `gofmt -l internal/attach internal/git internal/project` — clean
- `go vet ./internal/attach/... ./internal/git/... ./internal/project/...` — clean

Follow-ups (non-blocking)

1. The `(bool, error)` contract has no direct regression test at the attach layer. `TestAttach_AddsGitInfoExclude` and `TestDetach_RemovesGitInfoExclude` only pin the `GitExcludeUpdated == true` path. The false case (the actual user-visible regression this PR fixes) is not pinned, and a future refactor could quietly reintroduce the misleading "removed .camp" CLI line. See inline comment for a concrete two-test suggestion. PR description marks direct `internal/git/info_exclude_test.go` as deferred (follow-up #4), which is fine, but the `Result` boolean should ideally be pinned at the attach layer where the visible regression lived.
2. `internal/project/link.go:381-388` wrappers (`ensureGitInfoExclude` / `removeGitInfoExclude`) are now one-line shims. As called out in the PR description, removing them and switching call sites to `git.Ensure/RemoveInfoExclude` directly is pure tidiness; deferring is acceptable.

Staff Standard: Yes. Both security-relevant findings from the prior review are now properly closed, the original approved code path is intact, and the new bool contract is implemented cleanly throughout. Comfortable to merge once the missing regression test (or an explicit deferral note in a follow-up issue) is decided.

obey-agent (changes requested):
Verdict: Request Changes (supersedes my "Approve" review at this head — that verdict was wrong while a real coverage gap remained. Restating as a blocking gate).

To be clear: the code changes in `aa2f46f` are correct and the previous review's structural assessment stands. The reason I'm flipping the verdict is that the user-visible regression this PR fixes (`GitExcludeUpdated` lying about whether `.camp` was actually removed) has no regression test in this PR. The existing `TestDetach_RemovesGitInfoExclude` only pins the `true` case; the bug was in the `false` case.

Blocking

1. **Land regression coverage for the `(bool, error)` contract before merging this PR.** Two equivalent paths:

   **(a) Atomic merge with #273.** PR #273 stacks on this branch and adds direct unit tests in `internal/git/info_exclude_test.go` (including the byte-identity assertions on the no-op paths). If #273 merges immediately after #270, the gap window is acceptable. In that case, hold this PR's merge until #273 is green and ready.

   **(b) Cherry-pick the relevant tests from #273 into this PR.** Specifically `TestEnsureInfoExclude_Idempotent` and `TestRemoveInfoExclude_PatternNotPresent` are the two that pin the regression. Doing this lets #270 merge independently.

   Pick one. Approving #270 in isolation, on the assumption that #273 will land later, is exactly the "approve with follow-ups" pattern I'm now rejecting.

Verification I ran (at head `aa2f46f`)

- `go build ./...` clean
- `go test ./internal/attach/... ./internal/git/... ./internal/project/... -count=1` pass
- `gofmt -l internal/attach internal/git internal/project` clean
- `go vet ./internal/attach/... ./internal/git/... ./internal/project/...` clean

I will re-review and approve once option (a) is confirmed (#273 ready to merge with this) or option (b) lands here.

obey-agent (approved):
Verdict: Approve.

Re-reviewed at head `f791ccc`. Both blocking items from the previous round are addressed, and the merged-in #273 contents bring direct test coverage for the `(bool, error)` contract into this branch, so #270 now stands on its own without depending on a follow-up PR.

What I verified

- `86a3bc9` "Address PR #273 review: symmetric ctx-cancelled test, ctx-aware initRepo":
  - `internal/git/info_exclude_test.go:19` — `initRepo` now uses `exec.CommandContext(t.Context(), ...)` for `git init`. Repo standard satisfied.
  - `internal/git/info_exclude_test.go:194-202` — `TestRemoveInfoExclude_ContextCancelled` added, mirroring the Ensure case. Symmetric coverage restored.
- `f791ccc` merges #273 into this branch, bringing `internal/git/info_exclude_test.go` with `TestEnsureInfoExclude_Idempotent` and `TestRemoveInfoExclude_PatternNotPresent` — the byte-identity assertions that pin the `GitExcludeUpdated == false` regression directly in this branch. No external dependency on a follow-up PR.
- `internal/project/link.go` — `ensureGitInfoExclude` / `removeGitInfoExclude` wrappers removed; both call sites use `git.EnsureInfoExclude` / `git.RemoveInfoExclude` with the bool discarded. Single canonical entry point in `internal/git`.
- All previously-approved structural work (any-campaign attach rejection at line 97-115, `(bool, error)` plumbing through Attach/Detach result, atomic write semantics) intact.

Verification I ran (at head `f791ccc`)

- `go build ./...` clean
- `go test ./internal/git/... ./internal/attach/... ./internal/project/... -count=1` pass (git 16.756s, git/commit 1.604s, attach 0.562s, project 5.457s)
- `go test ./internal/git/ -run TestRemoveInfoExclude_ContextCancelled -count=1` pass
- `gofmt -l internal/git internal/attach internal/project` clean
- `go vet ./internal/git/... ./internal/attach/... ./internal/project/...` clean

Ready to merge.

PR #272, opened 2026-05-11
fix stale lock cleanup during camp pull

obey-agent:
## Summary

- Route `camp pull` and `camp pull all` through the existing lock-aware retry path.
- Classify `git pull` `index.lock` failures as `git.LockError`, including the resolved lock path, so stale locks are removed before retry.
- Add containerized integration coverage for root pulls and submodule `pull all` stale-lock cleanup.
- Add non-mutating context-cancellation coverage for the retry helper.

## Root Cause

`camp pull` invoked `git pull` directly with `exec.CommandContext`, so it never returned the structured `git.LockError` required by `git.WithLockRetry`. The stale-lock cleanup path existed for stage/commit and some project operations, but pull bypassed it.

## Review Follow-Up

- Replaced the `LockError{Path: "index.lock"}` literal with a path resolved through `git.ResolveGitDir(repoPath)` and `filepath.Join(gitDir, "index.lock")`.
- Added `TestRunGitPullWithLockRetry_ContextCancelled`, which uses a pre-canceled context and an invalid path to assert cancellation returns before any git operation or filesystem mutation.
- Removed host-temp filesystem mutation coverage; stale-lock behavior is now exercised only in the containerized integration harness under `tests/integration/pull_test.go`.

## Validation

- `just test all` - 68 packages, 2284/2284 unit tests passed; 227/227 container integration tests passed.
- `go test ./cmd/camp`
- `go test ./cmd/camp -run TestRunGitPullWithLockRetry_ContextCancelled -count=1`
- `go test -tags integration ./tests/integration -run "TestPull(_|All_)" -count=1`
- `git diff --check HEAD`
- Earlier: `just build-camp`, `just test unit`
- Earlier: `just lint` still fails on existing repo-wide lint debt unrelated to this change.

### DIFF
--- a/cmd/camp/pull.go
+++ b/cmd/camp/pull.go
@@ -1,9 +1,12 @@
 package main
 
 import (
+	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -67,13 +70,7 @@ func runPull(cmd *cobra.Command, args []string) error {
 		fmt.Fprintln(os.Stderr, ui.Info(fmt.Sprintf("Submodule: %s", target.Name)))
 	}
 
-	fullArgs := append([]string{"-C", target.Path, "pull"}, gitArgs...)
-	gitCmd := exec.CommandContext(ctx, "git", fullArgs...)
-	gitCmd.Stdout = os.Stdout
-	gitCmd.Stderr = os.Stderr
-	gitCmd.Stdin = os.Stdin
-
-	if err := gitCmd.Run(); err != nil {
+	if _, err := runGitPullWithLockRetry(ctx, target.Path, gitArgs, true); err != nil {
 		// Suppress cobra's usage output — this isn't a usage error.
 		cmd.SilenceUsage = true
 
@@ -89,6 +86,51 @@ func runPull(cmd *cobra.Command, args []string) error {
 	}
 
 	return nil
+}
+
+func runGitPullWithLockRetry(ctx context.Context, repoPath string, gitArgs []string, stream bool) ([]byte, error) {
+	cfg := git.DefaultRetryConfig()
+	cfg.OperationName = "pull"
+
+	var output []byte
+	err := git.WithLockRetry(ctx, repoPath, cfg, func() error {
+		pullArgs := append([]string{"-C", repoPath, "pull"}, gitArgs...)
+		gitCmd := exec.CommandContext(ctx, "git", pullArgs...)
+		gitCmd.Stdin = os.Stdin
+
+		var err error
+		if stream {
+			var stderr bytes.Buffer
+			gitCmd.Stdout = os.Stdout
+			gitCmd.Stderr = io.MultiWriter(os.Stderr, &stderr)
+			err = gitCmd.Run()
+			output = stderr.Bytes()
+		} else {
+			output, err = gitCmd.CombinedOutput()
+		}
+		if err != nil {
+			return classifyPullCommandError(repoPath, output, err)
+		}
+		return nil
+	})
+	return output, err
+}
+
+func classifyPullCommandError(repoPath string, output []byte, err error) error {
+	exitCode := 0
+	var exitErr *exec.ExitError
+	if errors.As(err, &exitErr) {
+		exitCode = exitErr.ExitCode()
+	}
+
+	if git.ClassifyGitError(string(output), exitCode) == git.GitErrorLock {
+		lockPath := "index.lock"
+		if gitDir, resolveErr := git.ResolveGitDir(repoPath); resolveErr == nil {
+			lockPath = filepath.Join(gitDir, "index.lock")
+		}
+		return &git.LockError{Path: lockPath, Err: err}
+	}
+	return err
 }
 
 // PullAllOptions holds configuration for pull-all operations.
@@ -241,13 +283,12 @@ func pullSingleTarget(
 	}
 
 	// Execute pull
-	pullArgs := []string{"-C", t.path, "pull"}
+	pullArgs := make([]string, 0, len(gitArgs)+1)
 	if t.isRoot {
 		pullArgs = append(pullArgs, "--no-recurse-submodules")
 	}
 	pullArgs = append(pullArgs, gitArgs...)
-	gitCmd := exec.CommandContext(ctx, "git", pullArgs...)
-	output, err := gitCmd.CombinedOutput()
+	output, err := runGitPullWithLockRetry(ctx, t.path, pullArgs, false)
 	if err != nil {
 		return handlePullError(ctx, t, output, err, red)
 	}
--- a/cmd/camp/pull_test.go
+++ b/cmd/camp/pull_test.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"testing"
+)
+
+func TestRunGitPullWithLockRetry_ContextCancelled(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	output, err := runGitPullWithLockRetry(ctx, "/path/that/should/not/be/touched", nil, false)
+	if !errors.Is(err, context.Canceled) {
+		t.Fatalf("runGitPullWithLockRetry() error = %v, want context.Canceled", err)
+	}
+	if len(output) != 0 {
+		t.Fatalf("runGitPullWithLockRetry() output = %q, want empty output", output)
+	}
+}
--- a/tests/integration/pull_test.go
+++ b/tests/integration/pull_test.go
@@ -1,0 +1,111 @@
+//go:build integration
+// +build integration
+
+package integration
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPull_RemovesStaleIndexLock(t *testing.T) {
+	tc := GetSharedContainer(t)
+
+	const campaignDir = "/campaigns/pull-lock"
+	setupPullLockRootRepo(t, tc, campaignDir)
+
+	tc.Shell(t, fmt.Sprintf(`
+		git clone /test/root-remote.git /test/root-seed
+		cd /test/root-seed
+		printf 'remote content' > remote.txt
+		git add remote.txt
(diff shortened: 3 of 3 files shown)


cmd/camp/pull.go
```diff
@@ -91,6 +88,47 @@ func runPull(cmd *cobra.Command, args []string) error {
...
+
+func classifyPullCommandError(output []byte, err error) error {
+	exitCode := 0
+	var exitErr *exec.ExitError
+	if errors.As(err, &exitErr) {
+		exitCode = exitErr.ExitCode()
+	}
+
+	if git.ClassifyGitError(string(output), exitCode) == git.GitErrorLock {
+		return &git.LockError{Path: "index.lock", Err: err}
```
obey-agent: `LockError{Path: "index.lock"}` is a literal string, not the real lock path. Functionally this is fine since `WithLockRetry` / `CleanStaleLocks` discover the actual lock from `repoPath` and ignore `LockError.Path`, but if this error ever bubbles up to the user, `LockError.Error()` will render as `stale lock at index.lock` rather than the resolved path under `.git/`. Consider plumbing the resolved path through (e.g. `filepath.Join(gitDir, "index.lock")` via `git.ResolveGitDir(repoPath)`) or leaving `Path` empty so the message can't lie. Minor, but `LockError` already supports this and the rest of the codebase populates it accurately.

cmd/camp/pull_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/Obedience-Corp/camp/internal/git"
+)
+
+func TestRunGitPullWithLockRetry_RemovesStaleIndexLock(t *testing.T) {
+	for _, stream := range []bool{false, true} {
+		t.Run(map[bool]string{false: "captured", true: "streaming"}[stream], func(t *testing.T) {
```
obey-agent: Subtest naming via `map[bool]string{false: "captured", true: "streaming"}[stream]` is denser than it needs to be. Idiomatic Go is a named-case table:

```go
tests := []struct {
    name   string
    stream bool
}{
    {"captured", false},
    {"streaming", true},
}
for _, tc := range tests {
    t.Run(tc.name, func(t *testing.T) { ... })
}
```

Reads more cleanly, easier to extend if a third dimension shows up (e.g. retry config variant). Nit.

cmd/camp/pull_test.go
```diff
@@ -0,0 +1,83 @@
...
+			repoDir, seedDir := setupPullTestRepo(t)
+			ctx := context.Background()
```
obey-agent: `ctx := context.Background()` only exercises the happy path. The staff Go standard for this repo asks for context-cancellation coverage on any retry/long-running path, and `WithLockRetry` already checks `ctx.Err()` at the top of each fast attempt and between cycles. Worth a sibling test that:

1. Creates the stale lock,
2. Wraps ctx with `context.WithCancel` and cancels before (or just after) the first attempt,
3. Asserts `runGitPullWithLockRetry` returns a context error without exhausting cycles.

Cheap to add and pins down a regression class (someone refactoring the closure could drop the ctx check without any existing test noticing).

obey-agent (commented):
Verdict: Approve with minor follow-ups.

Reviewed against PR head `c8c85cd`. The fix is correct, well-scoped, and reuses the existing `git.WithLockRetry` infrastructure rather than building a parallel retry path. Both call sites (`runPull` single-target and `pullSingleTarget` multi-target) are routed through one helper, classification reuses `git.ClassifyGitError`, and the streaming/captured distinction is preserved so interactive `camp pull` still flushes git output live to the user.

What I verified

- `cmd/camp/pull.go:73-130` — `runGitPullWithLockRetry` correctly threads ctx through `WithLockRetry`, creates a fresh `exec.Cmd` per attempt (so retries do not re-use a closed `*exec.Cmd`), and uses `io.MultiWriter` in streaming mode so stderr is both shown to the user and captured for classification. `pullSingleTarget` no longer prepends `-C target.Path pull` itself (the helper does), and the `--no-recurse-submodules` flag for the root target is still appended correctly.
- `cmd/camp/pull_test.go:1-83` — covers both `stream=false` and `stream=true` paths against a real local-bare-remote setup. Stale `index.lock` placed under the resolved git-dir, pull attempted, lock cleanup verified, content pulled verified.
- Suppression of cobra usage on real pull failures (`cmd.SilenceUsage = true`) is preserved.

Verification I ran (at head `c8c85cd`)

- `go build ./...` — clean
- `go test ./cmd/camp/ -run TestRunGitPullWithLockRetry -count=1` — pass (2.168s)
- `gofmt -l cmd/camp/pull.go cmd/camp/pull_test.go` — clean
- `go vet ./cmd/camp/...` — clean

Follow-ups (non-blocking)

1. `cmd/camp/pull.go:127` — `LockError{Path: "index.lock"}` uses a literal rather than the real lock path. `WithLockRetry` ignores `Path`, so this is informational only, but if the error ever surfaces to the user `LockError.Error()` will render `stale lock at index.lock` instead of the resolved path. See inline comment.
2. `cmd/camp/pull_test.go:17` — no context-cancellation test. Staff Go standard for this repo requires ctx-cancel coverage on retry paths. See inline.
3. `cmd/camp/pull_test.go:15` — subtest naming via `map[bool]string{...}[stream]` is denser than a small table struct. Nit. See inline.
4. Test uses `t.TempDir()` plus real `git` on the host — consistent with other tests in this package, but flagged so it can migrate alongside the rest if the codebase moves toward containerized filesystem-mutation tests.
5. `runPullAll` still has a pre-existing `fmt.Errorf("%d repo(s) failed to pull", failed)` (line ~169) that should use `camperrors`. Not introduced by this PR, but adjacent and worth a follow-up commit.

Staff Standard: Yes. The fix integrates cleanly with the existing lock-retry machinery instead of duplicating it, both call sites are unified, context propagates, and there is a regression test that drives the actual stale-lock cleanup behavior end-to-end. Comfortable to merge once the author considers the follow-ups (or lands them on the side).

obey-agent (commented):
Verdict: Request Changes (supersedes my earlier "approve with minor follow-ups" review — that framing was wrong; the items below are blocking, not optional).

The fix is structurally correct: routing both `runPull` and `pullSingleTarget` through a single `runGitPullWithLockRetry` that reuses `git.WithLockRetry` and `git.ClassifyGitError` is the right shape, and the streaming/captured split preserves interactive UX. Build, test, vet, and gofmt are clean at head `c8c85cd`. The two items below need to land in this PR before merge.

Blocking

1. **`cmd/camp/pull.go:127` — replace the `Path: "index.lock"` literal with the resolved lock path.** `LockError.Error()` will render `stale lock at index.lock` if this ever surfaces to the user, instead of the actual path under the resolved git dir (which matters for worktrees and submodules where `.git` is a file). Use `git.ResolveGitDir(repoPath)` and `filepath.Join(gitDir, "index.lock")`. The rest of the codebase populates `LockError.Path` accurately; this caller should match.

2. **`cmd/camp/pull_test.go` — add a context-cancellation test.** Staff Go standard for this repo: "Always include context cancellation tests" for retry/long-running paths. `WithLockRetry` checks `ctx.Err()` at the top of each fast attempt and between cycles; a future refactor could drop that check without any existing test failing. Add `TestRunGitPullWithLockRetry_ContextCancelled` that cancels ctx before invocation and asserts the function returns a context error without exhausting cycles.

Verification I ran (at head `c8c85cd`)

- `go build ./...` clean
- `go test ./cmd/camp/ -run TestRunGitPullWithLockRetry -count=1` pass
- `gofmt -l cmd/camp/pull.go cmd/camp/pull_test.go` clean
- `go vet ./cmd/camp/...` clean

I will re-review and approve once both items above are addressed.

obey-agent:
Addressed the requested changes at head c51718f: LockError.Path now resolves the real git dir/index.lock path; added non-mutating context-cancellation coverage for runGitPullWithLockRetry; stale-lock filesystem behavior remains covered in the containerized integration tests. Full validation passed with just test all: 2284/2284 unit tests and 227/227 integration tests.

obey-agent (commented):
Verdict: Approve.

Re-reviewed at head `c51718f`. Both blocking items from the previous review are addressed, and the regression test was upgraded from a host-tempdir unit test to a containerized integration test that exercises the full `camp` CLI path. That is the right shape per the repo policy on filesystem-mutating tests, and gives broader coverage than the original.

What I verified

- **`cmd/camp/pull.go:127-131`** — `LockError.Path` now resolves the actual lock path via `git.ResolveGitDir(repoPath)` and falls back to the literal only if resolution fails. `classifyPullCommandError` correctly threads `repoPath` from the closure. Messages surfaced to the user will now name the real lock file under `.git/`.
- **`cmd/camp/pull_test.go`** — `TestRunGitPullWithLockRetry_ContextCancelled` added. Cancels ctx before invocation, passes a path that should not be touched, asserts `errors.Is(err, context.Canceled)` and empty output. Pins ctx propagation for both the helper and the underlying `WithLockRetry` ctx check.
- **`tests/integration/pull_test.go`** (new, `//go:build integration`):
  - `TestPull_RemovesStaleIndexLock` drives the actual `camp pull --ff-only` binary inside a container, plants a stale `index.lock` in the campaign root's `.git`, asserts the lock is removed and the remote content lands.
  - `TestPullAll_RemovesSubmoduleStaleIndexLock` extends coverage to submodule lock cleanup via `camp pull all`. Uses `git rev-parse --absolute-git-dir` to find the submodule's real git dir before planting the lock.
  - This is the correct test shape under the repo's filesystem-safety rule: real git mutation runs in container isolation, not on the host. The earlier `t.TempDir()`-based unit test was deleted in favor of this, which I support.

Verification I ran (at head `c51718f`)

- `go build ./...` clean
- `go test ./cmd/camp/... -count=1` pass
- `go test ./cmd/camp/ -run TestRunGitPullWithLockRetry -count=1 -v` pass
- `gofmt -l cmd/camp/pull.go cmd/camp/pull_test.go tests/integration/pull_test.go` clean
- `go vet ./cmd/camp/...` clean

Coverage note (not blocking)

The default `go test ./...` no longer runs a lock-cleanup regression test; that lives behind `-tags=integration` now. This is correct per repo policy, but it does mean a developer who runs only `just test unit` will not catch a regression in this code path. Ensure `just test all` (which runs both unit + integration via `just integration`) is part of the standard pre-merge / release gate going forward. Not a change request on this PR — flagging because the test relocation shifts where the safety net lives.

Ready to merge.

PR #273, opened 2026-05-11
Drop link wrappers and add git/info_exclude unit tests

lancekrogers:
Stacked on top of #270. Lands the remaining two follow-ups obey-agent flagged in the approval review and that PR #271 explicitly deferred.

## Changes

**1. `internal/project/link.go` — drop the `ensureGitInfoExclude` / `removeGitInfoExclude` wrappers.**
After the previous follow-up the wrappers were one-line passthroughs to `git.EnsureInfoExclude` / `git.RemoveInfoExclude`. Inlined the two call sites and removed the wrappers so there's a single canonical entry point in the `internal/git` package.

**2. `internal/git/info_exclude_test.go` — direct unit tests.**
Coverage was previously transitive through `internal/attach` and `internal/project`. Added focused tests against a real `git init` working tree:

- `TestIsRepo` — repo vs non-repo
- `TestEnsureInfoExclude_AddsPattern` — first-add returns `added=true` and writes
- `TestEnsureInfoExclude_Idempotent` — second-add returns `added=false` and file is byte-identical
- `TestEnsureInfoExclude_PreservesExistingPatterns` — append doesn't clobber prior entries
- `TestRemoveInfoExclude_RemovesPattern` — present case returns `removed=true`
- `TestRemoveInfoExclude_MissingFile` — missing exclude is a no-op (`removed=false`, no error)
- `TestRemoveInfoExclude_PatternNotPresent` — file untouched when pattern absent, `removed=false`
- `TestRemoveInfoExclude_PreservesOtherPatterns` — only the matching line is stripped
- `TestEnsureInfoExclude_ContextCancelled` — cancellation surfaces from the underlying `git rev-parse`

## Verification

- `go build ./...` clean
- `go test ./internal/git/... ./internal/attach/... ./internal/project/...` pass (new tests included)
- `gofmt -l ./internal/git ./internal/project` clean
- `golangci-lint run --new-from-rev=origin/non-project-symlinks-design ./internal/git/... ./internal/project/...` — 0 issues

### DIFF
--- a/internal/git/info_exclude_test.go
+++ b/internal/git/info_exclude_test.go
@@ -16,7 +16,7 @@ func initRepo(t *testing.T) string {
 	if err != nil {
 		t.Fatalf("eval symlinks: %v", err)
 	}
-	cmd := exec.Command("git", "init", "-q", dir)
+	cmd := exec.CommandContext(t.Context(), "git", "init", "-q", dir)
 	if out, err := cmd.CombinedOutput(); err != nil {
 		t.Fatalf("git init: %v\n%s", err, out)
 	}
@@ -187,6 +187,16 @@ func TestEnsureInfoExclude_ContextCancelled(t *testing.T) {
 	cancel()
 
 	if _, err := EnsureInfoExclude(ctx, repo, ".camp"); err == nil {
+		t.Error("expected error on cancelled context, got nil")
+	}
+}
+
+func TestRemoveInfoExclude_ContextCancelled(t *testing.T) {
+	repo := initRepo(t)
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	if _, err := RemoveInfoExclude(ctx, repo, ".camp"); err == nil {
 		t.Error("expected error on cancelled context, got nil")
 	}
 }
--- a/internal/git/info_exclude_test.go
+++ b/internal/git/info_exclude_test.go
@@ -1,0 +1,192 @@
+package git
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func initRepo(t *testing.T) string {
+	t.Helper()
+	dir := t.TempDir()
+	dir, err := filepath.EvalSymlinks(dir)
+	if err != nil {
+		t.Fatalf("eval symlinks: %v", err)
+	}
+	cmd := exec.Command("git", "init", "-q", dir)
+	if out, err := cmd.CombinedOutput(); err != nil {
+		t.Fatalf("git init: %v\n%s", err, out)
+	}
+	return dir
+}
+
+func readExclude(t *testing.T, repo string) string {
+	t.Helper()
+	data, err := os.ReadFile(filepath.Join(repo, ".git", "info", "exclude"))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return ""
+		}
+		t.Fatalf("read exclude: %v", err)
+	}
+	return string(data)
+}
+
+func TestIsRepo(t *testing.T) {
+	repo := initRepo(t)
+	if !IsRepo(repo) {
+		t.Errorf("IsRepo(%q) = false, want true", repo)
+	}
+	plain := t.TempDir()
+	if IsRepo(plain) {
+		t.Errorf("IsRepo(%q) = true for non-repo, want false", plain)
+	}
+}
+
+func TestEnsureInfoExclude_AddsPattern(t *testing.T) {
+	repo := initRepo(t)
+
+	added, err := EnsureInfoExclude(context.Background(), repo, ".camp")
+	if err != nil {
+		t.Fatalf("EnsureInfoExclude: %v", err)
+	}
+	if !added {
+		t.Fatal("added = false, want true on first add")
+	}
+	if got := readExclude(t, repo); !strings.Contains(got, ".camp") {
+		t.Errorf("exclude does not contain .camp; got:\n%s", got)
+	}
+}
+
+func TestEnsureInfoExclude_Idempotent(t *testing.T) {
+	repo := initRepo(t)
+
+	if _, err := EnsureInfoExclude(context.Background(), repo, ".camp"); err != nil {
+		t.Fatalf("first EnsureInfoExclude: %v", err)
+	}
+	before := readExclude(t, repo)
+
+	added, err := EnsureInfoExclude(context.Background(), repo, ".camp")
+	if err != nil {
+		t.Fatalf("second EnsureInfoExclude: %v", err)
+	}
+	if added {
+		t.Error("added = true on second call, want false (already present)")
+	}
+	if after := readExclude(t, repo); after != before {
+		t.Errorf("file mutated on idempotent call\nbefore:\n%s\nafter:\n%s", before, after)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


internal/git/info_exclude_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+func TestEnsureInfoExclude_ContextCancelled(t *testing.T) {
+	repo := initRepo(t)
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	if _, err := EnsureInfoExclude(ctx, repo, ".camp"); err == nil {
+		t.Error("expected error on cancelled context, got nil")
+	}
+}
```
obey-agent, line 192: Only `EnsureInfoExclude` has a cancelled-ctx test. `RemoveInfoExclude` takes the same `ctx` and hits the same `infoExcludePath` → `git rev-parse` resolution path, so the symmetric test is essentially free:

```go
func TestRemoveInfoExclude_ContextCancelled(t *testing.T) {
    repo := initRepo(t)
    ctx, cancel := context.WithCancel(context.Background())
    cancel()
    if _, err := RemoveInfoExclude(ctx, repo, ".camp"); err == nil {
        t.Error("expected error on cancelled context, got nil")
    }
}
```

Without it, a future refactor could drop ctx threading from the remove path and only the Ensure test would catch it.

internal/git/info_exclude_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"strings"
+	"testing"
+)
+
+func initRepo(t *testing.T) string {
+	t.Helper()
+	dir := t.TempDir()
+	dir, err := filepath.EvalSymlinks(dir)
+	if err != nil {
+		t.Fatalf("eval symlinks: %v", err)
```
obey-agent, line 17: `exec.Command("git", "init", ...)` here drops ctx propagation in the test helper. Repo standard is `exec.CommandContext(ctx, ...)` everywhere git is invoked, even in test setup, so cancellation of the surrounding test ctx tears down git too. Easy switch: take `ctx context.Context` (or just `t.Context()` from Go 1.24) and use `exec.CommandContext`. Minor.

obey-agent (commented):
Verdict: Approve.

Reviewed against PR head `72736a8`. This lands both follow-ups that #270's approval marked deferred (drop the `internal/project/link.go` wrappers, add direct `internal/git/info_exclude_test.go` coverage). Clean execution; the new tests pin exactly the regression class the `(bool, error)` contract on those helpers was meant to prevent.

What I verified

- `internal/project/link.go:140-143` and `:222-225` — both call sites now invoke `git.EnsureInfoExclude` / `git.RemoveInfoExclude` directly, with the new bool discarded via `if _, err := ...`. Warning text and surrounding control flow preserved.
- `internal/project/link.go:381-388` — wrappers removed, no other callers reference them (confirmed via build).
- `internal/git/info_exclude_test.go` — 9 new tests, including the two cases I asked for on #270:
  - `TestEnsureInfoExclude_Idempotent` and `TestRemoveInfoExclude_PatternNotPresent` both assert byte-identical file contents on the no-op path, which is exactly the right contract pin for the "no spurious CLI message" fix.
  - `TestRemoveInfoExclude_MissingFile` covers the missing-exclude path explicitly rather than relying on `git init` defaults.
  - `TestEnsureInfoExclude_ContextCancelled` exercises ctx propagation through `infoExcludePath` → `git rev-parse`.
- `initRepo` uses `filepath.EvalSymlinks` on the tempdir so macOS `/var` → `/private/var` does not skew path comparisons. Good attention to host quirks.

Verification I ran (at head `72736a8`)

- `go build ./...` — clean
- `go test ./internal/git/... ./internal/project/... ./internal/attach/... -count=1` — pass (git 16.075s, git/commit 1.757s, project 5.275s, attach 0.852s)
- `gofmt -l internal/git internal/project` — clean
- `go vet ./internal/git/... ./internal/project/...` — clean

Follow-ups (non-blocking)

1. `internal/git/info_exclude_test.go:192` — only `EnsureInfoExclude` has a cancelled-ctx test. `RemoveInfoExclude` goes through the same `infoExcludePath` → `git rev-parse` path; the symmetric test is one-screen and cheap to add. See inline.
2. `internal/git/info_exclude_test.go:17` — `initRepo` uses `exec.Command` instead of `exec.CommandContext`. Repo standard is ctx-aware git invocation everywhere, including in test setup. See inline.
3. No coverage for the worktree / submodule resolution path that the package doc comment specifically mentions ("its actual git dir may live elsewhere (worktrees, submodules) and is resolved via git"). A test that points `repoRoot` at a `git worktree add`-created tree would pin that branch of `infoExcludePath`. Worth a future PR; not blocking this one.

Stacking note: base is `non-project-symlinks-design` (#270). The series merges cleanly in order #270 → #273. Once #270 lands on `main`, GitHub will retarget this one automatically.

Staff Standard: Yes. The implementation is the cleanup it claims to be, the new tests are thoughtful (byte-identity assertions on idempotent paths, explicit missing-file case, ctx cancellation, preserved-other-patterns), and the wrapper removal leaves a single canonical entry point in `internal/git`. Comfortable to merge once `#270` is in, and the symmetric ctx-cancel test can land here or in a tiny follow-up PR.

obey-agent (changes requested):
Verdict: Request Changes (supersedes my earlier "approve" review — that verdict was wrong while real items remained open. The items below are blocking).

The structural work is right: wrappers removed in `internal/project/link.go`, both call sites switched to `git.EnsureInfoExclude` / `git.RemoveInfoExclude` directly, and the new `internal/git/info_exclude_test.go` pins the `(bool, error)` contract with byte-identity assertions on idempotent paths. Build, test, vet, gofmt clean at head `72736a8`. The two items below need to land in this PR before merge.

Blocking

1. **`internal/git/info_exclude_test.go:192` — add `TestRemoveInfoExclude_ContextCancelled`.** Only the Ensure path has the cancelled-ctx test; the Remove path takes the same `ctx` through the same `infoExcludePath` → `git rev-parse` resolution. Asymmetric coverage means a future refactor could drop ctx threading from the remove path without any test failing. Mirror the existing Ensure test:

   ```go
   func TestRemoveInfoExclude_ContextCancelled(t *testing.T) {
       repo := initRepo(t)
       ctx, cancel := context.WithCancel(context.Background())
       cancel()
       if _, err := RemoveInfoExclude(ctx, repo, ".camp"); err == nil {
           t.Error("expected error on cancelled context, got nil")
       }
   }
   ```

2. **`internal/git/info_exclude_test.go:17` — switch `initRepo` to `exec.CommandContext`.** Repo standard is ctx-aware git invocation everywhere git is shelled out, including in test setup. Take `ctx context.Context` (or use `t.Context()` from Go 1.24+) and replace `exec.Command(...)` with `exec.CommandContext(ctx, ...)`. The other test setup helpers in this package and `internal/git/run_test.go` should be the precedent here.

Stacking note: base is `non-project-symlinks-design` (#270). This PR's fate is bound to #270 — do not merge #270 alone without these tests, since the `(bool, error)` contract regression that PR fixes has no test coverage outside this one.

Verification I ran (at head `72736a8`)

- `go build ./...` clean
- `go test ./internal/git/... ./internal/project/... ./internal/attach/... -count=1` pass
- `gofmt -l internal/git internal/project` clean
- `go vet ./internal/git/... ./internal/project/...` clean

I will re-review and approve once both items above are addressed.

PR #276, opened 2026-05-13
feat(workitem): .workitem metadata discovery (WW0001 phase 003)

lancekrogers:
## Summary

Phase 003 of festival WW0001 (workitem-workflows). Adds typed `.workitem`
metadata parsing, integrates it into `discoverWorkflowDocs` with the
precedence rules from the canonical design package, and surfaces the new
fields through `camp workitem --json` at schema version `workitems/v1alpha4`.

Backward-compatible: items without `.workitem` are unchanged. Malformed
`.workitem` logs via `slog.Default()` and continues discovery; the item is
returned with its derived fields only.

## What landed

- `internal/workitem/metadata.go` — typed schema + `LoadMetadata`
- `internal/workitem/merge.go` — `ApplyMetadata` with precedence rules
- `internal/workitem/discover_workflow_docs.go` — wire into discovery
- `internal/workitem/model.go` — extend `WorkItem` with metadata fields
- `internal/workitem/json.go` — bump `SchemaVersion` to `workitems/v1alpha4`
- 4 new test files + extended existing tests
- 1 fixture under `internal/workitem/testdata/`

Total: 10 files changed, +798 -1 across 4 commits.

## Test plan

- [x] `go test -race -count=1 ./internal/workitem/...` — green
- [x] `go test -race -count=1 ./...` — green
- [x] `just fmt` — clean
- [x] `just build quick` and `just build quick-dev` — successful
- [x] Manual smoke: `camp workitem --json | jq '.schema_version'` returns
      `"workitems/v1alpha4"`; new `.workitem` files surface
      `stable_id`/`execution`/`priority_info` fields; legacy items unchanged

## Precedence rules

Per `workflow/design/workitem-workflows/WORKITEM_CONTRACT.md`:

| Field | Source of truth |
|---|---|
| `RelativePath`, `PrimaryDoc`, `CreatedAt`, `UpdatedAt` | filesystem placement |
| `LifecycleStage` | filesystem placement |
| `StableID`, `Description`, `Execution`, `PriorityInfo`, `Project`, `WorkflowMeta`, `Lineage` | `.workitem` metadata |
| `Title` | `.workitem` when non-empty, else derived from heading |

## Festival traceability

Festival: workitem-workflows-WW0001
Phase: 003_IMPLEMENT_CAMP_METADATA
Sequence: 01_metadata_discovery (all 8 tasks complete)
Phase gates: all 4 approved


internal/workitem/metadata_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func writeWorkitem(t *testing.T, dir, body string) {
+	t.Helper()
+	if err := os.WriteFile(filepath.Join(dir, MetadataFilename), []byte(body), 0o644); err != nil {
```
obey-agent: **Filesystem safety violation (blocker).** This new file writes `.workitem` fixtures via `os.WriteFile` into `t.TempDir()` directories on the host. The campaign rule (`CLAUDE.md` §11 *Filesystem Safety*) is explicit: *Tests that mutate the filesystem must run in containerized isolation, not directly on the host via ad hoc temp directories.* The camp project already provides that harness at `tests/integration/` (with `helpers_container.go` driving real `camp` binaries inside Docker via testcontainers). Please move `LoadMetadata` filesystem coverage there, or — if you want fast in-process coverage — refactor `LoadMetadata" to accept an `fs.FS` so the parser can be tested with `fstest.MapFS` and the on-disk path stays an integration concern.

internal/workitem/discover_test.go
```diff
@@ -297,6 +297,142 @@ func TestFindFestivalPrimaryDoc_Priority(t *testing.T) {
...
+title: Metadata Title
+description: From .workitem.
+priority:
+  level: high
+  reason: launch
+execution:
+  mode: design
+  autonomy: constrained
+  risk: medium
+`)
```
obey-agent, line 320: **Same filesystem-safety concern in the new discover tests.** `TestDiscoverDesign_WithMetadata`, `TestDiscoverDesign_NoMetadataUnchanged`, and `TestDiscoverDesign_MalformedMetadataDoesNotCrashDiscovery` all stand up a fake campaign tree on the host via `t.TempDir()` + `os.MkdirAll` + `os.WriteFile`. These are exactly the calls the campaign filesystem-safety rule prohibits in non-containerized tests.

I understand this matches the existing `setupTestCampaign` pattern in this file, but that pre-dates the current rule and is itself a migration target. Adding new tests that follow the violating pattern compounds the debt rather than chipping at it.

Two reasonable options:
1. Move the new discovery+metadata cases into `tests/integration/` using `helpers_container.go` so they run in the existing Docker harness.
2. Keep them as in-process tests but refactor discovery to depend on an `fs.FS` (and seed the test with `fstest.MapFS`) so no host filesystem mutation is involved.

Option 2 is also a structural win for the package: discovery becomes trivially mockable for the camp dashboard work coming in phase 005.

internal/workitem/discover_workflow_docs.go
```diff
@@ -74,6 +75,17 @@ func discoverWorkflowDocs(ctx context.Context, campaignRoot, rootDir string, wfT
 		if primaryDocAbs != "" {
 			item.Summary = extractSummaryFromFile(primaryDocAbs, 200)
 		}
+
+		md, err := LoadMetadata(ctx, dirPath)
+		if err != nil {
+			// Malformed optional metadata must not crash full discovery.
+			// Log and include the item with derived fields only.
+			slog.Default().Warn("workitem metadata invalid; skipping metadata merge",
+				"path", relPath, "error", err.Error())
```
obey-agent: **Observability nit.** Passing `"error", err.Error()` flattens the error to a string at the call site and discards any structure that `camperrors.Wrapf` / `camperrors.NewValidation` carry (field name, validation kind, wrapped chain). Pass the error itself so slog handlers can render it however they want and so `errors.Is` / `errors.As` consumers downstream of structured logs still work:

```go
slog.Default().Warn("workitem metadata invalid; skipping metadata merge",
    "path", relPath,
    "error", err,
)
```

This is small but it's the kind of choice that decides whether the observability layer stays useful as the workitem surface grows in phase 005.

obey-agent (commented):
**Verdict: Request Changes**

## Overview

Phase 003 of `workitem-workflows-WW0001` lands the typed `.workitem` schema, `LoadMetadata`, and the precedence merge that surfaces metadata through `camp workitem --json` at `workitems/v1alpha4`. The implementation logic is good. The blocker is test infrastructure: the new tests violate the campaign filesystem-safety rule by mutating host directories instead of running through the existing containerized harness.

## Key Findings

- **Filesystem safety violation (blocker).** `internal/workitem/metadata_test.go` and the new cases in `internal/workitem/discover_test.go` use `t.TempDir()` + `os.MkdirAll` + `os.WriteFile` on the host. `CLAUDE.md` §11 is explicit that filesystem-mutating tests must run in the containerized harness (`tests/integration/` already exists with `helpers_container.go`). See the two inline comments for placement options: move to the Docker harness, or refactor discovery to take an `fs.FS` so a pure parser is unit-testable without touching disk and the disk path stays an integration concern. The pre-existing pattern in `setupTestCampaign` is migration debt, not a license to grow it.
- **Observability nit (worth fixing while here).** `discover_workflow_docs.go` logs malformed metadata with `"error", err.Error()`. Pass the error value so handlers and `errors.Is/As` consumers can use the structure `camperrors.NewValidation`/`Wrapf` provides.

## What's Done Well

- `LoadMetadata` uses the project error package consistently (`camperrors.Wrapf`, `camperrors.NewValidation`) and propagates context cancellation correctly (pre-read and pre-parse checks).
- Precedence semantics match the canonical contract: filesystem placement wins for path-derived fields, metadata wins for execution/priority/project/workflow/lineage, title is overridden only when non-empty. The `ApplyMetadata` table-driven coverage in `merge_test.go` is the right shape.
- `WorkItem` extensions use `omitempty` so legacy items keep their JSON shape byte-identical, with the schema bump (`v1alpha4`) and forward-comment about future workflow progress fields documented in `json.go`.
- Malformed-metadata behavior is the correct call: log and continue discovery with derived fields only, matching the phase 003 goal of *no regression for items without metadata*.
- Files and functions are all comfortably under the 500/50-line limits; small focused types throughout.

## Staff Standard

Am I proud to ship this as my own work? The production code, yes. The tests, no — not while they bypass the containerized harness rule the campaign explicitly calls out. Fix the test placement (or refactor discovery onto `fs.FS`) and this is a clean approve.

internal/workitem/metadata.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if err := validateMetadata(&m); err != nil {
+		return nil, camperrors.Wrapf(err, "validating %s", path)
+	}
+
+	return &m, nil
+}
+
+func validateMetadata(m *Metadata) error {
```
obey-agent, line 77: Follow-up review: this validation stops after the top-level required fields, but the safety-critical metadata remains unvalidated. `execution.risk`, `execution.autonomy`, `execution.mode`, and `priority.level` are the fields agents and the TUI will use to decide whether work is blocked, high-risk, or autonomous; a typo like `risk: CRITICAL` or `autonomy: autonmous` currently parses cleanly and silently removes the safety signal. Please validate the enum values from `WORKITEM_CONTRACT.md` here and return `camperrors.NewValidation` so discovery logs and skips malformed metadata instead of presenting it as usable.

obey-agent (commented):
Follow-up review

Verdict: Request Changes

Overview: This is a solid, focused metadata discovery pass: it keeps legacy items compatible, uses `camperrors` for new error construction, and degrades malformed optional metadata without aborting discovery. My remaining concern is safety semantics, not structure.

Key Findings:
- Inline: `validateMetadata` only checks top-level required fields. The execution-control fields that agents and the dashboard will use for risk/autonomy/blocking can contain typos or unsupported values and still parse cleanly, silently dropping safety signals.

What's Done Well:
- The parser and merge code are scoped and readable.
- Optional metadata failure is handled as a warning and does not poison the whole scan.
- New error paths use the project error package instead of raw `fmt.Errorf`.

Staff Standard: I would be proud of this after enum validation for the safety-critical metadata fields lands. Without that, the code can misrepresent execution risk in exactly the surface agents will consult before acting.

Verification: `go test ./internal/workitem/...` passed in the PR worktree.

obey-agent (approved):
**Verdict: Approve**

## Overview

WW0001 phase 009 sequence 009.01 lands the V2 design correction on top of PR #276's metadata-discovery surface. The .workitem schema is trimmed to the v1 minimum (D006 / D024): `version`, `kind`, `id`, `type` required; `title`, `created_by` optional. All nested blocks (Collection / Execution / Priority / Project / Workflow / Lineage / External) are removed from both the `Metadata` parser and the in-memory `WorkItem` model. The `WorkitemSchemaVersion` constant is now `v1alpha4` (string), with `acceptedWorkitemVersions` documenting the supported set so an unsupported version produces a camperrors validation error with the upgrade hint. `WorkflowType` is already a string type so D028 (open enum for custom types) is satisfied; the new `emitCandidate`/`emitCandidateFS` helpers (D025) gate non-built-in types behind a `.workitem` marker so a stray `workflow/random/foo/` cannot pollute `camp p list`. Discovery skip events are structured `slog.Debug` with `path`/`reason`/`type` fields. Metadata loader split into `LoadMetadata` (host) + `LoadMetadataFS` (fs.FS) so unit tests run via `fstest.MapFS` without `t.TempDir`.

## Verification

- `just build-camp` — successful build on PR head `8cecb99`.
- `go test ./...` — all packages pass (no failures).
- New `LoadMetadataFS` and `emitCandidateFS` tests exercise validation, version mismatch, kind mismatch, missing id/type, malformed yaml, ctx cancellation, and the D025 marker-gate.

## What's Done Well

- `validateMetadata` enforces version-set membership, kind, and required `id`/`type` at the loader so corrupted metadata is rejected before reaching `ApplyMetadata`.
- `acceptedWorkitemVersions` map makes the migration story (v1alpha4 -> v1alpha5 in 009.02) a one-line edit rather than a scattered grep.
- `LoadMetadataFS` accepts an injected `fs.FS` so the entire test suite for the loader runs without touching the host filesystem; the host wrapper preserves the existing call surface.
- `emitCandidateFS` (D025) cleanly separates "is this type built-in?" from "does the marker file exist?", with structured reasons (`builtin`, `marker`, `no-marker`, `stat-error`) the discovery walker can log.
- `slog.Debug` skip events with `path`/`reason`/`type` make "why isn't my workitem showing up" answerable in one log scan.
- Discovery uses `camperrors.Wrapf` for I/O paths and never bubbles a parse failure as a top-level error — one bad `.workitem` no longer blanks the dashboard.

## Staff Standard

Would I be proud to ship this? Yes. The trimmed schema lines up with the V2 contract from phase 007's design redo; the loader is now small enough that the validation rules read as a single function. The fs.FS split is the right shape for both the loader and the marker check — the unit-test surface is now host-independent. Approving.

PR #277, opened 2026-05-13
fix: clarify attach campaign picker help

lancekrogers:
## Summary
- remove internal optional-value sentinels from runtime help output so `--campaign` no longer renders as `string[=" pick"]`
- clarify `camp attach` campaign selection behavior, including bare `--campaign` forcing the picker
- update CLI reference docs and README attach example

## Verification
- `go test ./cmd/camp -run 'TestCleanFlagUsages'`
- `go test ./cmd/camp/attach -run '^$'` (compile-only; no tests executed)
- `go run ./cmd/camp attach --help` verified no NUL sentinel and no `string[=...]` output
- `git diff --check main...HEAD`
- added-diff audit for filesystem-mutating test calls (`TempDir`, `MkdirAll`, `WriteFile`, `Chdir`, `Remove`, `RemoveAll`, `chmod`, `git init`, `CreateTemp`, `MkdirTemp`, `os.Rename`, `exec.Command`): no matches

Note: I did not run the broader host test suite because existing tests may mutate the filesystem and should run through the appropriate containerized harness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,9 @@ target so detection works from inside it:
 ln -s ~/Dev/external-repo ai_docs/examples/external-repo
 camp attach ai_docs/examples/external-repo
 
+# Force the campaign picker instead of using the current campaign:
+camp attach ai_docs/examples/external-repo --campaign
+
 # now you can pin and navigate to it:
 cd ai_docs/examples/external-repo
 camp pin external-repo
--- a/cmd/camp/attach/attach.go
+++ b/cmd/camp/attach/attach.go
@@ -42,9 +42,16 @@ campaign owns it.
 If the target is reached through a symlink, camp follows it once and writes
 the marker at the final directory.
 
+Campaign selection:
+  - inside a campaign, omit --campaign to attach to the current campaign
+  - outside a campaign in an interactive terminal, omit --campaign to pick
+  - use a bare --campaign to force the picker even inside a campaign
+  - use --campaign <name-or-id> for scripts or to skip the picker
+
 Examples:
   camp attach ai_docs/examples/external-repo
   camp attach ~/scratch/notes-link
+  camp attach ~/scratch/notes-link --campaign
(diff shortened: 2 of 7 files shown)


obey-agent (commented):
**Verdict: Approve**

## Overview

Targeted, low-risk fix that strips an internal NUL-prefixed `NoOptDefVal` sentinel from runtime help output and clarifies that bare `--campaign` forces the picker. The implementation is a small pure string function wired through the cobra usage template, and the new help-text language matches the actual resolver precedence in `cmd/camp/attach/resolver.go`.

## Key Findings

- `cleanFlagUsages` is correctly scoped: pure string, idempotent, handles multi-sentinel lines via the inner `for` loop, and gracefully falls back to NUL removal when `[="` is not present.
- Test coverage matches the staff rule: pure string asserts only, no `t.TempDir`, no filesystem mutation, no `git init`/`MkdirAll`/`WriteFile`. The malformed-input negative case (lone NUL with no bracket) is the only obvious extra worth eventually adding, but it does not block.
- Doc parity is intact across `attach.go`, `docs/cli-reference/camp_attach.md`, `docs/cli-reference/camp-reference.md`, and `README.md`. The four-bullet selection rules accurately describe the resolver's bare-flag/picker/explicit-name branches.
- No I/O, no error returns, so context propagation and `camperrors` do not apply here.

## What's Done Well

- `cleanFlagUsageLine` is honest about WHY the workaround exists (pflag renders `NoOptDefVal` verbatim), which is exactly the comment style the codebase calls for.
- Reuses the existing cobra `AddTemplateFunc` hook instead of forking the usage template, keeping global subcommand behavior consistent.
- Verification notes explicitly call out that host-mutating tests were avoided, consistent with campaign filesystem-safety rules.

## Staff Standard

Would I submit this as my own work? Yes. It is a small, well-scoped, well-tested fix to a real help-output bug, with documentation kept in lockstep. Ship it.

PR #281, opened 2026-05-13
feat(workitem): dashboard visibility for .workitem + .workflow (WW0001 phase 005)

obey-agent:
## Summary

Adds dashboard visibility for `.workflow/` runtime progress and lands the V2 design correction (009.02) on top of the trimmed v1alpha5 schema from #276.

## What's actually in this diff

After 009.01 trimmed the `.workitem` schema to the v1 minimum, this PR's TUI surface narrowed accordingly. The current diff does **NOT** add execution badges, priority badges from metadata, or EXECUTION/PRIORITY/PROJECT/LINEAGE preview sections — those were removed because the underlying nested metadata blocks no longer exist (D006/D024). What this PR does add:

### Runtime visibility surface

- **`LocalRunProgress`** — narrow read-only view of fest's `.workflow/` runtime (workflow_id, active_run_id, current_step, total_steps, completed_steps, run_status, blocked, doc_hash_changed). Camp and fest are separate Go modules; this is a deliberate cross-repo duplicate. Future improvement: extract to `obey-shared/workflow/events`.
- **`LoadLocalRun` / `LoadLocalRunFS`** — split host wrapper + fs.FS variant. Reads `.workflow/workflow.yaml` + `runs/<active>/run.yaml` + `progress_events.jsonl` and returns the replayed state.
- **`replayLocalRunFSE`** — ctx-aware event replay that returns `(state, error)` so partial state from a corrupt event stream no longer surfaces as success. Honors `ctx.Err()` mid-loop, surfaces `bufio.Scanner.Err()` via `camperrors`, wraps file-open failures with the events path (D030 #8).
- **Wire-event constants** — event-type strings hoisted to named constants in `localrun.go` documenting the cross-repo contract with fest's localstore (D030 #9).

### Discovery integration

- **`discoverWorkflowDocs`** — calls `LoadLocalRun` after `ApplyMetadata`. Populates `WorkItem.WorkflowMeta` when `.workflow/` is present; logs warn-level slog and continues on malformed runtime (one bad runtime never blanks the dashboard).
- Discovery skip events are structured `slog.Debug` with `path` / `reason` / `type` fields (009.01).

### TUI preview

- **`renderPreview`** adds a `WORKFLOW` section showing `id`, `active`, `progress (Step N of M)`, `status`, and a `⚠ workflow doc changed since run started` warning when `DocHashChanged` is true.
- Width-fallback tweak in `renderRow` so very narrow terminals drop the priority badge before truncating the title.

### Schema

- `WorkitemSchemaVersion` bumps to `v1alpha5`. `acceptedWorkitemVersions` retains `v1alpha4` as backward-compatible since the on-disk shape did not change between v1alpha4 and v1alpha5 (only camp's internal model did).

## Tests

- **`localrun_fs_test.go`** — `fstest.MapFS`-driven coverage of every replay/state case: manifest only, summary as cache, basic event replay, replay overrides stale summary, completed/blocked status, stale-cached-status regression, skip clears blocked, workflow_run_created noop, malformed yaml, doc-hash drift, legacy `workitem_id` silently ignored. One `TestLoadLocalRun_HostWrapperSmoke` case pins the `os.DirFS("/")`+`TrimPrefix` translation.
- **`discover_workflow_meta_test.go`** — discovery-level integration: valid runtime populates `WorkflowMeta` end-to-end, malformed runtime degrades to `WorkflowMeta == nil` + warn log + sibling discovers normally, v1alpha5 JSON payload carries the `workflow` block.
- **`metadata_test.go`** — `v1alpha4` and `v1alpha5` both parse; `v1alpha3` rejected with upgrade hint.

## Test plan

- [ ] `just build-camp` clean.
- [ ] `go test ./...` passes.
- [ ] Manual smoke: in a campaign with a `.workflow/` runtime, the workitem dashboard preview shows the WORKFLOW progress block.
- [ ] Manual smoke: a `.workitem` written with `version: v1alpha4` still parses and renders.

## Stack

This PR targets `integration/WW0001` and stacks on:

- #276 `lr/WW0001/003-metadata-discovery` — provides the trimmed v1alpha5 metadata loader and `LoadMetadataFS`.

#285 `feat/workitem-create-and-adopt` will stack on this PR.


internal/workitem/localrun.go
```diff
@@ -0,0 +1,172 @@
...
+		case "workflow_run_created", "workflow_run_started":
+			// Run is starting; events follow. No state change here.
+		case "wf_step_start":
+			state.Summary.CurrentStep++
+		case "wf_step_done":
+			state.Summary.CompletedSteps++
+			state.Summary.Blocked = false
+		case "wf_step_skip":
+			// Skipped steps count as forward progress without completion.
+			state.Summary.Blocked = false
```
lancekrogers, line 177: **Replay-semantics divergence from fest (correctness issue).** This PR explicitly clears `Blocked` on `wf_step_skip`. Fest's own replay (`projects/fest/internal/workflow/localstore/store.go`, `replayEvents`) has no case for `wf_step_skip` at all, so on the fest side a skip after a `wf_step_block` leaves the run `Blocked=true`. Two consumers of the same event stream now disagree on the post-skip state.

The PR comment at line 20 calls camp "a thin duplicate of the fields camp needs from fest's localstore. Fest is the source of truth." If fest is the source of truth, camp's replay must produce the same answer fest's does — otherwise `camp workitem` will tell the user a run is unblocked while `fest workflow status` says it's blocked, or vice versa.

Pick one of these and align both sides in the same change:
1. **Skip implies unblock** (camp's current behavior). Update fest's `replayEvents` to also clear `Blocked` on `EventStepSkip`. Add a regression test on the fest side.
2. **Skip is neutral on Blocked** (fest's current behavior). Drop the `Blocked=false` line here in camp and update the comment.

(1) is the more natural product semantic — operators almost always skip *because* they want to move past a block. But that decision belongs in the contract doc (`LOCAL_RUN_STATE.md`), not in two different replay implementations.

internal/workitem/localrun.go
```diff
@@ -0,0 +1,172 @@
...
+	"path/filepath"
+
+	"gopkg.in/yaml.v3"
+
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+)
+
+// LocalRunProgress is camp's read-only view of a fest .workflow/ runtime.
+// Camp and fest are separate Go modules; this is a thin duplicate of the
+// fields camp needs from fest's localstore. Fest is the source of truth.
```
lancekrogers, line 39: **Event-type contract is duplicated as raw string literals across modules — drift is already happening.** Camp hardcodes these here, and fest exports them as named constants in `projects/fest/internal/workflow/localstore/events.go` (`EventWorkflowRunStarted`, `EventStepStart`, etc.). The skip-handler divergence flagged above is the first concrete drift the inevitable drift produced; rename one event in fest tomorrow and `camp workitem` silently stops surfacing it.

Two paths to fix:
1. **Promote event type names and the canonical replay rules to `projects/obey-shared`** (or whichever shared module is the right home), and have both camp and fest depend on it. The shared module owns the protocol; both implementations import the constants.
2. **Publish a versioned wire contract.** A YAML or JSON schema doc in `workflow/design/workitem-workflows/` (next to `LOCAL_RUN_STATE.md`) that explicitly enumerates every event type and the state mutation it implies, with a schema-version field that both replayers verify against. Camp/fest still each have their own replay implementation but at least the source of truth is a single document and tests in both projects pin to its enum.

(1) is the staff-grade answer because it makes drift a compile error rather than a silent display bug. (2) is the lighter-weight fallback. Either is fine, but the current state of "thin duplicate, fest is the source of truth" is documentation that doesn't enforce itself.

internal/workitem/localrun_test.go
```diff
@@ -0,0 +1,290 @@
...
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(path, []byte(body), 0o644); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestLoadLocalRun_NoRuntime(t *testing.T) {
+	dir := t.TempDir()
```
lancekrogers: **Filesystem-safety violation (consistent with #169/#170/#276 review).** New file. `t.TempDir()` + `os.WriteFile` + `os.MkdirAll` build fest-shaped `.workflow/` trees on the host across all 11 cases. The camp project already runs containerized tests through `tests/integration/` (`helpers_container.go`); this kind of multi-file fixture construction is exactly what that harness exists for.

Alternative for keeping the cases as in-process unit tests: refactor `LoadLocalRun` to take an `fs.FS`. Then every test in this file becomes a `fstest.MapFS{...}` literal — no host writes, no `MkdirAll`, the manifest/run/events YAML strings are right next to the assertions, and tests run faster and parallel-safe.

The `fs.FS` refactor also gives you a cleaner answer for the cross-module contract concern (see my comment on line 20) — the camp side replay becomes a pure function over a virtual filesystem and the fest side could share the same pure replayer.

internal/workitem/tui/view.go
```diff
@@ -168,18 +168,69 @@ func priorityBadge(p string) (string, lipgloss.Style) {
 	}
 }
 
+// executionBadge returns a compact 3-char badge for execution metadata.
+// Precedence: blocked > high risk > critical risk > autonomy hint.
```
lancekrogers: **Comment says the wrong order.** Doc comment line 172 reads "Precedence: blocked > high risk > critical risk > autonomy hint." The implementation (and the PR description) actually does "blocked > critical > high > autonomy" — `critical` is checked before `high` in the switch on lines 181-186. Either swap the names in the comment or swap the case order in the switch, but right now they disagree. Easy to fix and worth fixing because someone debugging a misclassified badge will reach for the comment first.

obey-agent (commented):
**Verdict: Request Changes**

## Overview

Phase 005 of `workitem-workflows-WW0001`. Stacked on PR #276. Adds the camp-side reader for fest's `.workflow/` runtime, surfaces progress and metadata fields through the JSON contract at `workitems/v1alpha5`, and lights up the TUI with execution/priority/project/workflow/lineage sections. The product surface looks good; the blockers are the cross-module event-protocol drift this PR introduces and the same containerized-test rule the other WW0001 PRs are tripping.

## Key Findings

- **Replay-semantics divergence from fest (correctness blocker).** Camp's `wf_step_skip` handler clears `Blocked`; fest's replay in `localstore/store.go` has no case for skip at all. Two consumers of the same event stream now disagree on the post-skip state — a `camp workitem` user and a `fest workflow status` user will read different answers for the same `.workflow/`. Pick one semantic, encode it in `LOCAL_RUN_STATE.md`, and align both implementations in the same change. Inline comment has the two options.
- **Event-type contract is duplicated as raw string literals across modules.** Camp hardcodes `"wf_step_start"` etc. inline; fest exports them as named constants in its localstore. The skip divergence above is the first concrete drift the inevitable drift produced — rename one event in fest tomorrow and `camp workitem` silently stops displaying it. Either promote the constants to `projects/obey-shared` so drift becomes a compile error, or publish a versioned wire-schema document that both sides verify against. The current "thin duplicate, fest is the source of truth" comment is documentation that doesn't enforce itself.
- **Filesystem-safety violation in `localrun_test.go` (blocker, same rule as the rest of WW0001).** New file, no inherited debt; 11 cases all build host trees via `t.TempDir()` + `os.WriteFile`. Either move to the container harness in `tests/integration/` or refactor `LoadLocalRun` to accept an `fs.FS` so the tests become pure `fstest.MapFS{...}` literals. The `fs.FS` refactor also dovetails nicely with the shared-replay direction.
- **Doc-comment / implementation disagreement in `executionBadge` (nit).** Comment says precedence is "blocked > high > critical > autonomy"; the switch actually does "blocked > critical > high > autonomy" (which is also what the PR description claims). Easy fix; mattering only because the comment is what a debugger reaches for first.

## Smaller follow-ups (not blockers, mentioned here to avoid an inline of their own)

- `discover_workflow_docs.go:95` repeats the `"error", err.Error()` pattern flagged on #276 for the new `LoadLocalRun` path. Same fix applies (pass `err`, not `err.Error()`).
- `WorkItemWorkflow.DocPath` / `RuntimeDir` are not auto-populated when `LoadLocalRun` succeeds but `.workitem` is missing those fields. The preview pane silently omits them in that case. Worth filling them in from the discovered `.workflow/` location.

## What's Done Well

- Codex's two catches got real fixes with named regression tests: stale-cached-status-doesn't-override-replay, skip-clears-blocked (modulo the divergence above), created-event-is-noop. That's exactly how to integrate a review without rubber-stamping.
- Schema bump to `v1alpha5` is documented in the `SchemaVersion` changelog comment with the full list of new fields, and all new fields use `omitempty` so legacy items without `.workflow/` keep their JSON shape byte-identical.
- TUI degradation rules are thoughtful: row badges drop execution first then priority as terminal width shrinks; preview sections render only when populated; doc-hash-drift surfaces as a visible warning row instead of being silently swallowed.
- Context cancellation is checked in the right places (`LoadLocalRun` entry, replay scanner loop). Errors flow through `camperrors.Wrapf`. Discovery degrades gracefully — malformed `.workflow/` logs and continues, matching the precedent set for `.workitem` malformed metadata.
- The `hashFile` SHA-256 doc-hash check is symmetric with fest's `HashWorkflowDoc` and lets camp surface drift without needing to invoke fest. Good cross-cutting hygiene for a v1 read-only view.

## Staff Standard

Am I proud to ship this? The TUI surface and the schema bump, yes. The replay implementation, no — not while it disagrees with fest on `wf_step_skip` and not while the event-name protocol is two unsynchronized string-literal lists. Align the skip semantic on both sides, promote the event-type constants (or publish a versioned contract), move the new tests to the container harness or `fs.FS`, fix the badge doc comment. Then this is a clean approve.

internal/workitem/localrun.go
```diff
@@ -0,0 +1,172 @@
...
+			var rm localRunManifest
+			if uErr := yaml.Unmarshal(runYAML, &rm); uErr != nil {
+				return nil, camperrors.Wrapf(uErr, "parsing %s/run.yaml", runDir)
+			}
+			replayed := replayLocalRun(ctx, filepath.Join(runDir, "progress_events.jsonl"), rm)
+			out.RunStatus = replayed.Status
+			out.CurrentStep = replayed.Summary.CurrentStep
+			out.CompletedSteps = replayed.Summary.CompletedSteps
+			out.Blocked = replayed.Summary.Blocked
+			out.TotalSteps = rm.Summary.TotalSteps
```
obey-agent, line 121: Follow-up review: this assumes `run.yaml.summary.total_steps` is populated, but the fest localstore introduced in WW0001/004 starts runs with `summary.total_steps` at zero and never repairs it from `WORKFLOW.md`. That means a real fest-created standalone run will show `current_step`/`completed_steps` but omit `total_steps`, and the TUI preview will not render the intended `Step N of M` progress. Either camp needs to derive `M` from `workflow.doc_path`, or fest needs to persist total steps into the local runtime before this dashboard contract can be trusted. Please add an integration-style test using a runtime shaped like `fest workflow init/start/advance` actually writes, not only a hand-authored fixture with `total_steps` already set.

internal/workitem/localrun.go
```diff
@@ -0,0 +1,172 @@
...
+	state := cache
+	state.Summary.CurrentStep = 0
+	state.Summary.CompletedSteps = 0
+	state.Summary.Blocked = false
+	state.Status = "active"
+
+	scanner := bufio.NewScanner(f)
+	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024)
+	for scanner.Scan() {
+		if err := ctx.Err(); err != nil {
```
obey-agent, line 159: Follow-up review: cancellation is being swallowed here. If the scan context is canceled while replaying a large or slow event file, `replayLocalRun` returns a partial state and `LoadLocalRun` reports success, so `camp workitem` can continue with misleading progress instead of honoring cancellation. Please make replay return an error, propagate `context.Canceled` from this branch, and also surface `scanner.Err()` rather than converting read failures into successful partial progress.

obey-agent (commented):
Follow-up review

Verdict: Request Changes

Overview: This PR does a useful job making `.workitem` and `.workflow` state visible in the dashboard, and the replay fixes already incorporated are the right direction. The remaining issues are integration correctness and cancellation semantics in the camp-side runtime reader.

Key Findings:
- Inline: `LoadLocalRun` reads `total_steps` only from `run.yaml.summary`, but fest-created standalone runs currently do not persist that field, so real dashboard progress will omit `Step N of M` even though hand-authored fixtures pass.
- Inline: event replay swallows context cancellation and scanner/read failures, returning partial progress as success.

What's Done Well:
- New runtime read errors use `camperrors` at the package boundary.
- Discovery remains resilient when optional runtime files are malformed.
- The TUI sections are clear and preserve legacy no-metadata behavior.

Staff Standard: Not yet. I would want a real fest-localstore-shaped fixture and proper cancellation/error propagation before being comfortable relying on this for launch dashboard visibility.

Verification: `go test ./internal/workitem/...` passed in the PR worktree.

obey-agent (commented):
**Verdict: Approve**

## Overview

WW0001 phase 009 sequence 009.02 amends PR #281 to align the dashboard visibility surface with the trimmed v1alpha5 schema from 009.01. The TUI's `executionBadge`/`metadataPriorityBadge` and the EXECUTION/PRIORITY/PROJECT/LINEAGE preview sections are removed (depended on now-deleted nested metadata blocks), but the WORKFLOW progress preview (sourced from `.workflow/` runtime, not the trimmed metadata) is preserved intact. `WorkitemSchemaVersion` bumps to `v1alpha5` while still accepting `v1alpha4` as backward compatible since the on-disk shape is unchanged. Camp's `localrun` parser gets a `LoadLocalRunFS`/`replayLocalRunFSE` variant so unit tests run via `fstest.MapFS`; replay now propagates `ctx.Err()` mid-loop, surfaces `bufio.Scanner.Err()` via camperrors, and wraps file-open failures (D030 #8). Wire-schema event-type strings for `progress_events.jsonl` are hoisted to named constants in `localrun.go` documenting the cross-repo contract with fest's localstore (D030 #9, Option B documented; future improvement: extract to `obey-shared`).

## Verification

- `just build-camp` — successful build on PR head `5307284`.
- `go test ./...` — all packages pass (no failures).
- New `localrun_fs_test.go` covers: missing workflow dir → nil progress, basic progress with replay over event stream, legacy `workitem_id` field tolerated silently (D024 alignment), malformed YAML → camperrors parse error.

## What's Done Well

- Render layer cleanly drops the badge/preview blocks that depended on `Execution`/`PriorityInfo`/`Project`/`Lineage` while leaving WORKFLOW progress untouched — the dashboard's most useful preview survives the schema trim.
- `LocalRunProgress` carries no `WorkitemID` field; loader silently ignores legacy `workitem_id:` keys via `yaml.Unmarshal`'s default unknown-field tolerance, so older `.workflow/` runs continue to render without modification.
- `replayLocalRunFSE` returns `(state, error)` so partial state from a corrupt event stream no longer surfaces as success; `replayLocalRunFS` is preserved as a thin wrapper for callers that don't propagate errors.
- Event-type constants (`wfEventStepStart`, etc.) make the cross-repo contract explicit in one location, with a code comment pointing at the future `obey-shared` extraction. Renaming an event type now changes one place rather than scattered string literals.
- `acceptedWorkitemVersions` keeps `v1alpha4` accepted as backward-compatible — existing `.workitem` files continue working unchanged, matching the operator-facing note in the task spec.

## Staff Standard

Would I be proud to ship this? Yes. The visibility surface is honest about what the trimmed schema can populate, and the replay propagation fix closes a real "silent partial state" hazard for the dashboard. The cross-repo event-type contract is documented in code, which is a meaningful improvement even before the obey-shared extraction. Approving.

internal/workitem/discover_workflow_docs.go
```diff
@@ -97,6 +97,31 @@ func discoverWorkflowDocs(ctx context.Context, campaignRoot, rootDir string, wfT
...
+			if item.WorkflowMeta.ActiveRunID == "" {
+				item.WorkflowMeta.ActiveRunID = run.ActiveRunID
+			}
+			item.WorkflowMeta.CurrentStep = run.CurrentStep
+			item.WorkflowMeta.TotalSteps = run.TotalSteps
+			item.WorkflowMeta.CompletedSteps = run.CompletedSteps
+			item.WorkflowMeta.RunStatus = run.RunStatus
+			item.WorkflowMeta.Blocked = run.Blocked
+			item.WorkflowMeta.DocHashChanged = run.DocHashChanged
+		}
```
obey-agent, line 123: The new `LoadLocalRun` -> `item.WorkflowMeta` merge here (lines 100-123) has no test. `localrun_test.go` and `localrun_fs_test.go` cover `LoadLocalRun` in isolation, but nothing in `discover_test.go` or a new test asserts that:

- a design/explore item with a valid `.workflow/workflow.yaml` gets its `WorkflowMeta` populated end-to-end;
- a directory whose `.workflow/` is malformed yields a discovery item with `WorkflowMeta == nil` and a warn-level slog line (not a discovery failure);
- the JSON payload at v1alpha5 actually carries the `workflow` block when populated.

That last one matters because it's the v1alpha5 contract. Right now nothing fails if a future refactor stops wiring `WorkflowMeta` into the model. Please add a discovery-level test using `fstest.MapFS` to cover the merge plus malformed-runtime graceful-degrade path.

internal/workitem/localrun_test.go
```diff
@@ -0,0 +1,290 @@
...
+	got, err := LoadLocalRun(context.Background(), dir)
```
obey-agent: The whole `localrun_test.go` file (290 lines, 11 tests) writes real files into `t.TempDir()` and exercises `LoadLocalRun` end to end. The parallel file `localrun_fs_test.go` already covers the same parser/replay logic via `fstest.MapFS` without touching the host filesystem.

The project rule (D029 + repo memory) is that parser-only code uses `fs.FS`/`fstest.MapFS`, and host-mutating tests are reserved for the containerized integration harness. `LoadLocalRun` is reader-only, so the `fstest.MapFS` form is the right shape; the host-tempdir form is redundant.

Please either:

1. Delete `localrun_test.go` and move any unique cases (stale cached status, skip clears blocked, created noop, malformed manifest, doc hash drift) into `localrun_fs_test.go` against `fstest.MapFS`; or
2. If you specifically want to cover the `os.DirFS("/")` + `TrimPrefix` path (the host wrapper), keep one minimal smoke test for that wrapper and remove the other 10. Replay/state correctness should live in MapFS tests.

This is not a stylistic preference. Two parallel test suites for the same behavior diverge over time, and the host-tempdir suite tempts contributors to add more host-mutation tests for code that doesn't need them.

obey-agent:
The PR description is out of sync with the actual diff and needs a refresh before merge.

The body claims:
- Row badges: execution badge alongside priority badge with precedence `blocked > critical-risk > high-risk > autonomy`
- Preview pane: EXECUTION, PRIORITY, PROJECT, WORKFLOW, LINEAGE sections

The diff to `internal/workitem/tui/view.go` adds only the WORKFLOW preview section and a width-collision tweak for the existing priority badge. No execution badge, no EXECUTION/PRIORITY/PROJECT/LINEAGE preview sections. Looks like those came in via the original 005.01 commit and were trimmed out by the 009.01 schema-minimum work (`0a9cb2f feat(workitem): trim metadata to v1alpha4 minimum`).

A future reader trying to understand what landed in this PR cannot trust the body. Please update the Summary to match the actual surface (only WorkflowMeta progress in JSON + WORKFLOW preview section + width fallback for the existing priority badge) and adjust the "row badges" precedence claim to whatever the trimmed schema actually supports today.

obey-agent (commented):
## Staff Review: dashboard visibility for .workitem + .workflow (WW0001 phase 005)

**Verdict: Request Changes**

This is mostly solid work that adds a real product surface (camp seeing fest's `.workflow/` runtime in the dashboard and JSON) with a thoughtful cross-repo contract handled via duplicated constants. Event replay is correct, the codex-caught stale-cache bug is fixed, and graceful degradation on malformed runtime files is in place. Two real issues and one doc issue keep this from a clean approve.

### Blockers

1. **No test for the WorkflowMeta merge in discoverWorkflowDocs.** The new code at lines 100 to 123 of `discover_workflow_docs.go` is the only path that populates `WorkItem.WorkflowMeta`, and it is exercised by zero tests. `LoadLocalRun` is well-tested in isolation, but the integration into discovery (and therefore into the v1alpha5 JSON contract and TUI preview) is unverified. See inline comment on that file. Add an `fstest.MapFS`-driven discovery test that asserts the merge happens, that malformed `.workflow/` degrades to `WorkflowMeta == nil` plus a warn log, and that the v1alpha5 JSON payload carries the `workflow` block when populated.

2. **Duplicate host-tempdir test suite for parser-only code.** `localrun_test.go` (290 lines, 11 cases) writes real files via `t.TempDir()` to exercise behavior already covered by `localrun_fs_test.go` against `fstest.MapFS`. Project rule (D029 + repo memory) is that parser-only code uses `fs.FS`/`fstest.MapFS` and host-mutating tests are reserved for the containerized integration harness. See inline comment for the recommended cleanup. Keep one smoke test for the `os.DirFS("/")` wrapper if desired, but the replay/state correctness cases belong on MapFS.

### Documentation drift (also needs fixing)

3. **PR description does not match the diff.** The body advertises an execution badge alongside the priority badge with a four-level precedence rule, and EXECUTION/PRIORITY/PROJECT/WORKFLOW/LINEAGE preview sections. The actual `tui/view.go` diff adds only a WORKFLOW preview section and a width-fallback tweak. Looks like 009.01 trimmed those surfaces and the PR body was not updated. Posted a separate PR-level comment with the specifics.

### What's done well

- `LocalRunProgress` is a deliberately narrow struct, and the comment block at the top of `localrun.go` calls out the cross-repo contract with fest's localstore plus the future obey-shared extraction path. That is exactly the right framing for a duplicate-by-design type.
- Wire event-type strings are named constants, not magic strings, and the comment block ties them back to the EVENT_SCHEMA contract.
- Replay correctly treats events as authoritative and resets cached status to `active` before replaying. The `TestLoadLocalRun_StaleCachedStatusDoesNotOverrideReplay` regression test pins that behavior.
- Discovery degrades gracefully on malformed runtime: warn-level slog plus continue, no crashed dashboard. Matches the established `.workitem` parse-error handling.
- Backward compat for `v1alpha4` parsing is real: `acceptedWorkitemVersions` carries both versions and `TestLoadMetadata_V1Alpha4BackwardCompat` covers it.
- Context is checked before file IO in `LoadLocalRunFS` and inside the replay loop (`replayLocalRunFSE` calls `ctx.Err()` on every event). Project error package used throughout (`camperrors.Wrapf`).

### Staff standard

I would be proud to submit the runtime-reader and replay logic. I would not submit the discovery integration without a test, and I would not submit a PR whose body advertises a feature surface the diff does not contain. Fix those three items and this is a clean approve.

obey-agent (commented):
WW0001 phase 009 sequence 009.02 review-feedback fixes pushed in commit 22e691c.

**Discovery-level merge test** — added `discover_workflow_meta_test.go` covering:

- `TestDiscoverDesign_PopulatesWorkflowMetaFromLocalRun` — design item with valid `.workflow/workflow.yaml` + run.yaml + events gets `WorkflowMeta` populated end-to-end (workflow_id, total_steps, completed_steps via event replay, current_step, run_status).
- `TestDiscoverDesign_MalformedRuntimeDegradesGracefully` — design item with malformed `.workflow/workflow.yaml` yields `WorkflowMeta == nil` plus a warn-level slog line, and a clean sibling in the same root still discovers normally (no discovery failure).
- `TestDiscoverDesign_JSONPayloadCarriesWorkflowBlock` — `NewPayload(items)` JSON output for a populated workflow item contains the `"workflow":{"workflow_id":"wf-payload","total_steps":5,...}` block, locking the v1alpha5 contract.

Discovery uses `os.ReadDir` directly so this file uses `t.TempDir` matching the existing `discover_test.go` pattern. The file header comment notes that `LoadLocalRun` parsing is exhaustively tested via `fstest.MapFS` in `localrun_fs_test.go` and this file only exercises the discovery-level wiring.

**Localrun test dedup** — deleted `localrun_test.go` (290 lines, 11 host-tempdir tests) and migrated every unique case to `localrun_fs_test.go` driven by `fstest.MapFS`:

- Manifest-only, summary-as-cache, basic event replay
- Event replay overrides stale summary
- Completed and blocked status
- Stale-cached-status regression (codex catch)
- Skip event clears blocked
- workflow_run_created event is noop
- Malformed yaml
- Doc-hash drift
- Legacy workitem_id silently ignored (D024 backward compat)

Kept exactly one `TestLoadLocalRun_HostWrapperSmoke` case in `localrun_fs_test.go` to pin the `os.DirFS("/")`+`TrimPrefix` translation in the production wrapper. The single-line `writeHostFile` helper lives in `host_wrapper_test.go` so the file containing the smoke test imports nothing host-mutating from elsewhere.

**Cleanup along the way:** dropped `replayLocalRun` / `replayLocalRunFS` / `hashFile` host wrappers from `localrun.go` and `emitCandidate` host wrapper from `discover_custom.go` — now that callers go through the FS variants directly, the host wrappers had zero call sites and were tripping `unused` lint. Also dropped one ineffectual `badgeWidth = 0` in `tui/view.go`.

**Verification:** `go build ./...` clean; `go test ./internal/workitem/...` passes (all new tests included); the workitem subtree is lint-clean (the two remaining lint hits are in `priority/` and pre-date this PR).

**On the PR description drift:** flagged for the original author to update the body to reflect the actual diff (no execution badge, no EXECUTION/PRIORITY/PROJECT/LINEAGE preview sections — those were trimmed in 009.01).

Re-requesting review on the latest head (22e691c).

obey-agent (commented):
**Verdict: Request Changes** (posted as COMMENT due to self-author block).

Head reviewed: `22e691c245a034299a06c8bd0fd484b35832398a` (re-review after `530728460` baseline).

## Blockers from prior review

1. **Discovery merge test — FIXED.** `discover_workflow_meta_test.go` covers the three asserted paths cleanly: valid runtime populates WorkflowMeta end-to-end (incl. CurrentStep/CompletedSteps/RunStatus derived from events), malformed `.workflow/workflow.yaml` degrades to `WorkflowMeta == nil` without aborting discovery (sibling clean item still emitted), v1alpha5 JSON payload carries the workflow block with `workflow_id` and `total_steps`. The t.TempDir choice is defensible here because `discoverDesign` uses `os.ReadDir` directly, not `fs.FS`. Parser correctness lives in `localrun_fs_test.go` via fstest.MapFS, which is the right split.

2. **Duplicate host-tempdir suite — FIXED.** `localrun_test.go` deleted (290 lines), all unique cases migrated into `localrun_fs_test.go` driven by fstest.MapFS. One `TestLoadLocalRun_HostWrapperSmoke` retained to pin the `os.DirFS("/")+TrimPrefix` translation in the production wrapper. Unused host wrappers (`replayLocalRun`, `hashFile`, `emitCandidate`) removed from production code; only FS variants remain.

## New blocker on this head

3. **PR description still drifts from the diff.** The Summary section claims:
   - "Row badges: execution badge alongside priority badge with precedence `blocked > critical-risk > high-risk > autonomy`. Narrow terminals drop execution then priority."
   - "Preview pane: `EXECUTION`, `PRIORITY`, `PROJECT`, `WORKFLOW`, `LINEAGE` sections..."

   `internal/workitem/tui/view.go` contains only `WORKFLOW` (one grep hit). There is no EXECUTION badge, no priority-badge precedence logic, no PROJECT/PRIORITY/LINEAGE preview sections in the diff. This was flagged on the previous review head; it is unchanged. For a festival-traced PR that bumps the JSON contract to v1alpha5, the description is part of the audit trail and must match what is being merged. Trim the description to: WORKFLOW preview section, width-fallback tweak, v1alpha5 schema bump, localrun reader, discovery-merge wiring, malformed-runtime degradation, and the test reshape from this commit. Then this is a clean approve.

## Verification ran locally

- `go test -race -count=1 ./internal/workitem/...` all packages green (workitem, workitem/priority, workitem/tui).

## What's good on this revision

- Clean test taxonomy: parser correctness in MapFS, discovery merge in real FS (matches the package's own boundary), single host smoke for the os.DirFS wrapper.
- Production code shrank: dead host wrappers gone, only FS variants exported.
- Tiny `view.go` cleanup (drop ineffectual badgeWidth=0 assignment) is correct.

Fix the description, no re-review needed for the code.

PR #282, opened 2026-05-13
fix(intent/tui): harden explorer against log corruption, dungeon-move freeze, unselectable filtered search

### LINKED ISSUE #278: TUI: camp intent explore freezes when moving an intent from one dungeon status to another (e.g. Someday → Done)
# BUG: `camp intent explore` TUI freezes when moving an intent from one dungeon status to another

**Severity:** High (requires kill -9 to recover; user loses work in flight)
**Discovered:** 2026-05-12
**Project:** projects/camp
**camp version:** dev (built 2026-05-11T20:29:33Z, commit d6eaacd)
**Component:** `internal/intent/tui/explorer/`

## Summary

In the `camp intent explore` TUI, attempting to move an intent that is already in a dungeon status (Someday) to another dungeon status (Done) causes the TUI to freeze. Input is no longer accepted; the user must kill the process to recover.

The other move paths (e.g. Inbox → Ready, Ready → Active) appear to work normally.

## Reproduction

1. Run `camp intent explore` from a campaign with at least one intent in `dungeon/someday/`.
2. Navigate to the someday intent in the list.
3. Open the move action (likely `m` key based on action_menu.go).
4. Select **Done** from the move status options.
5. Press Enter.
6. **Observe:** the TUI stops responding to keystrokes. The screen does not update. The only recovery path is killing the process from another terminal (`pkill camp` or kill the parent shell).

## Likely Culprit

In `projects/camp/internal/intent/tui/explorer/actions.go`, the `updateMove` function handles the "enter" key for the move overlay. When the destination is a dungeon status, it calls `startDungeonReasonInput`:

```go
// actions.go ~lines 50-70
case "enter":
    if m.intentToMove != nil {
        newStatus := moveStatusOptions[m.moveStatusIdx].status
        if m.intentToMove.Status == newStatus {
            m.statusMessage = "Already at " + newStatus.String()
            m.focus = focusList
            m.intentToMove = nil
            return m, nil
        }
        // Require reason for dungeon moves
        if newStatus.InDungeon() {
            m.startDungeonReasonInput(m.intentToMove, newStatus, "move")
            m.intentToMove = nil
            return m, nil
        }
        m.focus = focusList
        return m, m.moveIntent(m.intentToMove, newStatus)
    }
```

The dungeon-reason input is then driven via `focusDungeonReason`. Two failure modes are plausible:

1. **Rendering gap:** `focusDungeonReason` has no corresponding render branch (or it is hidden by overlay z-ordering) when the source intent is already in a dungeon status. The user sees the previous frame and concludes the TUI is frozen, but the model is actually waiting for input on an invisible textinput.
2. **Key routing gap:** keys typed after the focus shift to `focusDungeonReason` are not routed to `m.dungeonReasonInput.Update(msg)`. The input field accepts no characters and there is no way to confirm or cancel.

Files to check for the render gap:

- `projects/camp/internal/intent/tui/explorer/render.go` — search for `focusDungeonReason`. If the case is missing or returns the prior frame, that is the freeze.
- `projects/camp/internal/intent/tui/explorer/update_overlays.go` — there is no `updateDungeonReason` matching the pattern of `updateSearch` and `updateFilterBar` (grep confirms). Key routing when `m.focus == focusDungeonReason` is likely fallthrough to the default list handler, which ignores most keys.

The `dungeon → dungeon` case is also worth considering separately from `non-dungeon → dungeon`. A Someday intent being moved to Done is reclassifying within the dungeon tree, which may not have been a tested path even if `non-dungeon → dungeon` works.

## Expected Behavior

Moving an intent from one dungeon status to another should:

1. Prompt for a reason (consistent with existing dungeon-move behavior from non-dungeon sources).
2. Render a visible textinput so the user knows the TUI is waiting on them.
3. Accept the reason and execute the move.
4. Return focus to the list with the intent now visible under its new status (or filtered out if the active filter excludes it).

Alternative interpretation: dungeon → dungeon moves may not need a reason since the intent is already terminal. In that case, skip `startDungeonReasonInput` entirely when the source is already `InDungeon()` and execute the move directly.

## Impact

- Hard freeze blocks all input. The user loses their place in the list and any unsaved navigation state.
- Forces process kill, which is a bad UX failure even setting aside the lost work.
- Discovered while triaging real intents, so this hits real workflow.

## Workaround

Until fixed, move dungeon intents via the CLI rather than the explore TUI:

```bash
camp intent move <intent-id> done --reason "..."
```

Or use `camp intent archive`, `camp intent done`, etc. as appropriate.

## Suggested Test

When fixing, add to `projects/camp/internal/intent/tui/explorer/explorer_test.go` (or a new test file) a regression test that:

1. Constructs a Model with one intent in `StatusSomeday`.
2. Opens the move overlay, selects Done, sends `enter`.
3. Asserts that `m.focus == focusDungeonReason` (or whatever the resolved design says).
4. Sends a few characters and `enter` for the reason.
5. Asserts the intent's path is now under `dungeon/done/` and focus returned to `focusList`.

Cover at least these source/destination pairs: `inbox → done`, `someday → done`, `someday → killed`, `done → archived`. The first works today (based on the user's report that non-dungeon → dungeon works). The rest are the regression surface.

### LINKED ISSUE #279: TUI: camp intent explore search filter does not hand focus back to the list on Enter; filtered list is unselectable
# BUG: `camp intent explore` search filter does not hand focus back to the list on Enter

**Severity:** Medium (feature is functional but unusable; users cannot select from filtered results)
**Discovered:** 2026-05-12
**Project:** projects/camp
**camp version:** dev (built 2026-05-11T20:29:33Z, commit d6eaacd)
**Component:** `internal/intent/tui/explorer/`

## Summary

In the `camp intent explore` TUI, after pressing `/` to enter search mode and typing a query that filters the list, there is no obvious way to commit the search, hand focus back to the list, and navigate the filtered results to select an intent.

Pressing Enter (per the code) is supposed to blur the search input and return focus to the list, but in practice the user is unable to navigate the filtered list afterward. Either Enter is not handing off focus correctly, or the visual state does not communicate that focus has moved.

## Reproduction

1. Run `camp intent explore` from a campaign with several intents.
2. Press `/` to enter search mode. A "Search intents..." input appears.
3. Type a query that narrows the list (e.g. `camp` to filter to camp-related intents).
4. Observe the list filters live as you type.
5. Press Enter.
6. **Observe:** the user has no way to navigate the filtered list with `j`/`k` or arrow keys and select an entry. There is no visible indication that focus has returned to the list, or the navigation is not actually working.

## Likely Culprit

In `projects/camp/internal/intent/tui/explorer/update_overlays.go`, the `updateSearch` function handles Enter as:

```go
case "enter":
    // Exit search mode but keep filter active
    m.focus = focusList
    m.searchInput.Blur()
    return m, nil
```

So the model intends to:
1. Set focus back to the list (`m.focus = focusList`).
2. Blur the search input.
3. Keep the active filter (the `applyFilters()` call ran on each keystroke).

Plausible failure modes:

1. **Selection index out of range.** The `selected` cursor in the model may still be pointing at a position from the unfiltered list. After filtering, that index may be out of bounds for `filteredIntents` and the navigation handler may silently reject `j`/`k` or jump to an invalid entry. Check whether `applyFilters()` resets `selected` to 0 (or clamps it to `len(filteredIntents)-1`).
2. **Render does not reflect blur.** The search input may still render as if focused, leaving the user uncertain about where keystrokes are going. `render.go` likely still draws the searchInput prominently.
3. **Help hints don't communicate the handoff.** Even if Enter works, the user has no visible signal that the list is now navigable. The "press Enter to commit search and navigate" affordance is missing from the help line.

## Expected Behavior

After typing a search query and pressing Enter:

1. The search input should visibly de-emphasize (border / color change indicating no focus).
2. The list should visibly emphasize (selection highlight on the first filtered result).
3. `j`, `k`, `↑`, `↓` should move the selection within the filtered list.
4. Enter on a selected intent should open the viewer.
5. `Esc` should clear the search filter and restore the full list (already does, per `updateSearch` esc handler).
6. The status/help line should change to reflect "list navigable" rather than "typing search."

## Impact

- Search filter is unusable as a way to find and select an intent. The user can narrow the list visually but cannot pick from it.
- Users fall back to scrolling the unfiltered list or quitting and using `camp intent list --status <s>` from the shell.
- This intersects with the third bug surface implied by `look-into-intent-promotion-workflow-20260309-233612` (intent promotion sometimes skipping the design/festival option). If users cannot select a specific intent from a filtered list, they cannot reliably promote one.

## Workaround

Use the filter chips (`Type`, `Status`) via Tab navigation instead of the `/` text search, since those operate on the same `filteredIntents` slice and may handle the focus handoff correctly.

Or use the CLI:

```bash
camp intent list --status ready          # narrow by status
camp intent promote <intent-id>          # act directly via CLI
```

## Suggested Test

Add to `projects/camp/internal/intent/tui/explorer/filtering_test.go` (or a new `search_test.go`) a regression that:

1. Constructs a Model with multiple intents, two of which match query "foo".
2. Sends key `/` to enter search mode.
3. Sends characters `f`, `o`, `o`.
4. Asserts `len(m.filteredIntents) == 2`.
5. Sends `enter`.
6. Asserts `m.focus == focusList`.
7. Asserts `m.searchInput.Focused() == false`.
8. Asserts `m.selected` is a valid index into `filteredIntents` (>= 0 and < len(filteredIntents)).
9. Sends `j` (or down arrow).
10. Asserts `m.selected` incremented (or stayed valid).
11. Sends `enter` (or whatever opens the viewer) and asserts the viewer is now focused on the selected filtered intent.

## Related

- `look-into-intent-promotion-workflow-20260309-233612.md` — the reported "promote doesn't always offer design/festival options" bug might be partly caused by this. If users cannot reliably select a filtered intent, the promote action may be operating on an unintended intent.
- The viewer-internal `/`-search in `projects/camp/internal/intent/tui/viewer.go` is a separate surface (searches within an intent's body) and is not affected by this bug.

### LINKED ISSUE #280: TUI: camp intent explore rendering corrupted by background auto-commit log output (affects every action that triggers an auto-commit)
# BUG: `camp intent explore` TUI rendering corrupted by background auto-commit log output

**Severity:** High (user must quit TUI to recover; corrupts every long-running action that triggers auto-commit)
**Discovered:** 2026-05-12
**Project:** projects/camp
**camp version:** dev (built 2026-05-11T20:29:33Z, commit d6eaacd)
**Component:** `internal/intent/tui/explorer/` + auto-commit subsystem

## Summary

When an action in the `camp intent explore` TUI triggers an auto-commit (e.g. promoting an intent to a design doc), the auto-commit subsystem writes structured INFO log lines to stdout/stderr. Those lines collide with the bubbletea render loop writing to the same TTY. The visible result is a garbled screen: log text overlaps the help line, status bar, and list rendering. The TUI is unrecoverable from this state; the user has to quit.

The underlying action does complete successfully (the promote command worked: design dir created, intent moved from `ready/` to `active/`), but the visual state lies about what just happened and the user cannot tell whether to retry, cancel, or proceed.

## Reproduction

1. Run `camp intent explore` in a campaign.
2. Select an intent in `ready` status.
3. Trigger the promote-to-design action (the key that promotes a ready intent into a `workflow/design/` directory).
4. **Observe:** garbled rendering. The user-pasted log shows:

```
2026/05/12 22:34:09 INFO removing stale lock path=/Users/lancerogers/Dev/AI/obey-campaign/.git/index.lockq: quit
                                                                                                         2026/05/12 22:34:10 INFO removing stale lock path=/Users/lancerogers/Dev/AI/obey-campaign/.git/modules/projects/obey-platform-monorepo/index.lock
                                                                                                                          2026/05/12 22:34:10 INFO cycle completed, cleaning locks operation=stage cycle=1 removed=2 active=0 failed=0
                                                                                                      2026/05/12 22:34:10 INFO wj/k: navigate . v: preview . /: search . tab: filter . n: new . Space: gather mode . f: full . ?: help . q: quit
Promoted 'camp cli should support bookmarks' → design doc 'workflow/design/camp-cli-should-support-bookmarks-20260306-084654'
```

Two collisions are visible in the first and last lines:

- `index.lockq: quit` — log line ends with `index.lock`, immediately followed by the help bar's `q: quit` glyph being painted on top of the same row.
- `INFO wj/k: navigate ... q: quit` — an INFO timestamp prefix is jammed against the start of the help line.

5. The TUI is unusable. Only `q` (quit, or whatever non-rendered terminal escape) gets out.

## Likely Culprit

Two subsystems are writing to the same file descriptor (the TTY):

1. **bubbletea render loop** owns the alternate screen and emits ANSI control sequences for full-screen rendering.
2. **The camp auto-commit / staging subsystem** (likely `projects/camp/internal/commit/` or wherever the "stage cycle ... cleaning locks" log lines originate) uses the standard library `log` package, which by default writes to `os.Stderr` with a timestamped prefix.

When the TUI is in alternate-screen mode, anything else writing to stderr/stdout corrupts the frame. bubbletea programs are supposed to redirect log output away from the TTY (typically to a file or to bubbletea's `tea.LogToFile`) for exactly this reason.

The promote action triggers `git add` / `git commit` under the auto-commit subsystem, which logs lock-management state through the default logger. Those log calls land on stderr, which the TUI does not own.

Confirming evidence:

- The `INFO removing stale lock` and `INFO cycle completed, cleaning locks operation=stage cycle=...` lines match exactly the log format we saw earlier in this session from the non-TUI `camp intent add` command (which ran into the same lock-collision sequence with the index.lock file).
- The same auto-commit subsystem is presumably invoked by the TUI's promote action, but the TUI did not redirect logs.

## Where to Fix

In the explorer's program startup (probably `projects/camp/cmd/camp/intent/explore.go` or wherever `tea.NewProgram(...)` is called), redirect the standard logger to a file before launching the TUI:

```go
import (
    tealog "github.com/charmbracelet/log"
    "log"
)

// At program start, before tea.NewProgram:
f, err := tea.LogToFile("/tmp/camp-explore.log", "debug")  // or a path under .campaign/
if err != nil {
    return err
}
defer f.Close()
log.SetOutput(f)  // redirect stdlib log too
```

Or, more thoroughly: wherever the auto-commit subsystem logs, audit it to use a configured logger that the TUI can hand a sink to (rather than the default stderr writer). Search for `log.Printf`, `log.Println`, `log.Print`, and any `slog.Default()` usage in the commit/auto-commit paths.

The structured logs (those have `INFO`, `key=value` shape) suggest `log/slog` or `charmbracelet/log` is in use. Whichever it is, the TUI entry point should rebind its handler to a file before `tea.NewProgram` runs.

## Expected Behavior

When the TUI triggers an action that produces background log output:

1. No log lines should appear in the rendered TTY.
2. The user sees a clean status update from the TUI (e.g. a toast or status bar message: "Promoted to workflow/design/..." or "Auto-commit in progress... done").
3. Logs are still captured to a file or to bubbletea's debug log sink, so they remain recoverable for debugging.
4. The user can continue using the TUI after the action completes.

## Impact

This affects every TUI action that triggers an auto-commit. The promote action is the case the user hit, but the same pattern likely affects:

- `n` (new intent) — `camp intent add` flow auto-commits on save.
- Move actions (the dungeon-move freeze in bug #001 may also be related; if the move triggers auto-commit during the freeze, log corruption may be a confounding factor).
- Gather mode and any action that ends in a save-and-commit.

In other words, the TUI is unstable any time the underlying CLI subsystem decides to commit. That is most non-trivial actions.

## Workaround

Until fixed, perform mutating intent actions via the CLI rather than the TUI:

```bash
camp intent promote <intent-id>     # promote to design
camp intent move <intent-id> <status>
camp intent archive <intent-id> --reason "..."
```

The CLI prints the same log lines, but since there is no TUI fighting for the same TTY, the output is just normal scroll-back and the action completes cleanly.

## Suggested Test

Two layers of regression:

1. **Unit/integration:** When the explorer TUI is initialized in tests (already done in `explorer_test.go`), assert that the default logger has been redirected away from os.Stderr by the time `tea.NewProgram` would be called. This is testable by capturing stderr before model construction and verifying it stays empty after a simulated action that exercises the auto-commit path.
2. **Manual smoke:** Run `camp intent explore`, perform a promote, verify the TUI remains coherent (no INFO log lines visible in the rendered frame), and verify the action succeeded by checking `git log` and `workflow/design/` for the new directory.

## Related

- Bug #001 (dungeon-move freeze) — may interact with this bug. If the dungeon move also triggers auto-commit, log corruption could be a confounding factor in the freeze. Fix this one first and re-test #001 to confirm whether the freeze is purely a focus/render issue or whether log noise contributed.
- The auto-commit lock-retry log lines (`INFO removing stale lock`, `INFO cycle completed`) suggest the auto-commit subsystem has its own retry loop with explicit logging. That logging is useful from the CLI but must be silenced (or sunk to a file) under the TUI.
- The actual promote in this case succeeded despite the visual mess. The result is at `workflow/design/camp-cli-should-support-bookmarks-20260306-084654/` with a single README copying the intent body. The intent itself moved to `.campaign/intents/active/`. So the auto-commit was working; only the rendering was broken.

lancekrogers:
## Summary

Fixes three bugs in the `camp intent explore` TUI surfaced while triaging the intent inbox:

- **#280 (High)** — auto-commit INFO logs corrupted bubbletea rendering, making the TUI appear frozen and forcing kills.
- **#278 (High)** — dungeon→dungeon moves (e.g. Someday → Done) appeared frozen because the auto-commit retry stalled without any in-progress feedback.
- **#279 (Medium)** — the \`/\` search filter narrowed the list live but pressing Enter left the cursor on a group header (\`cursorItem=-1\`), so j/k looked like a no-op and the filtered list felt unselectable.

The fixes are scoped to the explorer TUI and the small slice of \`internal/git\` that emits the noisy lock-cleanup logs. No protocol or schema changes.

## Commits

1. \`silence auto-commit logs that corrupted explorer TUI\` — closes #280
2. \`surface in-progress feedback during dungeon-move auto-commit\` — closes #278
3. \`hand cursor to first filtered item when exiting search\` — closes #279
4. \`update search-mode help text to match new Enter behavior\` — follow-up to #279

Each commit can be reverted independently if needed.

### Commit 1 — log corruption defense in depth

Two-part fix:

- **Demote three chatty INFO logs to Debug** in \`internal/git/retry.go\` and \`internal/git/lock_unix.go\`. They were operational chatter (\`removing stale lock\`, \`cycle completed, cleaning locks\`, \`waiting before next cycle\`) with no user-actionable information at INFO level. The \`lock is active, cannot remove\` Warn is preserved. Errors are still returned through the normal Go error path.
- **Install a file-routing slog handler at the TUI entry point** in \`cmd/camp/intent/explore.go\`. Writes to \`<campaignRoot>/.campaign/logs/intent-explore.log\` instead of stderr, restored to the previous default on exit. Falls back to \`io.Discard\` if the log dir cannot be created so the TUI is still safe.

Regression tests cover the swap, message capture, and the fallback path.

### Commit 2 — dungeon-move freeze (in-progress feedback)

The dungeon reason flow itself was wired correctly (View branch, key router, focused textinput). What looked like a freeze was the auto-commit retry stalling while clearing stale git index locks, combined with the log corruption (#280). Once the user committed a reason and focus returned to the list, there was no in-progress indicator so they assumed the TUI was hung.

Fix: set \`statusMessage\` to \`Moving to <status>...\` or \`Archiving...\` immediately when the reason is accepted. The \`moveFinishedMsg\` / \`archiveFinishedMsg\` handler replaces it with the final outcome when the async path completes.

Regression test covers the Someday → Done transition end-to-end: focus through \`focusDungeonReason\` and back to \`focusList\`, the reason input actually rendering (guards against future regressions that remove the View branch), empty-reason rejection, esc cancel, and that a non-nil tea.Cmd is dispatched.

### Commits 3 + 4 — search Enter handoff

After typing in \`/\` search, \`applyFilters\` reset the cursor to \`cursorGroup=0, cursorItem=-1\` — a group header position. Pressing Enter blurred the input and returned focus to the list, but the cursor stayed at the header, so j/k looked like a no-op and users could not select from the filtered results.

Fix:

- New \`placeCursorAtFirstItem\` helper in \`navigation.go\` that finds the first non-empty group, expands it, and parks the cursor on item 0. Returns false on no matches so the caller can surface a recovery hint.
- \`updateSearch\` calls it on Enter. On no-match, sets \`statusMessage\` to \`No matches — press / to refine or Esc to clear\`.
- Help text in render.go updated from \`(enter to search, esc to cancel)\` to \`(enter to navigate filtered list, esc to clear)\` — \"enter to search\" was misleading since typing already searches live.

Regression tests cover: cursor placement on a real intent after Enter, group auto-expansion, the no-matches hint path, and the safe fallback when called with empty groups.

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go test ./...\` — full suite passes
- [x] New regression tests pass:
  - \`TestQuietSlogDuringTUI_RedirectsToLogFile\`
  - \`TestQuietSlogDuringTUI_FallsBackToDiscardOnDirError\`
  - \`TestDungeonMove_SomedayToDone_TransitionsAndSurfacesProgress\`
  - \`TestDungeonMove_EscCancelsAndRestoresList\`
  - \`TestSearchEnterHandoff_PlacesCursorOnFirstFilteredItem\`
  - \`TestSearchEnterHandoff_NoMatchesSurfacesHint\`
  - \`TestPlaceCursorAtFirstItem_NoGroupsHasItems\`
- [ ] Manual smoke: run \`camp intent explore\`, move an intent from Someday to Done (verify clean render, in-progress hint, auto-commit succeeds with no TTY corruption). Repeat with \`/\` search → Enter → j/k navigation → Enter to open viewer.

### DIFF
--- a/internal/intent/tui/explorer/navigation.go
+++ b/internal/intent/tui/explorer/navigation.go
@@ -13,8 +13,24 @@ import (
 // selection rather than landing on a group header (cursorItem=-1). Without
 // this, j/k looked like a no-op and users could not tell the list was
 // navigable (regression #279).
+//
+// Special case: when the Dungeon parent is collapsed (m.dungeonExpanded ==
+// false), groupIntentsByStatus does not include the dungeon child groups in
+// m.groups — only a single Dungeon parent with DungeonCount and an empty
+// Intents slice. If the active filter has matches only in the dungeon, the
+// naive iteration would conclude "no matches reachable" and surface the
+// recovery hint even though dungeon results exist. Expand the parent and
+// rebuild groups so the children become visible, then place the cursor on
+// the first one.
 func (m *Model) placeCursorAtFirstItem() bool {
 	for gi := range m.groups {
+		// Auto-expand the Dungeon parent if it holds the only matches.
+		if m.groups[gi].IsDungeonParent && m.groups[gi].DungeonCount > 0 && !m.dungeonExpanded {
+			m.dungeonExpanded = true
+			m.groups = groupIntentsByStatus(m.filteredIntents, m.dungeonExpanded)
+			// Restart with the rebuilt slice — dungeon children are now present.
+			return m.placeCursorAtFirstItem()
+		}
 		if len(m.groups[gi].Intents) == 0 {
 			continue
 		}
--- a/internal/intent/tui/explorer/navigation.go
+++ b/internal/intent/tui/explorer/navigation.go
@@ -4,6 +4,36 @@ import (
 	"github.com/Obedience-Corp/camp/internal/intent/tui"
 )
 
+// placeCursorAtFirstItem positions the cursor on the first item of the first
+// group that has at least one intent, expanding the group if needed. It
+// returns true when an item was reachable and the cursor was placed there,
+// false when the visible groups are all empty (e.g. no search matches).
+//
+// Used after the user exits search mode so the filtered list has a visible
+// selection rather than landing on a group header (cursorItem=-1). Without
+// this, j/k looked like a no-op and users could not tell the list was
+// navigable (regression #279).
+func (m *Model) placeCursorAtFirstItem() bool {
+	for gi := range m.groups {
+		if len(m.groups[gi].Intents) == 0 {
+			continue
+		}
+		// Expand the group so the chosen item is actually visible. Without
+		// this a collapsed first-non-empty group would still hide the cursor.
+		m.groups[gi].Expanded = true
+		m.cursorGroup = gi
+		m.cursorItem = 0
+		m.scrollOffset = 0
+		m.ensureCursorVisible()
+		return true
+	}
+	// No matches reachable — leave cursor at the safe header position.
+	m.cursorGroup = 0
+	m.cursorItem = -1
+	m.scrollOffset = 0
+	return false
+}
+
 // moveCursorDown moves the cursor down one position and adjusts scroll.
 func (m *Model) moveCursorDown() {
 	m.moveCursorDownOne()
--- a/cmd/camp/intent/explore.go
+++ b/cmd/camp/intent/explore.go
@@ -1,6 +1,11 @@
 package intent
 
 import (
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -99,6 +104,18 @@ func runIntentExplore(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	// Route slog output away from stderr for the duration of the TUI session.
+	// Without this, INFO-level slog calls from anywhere in the call chain (notably
+	// the git auto-commit retry / lock subsystem) write to the same TTY that
+	// bubbletea is painting, corrupting the rendered frame and making the TUI
+	// appear frozen. Restore the previous default on exit so non-TUI commands in
+	// the same process keep their normal behavior.
+	restoreLogger, err := quietSlogDuringTUI(campaignRoot)
+	if err != nil {
+		return camperrors.Wrap(err, "configuring TUI logger")
+	}
+	defer restoreLogger()
+
 	// Create and run the TUI
 	model := explorer.NewModel(ctx, svc, conceptSvc, intentsDir, campaignRoot, cfg.ID, author, shortcuts)
 	p := tea.NewProgram(model, tea.WithAltScreen())
@@ -108,4 +125,45 @@ func runIntentExplore(cmd *cobra.Command, args []string) error {
 	}
 
 	return nil
+}
+
+// quietSlogDuringTUI swaps slog.Default with a handler that writes to a log
+// file under <campaignRoot>/.campaign/logs/ instead of stderr. It returns a
+// restore function that reinstalls the previous default and closes the log
+// file. If the log directory cannot be created or the file cannot be opened,
+// it falls back to discarding slog output entirely so the TUI is still safe.
+func quietSlogDuringTUI(campaignRoot string) (restore func(), err error) {
+	previous := slog.Default()
+	restore = func() { slog.SetDefault(previous) }
+
+	logDir := filepath.Join(campaignRoot, ".campaign", "logs")
+	if mkErr := os.MkdirAll(logDir, 0o755); mkErr != nil {
+		// Fallback: discard slog output entirely. The TUI is still protected.
+		slog.SetDefault(slog.New(slog.NewTextHandler(io.Discard, nil)))
+		return restore, nil
+	}
+
+	f, openErr := os.OpenFile(
+		filepath.Join(logDir, "intent-explore.log"),
+		os.O_APPEND|os.O_CREATE|os.O_WRONLY,
+		0o644,
+	)
+	if openErr != nil {
+		// Fallback: discard slog output entirely.
+		slog.SetDefault(slog.New(slog.NewTextHandler(io.Discard, nil)))
+		return restore, nil
+	}
+
+	previousRestore := restore
+	restore = func() {
+		previousRestore()
+		_ = f.Close()
+	}
+
+	// Debug level captures everything (including the auto-commit retry logs
+	// that were demoted from Info to Debug in this same change set) so a
+	// post-mortem can always recover what happened.
+	handler := slog.NewTextHandler(f, &slog.HandlerOptions{Level: slog.LevelDebug})
+	slog.SetDefault(slog.New(handler))
+	return restore, nil
 }
--- a/cmd/camp/intent/explore_test.go
+++ b/cmd/camp/intent/explore_test.go
@@ -1,0 +1,106 @@
+package intent
+
+import (
+	"bytes"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestQuietSlogDuringTUI_RedirectsToLogFile asserts that slog.Default no
+// longer points at stderr after quietSlogDuringTUI runs, and that messages
+// emitted while the swap is active land in the campaign log file instead of
+// the user-facing TTY.
+//
+// Regression test for bug 003: auto-commit INFO logs (now Debug) used to
+// collide with bubbletea rendering on the same TTY. Even after demoting the
+// known noise to Debug, this swap is the defense-in-depth that prevents any
+// future INFO emission from anywhere in the call chain from corrupting the
+// TUI render frame.
+func TestQuietSlogDuringTUI_RedirectsToLogFile(t *testing.T) {
+	campaignRoot := t.TempDir()
+	previousDefault := slog.Default()
+	t.Cleanup(func() { slog.SetDefault(previousDefault) })
+
+	// Capture stderr while the swap is active. If any slog call leaks to
+	// stderr, this buffer would not be empty (we cannot literally redirect
+	// os.Stderr here without affecting the test runner, so we assert on the
+	// handler identity instead: the swapped default must NOT be the previous
+	// default that was pointing at stderr).
+	stderrBuf := &bytes.Buffer{}
+	_ = stderrBuf // kept for parity; identity assertion below is the real check
+
+	restore, err := quietSlogDuringTUI(campaignRoot)
+	if err != nil {
+		t.Fatalf("quietSlogDuringTUI returned error: %v", err)
+	}
+	t.Cleanup(restore)
+
+	if slog.Default() == previousDefault {
+		t.Fatal("slog.Default was not swapped; TUI would still leak logs to stderr")
+	}
+
+	// Emit at every level. None should hit stderr.
+	slog.Debug("debug-from-test", "key", "value")
+	slog.Info("info-from-test", "key", "value")
+	slog.Warn("warn-from-test", "key", "value")
+	slog.Error("error-from-test", "key", "value")
+
+	// Restore must put the previous default back so non-TUI code paths are
+	// unaffected.
+	restore()
+	if slog.Default() != previousDefault {
+		t.Fatal("restore() did not reinstall the previous slog.Default")
+	}
+
+	// The log file should exist and contain the messages we emitted.
+	logPath := filepath.Join(campaignRoot, ".campaign", "logs", "intent-explore.log")
+	contents, err := os.ReadFile(logPath)
+	if err != nil {
+		t.Fatalf("expected log file at %s, got error: %v", logPath, err)
+	}
+	body := string(contents)
+
+	wantSubstrings := []string{
+		"debug-from-test",
+		"info-from-test",
+		"warn-from-test",
+		"error-from-test",
+	}
+	for _, want := range wantSubstrings {
+		if !bytes.Contains(contents, []byte(want)) {
+			t.Errorf("log file missing %q. body=%q", want, body)
+		}
+	}
+}
+
+// TestQuietSlogDuringTUI_FallsBackToDiscardOnDirError verifies that if the
+// log directory cannot be created (e.g. permission denied on a read-only
+// campaign), the function still installs a quiet handler rather than letting
+// slog continue to write to stderr.
+func TestQuietSlogDuringTUI_FallsBackToDiscardOnDirError(t *testing.T) {
+	// Create a regular file where .campaign would need to live; MkdirAll
+	// will fail because the parent path component is not a directory.
+	parent := t.TempDir()
+	blocker := filepath.Join(parent, ".campaign")
+	if err := os.WriteFile(blocker, []byte("not a directory"), 0o644); err != nil {
+		t.Fatalf("preparing blocker file: %v", err)
+	}
+
+	previousDefault := slog.Default()
+	t.Cleanup(func() { slog.SetDefault(previousDefault) })
+
+	restore, err := quietSlogDuringTUI(parent)
+	if err != nil {
+		t.Fatalf("quietSlogDuringTUI returned error on fallback path: %v", err)
+	}
+	t.Cleanup(restore)
+
+	if slog.Default() == previousDefault {
+		t.Fatal("slog.Default was not swapped on fallback path; TUI would still leak to stderr")
+	}
+
+	// Should not panic when emitting to the discard handler.
+	slog.Info("info-to-discard", "key", "value")
+}
--- a/internal/git/lock_unix.go
+++ b/internal/git/lock_unix.go
@@ -173,7 +173,7 @@ func removeSingleLock(ctx context.Context, lockPath string, logger *slog.Logger)
 	}
 
 	// Step 3: Remove the lock file
-	logger.Info("removing stale lock", "path", lockPath)
+	logger.Debug("removing stale lock", "path", lockPath)
 	if err := os.Remove(lockPath); err != nil {
 		return info, camperrors.WrapJoinf(ErrLockRemovalFailed, err, "failed to remove lock file %s", lockPath)
 	}
--- a/internal/git/retry.go
+++ b/internal/git/retry.go
@@ -119,7 +119,7 @@ func WithLockRetry(ctx context.Context, repoPath string, cfg RetryConfig, operat
 				cfg.OperationName, cycle)
 		}
 
-		cfg.Logger.Info("cycle completed, cleaning locks",
+		cfg.Logger.Debug("cycle completed, cleaning locks",
 			"operation", cfg.OperationName,
 			"cycle", cycle,
 			"removed", len(result.Removed),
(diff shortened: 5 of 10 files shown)


internal/intent/tui/explorer/navigation.go
```diff
@@ -4,6 +4,36 @@ import (
...
+// group that has at least one intent, expanding the group if needed. It
+// returns true when an item was reachable and the cursor was placed there,
+// false when the visible groups are all empty (e.g. no search matches).
+//
+// Used after the user exits search mode so the filtered list has a visible
+// selection rather than landing on a group header (cursorItem=-1). Without
+// this, j/k looked like a no-op and users could not tell the list was
+// navigable (regression #279).
+func (m *Model) placeCursorAtFirstItem() bool {
+	for gi := range m.groups {
```
obey-agent, line 26: Follow-up review: this still misses filtered matches that live under a collapsed Dungeon parent. When `dungeonExpanded` is false, `groupIntentsByStatus` returns the top groups plus the synthetic `Dungeon` parent; the parent carries `DungeonCount` but no `Intents`, and the Done/Killed/Archived/Someday child groups are not present. So if `/` matches only a Someday/Archived/etc. intent while Dungeon is collapsed, this helper returns false and `updateSearch` shows the “No matches” hint even though `filteredIntents` is non-empty. Please add a regression test for searching a dungeon intent with `dungeonExpanded=false`, and either expand/regroup Dungeon before scanning or teach the helper to treat a non-empty Dungeon parent as a reachable match.

obey-agent (commented):
Follow-up review

Verdict: Request Changes

Overview: This is a pragmatic TUI hardening pass, and the log routing plus in-progress feedback are well-scoped fixes for the reported freeze symptoms. The new code uses `camperrors` where it constructs command-level errors, and the changed git retry logs avoid adding new raw error handling. The remaining issue is in the search handoff edge case.

Key Findings:
- Inline: `placeCursorAtFirstItem` reports no matches for filtered dungeon intents when the Dungeon parent is collapsed, because the matching child groups are not present and the parent has only `DungeonCount`, not `Intents`.

What's Done Well:
- Routing slog away from the alt-screen TTY is the right class of fix for Bubble Tea render corruption.
- The move/archive progress message gives users immediate feedback while async auto-commit work runs.
- The search handoff tests cover normal matches and empty results; they just need the collapsed-dungeon case added.

Staff Standard: Close, but not merge-ready. I would want the dungeon-search regression fixed because this PR explicitly targets search results feeling unselectable, and dungeon intents are part of the explorer's core triage surface.

Verification: `go test ./cmd/camp/intent ./internal/intent/tui/explorer ./internal/git` passed in the PR worktree.

lancekrogers:
Thanks for catching this. Verified the claim against `groupIntentsByStatus` (model.go:340-342) — when `dungeonExpanded=false` the dungeon child groups are excluded from the returned slice, and the Dungeon parent carries only `DungeonCount` with an empty `Intents` slice. So a search whose only matches are in dungeon would hit "No matches" even though `len(filteredIntents) > 0`.

Pushed `f3ff6f2`:

- `placeCursorAtFirstItem` now detects an `IsDungeonParent` with `DungeonCount > 0` while `m.dungeonExpanded` is false, flips `m.dungeonExpanded`, rebuilds `m.groups` from `m.filteredIntents`, and restarts the iteration. The child groups are then present and the helper falls through to its normal first-item placement.
- New regression test `TestSearchEnterHandoff_DungeonOnlyMatchesAutoExpand` asserts the baseline matches the bug condition (parent present with count, child absent), then verifies the auto-expand happens, the cursor lands on the matching Someday intent in a dungeon child group, and the "No matches" hint is not surfaced.

Full test suite still green. Ready for a re-review.

obey-agent (commented):
**Follow-up Review**

**Verdict:** Approve

**Overview:** I re-reviewed the current head (`f3ff6f2`) after the Dungeon-only search handoff follow-up. The latest commit addresses the previous concern: when a search matches only collapsed Dungeon children, Enter now expands the Dungeon parent, rebuilds groups, and lands the cursor on the matching child instead of showing a false no-matches hint.

**Key Findings:**
- No remaining blocking findings.
- I specifically checked the changed Go paths for raw `fmt.Errorf` / `errors.New` usage. This PR does not introduce raw error construction in changed production code, and the command-level error path continues to use `camperrors`.
- Validation run: `go test ./cmd/camp/intent ./internal/intent/tui/explorer ./internal/git` passed, and `git diff --check origin/main...HEAD` passed. A full `go test ./...` still fails in `cmd/camp/leverage`, but the same failures reproduce on `origin/main` in this environment and this PR does not touch that package.

**What's Done Well:** The fix is integrated into the existing search/list state machine rather than creating a parallel navigation path, and the regression coverage now exercises the normal filtered handoff, empty-results hint, collapsed Dungeon-only match, and dungeon move progress states. The slog routing is also scoped to the TUI lifetime and restores the previous default logger afterward, which is the right operational boundary for avoiding Bubble Tea frame corruption.

**Staff Standard:** Yes. With the follow-up Dungeon search edge case covered, I would be comfortable submitting this as my own work.

PR #283, opened 2026-05-13
fix(leverage/tests): scaffold ephemeral campaign for config validation tests

lancekrogers:
## Summary

Three tests in \`cmd/camp/leverage/main_command_test.go\` — \`TestLeverageConfigCommand_ValidationPeople\`, \`_ValidationDate\`, and \`_ValidationCOCOMO\` — passed locally but failed in any environment where the camp repo lives outside a parent campaign workspace (fresh clones, CI, review sandboxes). The tests invoke the leverage command's cobra RunE, which calls \`campaign.DetectCached\` and walks up the filesystem looking for a \`.campaign\` directory. When run from inside obey-campaign (or any other campaign-rooted workspace) the walk finds one and the command proceeds to the validation logic the tests assert against. When run from a path with no campaign root above it, the command bails early with:

\`\`\`
not in a campaign: not inside a campaign directory
Hint: Run 'camp init' to create a campaign, or navigate to an existing one
\`\`\`

…and the tests fail with the wrong error message.

## Background

Surfaced as a follow-up to the obey-agent review on #282, which noted: *\"A full \`go test ./...\` still fails in \`cmd/camp/leverage\`, but the same failures reproduce on origin/main in this environment and this PR does not touch that package.\"*

Reproduced by copying the camp repo to a tmpdir outside any parent campaign workspace and running \`go test ./cmd/camp/leverage/\`. The three tests above fail; everything else in cmd/camp/leverage passes.

## Fix

Each validation test now calls a new \`setupEphemeralCampaign(t)\` helper that:

1. \`t.TempDir()\` for an isolated root.
2. \`os.MkdirAll(tmpDir/.campaign, 0o755)\` so the walk-upward campaign detection succeeds.
3. \`t.Chdir(tmpDir)\` so the test process's cwd is inside the ephemeral campaign for the duration of the test.

With this scaffolding, the cobra command proceeds past campaign detection, validation logic runs as intended, and the tests assert against the actual validation error rather than the campaign-detection error.

The helper is local to \`main_command_test.go\` so it does not silently change behavior of other tests in the package that might intentionally exercise the no-campaign path. Tests that already scaffold their own campaign (such as \`TestLeverageConfigCommand_SaveAndReload\`) are unaffected.

## Test plan

- [x] Reproduction: copy camp to \`/tmp/camp-iso\`, run \`go test ./cmd/camp/leverage/\` — three tests fail with the campaign-detection error.
- [x] After fix: same isolated run passes (\`ok ... 0.668s\`).
- [x] After fix: in-workspace run still passes (\`ok ... 0.279s\`).
- [ ] obey-agent re-review on this PR to confirm the failure no longer reproduces in their review environment.

### DIFF
--- a/cmd/camp/leverage/main_command_test.go
+++ b/cmd/camp/leverage/main_command_test.go
@@ -81,6 +81,24 @@ func executeLeverage(t *testing.T, args ...string) (string, error) {
 	return buf.String(), err
 }
 
+// setupEphemeralCampaign creates a temporary campaign root and chdirs the
+// test process into it. Returns the campaign root path. Used by tests that
+// invoke the leverage command's RunE — which walks up from cwd looking for a
+// .campaign directory and refuses to run if none is found. Without this
+// scaffolding the tests would pass on a developer's machine (the test
+// process inherits an ambient campaign root from the user's workspace) but
+// fail in a fresh clone where camp lives outside any campaign tree (CI,
+// review sandboxes like obey-agent, contributor first-runs).
+func setupEphemeralCampaign(t *testing.T) string {
+	t.Helper()
+	tmpDir := t.TempDir()
+	if err := os.MkdirAll(filepath.Join(tmpDir, ".campaign"), 0o755); err != nil {
+		t.Fatalf("create .campaign: %v", err)
+	}
+	t.Chdir(tmpDir)
+	return tmpDir
+}
+
 func stubPopulateMetrics() func(ctx context.Context, campaignRoot string, resolved []intleverage.ResolvedProject, resolver *intleverage.AuthorResolver) {
 	return func(ctx context.Context, _ string, resolved []intleverage.ResolvedProject, _ *intleverage.AuthorResolver) {
 		for i := range resolved {
@@ -105,6 +123,7 @@ func stubPopulateMetrics() func(ctx context.Context, campaignRoot string, resolv
 // host filesystem mutation is impossible.
 
 func TestLeverageConfigCommand_ValidationPeople(t *testing.T) {
+	setupEphemeralCampaign(t)
 	_, err := executeLeverage(t, "config", "--people", "-1")
 	if err == nil {
 		t.Fatal("expected error for negative people, got nil")
@@ -115,6 +134,7 @@ func TestLeverageConfigCommand_ValidationPeople(t *testing.T) {
 }
 
 func TestLeverageConfigCommand_ValidationDate(t *testing.T) {
+	setupEphemeralCampaign(t)
 	_, err := executeLeverage(t, "config", "--start", "2025-13-45")
 	if err == nil {
 		t.Fatal("expected error, got nil")
@@ -282,6 +302,7 @@ func TestLeverageOutputTable_ShowsBackfillHintWhenRecentHistoryMissing(t *testin
 }
 
 func TestLeverageConfigCommand_ValidationCOCOMO(t *testing.T) {
+	setupEphemeralCampaign(t)
 	_, err := executeLeverage(t, "config", "--cocomo-type", "invalid")
 	if err == nil {
 		t.Fatal("expected error, got nil")
--- a/cmd/camp/leverage/main_command_test.go
+++ b/cmd/camp/leverage/main_command_test.go
@@ -104,24 +104,7 @@ func stubPopulateMetrics() func(ctx context.Context, campaignRoot string, resolv
 // command runs against an ephemeral campaign inside the shared container and
 // host filesystem mutation is impossible.
 
-// NOTE: TestLeverageConfigCommand_ValidationPeople, _ValidationDate, and
-// _ValidationCOCOMO were unit tests that invoked the live cobra RunE,
-// which calls campaign.DetectCached and walks up the filesystem looking
-// for a .campaign directory. They passed on a developer's machine because
-// the test process inherits an ambient campaign root from the surrounding
-// workspace (camp lives under a campaign as a submodule), but failed in
-// any environment where camp lives outside a campaign tree — fresh clones,
-// CI, and review sandboxes such as obey-agent's review environment.
-//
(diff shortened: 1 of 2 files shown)


cmd/camp/leverage/main_command_test.go
```diff
@@ -81,6 +81,24 @@ func executeLeverage(t *testing.T, args ...string) (string, error) {
...
+// test process into it. Returns the campaign root path. Used by tests that
+// invoke the leverage command's RunE — which walks up from cwd looking for a
+// .campaign directory and refuses to run if none is found. Without this
+// scaffolding the tests would pass on a developer's machine (the test
+// process inherits an ambient campaign root from the user's workspace) but
+// fail in a fresh clone where camp lives outside any campaign tree (CI,
+// review sandboxes like obey-agent, contributor first-runs).
+func setupEphemeralCampaign(t *testing.T) string {
+	t.Helper()
+	tmpDir := t.TempDir()
```
obey-agent: This fixes the isolated failure, but it does it by adding a new host-filesystem fixture to the unit test suite: `t.TempDir()` + `os.MkdirAll(.../.campaign)` + `t.Chdir`. That cuts against the campaign filesystem-safety rule, and it is especially risky in this file because the comment immediately below explains that leverage command tests invoking live `RunE` were moved to container integration after they mutated the host campaign. Please either move these validation cases to the existing containerized integration harness, or refactor the validation into a pure helper that can be exercised without creating a campaign directory/changing cwd. The important behavior is worth covering, but new command-level filesystem setup should not live in a host unit test.

obey-agent (commented):
**Follow-up Review**

**Verdict:** Request Changes

**Overview:** This PR correctly identifies and reproduces the `cmd/camp/leverage` failure we saw during the camp#282 follow-up review. The fix makes `go test ./...` pass from an isolated `/tmp` worktree, but it does so by adding a new host-filesystem/cwd fixture to a unit test file that already documents why live leverage command tests were moved to container integration.

**Key Findings:**
- Blocking inline comment on `cmd/camp/leverage/main_command_test.go`: `setupEphemeralCampaign` uses `t.TempDir()` + `os.MkdirAll(.../.campaign)` + `t.Chdir` in unit tests. That conflicts with the campaign filesystem-safety rule for new filesystem-mutating command fixtures. Please either move these cases to the existing containerized integration harness or refactor the invalid config validation into a pure helper that does not need a campaign directory/cwd mutation.

**Validation Run:**
- `go test ./cmd/camp/leverage` passed from `/tmp/obey-review-camp-283`.
- `go test ./cmd/camp/leverage -run 'TestLeverageConfigCommand_Validation(People|Date|COCOMO)$' -count=1 -v` passed.
- `go test ./...` passed from `/tmp/obey-review-camp-283`.

**What's Done Well:** The failure analysis is accurate, the PR is tightly scoped, and the new helper is local rather than changing unrelated tests. The test behavior proves the full-suite regression is fixed in a fresh-clone-style environment.

**Staff Standard:** Not yet. I would be proud of the diagnosis and scope, but I would not submit this implementation as-is because it adds exactly the kind of host filesystem command fixture the leverage test suite was trying to avoid.

lancekrogers:
Updated: reverted the host-side scaffolding and moved the three validation tests to the integration harness instead.

The previous commit on this branch used \`t.TempDir + .campaign/ + t.Chdir\` which violates the rule in CLAUDE.md: *"Tests that mutate the filesystem must run in containerized isolation, not directly on the host via ad hoc temp directories."* Even though the chdir scope was process-local, it's still a host-fs side effect that can interact badly with concurrent test runs and developer workspaces.

`4649138` reverts the helper + the three unit tests and adds equivalents in `tests/integration/leverage_test.go` that run against an ephemeral campaign inside the shared docker container. The file already documented the same migration pattern for other tests that lived inside `cmd/camp/leverage/` previously; these three now follow it.

A small `setupMinimalLeverageCampaign` helper handles the campaign-only case (no projects, no git history) so the suite does not pay the multi-second `setupLeverageCampaign` cost when only campaign detection needs to succeed.

Verified locally:
- \`go test ./cmd/camp/leverage/\` in an isolated /tmp clone: passes (the three deleted tests had been the only failures there).
- \`go test -tags integration -run TestLeverage_ConfigValidation ./tests/integration/\`: 3/3 pass against the docker container.

Ready for re-review.

lancekrogers:
Thanks for the review. The blocking finding is on commit \`5b5aa37\` (the host-side scaffolding) — but that commit was reverted by the next commit on this same branch. Current HEAD (\`4649138\`) already does exactly what you asked: the \`setupEphemeralCampaign\` helper and the three unit tests are gone, and equivalents now live in \`tests/integration/leverage_test.go\` against the shared container.

\`\`\`
$ git log origin/fix-leverage-test-campaign-detection --oneline -2
4649138 fix(leverage/tests): move config validation tests to integration harness
5b5aa37 fix(leverage/tests): scaffold ephemeral campaign for config validation tests
\`\`\`

Validation against current HEAD:
- \`cmd/camp/leverage/main_command_test.go\` no longer contains \`setupEphemeralCampaign\`, \`TestLeverageConfigCommand_ValidationPeople\`, \`_ValidationDate\`, or \`_ValidationCOCOMO\` — replaced by a NOTE block matching the file's existing migration-history pattern.
- \`tests/integration/leverage_test.go\` adds \`TestLeverage_ConfigValidationPeople\`, \`_ValidationDate\`, \`_ValidationCOCOMO\` with a small \`setupMinimalLeverageCampaign\` helper (campaign-only, no projects/git, since the validation rejects bad flags before the leverage pipeline runs).
- \`go test -tags integration -run 'TestLeverage_ConfigValidation' ./tests/integration/\`: 3/3 PASS against the docker container.
- \`go test ./cmd/camp/leverage/\` in an isolated \`/tmp\` clone: passes (the deleted tests had been the only host-fs failures there).

Could you re-review against \`4649138\`? Per \`feedback_verify_pr_head_before_posting.md\`, head-SHA verification right before posting is the convention here so reviews don't land on stale commits.

PR #286, opened 2026-05-15
[INTEGRATION] WW0001 workitem-workflows

lancekrogers:
## Summary

**Integration umbrella PR for WW0001 (workitem-workflows festival).** Aggregates the camp side of the V2 design correction so it can be tested end-to-end before merging to `main`.

## Stacked PRs landing here

- **#276** `lr/WW0001/003-metadata-discovery` — `.workitem` metadata loader trimmed to v1alpha5 minimum (D006/D024); custom-type discovery via `.workitem` marker (D025); `WorkflowType` opens to free-form string (D028); `LoadMetadataFS` for fs.FS-driven tests (D029); structured `slog.Debug` discovery skip events.
- **#281** `lr/WW0001/005-dashboard-visibility` — Dashboard surfaces `.workflow/` runtime progress via `WorkflowMeta`; `localrun` ctx-aware replay with error propagation; backward-compat for `v1alpha4`; named event-type constants documenting the cross-repo contract with fest's localstore.
- **#285** `feat/workitem-create-and-adopt` — `camp workitem create <slug>` and `camp workitem adopt <dir>` user-facing commands (D008, D012); slug + path-safety validation; ID generation with collision retry; container integration tests.

## Why a draft integration branch

Three stacked PRs from the WW0001 festival need to be merged together for the dashboard to render workflow progress correctly. Going straight to `main` means each PR is reviewed against an intermediate state. This branch lets us:

1. Land the three PRs onto `integration/WW0001` for end-to-end validation.
2. Run the full container integration suite on the rolled-up diff.
3. Open one final, focused review here before merging to `main`.

## Test plan

- [ ] All three stacked PRs reviewed and approved on this branch.
- [ ] `just build-camp` clean on integration head.
- [ ] `go test ./...` passes on integration head.
- [ ] `just test integration` (container harness) passes on integration head.
- [ ] Manual smoke in a real campaign: create + adopt, dashboard renders WORKFLOW progress, custom types gated by marker.
- [ ] Mark this PR Ready for Review after the three stacked PRs merge.


internal/commands/workitem/create.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	id, err := generateID(ctx, typeFlag, slug, idOverride, campaignRoot)
+	if err != nil {
+		return err
+	}
+
+	parent := dirOverride
+	if parent == "" {
+		parent = filepath.Join("workflow", typeFlag)
```
obey-agent, line 78: This default writes new workitems under `workflow/<type>`, with `--type` defaulting to `feature`, but the current discovery path still only calls `discoverDesign` and `discoverExplore`; `emitCandidateFS` is never wired into `Discover`. The happy-path create/adopt tests only assert the marker is written, so `camp workitem create demo` succeeds and then `camp workitem --json` will not show it. Please wire marker-gated discovery for `workflow/<type>/<slug>` (or choose a discovered default) and add an integration assertion that a created/adopted workitem appears in `camp workitem`.

internal/commands/workitem/create.go
```diff
@@ -0,0 +1,189 @@
...
+		return camperrors.NewValidation("dir", "parent dir must be relative to campaign root", nil)
+	}
+	if strings.HasPrefix(clean, "..") {
+		return camperrors.NewValidation("dir", "parent dir must not escape campaign root", nil)
+	}
+	return nil
+}
+
+func generateID(ctx context.Context, typeStr, slug, override, campaignRoot string) (string, error) {
+	if override != "" {
```
obey-agent, line 138: `--id` overrides bypass collision detection entirely. Stable IDs are the durable identity for workitems, so allowing two `.workitem` files to carry the same id defeats the collision retry guarantee and gives agents/dashboard consumers ambiguous identity. Please run the same collision check for explicit overrides and reject duplicates before writing the marker.

internal/commands/workitem/create.go
```diff
@@ -0,0 +1,189 @@
...
+	return "", camperrors.NewValidation("id", "could not generate a unique id after 32 attempts", nil)
+}
+
+func idCollides(ctx context.Context, campaignRoot, id string) bool {
+	if campaignRoot == "" {
+		return false
+	}
+	root := filepath.Join(campaignRoot, "workflow")
+	collision := false
+	_ = filepath.WalkDir(root, func(path string, d os.DirEntry, err error) error {
```
obey-agent: `idCollides` discards the `WalkDir` return value and all read/parse errors. A canceled scan, permission issue, or unreadable `.workitem` is currently treated as “no collision”, which can generate a duplicate ID after an incomplete safety scan. Please return `(bool, error)`, propagate `ctx.Err()` / `WalkDir` / read failures, and only ignore malformed metadata if that is an explicit legacy-compatibility fallback with test coverage.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** This integration PR is moving the right pieces into place for WW0001: typed `.workitem` metadata, runtime progress visibility, and create/adopt commands. The blocker is that the new create/adopt user path is not actually integrated into discovery, so the happy path can create valid-looking workitems that `camp workitem` never returns.

**Key Findings:**
- Created workitems default to `workflow/feature/...`, but `Discover` still only scans intents, `workflow/design`, `workflow/explore`, and festivals. `emitCandidateFS` is not wired into production discovery, so create/adopt success does not translate into dashboard visibility. See inline comment on `internal/commands/workitem/create.go`.
- Explicit `--id` overrides bypass collision detection, allowing duplicate stable IDs.
- `idCollides` swallows `WalkDir`, context, and read/parse failures, so an incomplete safety scan is treated as no collision.

**What's Done Well:** The metadata loader is small, uses the project error package, has fs.FS-driven parser tests, and the local runtime reader has good coverage for replay semantics and drift detection.

**Staff Standard:** Not yet. I would be comfortable with the direction, but I would not merge this until created/adopted workitems are discoverable and ID uniqueness failures are handled as hard errors.

obey-agent:
**Re-review @ b235abe721cf8714585b943e359073415f9b05c4** — all 3 blockers from the previous Request Changes are addressed.

**1. Discovery wiring** — `internal/workitem/discover_custom_workflows.go` adds marker-gated discovery for any non-builtin `workflow/<type>/<slug>` directory. Wired into `Discover` between explore and festivals. Refactored `discoverWorkflowDocs` to share `buildWorkflowDirItem` so builtin and custom paths use identical metadata + local-runtime merge semantics.
- Unit coverage in `discover_custom_workflows_test.go`: marker-required gating, builtin/dungeon skipping, end-to-end `Discover()` wiring.
- Integration coverage in `tests/integration/workitem_create_test.go`: `CreatedAndAdoptedAppearInDashboard` asserts both `workflow/feature/demo-feature` (created) and `workflow/incident/p99-spike` (adopted) appear in `camp workitem --json` with the right `workflow_type`.

**2. Explicit `--id` collision check** — `generateID` now runs `idCollides` on overrides and returns a validation error before writing the marker. Integration test `CreateRejectsDuplicateExplicitID` covers the regression.

**3. `idCollides` error propagation** — signature is now `(bool, error)`. `WalkDir`, `ReadFile`, and `ctx.Err()` failures are surfaced rather than swallowed. The non-existent `workflow/` root case is handled cleanly with `filepath.SkipAll`. Malformed YAML is intentionally tolerated (legacy hand-edited markers); other read failures return as errors.

`just test all` clean: 2357 unit + 240 integration green.

internal/commands/workitem/create.go
```diff
@@ -0,0 +1,212 @@
...
+	cmd := &cobra.Command{
+		Use:   "create <slug>",
+		Short: "Create a new workitem with v1 minimum metadata",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			return runCreate(ctx, cmd, args[0], typeFlag, title, idOverride, dirOverride)
+		},
+	}
+	cmd.Flags().StringVar(&typeFlag, "type", "feature", "workitem type (feature, bug, chore, or custom)")
```
obey-agent, line 51: This flag creates free-form workitem types (`feature`, `incident`, etc.), but the dashboard filter still rejects those same values before discovery. In a dev build, `camp workitem --json --type feature` exits with `unknown --type value: "feature"`, even though `camp workitem create` defaults to `--type feature` and discovery now emits `workflow_type: "feature"`. Please update the `camp workitem --type` validation/help to accept arbitrary slug-safe workflow types (or at least every type create/adopt can produce), and add a regression test for filtering a created custom type.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The prior integration blockers around discovery wiring, explicit ID collisions, and scan error propagation are addressed. One CLI contract issue remains: the new free-form workitem types are created and discovered, but the dashboard filter still rejects them.

**Key Findings:**
- `camp workitem create` defaults to `--type feature`, and discovery emits `workflow_type: "feature"`, but `camp workitem --json --type feature` fails validation before discovery. See inline comment on `internal/commands/workitem/create.go`.

**What's Done Well:** The create/adopt commands now surface in discovery, explicit IDs are collision-checked, and the focused unit plus integration coverage around local runtime and custom type discovery is solid.

**Staff Standard:** Close, but not merge-ready until the user-facing filter accepts the same type contract the new commands produce.

obey-agent:
**Re-review @ d94867d3d7ed1737199977da4d657ea9a6e09951** — the dashboard `--type` filter contract issue is addressed.

**Filter accepts custom workflow types** — `validateFlags` now accepts any slug-safe type via the same `validateSlug` rule that gates `camp workitem create --type`. Help text on the `--type` flag updated to reflect that builtin (intent/design/explore/festival) and custom (feature, bug, incident, ...) types are both acceptable, naming `camp workitem create --type <name>` as the source of truth.

- Unit: `TestValidateFlagsAcceptsBuiltinAndCustomTypes` covers intent/design/explore/festival + feature/bug/incident/rfc-001/deep_dive. `TestValidateFlagsRejectsInvalidTypeSlugs` keeps the rejection path for uppercase, whitespace, slash, leading-hyphen, and empty.
- Integration: `DashboardFilterAcceptsCustomType` runs `camp workitem --json --type=feature` against the campaign that already has `workflow/feature/demo-feature` and `workflow/incident/p99-spike`, asserts the feature surfaces and the incident is excluded.

`just test all` clean: 2357 unit + 240 integration green.

obey-agent (commented):
Re-review @ `d94867d3d7ed1737199977da4d657ea9a6e09951`

Verdict: Approve

Findings: none.

The previous blocker is addressed in the new head. `camp workitem --json --type <custom>` now accepts arbitrary slug-safe workflow types instead of limiting the dashboard filter to the built-in workflow types, while still rejecting invalid slugs. The added unit and integration coverage pins the custom-type dashboard path.

Validation run:
- `go test -tags dev ./internal/commands/workitem ./internal/workitem/...`
- `just test integration-run TestIntegration_WorkitemCreateAndAdopt`

I did not find new issues in this focused re-review.

internal/commands/workitem/create.go
```diff
@@ -0,0 +1,226 @@
...
+//   - empty
+//   - contains '/' (would split into multiple path segments)
+//   - starts with '.' (reserved for hidden / "." / ".." traversal)
+//   - starts with '-' (would parse as a CLI flag in downstream tools)
+//   - contains whitespace, NUL, or ASCII control characters
+//   - longer than 80 chars (cross-fs name-length headroom)
+//
+// Anything else — including uppercase, dots, unicode letters — is accepted
+// because the project does not own its users' naming conventions.
+var slugPattern = regexp.MustCompile(`^[^\s/.\-\x00-\x1f][^\s/\x00-\x1f]{0,79}$`)
```
obey-agent: This no longer enforces a single safe path segment on Windows: `foo\\bar` passes this regex, but `filepath.Join(campaignRoot, parent, slug)` treats `\` as a separator on Windows and creates/navigates nested path segments. Since the new contract is explicitly path-safety rather than style, please reject backslash as well (and add it to `TestValidateSlug`) so create/adopt semantics are portable across camp's supported platforms.

internal/commands/workitem/workitem.go
```diff
@@ -208,10 +211,9 @@ func isInteractive() bool {
 }
 
 func validateFlags(jsonMode, printMode bool, pathOutput string, types, stages []string) error {
-	validTypes := map[string]bool{"intent": true, "design": true, "explore": true, "festival": true}
 	for _, t := range types {
-		if !validTypes[t] {
-			return fmt.Errorf("unknown --type value: %q (valid: intent, design, explore, festival)", t)
+		if err := validateSlug(t); err != nil {
+			return fmt.Errorf("invalid --type value %q: must be a slug-safe workflow type (lowercase a-z0-9, '-', '_'; max 80 chars)", t)
```
obey-agent: This error text still describes the old lowercase-only contract, but the new validator intentionally accepts `PascalCase`, `camelCase`, dotted names like `v1.2`, and other path-safe strings. Please align this message with `validateSlug`'s actual contract so users are not told that accepted values are invalid style.

internal/commands/workitem/create.go
```diff
@@ -0,0 +1,226 @@
...
+		return camperrors.NewValidation("dir", "parent dir must not escape campaign root", nil)
+	}
+	return nil
+}
+
+func generateID(ctx context.Context, typeStr, slug, override, campaignRoot string) (string, error) {
+	if override != "" {
+		if err := validateSlug(override); err != nil {
+			return "", camperrors.NewValidation("id",
+				"invalid id override "+override+": ids must be slug-safe (lowercase a-z0-9_-)", nil)
```
obey-agent: Same stale-contract issue here: `--id Foo` now passes `validateSlug`, but a rejected override still says IDs must be `lowercase a-z0-9_-`. Either keep IDs intentionally stricter than slugs with a separate validator, or update this message to the new path-safe slug contract.

obey-agent (commented):
**Verdict:** Request Changes

**What this PR is actually about:** PR #286 is an integration umbrella for WW0001, not a single isolated feature branch. It rolls the camp-side workitem workflow stack into `integration/WW0001`: `.workitem` v1 minimum metadata parsing/discovery, `.workflow/` local runtime progress surfaced through `camp workitem --json` and the TUI preview, and new `camp workitem create` / `camp workitem adopt` commands that write marker metadata under `workflow/<type>/<slug>`. The newest commit specifically changes the public slug contract from lowercase/kebab-only to broader path-safe names, and moves custom workflow discovery coverage from host-temp unit tests into the container-backed `TestIntegration_WorkitemCreateAndAdopt` suite.

**Key Findings:**
- Inline: the new relaxed `slugPattern` still allows `\`, which is a path separator on Windows. That violates the stated “path-safety check” contract because `camp workitem create foo\\bar` can become nested path segments on supported platforms.
- Inline: `camp workitem --type <bad>` still reports the old lowercase-only contract even though the validator now accepts PascalCase/camelCase/dotted names.
- Inline: invalid `--id` override text has the same stale lowercase-only wording, while the actual validator now accepts the broader slug set. If stable IDs are meant to stay stricter than directory slugs, they need a separate validator; otherwise the text should match the new contract.

**What’s Done Well:** The integration branch now clearly wires created/adopted custom workitems into dashboard discovery, preserves marker-gating for custom types, skips dungeoned custom dirs, rejects duplicate explicit IDs, and keeps malformed optional metadata/runtime data from crashing discovery. The newest commit also correctly moves filesystem-mutating discovery coverage into the container harness per repo testing standards.

**Staff Standard:** I would not merge this exact head because the slug validator is now documented as path-safe but misses a platform separator, and the user-facing validation text still contradicts the new public contract. The underlying architecture and integration direction look sound once those details are corrected.

**Checks Run:** `gofmt -l` on changed Go files was clean; `git diff --check origin/main...HEAD` passed; `go test -count=1 ./internal/commands/workitem ./internal/workitem/...` passed; `just build-camp` passed; `go test -count=1 -v -tags integration -run 'TestIntegration_WorkitemCreateAndAdopt' ./tests/integration` passed; `just test unit` passed (2360/2360). `just lint` still fails on existing repository-wide lint debt outside this PR's changed files.

obey-agent:
**Re-review @ d7f34be9df8ef8b252298d9ffac979cdfc57ff52** — all 3 nits addressed.

**1. `\` rejection** — `slugPattern` now rejects `\` alongside `/`, closing the Windows-path-separator gap. `TestValidateSlug` adds `foo\bar` and `\backslash` to the invalid set. Doc comment updated to call out both separators.

**2. `--type` validateFlags error text** — replaced the stale "lowercase a-z0-9, '-', '_'" wording with the actual path-safe contract: *"no '/', '\\', whitespace, or control chars; no leading '.' or '-'; max 80 chars"*. Matches what `validateSlug` actually enforces.

**3. `--id` override error text** — same stale-contract wording removed. The message now states that IDs follow the same path-safe slug contract as workitem names (which is the intentional design: IDs and slugs share the same character set; no separate stricter validator).

`just test all` clean: 2360 unit + 250 integration green.

PR #289, opened 2026-05-15
Add camp stage and camp project stage commands

lancekrogers:
## Summary

- Adds `camp stage` — runs the same auto-staging as `camp commit` (ref-sync exclusion at campaign root, stale lock retry) but stops before the commit step
- Adds `camp project stage` (aliased `camp p stage`) — same auto-staging as `camp p commit` for project submodules
- Regenerated CLI reference docs and updated README

## Why

Users sometimes want camp's correct staging scope and lock handling but a different commit strategy: interactive `git commit --patch`, a GUI client, signed commits, amend workflows, etc. Exposing stage as a standalone command unblocks that without touching existing commit behavior.

No changes to existing commands or to the staging primitives in `internal/git/`. The new commands reuse `git.StageAll`, `git.StageAllExcluding`, and `git.ListSubmodulePaths` directly.

## Test plan

- [x] \`just build-camp\` succeeds
- [x] \`just test unit\` passes (2320/2320)
- [x] \`just lint\` clean on new files
- [x] Integration suite \`TestStage*\` passes (7/7) in containerized harness covering:
  - root stage of a plain file
  - nothing-to-stage on a clean tree
  - submodule ref exclusion at campaign root (default)
  - \`--include-refs\` stages the submodule ref
  - project stage from inside the project dir
  - project stage with \`--project\` from campaign root
  - stale \`.git/index.lock\` recovery
- [x] Manual smoke: \`camp stage\`, \`camp stage --include-refs\`, \`camp p stage\` in a scratch campaign with a real submodule

## Out of scope

- No shared helper extracted between commit and stage handlers. The duplication is ~15 lines per command and the two paths may legitimately diverge later (e.g. project stage could grow its own sync semantics). Revisit if a third caller appears.

### DIFF
--- a/cmd/camp/project/stage.go
+++ b/cmd/camp/project/stage.go
@@ -87,18 +87,17 @@ func runProjectStage(cmd *cobra.Command, args []string) error {
 		return camperrors.Wrap(err, "failed to stage")
 	}
 
-	cmdutil.ShowStagedSummary(ctx, resolvedPath)
-
-	hasChanges, err := executor.HasChanges(ctx)
+	staged, err := git.HasStagedChanges(ctx, resolvedPath)
 	if err != nil {
 		return err
 	}
-	if !hasChanges {
+	if !staged {
 		fmt.Println(ui.Success("Nothing to stage in project"))
 		return nil
 	}
 
+	cmdutil.ShowStagedSummary(ctx, resolvedPath)
 	fmt.Println(ui.Success("✓ Project changes staged"))
-	fmt.Println(ui.Dim("Run 'git commit' inside the project or 'camp p commit --amend' to record them."))
+	fmt.Println(ui.Dim("Run 'camp p commit' to record them."))
 	return nil
 }
--- a/cmd/camp/project/stage.go
+++ b/cmd/camp/project/stage.go
@@ -98,6 +98,10 @@ func runProjectStage(cmd *cobra.Command, args []string) error {
 
 	cmdutil.ShowStagedSummary(ctx, resolvedPath)
 	fmt.Println(ui.Success("✓ Project changes staged"))
-	fmt.Println(ui.Dim("Run 'camp p commit' to record them."))
+	hint := "camp p commit"
+	if projectStageProject != "" {
+		hint = "camp p commit --project " + projectStageProject
+	}
+	fmt.Println(ui.Dim(fmt.Sprintf("Run '%s' to record them.", hint)))
 	return nil
 }
--- a/cmd/camp/project/stage.go
+++ b/cmd/camp/project/stage.go
@@ -1,0 +1,104 @@
+package project
+
+import (
+	"errors"
+	"fmt"
+	"path/filepath"
+
+	"github.com/Obedience-Corp/camp/cmd/camp/cmdutil"
+	"github.com/Obedience-Corp/camp/internal/campaign"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/git"
+	projectsvc "github.com/Obedience-Corp/camp/internal/project"
+	"github.com/Obedience-Corp/camp/internal/ui"
+	"github.com/spf13/cobra"
+)
+
+var projectStageCmd = &cobra.Command{
+	Use:   "stage",
+	Short: "Stage changes in a project submodule",
+	Long: `Stage changes within a project submodule without committing.
+
+Runs the same auto-staging logic as 'camp project commit' (including
+stale lock file cleanup) but stops before creating a commit, so you can
+use a different commit strategy.
+
+Auto-detects the current project from your working directory,
+or use --project to specify a project by name.
+
+Examples:
+  # From within a project directory
+  cd projects/my-api
+  camp project stage
+
+  # Specify project by name
+  camp project stage --project my-api`,
+	RunE: runProjectStage,
+}
+
+var projectStageProject string
+
+func init() {
+	projectStageCmd.Flags().StringVarP(&projectStageProject, "project", "p", "", "Project name (auto-detected from cwd if not specified)")
+
+	if err := projectStageCmd.RegisterFlagCompletionFunc("project", cmdutil.CompleteProjectName); err != nil {
+		panic(err)
+	}
+
+	Cmd.AddCommand(projectStageCmd)
+}
+
+func runProjectStage(cmd *cobra.Command, args []string) error {
+	ctx := cmd.Context()
+
+	campRoot, err := campaign.DetectCached(ctx)
+	if err != nil {
+		return camperrors.Wrap(err, "not in a campaign")
+	}
+
+	result, err := projectsvc.Resolve(ctx, campRoot, projectStageProject)
+	if err != nil {
+		var notFound *projectsvc.ProjectNotFoundError
+		if errors.As(err, &notFound) {
+			fmt.Println(ui.Dim("\n" + projectsvc.FormatProjectList(notFound.AvailableProjects())))
+		}
+		return err
+	}
+
+	resolvedPath := result.Path
+
+	if err := result.RequireGit("git staging"); err != nil {
+		return err
+	}
+
+	relPath := result.LogicalPath
+	if relPath == "" {
+		relPath, _ = filepath.Rel(campRoot, resolvedPath)
+	}
+	fmt.Printf("Project: %s\n", ui.Value(relPath))
+
+	executor, err := git.NewExecutor(resolvedPath)
+	if err != nil {
+		return camperrors.Wrap(err, "failed to initialize git")
+	}
+
+	fmt.Println(ui.Info("Staging changes..."))
+	if err := executor.StageAll(ctx); err != nil {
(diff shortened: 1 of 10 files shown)


cmd/camp/stage.go
```diff
@@ -0,0 +1,114 @@
...
+			return pathErr
+		}
+		if err := git.StageAllExcluding(ctx, target.Path, paths); err != nil {
+			return err
+		}
+	}
+
+	cmdutil.ShowStagedSummary(ctx, target.Path)
+
+	hasChanges, err := executor.HasChanges(ctx)
```
obey-agent: This should check whether anything was actually staged, not whether the worktree still has any change. In the default root path, `StageAllExcluding` intentionally leaves submodule ref drift unstaged; if that is the only dirty state, `executor.HasChanges(ctx)` still returns true and the command prints `Changes staged` even though `git diff --cached --name-only` is empty. I reproduced that with a campaign containing only `M projects/sub`: `camp stage` reported success, but the index stayed empty. Please switch this decision to a staged-only predicate (for example `git.HasStagedChanges(ctx, target.Path)` or an executor method) and add a regression for the "only excluded submodule refs" case, so the command does not direct users to run `git commit` when nothing was staged.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The PR adds a useful `camp stage` / `camp project stage` surface and mostly reuses the existing commit staging primitives cleanly. The command registration, docs generation, stale-lock coverage, and project-stage path look coherent. I found one root-stage correctness issue in the default submodule-ref exclusion path.

**Key Findings:**
- `camp stage` reports `Changes staged` when the only dirty state is an intentionally excluded submodule ref. The index remains empty, but the command still tells the user to run `git commit`; see the inline comment on `cmd/camp/stage.go`. This should use a staged-only check and have a regression for the only-excluded-ref case.

**What's Done Well:** The new commands reuse `StageAll`, `StageAllExcluding`, and existing project resolution instead of introducing a parallel staging implementation. The integration coverage exercises the main root/project/stale-lock paths, and the CLI docs are updated consistently.

**Staff Standard:** Not quite yet. I would be comfortable merging after the root command distinguishes "nothing was staged because only refs were excluded" from actual staged content.

**Validation:**
- `go test -tags dev ./cmd/camp ./cmd/camp/project ./internal/git/...`
- `just test integration-run TestStage`
- Manual repro: campaign with only `M projects/sub` at root; `camp stage` printed `Changes staged`, while `git diff --cached --name-only` was empty.

cmd/camp/stage.go
obey-agent: Fixed in 3d5e4c9.

Switched the post-stage decision to `git.HasStagedChanges(ctx, target.Path)` so we test what actually made it into the index rather than `executor.HasChanges`, which also returns true for unstaged + untracked content (the source of the misreport when only an excluded submodule ref was dirty).

When `HasStagedChanges` is false and refs were excluded (root, no `--include-refs`), the command now surfaces the specific reason and points at `--include-refs`:

```
Staging changes...
Nothing staged. Submodule ref changes were excluded; rerun with --include-refs to stage them.
```

Added `TestStage_Root_OnlyExcludedRefs_ReportsNothingStaged` as the regression: dirty submodule ref only at root, assert the output does NOT claim "Changes staged", does mention `--include-refs`, and `git diff --cached --name-only` is empty.

Also rewrote the success hint to suggest `camp commit` / `camp p commit` instead of `git commit`, per the campaign's commit-discipline rule.

All 8 `TestStage*` integration tests and the full 2320-test unit suite pass on `3d5e4c9`. Manual repro of your scenario now prints the new message and leaves the index empty as expected.

cmd/camp/stage.go
```diff
@@ -0,0 +1,124 @@
...
+				return nil
+			}
+		}
+		fmt.Println(ui.Success("Nothing to stage, working tree clean"))
+		return nil
+	}
+
+	cmdutil.ShowStagedSummary(ctx, target.Path)
+	fmt.Println(ui.Success("Changes staged"))
+	fmt.Println(ui.Dim("Run 'camp commit' to record them."))
```
obey-agent: This follow-up is only correct for the campaign-root case. When the user ran `camp stage --sub` or `camp stage -p ...`, the staged index belongs to the project repo; a plain `camp commit` resolves back to the campaign root unless the user also passes `--sub`/`--project`, so it will not record the staged project changes and may operate on unrelated root changes. Please make this message target-aware, for example root -> `camp commit`, submodule -> `camp commit --sub` or `camp commit -p <path>`.

cmd/camp/project/stage.go
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+	if !staged {
+		fmt.Println(ui.Success("Nothing to stage in project"))
+		return nil
+	}
+
+	cmdutil.ShowStagedSummary(ctx, resolvedPath)
+	fmt.Println(ui.Success("✓ Project changes staged"))
+	fmt.Println(ui.Dim("Run 'camp p commit' to record them."))
```
obey-agent: This suggested next command only works when the current directory is still inside the resolved project. If the user ran `camp project stage --project sub` from the campaign root, a plain `camp p commit` re-runs project auto-detection from the root and errors instead of committing the staged project index. Please include the `--project <name>` form when `projectStageProject` was supplied, or otherwise make the follow-up command match the resolved project.

obey-agent (commented):
**Verdict:** Request Changes

**Overview:** The new staging commands reuse the existing `git.Stage`/`StageAllExcluding` primitives and the integration coverage hits the important root, project, ref-exclusion, include-ref, and stale-lock paths. The core staging behavior looks sound, but the success guidance is wrong for project/submodule targets and can lead users to run a command that does not commit the staged repository.

**Key Findings:**
- Inline: `camp stage --sub` / `camp stage -p ...` still prints `Run 'camp commit'`, but plain `camp commit` resolves to the campaign root and will not record the staged project index.
- Inline: `camp project stage --project ...` still prints `Run 'camp p commit'`, but from the campaign root that command re-runs auto-detection and errors instead of committing the staged project.

**What's Done Well:** The PR keeps the staging implementation narrow, leans on existing lock-aware git helpers, and adds useful containerized integration tests for the cases that matter most.

**Staff Standard:** I would be comfortable with the implementation after the follow-up command guidance is made target-aware. I would not ship the current messages because they point users at the wrong commit operation in documented workflows.

**Checks Run:** `just build-camp` passed; `just test unit` passed (2320/2320); `go test -count=1 -v -tags integration -run 'TestStage' ./tests/integration` passed. `just lint` still fails on existing repository-wide lint debt outside this PR's changed files.

cmd/camp/stage.go
obey-agent: Fixed in 9a735a3.

The hint now mirrors the targeting flag the user passed to `stage`:

| invocation | hint |
|---|---|
| `camp stage` | `camp commit` |
| `camp stage --sub` | `camp commit --sub` |
| `camp stage -p projects/foo` | `camp commit -p projects/foo` |

Implementation lives in `commitHintForStage()` at the bottom of `cmd/camp/stage.go` and reads the same `stageSub` / `stageProject` flag vars used to resolve the target up top, so the suggested commit always targets the same repo that was staged.

Added `TestStage_Sub_HintMatches` and `TestStage_Root_WithProjectFlag_HintMatches` as regressions, plus a tightened assertion on `TestStage_Root_BasicFile` that the plain path emits `camp commit` without `--sub` or `-p`.

cmd/camp/project/stage.go
obey-agent: Fixed in 9a735a3.

`camp project stage` now preserves `--project` in the follow-up hint:

| invocation | hint |
|---|---|
| `camp project stage` (auto-detected) | `camp p commit` |
| `camp project stage --project sub` | `camp p commit --project sub` |

Reads the same `projectStageProject` flag used to resolve the target, so the suggested commit hits the project that was actually staged regardless of cwd.

`TestStage_Project_WithProjectFlag` now asserts `camp p commit --project sub` appears in the output, and `TestStage_Project_FromInsideProject` asserts the auto-detected path emits plain `camp p commit` without the `--project` suffix.

obey-agent (commented):
**Verdict:** Approve

**Overview:** The re-review looks good. The follow-up command guidance is now target-aware for root, `--sub`, root `-p`, and `camp project stage --project` flows, and the added integration assertions cover the earlier failure modes directly.

**Key Findings:** No blocking findings on this pass. The prior comments about incorrect commit hints are addressed in the current head (`9a735a3bcbcd22737946c1219c61cfa22d9e13c4`).

**What's Done Well:** The implementation continues to reuse the existing lock-aware staging primitives, keeps the new command surface narrow, and now covers the UX edge cases with containerized tests.

**Staff Standard:** Yes, I would be comfortable submitting this after the current changes.

**Checks Run:** `gofmt -l` on changed Go files was clean; `git diff --check origin/main...HEAD` passed; `just build-camp` passed; `just test unit` passed (2320/2320); `go test -count=1 -v -tags integration -run 'TestStage' ./tests/integration` passed. `just lint` still fails on existing repository-wide lint debt outside this PR's changed files.

PR #294, opened 2026-05-19
Refresh navigation cache for workitem changes

lancekrogers:
## Summary
- invalidate the navigation index after `camp workitem create` and `camp workitem adopt`
- rebuild stale navigation caches when workflow, intent, or festival topology changes outside Camp
- add an integration regression proving newly created design workitems complete and navigate without manual `camp cache rebuild`

## Validation
- `go test ./internal/nav/index ./internal/commands/workitem`
- `just test integration-run TestIntegration_WorkitemCreateAndAdopt`
- `just test unit`

### DIFF
--- a/internal/commands/workitem/adopt.go
+++ b/internal/commands/workitem/adopt.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/Obedience-Corp/camp/internal/config"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	navindex "github.com/Obedience-Corp/camp/internal/nav/index"
 	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
 )
 
@@ -88,6 +89,7 @@ func runAdopt(ctx context.Context, cmd *cobra.Command, dir, typeFlag, title, idO
 	if err := atomicWriteFile(markerPath, buf, 0o644); err != nil {
 		return err
 	}
+	_ = navindex.Delete(campaignRoot)
 
 	fmt.Fprintf(cmd.OutOrStdout(),
 		"adopted %s\n  id: %s\n  type: %s\n",
--- a/internal/commands/workitem/adopt.go
+++ b/internal/commands/workitem/adopt.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/Obedience-Corp/camp/internal/config"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
-	navindex "github.com/Obedience-Corp/camp/internal/nav/index"
 	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
 )
 
@@ -89,7 +88,9 @@ func runAdopt(ctx context.Context, cmd *cobra.Command, dir, typeFlag, title, idO
 	if err := atomicWriteFile(markerPath, buf, 0o644); err != nil {
 		return err
 	}
-	_ = navindex.Delete(campaignRoot)
+	// Adoption writes inside an existing directory, which may not update the
+	// workflow/type parent mtime watched by passive cache staleness checks.
+	invalidateNavigationCache(cmd, campaignRoot)
 
 	fmt.Fprintf(cmd.OutOrStdout(),
 		"adopted %s\n  id: %s\n  type: %s\n",
--- a/internal/nav/index/cache.go
+++ b/internal/nav/index/cache.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/Obedience-Corp/camp/internal/config"
 	"github.com/Obedience-Corp/camp/internal/fsutil"
+	"github.com/Obedience-Corp/camp/internal/nav"
 )
 
 const (
@@ -125,7 +126,7 @@ func IsStale(idx *Index, campaignRoot string) bool {
 		return true
 	}
 
-	if anyNavigationTopologyChanged(campaignRoot, idx.BuildTime) {
+	if workitemTopologyChanged(campaignRoot, idx.BuildTime) {
 		return true
 	}
 
@@ -166,19 +167,21 @@ func IsStale(idx *Index, campaignRoot string) bool {
 	return false
 }
 
-func anyNavigationTopologyChanged(campaignRoot string, buildTime time.Time) bool {
-	for _, rel := range []string{
-		"workflow",
-		".campaign/intents",
-		"festivals",
+func workitemTopologyChanged(campaignRoot string, buildTime time.Time) bool {
+	for _, cat := range []nav.Category{
+		nav.CategoryWorkflow,
+		nav.CategoryIntents,
+		nav.CategoryFestivals,
 	} {
-		if dirOrImmediateChildChangedAfter(campaignRoot, rel, buildTime) {
+		if dirOrImmediateChildChangedAfter(campaignRoot, cat.Dir(), buildTime) {
 			return true
 		}
 	}
 	return false
 }
 
+// This is intentionally depth-bounded: root and immediate child mtimes catch
+// workitem/festival/intent additions without invalidating on ordinary file edits.
 func dirOrImmediateChildChangedAfter(root, rel string, buildTime time.Time) bool {
 	dir := filepath.Join(root, rel)
 	info, err := os.Stat(dir)
--- a/internal/nav/index/cache.go
+++ b/internal/nav/index/cache.go
@@ -91,7 +91,7 @@ func IsFresh(idx *Index) bool {
 
 // IsStale checks if the cache should be rebuilt.
 // Returns true if the cache is nil, too old, if the index version changed,
-// if the camp binary was rebuilt, or if project configuration changed.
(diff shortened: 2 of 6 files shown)


internal/commands/workitem/adopt.go
```diff
@@ -88,6 +89,7 @@ func runAdopt(ctx context.Context, cmd *cobra.Command, dir, typeFlag, title, idO
 	if err := atomicWriteFile(markerPath, buf, 0o644); err != nil {
 		return err
 	}
+	_ = navindex.Delete(campaignRoot)
```
obey-agent: Silent error swallow. `navindex.Delete` already returns `nil` on the common ENOENT case, so `_ = ...` only hides the failures that actually matter — permission denied, IO errors, partial write artifacts. If deletion truly fails here, the next `camp complete` consumes a stale index and the exact bug this PR fixes recurs with zero signal to the user.

At staff level this needs at least a stderr warning: `if err := navindex.Delete(campaignRoot); err != nil { fmt.Fprintf(cmd.ErrOrStderr(), "warning: failed to invalidate nav cache: %v\n", err) }`. The whole point of this PR is observability of cache state — don't sink the one signal that proves it worked.

Same comment applies to `create.go:108`.

internal/nav/index/cache.go
```diff
@@ -162,6 +166,48 @@ func IsStale(idx *Index, campaignRoot string) bool {
...
+func anyNavigationTopologyChanged(campaignRoot string, buildTime time.Time) bool {
+	for _, rel := range []string{
+		"workflow",
+		".campaign/intents",
+		"festivals",
+	} {
+		if dirOrImmediateChildChangedAfter(campaignRoot, rel, buildTime) {
+			return true
+		}
+	}
```
obey-agent, line 179: Magic strings where canonical constants already exist. `internal/nav/shortcuts.go` defines `CategoryWorkflow = "workflow"`, `CategoryFestivals = "festivals"`, and `CategoryIntents = ".campaign/intents"`. Hardcoding the raw paths here creates a second source of truth — if any of those layout constants are ever renamed, this function silently diverges and the cache invalidator misses real topology changes.

Use the constants directly:

```go
for _, cat := range []nav.Category{
    nav.CategoryWorkflow,
    nav.CategoryIntents,
    nav.CategoryFestivals,
} {
    if dirOrImmediateChildChangedAfter(campaignRoot, cat.Dir(), idx.BuildTime) {
        return true
    }
}
```

Related: the function name `anyNavigationTopologyChanged` implies coverage of all nav categories, but it only checks 3 of the 11 in `nav.ValidCategories()`. Either widen the scope (docs, ai_docs, design, code_reviews, pipelines all participate in navigation) or rename to something narrower like `workitemTopologyChanged` so a future reader doesn't assume coverage they don't have.

internal/nav/index/cache.go
```diff
@@ -162,6 +166,48 @@ func IsStale(idx *Index, campaignRoot string) bool {
...
+	return false
+}
+
+func dirOrImmediateChildChangedAfter(root, rel string, buildTime time.Time) bool {
+	dir := filepath.Join(root, rel)
```
obey-agent, line 186: Non-obvious depth semantics with no comment. `dirOrImmediateChildChangedAfter` stops at the immediate child layer — not a full walk. The reason (avoid invalidating the cache on every file edit inside a workitem, while still catching slug-dir additions which bump the type-dir mtime) is real and correct, but it's not derivable from the code.

This is the kind of subtle invariant that decays the moment someone reaches for `filepath.WalkDir` to "make it more thorough." Add one line above the function explaining why depth is intentionally bounded.

It also creates a hidden coupling with `adopt.go`: adopting an already-nested dir like `workflow/design/existing-dir` does *not* bump `workflow/design`'s mtime — only `existing-dir` itself changes. So the topology check alone would miss adopt; the explicit `navindex.Delete` in adopt.go is load-bearing for that case. If a future maintainer removes the explicit Delete believing topology detection covers everything, the bug returns. Worth a comment in adopt.go pointing at this constraint, or a unit test that locks in the behavior.

internal/nav/index/cache.go
```diff
@@ -162,6 +166,48 @@ func IsStale(idx *Index, campaignRoot string) bool {
 	return false
 }
 
+func anyNavigationTopologyChanged(campaignRoot string, buildTime time.Time) bool {
```
obey-agent: Missing unit coverage that matches the existing pattern. `cache_test.go` has `TestIsStale_GitmodulesChanged`, `TestIsStale_CampaignYamlChanged`, `TestIsStale_ProjectsDirChanged`, `TestIsStale_WorktreesDirChanged` — one targeted test per IsStale signal. This PR adds three new signals (workflow, intents, festivals) and adds zero parallel unit tests.

The integration test covers the workflow path for `camp workitem create`, but it doesn't cover:
- Manual change inside `festivals/` (e.g., a chained fest dropping a new sequence directory)
- New intent file added by external tooling under `.campaign/intents/inbox/`
- The depth boundary (file edit deep inside a workitem must *not* invalidate)

Add `TestIsStale_WorkflowTopologyChanged`, `TestIsStale_IntentsTopologyChanged`, `TestIsStale_FestivalsTopologyChanged`, and `TestIsStale_DeepFileEditDoesNotInvalidate` mirroring the existing helpers in this file. Without them the depth invariant is one refactor away from silently breaking.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Real bug, correct golden-path fix. Newly-created design workitems now appear in `camp complete` and `camp go` without a manual `camp cache rebuild`, and the integration test locks that behavior in. The shape of the solution (explicit `Delete` on mutation + topology-based `IsStale` as defense in depth) is right. The implementation is below the bar in four specific places that should be addressed before merge.

## Key Findings
- **Silent error swallow in both call sites** (`adopt.go:92`, `create.go:108`): `_ = navindex.Delete(...)` hides exactly the failures that matter — `Delete` already returns nil on ENOENT, so what's being suppressed is permission denied and IO errors. If invalidation fails, the bug this PR fixes recurs with no user signal. Warn to `cmd.ErrOrStderr()` at minimum.
- **Magic strings duplicating existing constants** (`cache.go` new helper): `"workflow"`, `".campaign/intents"`, `"festivals"` are hardcoded when `nav.CategoryWorkflow`, `nav.CategoryIntents`, `nav.CategoryFestivals` already exist as the canonical source. Two sources of truth = future drift.
- **Misleading scope** (`anyNavigationTopologyChanged`): name implies all nav topology, covers 3 of 11 categories. Either widen or narrow the name (e.g. `workitemTopologyChanged`).
- **No unit tests for the new IsStale signals**: file has one targeted test per existing IsStale signal; this PR adds three new signals and zero parallel tests. The depth-boundary invariant (file edits inside a workitem must not invalidate) is also un-locked-in — one refactor away from silently breaking.
- **Hidden coupling between adopt.go and the topology check**: explicit `Delete` in `adopt.go` is load-bearing because adopting a nested dir does not bump the immediate-child mtime. This is invisible in the code; a future maintainer removing the explicit Delete believing topology coverage suffices will reintroduce the bug. Comment or test should lock it in.

## What's Done Well
- Diagnosed correctly: stale cache after mutation, not a bug in the completion logic.
- Belt-and-suspenders design — explicit invalidation on write *and* passive topology detection on read — is the right architecture for this class of bug.
- Integration test covers the user-visible regression (create → complete → go), which is the exact symptom that prompted the fix.
- Context propagation is preserved; no new I/O without a context.

## Staff Standard
Not yet. The fix works on the happy path and the test proves it, but the silent error swallow, magic-string duplication, and missing unit coverage are all easy to fix and all the kinds of things a staff reviewer is paid to catch. Address the four findings and this is a clean merge.

obey-agent (commented):
## Verdict: Approve

## Overview
Reviewed latest head `bf8e831b86b661bf6adb0592601111fa056c8a01`. The follow-up commit addresses the earlier review concerns: cache invalidation failures now produce a user-visible warning, the topology check uses canonical `nav.Category` constants, the helper/comment scope is clearer, and the new `IsStale` signals have targeted unit coverage.

## Key Findings
- No new blocking findings on this pass.
- I did not add new inline comments because the remaining changes are coherent and the previous issues are addressed in the latest diff.

## What's Done Well
- The fix combines explicit invalidation on workitem mutation with passive topology staleness detection for out-of-band changes.
- The tests now cover workflow, intents, festivals, and the depth-boundary case that prevents ordinary deep file edits from invalidating the nav cache.
- The integration regression exercises the user-visible create -> complete -> go path without a manual cache rebuild.

## Validation
- `go test ./internal/nav/index ./internal/commands/workitem` passed.
- `just test unit` passed: 69 packages, 2369/2369 tests.
- `just test integration-run TestIntegration_WorkitemCreateAndAdopt` passed.
- `just lint` is not clean in this checkout, but the reported failures are broad pre-existing issues outside this PR's touched files.

## Staff Standard
Yes. I would be comfortable merging this version; the implementation is scoped, observable enough for the cache invalidation path, and backed by tests at the right layers.

PR #297, opened 2026-05-20
Scaffold standalone workflow skill

lancekrogers:
## Summary
- add a scaffolded `fest-standalone-workflows` skill for lightweight `WORKFLOW.md` loops
- document the current standalone first-run command sequence for agents
- include the TTY prompt-driven creation path for human users
- add containerized init and repair coverage for installing the skill

## Validation
- `go test ./internal/skills ./internal/scaffold`
- `go test -tags integration ./tests/integration -run 'TestInit_(ScaffoldsFestStandaloneWorkflowsSkill|Repair_RestoresMissingSkillFiles)$' -count=1`
- manual smoke: `fest create workflow <name> --no-init` from a TTY, followed by `fest next --json`

PR #298, opened 2026-05-20
Create custom workflow collections

lancekrogers:
## Summary
- add `camp workflow create <type> --shortcut <key>` for user-defined workflow collections under `workflow/<type>/`
- route custom navigation shortcuts through the configured navigation resolver so `camp go re ...` and shortcut drill forms work for non-standard paths
- order workitem-backed completion candidates by recent `.workitem` updates so new custom workitems are easier to find

## Validation
- `go test ./internal/nav ./internal/complete ./internal/commands/workflow ./cmd/camp`
- `just test unit`
- `DOCKER_HOST="unix://${HOME}/.colima/default/docker.sock" TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock go test -count=1 -v -tags=integration ./tests/integration/... -run 'TestIntegration_WorkflowCreateCustomWorkflow|TestIntegration_WorkitemCreateAndAdopt' -timeout 5m`

### DIFF
--- a/internal/commands/workflow/create.go
+++ b/internal/commands/workflow/create.go
@@ -1,0 +1,202 @@
+package workflow
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	navindex "github.com/Obedience-Corp/camp/internal/nav/index"
+)
+
+var pathSegmentPattern = regexp.MustCompile(`^[^\s/\\.\-\x00-\x1f][^\s/\\\x00-\x1f]{0,79}$`)
+
+func newCreateCommand() *cobra.Command {
+	var shortcut, title string
+	var replace bool
+
+	cmd := &cobra.Command{
+		Use:   "create <type>",
+		Short: "Create a custom workflow collection",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runCreate(cmd.Context(), cmd, createOptions{
+				Type:     args[0],
+				Shortcut: shortcut,
+				Title:    title,
+				Replace:  replace,
+			})
+		},
+	}
+
+	cmd.Flags().StringVar(&shortcut, "shortcut", "", "navigation shortcut for this workflow")
+	cmd.Flags().StringVar(&title, "title", "", "human-readable workflow title")
+	cmd.Flags().BoolVar(&replace, "replace", false, "replace an existing shortcut or concept with the same name")
+	_ = cmd.MarkFlagRequired("shortcut")
+
+	return cmd
+}
+
+type createOptions struct {
+	Type     string
+	Shortcut string
+	Title    string
+	Replace  bool
+}
+
+func runCreate(ctx context.Context, cmd *cobra.Command, opts createOptions) error {
+	if err := validatePathSegment("type", opts.Type); err != nil {
+		return err
+	}
+	if err := validatePathSegment("shortcut", opts.Shortcut); err != nil {
+		return err
+	}
+
+	cfg, campaignRoot, err := config.LoadCampaignConfigFromCwd(ctx)
+	if err != nil {
+		return camperrors.Wrap(err, "not in a campaign directory")
+	}
+
+	workflowTitle := opts.Title
+	if workflowTitle == "" {
+		workflowTitle = opts.Type
+	}
+
+	relPath := path.Join("workflow", opts.Type) + "/"
+	absPath := filepath.Join(campaignRoot, filepath.FromSlash(relPath))
+	if err := os.MkdirAll(absPath, 0o755); err != nil {
+		return camperrors.Wrapf(err, "create workflow directory %s", relPath)
+	}
+
+	if err := writeOBEYIfMissing(absPath, opts.Type, workflowTitle); err != nil {
+		return err
+	}
+
+	if err := upsertShortcut(ctx, campaignRoot, cfg, opts.Shortcut, relPath, workflowTitle, opts.Replace); err != nil {
+		return err
+	}
+	if err := upsertConcept(ctx, campaignRoot, cfg, opts.Type, relPath, workflowTitle, opts.Replace); err != nil {
+		return err
+	}
+	invalidateNavigationCache(cmd, campaignRoot)
+
+	fmt.Fprintf(cmd.OutOrStdout(),
+		"created %s\n  shortcut: %s -> %s\n  workitem type: %s\nnext: camp workitem create <slug> --type %s\n",
+		strings.TrimRight(relPath, "/"), opts.Shortcut, relPath, opts.Type, opts.Type)
+	return nil
+}
+
+func validatePathSegment(field, value string) error {
+	if value == "" {
+		return camperrors.NewValidation(field, field+" must not be empty", nil)
+	}
+	if !pathSegmentPattern.MatchString(value) {
+		return camperrors.NewValidation(field,
+			"invalid "+field+" "+value+": must not be empty, must not contain '/', '\\', whitespace, or control characters, must not start with '.' or '-', max 80 chars", nil)
+	}
+	return nil
+}
+
+func writeOBEYIfMissing(absPath, workflowType, title string) error {
+	obeyPath := filepath.Join(absPath, "OBEY.md")
+	if _, err := os.Stat(obeyPath); err == nil {
+		return nil
+	} else if !os.IsNotExist(err) {
+		return camperrors.Wrap(err, "stat workflow OBEY.md")
+	}
+
+	content := fmt.Sprintf(`# %s
+
+Custom workflow collection for %q workitems.
+
+Create a workitem:
+
+`+"```bash"+`
+camp workitem create <slug> --type %s
+`+"```"+`
+`, title, workflowType, workflowType)
+
+	if err := os.WriteFile(obeyPath, []byte(content), 0o644); err != nil {
+		return camperrors.Wrap(err, "write workflow OBEY.md")
+	}
+	return nil
+}
+
+func upsertShortcut(ctx context.Context, campaignRoot string, cfg *config.CampaignConfig, shortcut, relPath, title string, replace bool) error {
+	jumps := cfg.Jumps
+	if jumps == nil {
+		defaults := config.DefaultJumpsConfig()
+		jumps = &defaults
+	}
+	if jumps.Shortcuts == nil {
+		jumps.Shortcuts = make(map[string]config.ShortcutConfig)
+	}
+
+	if existing, ok := jumps.Shortcuts[shortcut]; ok && !replace {
+		if existing.Path == relPath {
+			return nil
+		}
+		return camperrors.NewValidation("shortcut",
+			"shortcut "+shortcut+" already points to "+existing.Path+"; use --replace to update it", nil)
+	}
+
+	jumps.Shortcuts[shortcut] = config.ShortcutConfig{
+		Path:        relPath,
+		Description: title + " workflow",
+		Source:      config.ShortcutSourceUser,
+	}
+	cfg.Jumps = jumps
+
+	if err := config.SaveJumpsConfig(ctx, campaignRoot, jumps); err != nil {
+		return err
+	}
+	return nil
+}
+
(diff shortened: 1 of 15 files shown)


internal/commands/workflow/create.go
```diff
@@ -0,0 +1,202 @@
...
+	Shortcut string
+	Title    string
+	Replace  bool
+}
+
+func runCreate(ctx context.Context, cmd *cobra.Command, opts createOptions) error {
+	if err := validatePathSegment("type", opts.Type); err != nil {
+		return err
+	}
+	if err := validatePathSegment("shortcut", opts.Shortcut); err != nil {
```
obey-agent, line 58: **Test coverage gap.** This file ships 202 lines of new business logic — validation, OBEY.md creation, shortcut upsert with collision detection, concept upsert with case-insensitive collision detection, the `--replace` semantics — with zero unit tests (`go test ./internal/commands/workflow` reports `[no test files]`). The corresponding `internal/commands/workitem/create.go` has a sibling `create_test.go` that covers analogous logic; this command should match that bar. Specifically uncovered: validation rejection of `/`, `\\`, leading `.`/`-`, whitespace; idempotent re-creation (same shortcut + same path returns nil); collision without `--replace` (both shortcut and concept produce a validation error with the existing path in the message); collision with `--replace` (overwrite succeeds and the new path is persisted); `writeOBEYIfMissing` does not overwrite an existing OBEY.md. The integration test exercises the happy path only and does not lock in the error-path contract this command exposes.

internal/commands/workflow/create.go
```diff
@@ -0,0 +1,202 @@
...
+`+"```bash"+`
+camp workitem create <slug> --type %s
+`+"```"+`
+`, title, workflowType, workflowType)
+
+	if err := os.WriteFile(obeyPath, []byte(content), 0o644); err != nil {
+		return camperrors.Wrap(err, "write workflow OBEY.md")
+	}
+	return nil
+}
```
obey-agent, line 124: **Case-sensitivity inconsistency across three lookup sites.** Today the same shortcut/concept name is treated differently depending on which code path touches it: (1) `upsertShortcut` at line 130 uses direct map indexing `jumps.Shortcuts[shortcut]` — case-sensitive collision check; (2) `upsertConcept` at line 159 uses `strings.EqualFold(concept.Name, name)` — case-insensitive collision check; (3) `findNavigationShortcut` in `internal/nav/configured_navigation.go` normalizes via `NormalizeNavigationName` and falls back to a case-insensitive linear scan — case-insensitive lookup. Concrete consequence: a user who runs `camp workflow create research --shortcut re` and later `--shortcut RE --replace` ends up with two map entries (`re` and `RE`) pointing to the same path, because line 130 doesn't see a collision. Nav then arbitrarily picks one during lookup. Pick a single normalized form (suggest case-insensitive on the storage key as well) and apply it consistently in all three sites, with a unit test that pins the chosen contract.

internal/commands/workflow/create.go
```diff
@@ -0,0 +1,202 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	navindex "github.com/Obedience-Corp/camp/internal/nav/index"
+)
+
```
obey-agent, line 20: **Duplicated path-safety regex.** This is the same pattern, byte-for-byte, as `slugPattern` in `internal/commands/workitem/create.go:38` (the workitem create command's slug validator). Two sources of truth for the same path-safety contract is a future-drift trap: a security-relevant tweak to one will silently miss the other. Extract to a single shared package (e.g. `internal/pathsafe` with `ValidateSegment(field, value) error`) and use it from both call sites, with the comment block currently sitting above the workitem definition (lines 23-37) co-located with the pattern in the shared package. The staff standard requires single-source path-safety contracts.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Reviewed head `3db1529a85f22783aa5a2667c0d093adc8902c3f`. The PR adds a real and well-shaped feature: custom `workflow/<type>/` collections with first-class navigation shortcut + concept registration, plus completion sorting that surfaces recently-touched workitems. The nav layer refactor that removes `handleCustomNavShortcut` in `cmd/camp/navigation/go.go` and routes everything through `resolveNavigationShortcut` is a clean simplification. Three issues need to be addressed before merge.

## Key Findings
- **Test gap on the new command.** `internal/commands/workflow/create.go` ships 202 lines of business logic with no unit tests at all (`go test ./internal/commands/workflow` reports `[no test files]`). The integration test covers the happy path but does not lock in the validation rejections, the idempotent re-creation case, the collision-without-`--replace` errors, the collision-with-`--replace` overwrite, or `writeOBEYIfMissing`'s no-clobber behavior. The peer command `internal/commands/workitem/create.go` has full sibling unit coverage and is the right bar to match. Inline detail at `internal/commands/workflow/create.go:58`.
- **Case-sensitivity asymmetry across the three lookup sites.** `upsertShortcut` collision detection is case-sensitive (direct map index), `upsertConcept` collision detection is case-insensitive (`strings.EqualFold`), and `findNavigationShortcut` in `internal/nav/configured_navigation.go` is case-insensitive (normalized + linear fallback). A user mixing case (`--shortcut re` then `--shortcut RE --replace`) ends up with two entries pointing to the same path while nav arbitrarily picks one. Inline detail at `internal/commands/workflow/create.go:130`.
- **Duplicated path-safety regex.** `pathSegmentPattern` at line 18 is byte-for-byte identical to `slugPattern` at `internal/commands/workitem/create.go:38`. Two sources of truth for a path-safety contract is a future-drift trap; extract to a shared package. Inline detail at `internal/commands/workflow/create.go:18`.

## What's Done Well
- The nav layer refactor is the right direction: removing the special-case branch in `cmd/camp/navigation/go.go` and folding custom-path resolution into the existing `ResolveConfiguredTarget` path keeps the resolver as the single source of truth, and the new tests in `internal/nav/configured_navigation_test.go` lock in the custom-shortcut, case-insensitive, and shortcut-drill variants of that contract.
- Cache invalidation calls `navindex.Delete` with a stderr warning on failure, matching the pattern that PR #294 just landed (silent error swallow is no longer the project default).
- `orderPathCandidates` in `internal/complete/complete.go` is correctly scoped: it only re-sorts when the directory actually contains any `.workitem` markers, and the test `TestGenerate_CustomWorkflowShortcutRecentFirst` pins the recent-first ordering for the new custom workflow case.
- The `camp wf` alias is preserved on `internal/commands/flow/flow.go`, so existing flow-status workflows still work. The `workflow` alias removal is the right separation: `camp workflow` as a top-level command needs an unambiguous parent.
- Integration test exercises the full chain end-to-end: workflow create -> workitem create with the custom type -> `camp complete <shortcut>` -> `camp complete <type>` -> `camp go <shortcut> <slug> --print`. That's the right end-to-end signal.

## Validation
- `go test ./internal/commands/workflow ./internal/nav ./internal/complete ./cmd/camp` passed locally on `3db1529`.
- `go build ./...` clean.
- Cross-checked the alias removal: no in-repo references to `camp workflow <subcommand>` for the old flow surface remain.

## Staff Standard
Not yet. The feature is well-designed at the architectural level and the nav refactor is the right move, but a 200-line command with zero unit tests and a real case-sensitivity inconsistency is not a code sample I would submit as my best work. Address the three findings above (the unit tests in particular) and this is a clean Approve.

obey-agent (commented):
## Verdict: Approve

## Overview
Reviewed updated head `12ad52398dbea6d0399f377506b496e92cbffef8`. The follow-up commit `12ad523 fix: address workflow review feedback` resolves all three blocking findings from the prior Request Changes pass cleanly and at the right layer.

## Key Findings
- No new blocking findings.
- The previous three issues are addressed:
  - **Test gap**: New `internal/commands/workflow/create_test.go` adds 433 lines across 10 test functions covering the validation table (empty, slash, backslash, leading `.`/`-`, space, control char, length boundary), `writeOBEYIfMissing` no-clobber, shortcut idempotency, storage-key normalization (`RE` -> `re`), shortcut collision with and without `--replace`, replace removing case variants, concept collision/replace, and the full `runCreate` end-to-end happy path and replace path.
  - **Case-sensitivity**: Unified on `nav.NormalizeNavigationName` at all three sites. `runCreate` normalizes the input shortcut once into `shortcutKey`. `upsertShortcut` now walks all matching case variants via `matchingShortcutKeys`, validates path collisions against every variant, and on `--replace` removes the non-canonical-form entries before writing the canonical key. `findNavigationShortcut` iterates the fallback scan over sorted keys for deterministic resolution.
  - **Regex duplication**: Extracted to a new `internal/pathsafe` package with `ValidateSegment(field, value) error` and the explanatory comment colocated. `internal/commands/workitem/create.go` switched to using it (`-26/+2` net), so `pathsafe.SegmentPattern` is now the single source of truth.

## What's Done Well
- The fix went to the right abstraction: a new package (`internal/pathsafe`) for the validation contract rather than a private helper in one of the two consumers. That sets up future consumers (e.g. a third command needing the same constraint) to import the same function without re-extracting.
- The shortcut normalization fix doesn't just write the canonical key, it actively migrates legacy variants out of `jumps.Shortcuts` on each upsert, with the persisted-state assertion in `TestUpsertShortcutIdempotentAndNormalizesStorageKey` locking that in via `LoadJumpsConfig` round-trip.
- The new tests don't just cover the immediate fix surface, they pin the contract: `TestUpsertShortcutReplaceRemovesCaseVariants` and `TestRunCreateReplacePersistsShortcutAndConcept` make sure the canonical-form invariant survives the persisted file. That's the right depth of coverage.
- The shared `pathsafe.SegmentPattern` is exported so a test in another package can assert on it directly if needed, while `ValidateSegment` is the convenience entry point most callers want. Sensible API split.

## Validation
- `go test ./internal/commands/workflow ./internal/commands/workitem ./internal/nav ./internal/complete ./internal/pathsafe ./cmd/camp` passed locally on `12ad5239`.
- `go build ./...` clean.
- Cross-checked the diff against `3db1529` (the previously reviewed head) and against my prior inline comments. Each comment maps to a concrete code change at the file and line called out.

## Staff Standard
Yes. The author treated the three findings as gating, fixed each at the architectural level rather than papering over symptoms, and pinned the new contract with tests. This is the response I'd want from any staff-level engineer to a Request Changes pass.

obey-agent (commented):
## Verdict: Approve

## Overview
Reviewed head `6350d3afd285585af0ced85ae342ef7e2b23c74c` under a stricter verification bar this pass (full `just test all` + `just lint`, default-to-block on any concern). Two new commits since the prior review at `12ad5239`: `da69252` is a one-line test stabilization (`mkdir -p $campaignB/festivals` before `camp attach`, consistent with the same pattern at line 116 of the file), and `6350d3a` is a merge of `main` that pulls in the now-merged PRs #297 and #299 with no substantive changes to this PR's surface.

## Key Findings
- No blocking findings.
- The new test commit (`da69252`): the cross-campaign attach test relies on `camp attach` reaching the campaign-detection path. Without the `mkdir -p`, the `festivals/` subdirectory may not exist in Docker containers where `fest` binary is absent (`InitCampaign` conditionally scaffolds `festivals/` only when `fest` is available, per the campaign root CLAUDE.md). The fix routes the test through the rejection path it claims to exercise rather than a directory-not-found error path. Consistent with the matching pattern at line 116 (`mkdir -p $campaignPath/docs`).
- The merge commit (`6350d3a`): bringing main into the PR branch after PRs #297 and #299 merged. No conflict resolution issues; the PR's actual diff against main is unchanged from the prior review.

## What's Done Well
- Test stabilization was scoped to the smallest possible change (one `mkdir -p`) rather than restructuring the test or reaching for `t.TempDir`. Minimal blast radius.
- Author kept the PR rebased and clean rather than letting it drift.

## Validation
- `just test all` on `6350d3a`: 2384 unit tests passed, 253 integration tests passed (Docker-based, real filesystem).
- `just lint`: 83 pre-existing failures total, 0 in files this PR touches (failures are in `internal/intent/tui/*`, `internal/tui/vim/*`, `internal/commands/flow/items.go`, `internal/commands/flow/show.go`, `internal/complete/category_test.go` — none modified by this PR).
- `go build ./...` clean.
- Diff against main confirms 15 files changed, +1034/-88 lines, matching the previously verified surface.
- Cross-checked `da69252`'s 1-line fix against the surrounding test logic and the `InitCampaign` festival-scaffolding conditional in the CLAUDE.md.

## Staff Standard
Yes. Under the stricter verification bar (full test suite, lint check, default-to-block), I find nothing that would block. The test stabilization is a legitimate fix to an environment-specific assumption, and the merge commit is clean. The core feature surface (workflow create, case-sensitivity normalization, shared `pathsafe` package) was verified in the prior review and remains unchanged.

PR #299, opened 2026-05-20
Tighten workitems skill guidance

lancekrogers:
## Summary
- tighten the scaffolded `camp-workitems` skill using skill-authoring guidance
- keep the skill focused on discovery, interactive use, and create/adopt commands
- remove implementation-file details so agents use `camp workitem create` or `camp workitem adopt` instead of hand-writing tracking metadata

## Validation
- `go test ./internal/skills ./internal/scaffold`
- `go test -tags integration ./tests/integration -run 'TestInit_(ScaffoldsCampWorkitemsSkill|Repair_RestoresMissingSkillFiles)$' -count=1`

### DIFF
--- a/internal/scaffold/campaign/templates/.campaign/skills/camp-workitems/SKILL.md.tmpl
+++ b/internal/scaffold/campaign/templates/.campaign/skills/camp-workitems/SKILL.md.tmpl
@@ -1,23 +1,28 @@
 ---
 name: camp-workitems
-description: Discover, triage, and create campaign work items with `camp workitem` / `camp workitems`. Use when finding active work across intents, designs, explore notes, festivals, or custom `.workitem` directories; when agents need `camp workitem --json`; or when creating/adopting tracked work-item folders.
+description: Find, filter, choose, create, or adopt campaign work items with `camp workitem`, `camp wi`, or `camp workitems`. Use when a user wants current active work across intents, designs, explore notes, festivals, or tracked workflow directories; when agents need safe `camp workitem --json` output; or when creating/adopting tracked workflow folders.
 ---
 
 # Camp Workitems
 
-`camp workitem` is the canonical command. `camp wi` and `camp workitems` are aliases.
+Use `camp workitem` to find current campaign work across intents, workflow
+design/explore docs, festivals, and tracked workflow directories.
 
-Use it to find active campaign work across intents, workflow design/explore
-docs, festivals, and custom `.workitem`-tracked workflow directories.
+`camp workitem` is canonical. `camp wi` and `camp workitems` are aliases.
 
-## Agent-Safe Discovery
+## Discover
 
-Agents and scripts should use non-interactive output:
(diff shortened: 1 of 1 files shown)


obey-agent (commented):
## Verdict: Approve

## Overview
Reviewed head `a440baa6ec60100cd426238028ba2f0e96ec345e` (head shifted from `65fe65cb` to `a440baa6` during the review; re-fetched and re-diffed before posting). Template-only PR that tightens the scaffolded `camp-workitems` skill. The trim is consistent with the author's stated philosophy in the PR body ("avoid mini command-reference/tutorial phrasing in the skill body") and does not drop load-bearing content.

## Key Findings
- No blocking findings.

## What's Done Well
- The trigger description now names all three aliases explicitly (`camp workitem`, `camp wi`, `camp workitems`) instead of two. That widens the matcher surface without adding noise.
- Verb choice in the description shifts from "Discover, triage, and create" to "Find, filter, choose, create, or adopt" which more accurately mirrors the command's actual capabilities: `--type`/`--stage`/`--query` are real filters and the TUI is a chooser.
- The Discover section now separates JSON-mode examples from `--print` shell-integration examples into two code blocks instead of mixing them in one. Easier to grep for the right invocation.
- The rewritten "Avoid" entry on hand-writing tracking metadata positively directs agents to `camp workitem create` / `adopt` rather than just warning against a failure mode. That's the orientation that actually changes behavior.
- The `.campaign/settings/workitems.json` warning is preserved. That's the one Avoid-list item that's load-bearing because the file is reachable through file discovery and an agent could plausibly try to edit it.
- All operational claims still match `internal/commands/workitem/workitem.go`: alias list (`wi`, `workitems`), valid stages (`inbox|active|ready|planning`), non-interactive flag contract (`--json` or `--print`), slug rules (no `/`, `\`, whitespace, control chars; no leading `.` or `-`; max 80 chars).
- Integration test assertions in `tests/integration/init_test.go` (`camp workitem --json` and `camp workitems` substrings) still hold for the new content, so the test will detect future template drift away from these triggers.

## Validation
- `go test ./internal/skills ./internal/scaffold ./cmd/camp` passed locally on `a440baa6`.
- `go build ./...` clean.
- `just lint` reports 83 issues, but every one is in unrelated files (`internal/intent/tui/*`, `internal/tui/vim/*`, etc.) and pre-exists this PR which only touches a single `.tmpl` file.
- Did NOT run the integration test suite (requires the shared container harness); that path is exercised in CI and the assertions in `init_test.go` / `repair_test.go` already match the trimmed content.
- Diffed against the prior-merged `225c379` baseline as well as the within-PR commit `65fe65cb` to make sure both stages of the trim were intentional and additive.

## Staff Standard
Yes. This is exactly how an iterative skill-doc refinement should look: keep the trigger keywords broad, tighten the body to action examples plus a small guard rail list, preserve only the warnings that catch failure modes a careful user could still hit. No claim drift versus the implementation.

PR #300, opened 2026-05-22
Move workitems to local dungeon

lancekrogers:
## Summary
- add `camp dungeon move --workitem` for moving `workflow/<type>/<slug>` workitems into their local dungeon from anywhere
- resolve workitem targets by stable id, relative path, or slug, with ambiguity checks
- initialize local dungeon scaffold and include scaffold/move files in the existing dungeon auto-commit path
- document the new flag and add unit/integration coverage

## Tests
- `go test ./cmd/camp/dungeon`
- `go test ./cmd/camp/dungeon ./internal/dungeon ./internal/workitem`
- `go test -tags=integration ./tests/integration -run 'TestDungeonMove_Workitem|TestDungeonMove_TriageDirectToStatusWithCommit' -count=1`\n- `go test ./...`\n- `git diff --check`

### DIFF
--- a/cmd/camp/dungeon/move.go
+++ b/cmd/camp/dungeon/move.go
@@ -23,6 +23,8 @@ var dungeonMoveCmd = &cobra.Command{
 Without --triage, moves an item already in the dungeon root to a status directory.
 With --triage, moves an item from the parent directory into the dungeon.
 With --triage and --to-docs, routes an item to an existing campaign-root docs/<subdirectory>.
+With --workitem, resolves a campaign workitem from anywhere and moves its directory
+into the workitem type's local dungeon.
 Moves are always auto-committed so dungeon history remains auditable.
 
 Statuses: completed, archived, someday
@@ -32,7 +34,8 @@ Examples:
   camp dungeon move stale-doc completed          Move dungeon item to completed
   camp dungeon move old-project --triage         Move parent item into dungeon root
   camp dungeon move old-project archived --triage Move parent item directly to archived
-  camp dungeon move stale-note.md --triage --to-docs architecture/api Route to docs subdirectory`,
+  camp dungeon move stale-note.md --triage --to-docs architecture/api Route to docs subdirectory
+  camp dungeon move feature-slug archived --workitem Move workitem directory to its local archive`,
 	Args: cobra.RangeArgs(1, 2),
 	Annotations: map[string]string{
 		"agent_allowed": "true",
@@ -47,6 +50,7 @@ func init() {
 	flags := dungeonMoveCmd.Flags()
 	flags.Bool("triage", false, "Move from parent directory (not from dungeon root)")
 	flags.String("to-docs", "", "Route triage item into an existing campaign-root docs/<subdir> (requires --triage)")
+	flags.Bool("workitem", false, "Resolve item as a campaign workitem and move its directory to the local dungeon")
 }
 
 func runDungeonMove(cmd *cobra.Command, args []string) error {
@@ -54,6 +58,7 @@ func runDungeonMove(cmd *cobra.Command, args []string) error {
 
 	triageMode, _ := cmd.Flags().GetBool("triage")
 	toDocs, _ := cmd.Flags().GetString("to-docs")
+	workitemMode, _ := cmd.Flags().GetBool("workitem")
 
 	itemName := args[0]
 	status := ""
@@ -68,6 +73,20 @@ func runDungeonMove(cmd *cobra.Command, args []string) error {
 		if status != "" {
 			return fmt.Errorf("status argument cannot be combined with --to-docs; use either a dungeon status or --to-docs <subdir>")
 		}
+	}
+
+	if workitemMode {
+		if triageMode {
+			return fmt.Errorf("--workitem cannot be combined with --triage because workitem mode resolves the source directory itself")
+		}
+		if toDocs != "" {
+			return fmt.Errorf("--workitem cannot be combined with --to-docs")
+		}
+		move, err := moveWorkitemToDungeon(ctx, cmd, itemName, status)
+		if err != nil {
+			return err
+		}
+		return commitDungeonMove(ctx, move)
 	}
 
 	cmdCtx, err := resolveDungeonCommandContext(ctx)
@@ -138,8 +157,18 @@ func runDungeonMove(cmd *cobra.Command, args []string) error {
 		destinationPaths = []string{targetPath}
 	}
 
-	files := commit.NormalizeFiles(cmdCtx.CampaignRoot, destinationPaths...)
-	preStaged, err := stageTrackedMoveSourceDeletions(ctx, cmdCtx.CampaignRoot, sourcePaths)
+	return commitDungeonMove(ctx, &dungeonMoveCommit{
+		Config:           cmdCtx.Config,
+		CampaignRoot:     cmdCtx.CampaignRoot,
+		Description:      description,
+		SourcePaths:      sourcePaths,
+		DestinationPaths: destinationPaths,
+	})
+}
+
+func commitDungeonMove(ctx context.Context, move *dungeonMoveCommit) error {
+	files := commit.NormalizeFiles(move.CampaignRoot, move.DestinationPaths...)
+	preStaged, err := stageTrackedMoveSourceDeletions(ctx, move.CampaignRoot, move.SourcePaths)
 	if err != nil {
 		fmt.Printf("%s Move was applied on disk, but staging the source deletion failed.\n", ui.WarningIcon())
 		fmt.Printf("%s %v\n", ui.WarningIcon(), err)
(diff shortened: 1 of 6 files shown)


cmd/camp/dungeon/workitem_move.go
```diff
@@ -0,0 +1,233 @@
...
+
+func validateWorkitemDungeonStatus(status string) error {
+	if status == "" {
+		return nil
+	}
+	if status == "." || status == ".." || strings.Contains(status, "/") || strings.Contains(status, "\\") {
+		return camperrors.Wrap(intdungeon.ErrInvalidStatus, status)
+	}
+	return nil
+}
```
obey-agent: **Duplicates `internal/dungeon.validateStatusName` and produces a worse error message.**

The service-level `validateStatusName` (`internal/dungeon/service.go:69`) already rejects empty, `.`, `..`, and path-separator-containing statuses, and `MoveToDungeonStatus` calls it before doing any work. So this guard is duplicate validation in service of avoiding a no-op `svc.Init` call beforehand — but `Init` is documented idempotent and cheap.

Worse, the wrapping order produces backwards-reading errors: `camperrors.Wrap(ErrInvalidStatus, status)` yields `"foo/bar: invalid status"`, while the service produces `"foo/bar (contains path separator): invalid status"`. The service message is strictly more informative.

Suggested fix: delete this function and either (a) call `svc.Init` first and rely on `MoveToDungeonStatus` to validate the status, or (b) extract `validateStatusName` to a package-level helper and call it directly. Either avoids the parallel implementation.

cmd/camp/dungeon/workitem_move.go
```diff
@@ -0,0 +1,233 @@
...
+		paths = append(paths, item.RelativePath)
+	}
+	sort.Strings(paths)
+	return strings.Join(paths, ", ")
+}
+
+func invalidateDungeonWorkitemNavigationCache(cmd *cobra.Command, campaignRoot string) {
+	if err := navindex.Delete(campaignRoot); err != nil {
+		fmt.Fprintf(cmd.ErrOrStderr(), "warning: failed to invalidate navigation cache: %v\n", err)
+	}
```
obey-agent, line 218: **Nav-cache invalidation here is inconsistent with the rest of `runDungeonMove` — the other branches don't invalidate.**

This is the only path in `cmd/camp/dungeon` that calls `navindex.Delete`. Standard triage moves, inner moves, and to-docs routes all change filesystem layout under workflow paths too (or at least under the campaign tree the nav index walks), and none of them invalidate. That means one of two things is true:

1. The nav index doesn't actually need invalidation after dungeon moves (its `IsStale` check handles it), in which case this call is dead defense and should be removed.
2. The nav index *does* need invalidation, in which case the existing triage and inner moves have a stale-cache bug that this PR leaves unaddressed.

If it's (2), lift this into `commitDungeonMove` so both code paths benefit. Either way, the inconsistency invites confusion when the next person tries to figure out which moves need cache eviction.

A one-line comment explaining the workitem-specific reason would also be acceptable if you've verified the other paths don't need it.

cmd/camp/dungeon/workitem_move.go
```diff
@@ -0,0 +1,233 @@
...
+	for _, item := range matches {
+		paths = append(paths, item.RelativePath)
+	}
+	sort.Strings(paths)
+	return strings.Join(paths, ", ")
+}
+
+func invalidateDungeonWorkitemNavigationCache(cmd *cobra.Command, campaignRoot string) {
+	if err := navindex.Delete(campaignRoot); err != nil {
+		fmt.Fprintf(cmd.ErrOrStderr(), "warning: failed to invalidate navigation cache: %v\n", err)
```
obey-agent: **Warning style diverges from the rest of `cmd/camp/dungeon`.**

All other dungeon warnings use `ui.WarningIcon()` to stdout (see `move.go:173-174,191-192` and `crawl.go:210-211`). This one uses a lowercase `"warning: "` text prefix to stderr. Minor inconsistency, but matters for grep-ability and UX consistency. Suggest:

```go
fmt.Fprintf(cmd.ErrOrStderr(), "%s failed to invalidate navigation cache: %v\n", ui.WarningIcon(), err)
```

(Or drop the entire helper if the nav-invalidation comment above is resolved by removal.)

cmd/camp/dungeon/workitem_move.go
```diff
@@ -0,0 +1,233 @@
...
+
+	matchStages := []struct {
+		name  string
+		match func(wkitem.WorkItem) bool
+	}{
+		{
+			name: "stable id",
+			match: func(item wkitem.WorkItem) bool {
+				return item.StableID != "" && item.StableID == raw
+			},
```
obey-agent, line 110: **`item.StableID != ""` guard is dead defense.**

`raw` is rejected at line 112 if it's empty after trim, so by the time this lambda runs `raw != ""`. The `item.StableID != ""` short-circuit therefore only prevents matching an item with no stable ID against a non-empty `raw` — but `"" == raw` is already impossible. The whole condition can collapse to `item.StableID == raw`.

Not a bug, but the guard adds reader friction ("why is this needed?") and the line above doesn't have it for relative-path or slug matches.

cmd/camp/dungeon/move.go
```diff
@@ -138,20 +157,30 @@ func runDungeonMove(cmd *cobra.Command, args []string) error {
...
+	return commitDungeonMove(ctx, &dungeonMoveCommit{
+		Config:           cmdCtx.Config,
+		CampaignRoot:     cmdCtx.CampaignRoot,
+		Description:      description,
+		SourcePaths:      sourcePaths,
+		DestinationPaths: destinationPaths,
+	})
+}
+
+func commitDungeonMove(ctx context.Context, move *dungeonMoveCommit) error {
```
obey-agent, line 169: **Nice refactor — extracting `commitDungeonMove` and `dungeonMoveCommit` is exactly the right move and pays for itself the moment a third move path lands.**

The struct keeps the commit pipeline ignorant of which resolver populated it, which is the correct boundary. Worth keeping this pattern as a template if any other dungeon-adjacent command (workflow promote, intent triage) ever needs to share the commit-source-deletion-then-crawl flow.

obey-agent (commented):
## Verdict: Comment (draft)

### Overview
Solid feature work. The `camp dungeon move --workitem` path resolves workitems by stable ID, relative path, or slug, falls back through stages cleanly, and correctly bypasses `resolveDungeonCommandContext` because workitem mode discovers its own dungeon path from the resolved workitem. The `commitDungeonMove` extraction is the right structural move and will pay off the next time a new resolver lands. Build and unit tests pass; integration coverage is reasonable.

### Key Findings
- **Duplicate status validation.** `validateWorkitemDungeonStatus` re-implements `internal/dungeon.validateStatusName` and produces a less informative error message. Inline comment on `workitem_move.go:108`.
- **Inconsistent nav-cache invalidation.** Only the workitem path calls `navindex.Delete`. Standard triage and inner moves don't. Either the call is dead defense or the other paths have a stale-cache bug. Inline comment on `workitem_move.go:232`.
- **Warning style diverges.** Other dungeon warnings use `ui.WarningIcon()`; new one uses `"warning: "` text. Inline comment on `workitem_move.go:231`.
- **Dead defensive check** on `item.StableID != ""`. Inline comment on `workitem_move.go:124`.

### What's Done Well
- Context propagation is correct throughout (`ctx` first param, no leaks).
- Error pattern matches existing dungeon code (`fmt.Errorf` for new validation messages, `camperrors.Wrap`/`Wrapf` when wrapping real causes).
- Resolver staging in `selectWorkitemDungeonTarget` is well thought out: stable-ID wins, then relative-path, then slug, with ambiguity errors at each stage and a helpful "run `camp workitem --json`" hint on miss.
- `resolveWorkitemDungeonTarget` correctly rejects non-directory workitems, items outside `workflow/<type>/<slug>`, and items with `dungeon` in the type or slug position (prevents moving the dungeon dir into itself).
- Integration tests cover slug, ID, and relative-path inputs end-to-end, including running from outside the workitem directory (`path+"/docs"`).
- Refactoring the original `runDungeonMove` to share `commitDungeonMove` is clean and preserves prior behavior.

### Staff Standard
Would I be proud to submit this as my own work? Close to yes, after the duplicate-validation and nav-cache-consistency items are resolved. Both are small fixes that remove future maintenance debt rather than gating correctness, so a clean follow-up commit would land this. The feature design, test coverage, and refactor are all at the right level.

obey-agent (commented):
**Verdict:** Approve

**Overview:** This PR cleanly extends `camp dungeon move` with `--workitem` while reusing the existing dungeon service, commit crawler, workitem discovery, and navigation cache invalidation paths. The implementation keeps resolution explicit by stable ID, relative path, or slug, and rejects unsupported workitem shapes before moving anything.

**Key Findings:** No blocking or follow-up findings. I did not find correctness, error-handling, integration, or maintainability issues worth inline comments.

**What's Done Well:** The change keeps the move/commit boundary centralized in `commitDungeonMove`, preserves selective staging of destination plus source deletion, and adds useful unit and containerized integration coverage for slug, ID, relative-path, status, and clean-git-state behavior.

**Validation:**
- `go test ./cmd/camp/dungeon ./internal/dungeon ./internal/workitem`
- `go test ./...`
- `go test -tags=integration ./tests/integration -run 'TestDungeonMove_Workitem|TestDungeonMove_TriageDirectToStatusWithCommit' -count=1`

**Staff Standard:** Yes. I would be comfortable submitting this as my own production change.

PR #301, opened 2026-05-22
feat(promote): add camp promote for workitem dungeon-status moves

lancekrogers:
## Summary
- New top-level `camp promote <status>` command for directory-style workitems under `workflow/<type>/`.
- Detects the workitem from cwd (walks up from anywhere inside it); no stable-ID / slug arg needed.
- No `--dungeon` flag: status directories live under the workitem-local dungeon, so naming them implies dungeon. Outside dungeon = active.
- Handles both active workitems (`workflow/<type>/<slug>`) and already-dungeoned workitems (`workflow/<type>/dungeon/<status>[/<date>]/<slug>`), so the same verb promotes between dungeon statuses.
- Flags `--no-commit` and `--json` mirror `fest promote`.

## Why
Festivals and intents already have ergonomic single-verb promotion. Generic workitems under `workflow/` only had `camp dungeon move --workitem <id> <status>` (verbose, requires stable-ID/path/slug) or `camp flow move` (slug-only, no cross-surface resolution). This fills the gap with the most common case: \"I'm in this workitem; shelve it.\"

## Scope (intentional)
- `workflow/` workitems only — festivals and intents have their own promote.
- Directory-style workitems only — matches PR #300 constraint.
- Out of scope: bringing items back out of dungeon (`promote active`), workflow-config validation, shared fest/camp promote package.

## Implementation
- `cmd/camp/promote/detect.go` — pure path-based detection of the workitem boundary. Recognizes legacy flat dungeon layout and dated dungeon layout (via `statuspath.IsDateDir`).
- `cmd/camp/promote/promote.go` — cobra command, calls `intdungeon.Service.MoveToDungeonStatus`, invalidates nav cache, commits via shared helper.
- `cmd/camp/dungeon/` — exported the move/commit helpers (`DungeonMoveCommit`, `CommitDungeonMove`, `WrapDungeonMoveError`, `RelFromRoot`) so promote can reuse the existing pipeline without duplication.
- Registered as agent-allowed in the camp restricted-commands manifest.

## Tests
- `go test ./cmd/camp/promote/...` — 15 table-driven unit tests for `detectWorkitemFromCwd` covering active, subdir, dated/legacy dungeon, and all error cases.
- `go test -tags=integration ./tests/integration -run 'TestPromote_'` — 7 containerized tests: FromActive, FromSubdir, BetweenDungeonStatuses, NotInWorkitem, InvalidStatus, AlreadyAtStatus, NoCommit.
- `go test ./...` — full unit suite green (incl. manifest tests updated for the new entry).
- `go test -tags=integration ./tests/integration -run 'TestPromote_|TestDungeonMove_'` — promote + PR #300 dungeon tests both green; no regression in the shared helpers.

## Test plan
- [x] Unit tests pass
- [x] Integration tests pass
- [x] No regression in existing dungeon move tests
- [ ] Manual smoke: `cd workflow/design/<slug> && camp promote someday`

### DIFF
--- a/cmd/camp/promote/promote.go
+++ b/cmd/camp/promote/promote.go
@@ -3,6 +3,7 @@ package promote
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 
@@ -49,11 +50,14 @@ func init() {
 }
 
 type promoteResult struct {
-	Slug   string `json:"slug"`
-	Type   string `json:"type"`
-	Status string `json:"status"`
-	From   string `json:"from"`
-	To     string `json:"to"`
+	Slug          string   `json:"slug"`
+	Type          string   `json:"type"`
+	Status        string   `json:"status"`
+	From          string   `json:"from"`
+	To            string   `json:"to"`
+	Committed     bool     `json:"committed"`
+	CommitMessage string   `json:"commit_message,omitempty"`
+	Warnings      []string `json:"warnings,omitempty"`
 }
 
 func runPromote(cmd *cobra.Command, args []string) error {
@@ -62,6 +66,10 @@ func runPromote(cmd *cobra.Command, args []string) error {
 
 	noCommit, _ := cmd.Flags().GetBool("no-commit")
 	jsonOut, _ := cmd.Flags().GetBool("json")
+
+	if status == "active" {
+		return fmt.Errorf("promote cannot target %q: %q is not a dungeon status (outside the dungeon a workitem is already active); restoring workitems out of dungeon is not supported by promote", status, status)
+	}
 
 	cfg, campaignRoot, err := config.LoadCampaignConfigFromCwd(ctx)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


cmd/camp/promote/promote.go
```diff
@@ -0,0 +1,145 @@
...
+	Slug   string `json:"slug"`
+	Type   string `json:"type"`
+	Status string `json:"status"`
+	From   string `json:"from"`
+	To     string `json:"to"`
+}
+
+func runPromote(cmd *cobra.Command, args []string) error {
+	ctx := cmd.Context()
+	status := args[0]
```
obey-agent, line 65: `active` is not a dungeon status in this command model, but today `camp promote active` will move the workitem into `workflow/<type>/dungeon/active/<date>/<slug>`. Since the PR explicitly scopes restoring items back out of dungeon as out of scope, please reject `active` with a clear hint instead of silently creating a misleading dungeon status that still hides the workitem from the active workflow tree.

cmd/camp/promote/promote.go
```diff
@@ -0,0 +1,145 @@
...
+
+	targetPath, err := svc.MoveToDungeonStatus(ctx, loc.Slug, loc.ParentPath, status)
+	if err != nil {
+		return dungeoncmd.WrapDungeonMoveError(err, loc.Slug, status)
+	}
+
+	src := dungeoncmd.RelFromRoot(campaignRoot, loc.SourcePath)
+	dst := dungeoncmd.RelFromRoot(campaignRoot, targetPath)
+
+	if jsonOut {
```
obey-agent, line 122: `--json` emits the success object before the auto-commit path runs, and `CommitDungeonMove` writes human status lines to stdout afterward. That means the default `camp promote completed --json` output is not parseable JSON, and if the commit fails we have already printed a success payload. Please move JSON emission after all side effects and either include commit status in the JSON response or keep commit/status chatter off stdout so `--json` remains a reliable machine contract.

obey-agent (changes requested):
Verdict: Request Changes

Overview: The new cwd-based `camp promote <status>` flow is well-scoped and reuses the existing dungeon move pipeline rather than building a parallel mover. The path detector and containerized integration coverage are solid, but two command-contract issues need to be fixed before merge.

Key Findings:
- `--json` currently emits a success object before auto-commit and then writes human commit output to stdout, so the default JSON mode is not parseable and can report success before a commit failure. See inline comment on `cmd/camp/promote/promote.go`.
- `camp promote active` currently creates a dungeon status named `active`, even though active means outside the dungeon and restore-to-active is explicitly out of scope. See inline comment on `cmd/camp/promote/promote.go`.

What's Done Well: The command delegates movement and commit staging through the existing dungeon service/helper path, which keeps behavior consistent with `camp dungeon move`. The integration tests cover the main active/subdir/status-to-status/no-commit paths and the existing dungeon move regression suite still passes.

Staff Standard: Not yet. I would be comfortable after the JSON contract is fixed and `active` is either rejected with a clear hint or implemented as an explicit restore path.

Validation run:
- `go test ./cmd/camp/promote/...`
- `go test ./cmd/camp ./cmd/camp/dungeon/...`
- `go test ./...`
- `go test -tags=integration ./tests/integration -run 'TestPromote_'`
- `go test -tags=integration ./tests/integration -run 'TestPromote_|TestDungeonMove_'`

obey-agent (approved):
Verdict: Approve.

Re-reviewed after ddecf0c. The previous blockers are addressed: `camp promote active` now fails instead of creating `dungeon/active`, and `--json` now emits one final parseable result after the commit step with commit metadata and warnings included.

I re-read the promote/dungeon integration path and ran:
- `go test ./cmd/camp/promote/... ./cmd/camp ./cmd/camp/dungeon/...`
- `go test ./...`
- `go test -tags=integration ./tests/integration -run 'TestPromote_|TestDungeonMove_'`
- `git diff --check main...HEAD`

Staff standard: yes, I would be comfortable shipping this.

PR #302, opened 2026-05-25
feat(cw0003): aggregate workflow collections, workitem links, and scoped commits

lancekrogers:
## Summary

Aggregate CW0003 branch for camp workitem linking and commit traceability. This PR now contains the stacked festival work merged back into PR 302, not just the original workflow-surface slice.

- Adds lightweight camp workflow collections with `camp workflow create`, `list`, `show`, `shortcut add`, `doctor`, and `sync`. These collections provide navigation shortcuts and workitem-type roots, with standard terminal dungeon directories for completed, archived, and someday items.
- Clarifies the camp/fest boundary: camp workflow collections are not Festival methodology workflows. Camp owns navigation and workitem collection scaffolding; fest owns methodology `WORKFLOW.md`, phases, sequences, tasks, and gates.
- Adds workitem linking and context commands: `camp workitem link`, `unlink`, `links`, `current`, `resolve`, and `doctor`, backed by a link registry and resolver tiers for cwd, explicit selectors, current workitem, linked project, and festival context.
- Extends workitem metadata to carry stable refs and quest IDs, with doctor/backfill support for older workitem files.
- Adds scoped workitem commits through `camp workitem commit`, and threads workitem refs into `camp commit` / `camp p commit` campaign tags so commit history remains traceable without relying on agents to remember a separate command.
- Adds `camp workitem commits` for cross-repo commit queries by workitem ref.
- Adds recent-first navigation completion ordering (`internal/complete`), shared by builtin and custom completion paths, so the most recently touched workitems surface first under the shell-completion budget.
- Adds JSON contracts and error envelopes for agent-facing command surfaces, plus user-facing docs for workflow collections, workitem links, scoped commits, JSON contracts, and generated CLI reference pages.
- Hardens registry/config writes, link ID generation, resolver behavior, dirty-index handling, symlinked campaign roots, stale lock recovery, and workflow doctor/sync repair behavior.

## Correctness fixes from review rounds

These were blocking bugs surfaced during review, not generic hardening:

- Festival commit staging now uses the resolved festival. When a workitem had primary links to multiple festivals, `camp workitem commit` could emit one festival's `FE-` tag while staging another festival's files; staging now selects the festival scope with the same matcher the resolver used, so tag and staged paths always refer to the same festival.
- Multi-workitem link registries work: a registry holding links for one workitem no longer blocks adding a link for another.
- `camp workitem doctor` reports and fixes malformed `current.yaml`, and no longer flags valid key-based current selections for legacy design/explore items as missing.
- Single-festival `camp workitem commit` reaches the festival matrix row and emits the `FE-` segment.
- `camp workflow create --json` reports `applied` based on real mutation; idempotent no-op reruns report `applied: false`.
- Link validation no longer ages valid historical `created_at` timestamps into schema violations after 24h; it rejects only zero and beyond-tolerance future timestamps.

## Test plan

Verified at head `8f164e5`:

- [x] `go test ./...` (all packages pass)
- [x] `just lint` (`0 issues`; `lint-no-host-fs-tests` clean, no new violators)
- [x] `just build-camp` (`22.18s`)
- [x] `just test integration` (`335/335` tests passed in `490.15s`)

## Reviewer notes

- The workflow-create scaffold is intentionally small: a workflow collection root plus the standard terminal dungeon directories. It does not model intent-style status queues.
- The PR preserves the stacked commit history from CW0003 instead of collapsing the work into a single commit.
- The previous PR body described the initial sequence only; this body reflects the final aggregate branch after the stacked PRs landed.


internal/commands/workflow/doctor.go
```diff
@@ -0,0 +1,320 @@
...
+	Target      string `json:"target,omitempty"`
+	Message     string `json:"message"`
+	FixHint     string `json:"fix_hint,omitempty"`
+	AutoFixable bool   `json:"auto_fixable"`
+}
+
+// errDoctorIssuesFound is returned by runDoctor when any finding has severity
+// error. cobra surfaces this as a non-zero exit code without re-printing the
+// findings (we already emitted them).
+var errDoctorIssuesFound = camperrors.NewValidation("doctor", "doctor reported error-severity findings", nil)
```
obey-agent: **The PR description claim "Exit 2 on any error severity" is incorrect.** Doctor exits with **1**, not 2.

`errDoctorIssuesFound` is a `*ValidationError`. `cmd/camp/main.go` only special-cases `*CommandError` (which has an `ExitCode` field); everything else falls through to `os.Exit(1)`:

```go
// cmd/camp/main.go
if err := Execute(); err != nil {
    if errors.As(err, &cmdErr) && cmdErr.ExitCode != 0 {
        os.Exit(cmdErr.ExitCode)
    }
    fmt.Fprintf(os.Stderr, "Error: %v\n", err)
    os.Exit(1)
}
```

The integration test at `tests/integration/workflow_test.go:250` uses `require.Error(t, err)` which cannot distinguish exit 1 from exit 2, so this slipped through. The same bogus claim is also stated in a comment at `show.go:21` for `errWorkflowNotFound`.

Two ways to fix:

1. Return a `CommandError` from doctor when findings are error-severity:
   ```go
   var errDoctorIssuesFound = &camperrors.CommandError{
       Err:      errors.New("doctor reported error-severity findings"),
       ExitCode: 2,
   }
   ```
2. Or update the PR description, the integration tests, and the show.go comment to reflect that the exit code is actually 1.

Whichever way you go, add an integration assertion on the actual `exitCode` (extend `RunCampInDir` or use a new harness method) so this contract is locked.

internal/commands/workflow/show.go
```diff
@@ -0,0 +1,199 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+)
+
+// errWorkflowNotFound triggers exit code 2 from cobra via RunE.
+var errWorkflowNotFound = camperrors.NewValidation("type", "workflow not found", nil)
```
obey-agent: Same problem as `doctor.go:55`: this comment claims `errWorkflowNotFound` triggers exit code 2 from cobra via RunE, but `cmd/camp/main.go` returns exit 1 for any non-`CommandError`. Either rewrite this as a `CommandError` with `ExitCode: 2` or correct the comment.

```go
var errWorkflowNotFound = &camperrors.CommandError{
    Err:      errors.New("workflow not found"),
    ExitCode: 2,
}
```

If you go with the rewrite, mirror it in `doctor.go:55` so both subcommands share the same exit-2 contract, and add an integration assertion on the actual exit code.

internal/commands/workflow/shortcut_add.go
```diff
@@ -0,0 +1,124 @@
...
+		NoChanges     bool      `json:"no_changes"`
+		Applied       bool      `json:"applied"`
+	}{
+		SchemaVersion: JSONSchemaVersion,
+		GeneratedAt:   time.Now().UTC(),
+		Type:          typeName,
+		Shortcut:      key,
+		Path:          path,
+		NoChanges:     noChange,
+		Applied:       true,
```
obey-agent: `Applied` is hardcoded to `true` here, but in the no-change path at line 74 the function returns without calling `upsertShortcut` or `invalidateNavigationCache`. So when `noChange == true`, no write happens, yet the JSON contract reports `applied: true`.

This is a silent JSON contract bug: a consumer that dispatches on `applied` (e.g., a script that re-invokes `camp doctor` when changes were made) cannot tell from this output whether anything actually changed.

Fix:

```go
Applied: !noChange,
```

Compare with the `create.go:376` JSON output which correctly threads `Applied: !opts.DryRun`. The two contracts should be consistent; if `applied` means "this invocation wrote to the filesystem", then `noChange` must flip it to `false` here.

Worth adding a regression test like the existing `TestShortcutAdd_NoChange` (if it does not assert on this field already).

internal/commands/workflow/sync.go
```diff
@@ -0,0 +1,276 @@
...
+		case codeCacheStale:
+			actions = append(actions, syncAction{Finding: f, Kind: syncDeleteNavCache, Target: "cache:nav"})
+		case codeShortcutDuplicate:
+			actions = append(actions, syncAction{Finding: f, Kind: syncDeduplicateShortcut, Target: strings.TrimPrefix(f.Target, "shortcut:")})
+		}
+	}
+	return actions
+}
+
+func applySyncActions(ctx context.Context, cmd *cobra.Command, campaignRoot string, cfg *config.CampaignConfig, actions []syncAction) ([]syncAction, error) {
```
obey-agent, line 100: `applySyncActions` is **128 lines** (99-226), well over the project standard of 50 lines per function. Each action kind (`syncRemoveShortcut`, `syncRemoveConcept`, `syncAddConcept`, `syncDeleteNavCache`, `syncDeduplicateShortcut`) is a self-contained block that can be extracted.

Suggested refactor:

```go
func applySyncActions(ctx context.Context, cmd *cobra.Command, root string, cfg *config.CampaignConfig, actions []syncAction) ([]syncAction, error) {
    state := newSyncState(cfg)
    applied := make([]syncAction, 0, len(actions))

    for _, action := range actions {
        ok, err := state.apply(action, root)
        if err != nil {
            return nil, err
        }
        if ok {
            applied = append(applied, action)
        }
    }

    return state.flush(ctx, cmd, root, applied)
}
```

Where `syncState` holds the dirty flags and a per-kind dispatch table. This also makes each action kind unit-testable in isolation, which the current shape obscures.

Separately: the in-place slice filter `next := concepts[:0]` at line 130 mutates the backing array of `cfg.ConceptList`. Defensive `next := make([]config.ConceptEntry, 0, len(concepts))` is safer and trivially cheap relative to the I/O that follows.

internal/commands/workflow/sync.go
```diff
@@ -0,0 +1,276 @@
...
+			for k := range jumps.Shortcuts {
+				if nav.NormalizeNavigationName(k) == normalized {
+					keys = append(keys, k)
+				}
+			}
+			sort.Strings(keys)
+			if len(keys) <= 1 {
+				continue
+			}
+			kept := keys[0]
```
obey-agent: The dedupe heuristic silently picks the lexicographically-first key and discards the others under `--apply`. For a user with shortcuts `Plan` and `plan` (both normalizing to `plan`), this keeps `Plan` and deletes `plan` along with its config entry. The user is not warned which key won; the human output just says `fixed workflow.shortcut.duplicate (plan)`.

Two concrete improvements:

1. **Surface the choice in the human output:**
   ```go
   fmt.Fprintf(w, "  fixed %s (%s, kept %q, removed %v)\n", a.Finding.Code, a.Target, kept, removed)
   ```
2. **Prefer the already-normalized key when present**, instead of `sort.Strings(keys)[0]`. If the user has both `Plan` and `plan` and the normalized form is `plan`, keeping `plan` (the exact-normalized variant) is less surprising than keeping `Plan`.

The current heuristic is documented in the fix-hint as "manually consolidate if you want control", but with `--apply` the user does not see the consolidation choice at all. At a minimum show what was kept.

internal/commands/workflow/create.go
```diff
@@ -58,42 +118,280 @@ func runCreate(ctx context.Context, cmd *cobra.Command, opts createOptions) erro
...
-	if err := upsertShortcut(ctx, campaignRoot, cfg, shortcutKey, relPath, workflowTitle, opts.Replace); err != nil {
+	if err := upsertShortcut(ctx, campaignRoot, cfg, plan.Shortcut.Key, plan.WorkflowRel, plan.Title, plan.Shortcut.Replaced); err != nil {
 		return err
 	}
-	if err := upsertConcept(ctx, campaignRoot, cfg, opts.Type, relPath, workflowTitle, opts.Replace); err != nil {
+	if err := upsertConcept(ctx, campaignRoot, cfg, plan.Type, plan.WorkflowRel, plan.Title, plan.Concept.Replaced); err != nil {
 		return err
 	}
 	invalidateNavigationCache(cmd, campaignRoot)
+	return nil
```
obey-agent: `applyCreatePlan` runs five sequential writes (mkdir, status scaffold, OBEY.md, shortcut, concept) with no rollback. If `upsertShortcut` succeeds and `upsertConcept` fails (e.g., disk full or a permissions glitch on `campaign.yaml`), the user is left with:

- `workflow/<type>/` directory created
- shortcut written to `jumps.yaml`
- concept missing from `campaign.yaml`
- nav cache not invalidated (line 296 only runs after both succeed)

Doctor will surface this as `workflow.dir.missing-concept` on next run, and `sync --apply` can repair it. So the partial state is recoverable, which is actually the right design. But it should be called out:

1. **Document the recovery path in the function comment** so the next maintainer does not feel compelled to add transactions.
2. **Invalidate the nav cache before returning errors too**, since shortcuts/concepts may have changed even on a partial failure:
   ```go
   defer invalidateNavigationCache(cmd, campaignRoot)
   ```
   (assuming `invalidateNavigationCache` is idempotent on missing cache, which it appears to be at `nav/index.Delete`)

This is borderline-nit, but for a feature that explicitly ships a doctor + sync pair, the partial-failure contract is the kind of thing a staff review should pin down.

internal/commands/workflow/create.go
```diff
@@ -18,9 +20,24 @@ import (
 	"github.com/Obedience-Corp/camp/internal/pathsafe"
 )
 
+// JSONSchemaVersion is the contract version for `camp workflow` JSON output.
+const JSONSchemaVersion = "workflow/v1"
+
+// statusDirs are the status sub-directories scaffolded inside every workflow
+// collection. They mirror the workitem-collection layout used by
+// `.campaign/intents/` and are documented in DESIGN.md §3.1.
+var statusDirs = []string{
```
obey-agent: `internal/commands/workflow/create.go` is now **513 lines** and `internal/complete/complete.go` is **515 lines**, both just over the project standard of 500 lines per file. Not a hard blocker, but a fork worth taking now while the surface is still being shaped.

Suggested split for `create.go`:

- `create.go` keeps `newCreateCommand`, `runCreate`, the options + plan types, and the top-level orchestration.
- `create_plan.go` for `computeCreatePlan`, `planShortcut`, `planConcept`.
- `create_apply.go` for `applyCreatePlan`, `writeStatusScaffold`, `writeOBEYIfMissing`, `upsertShortcut`, `upsertConcept`, `invalidateNavigationCache`.
- `create_emit.go` for `emitCreateResult` and the human/JSON renderers.

For `complete.go`, the natural cut is to move `completeInRelativePath` + `completeInRelativePathRich` + `orderPathCandidates` into a `complete_path.go` (the file already imports a lot of `index.*` types that other functions do not need).

Same staff-level discipline that drives the 50-line function limit: small files force you to draw a contract between sections, which keeps the next change reviewable.

internal/complete/complete.go
```diff
@@ -219,18 +219,24 @@ func completeInCategory(ctx context.Context, cat nav.Category, query string) ([]
...
 	}
 
-	// Has query - filter by prefix
-	candidates := q.Complete(query, cat)
+	// Recent-first ordering: when the category resolves to a known directory
+	// under the campaign root, boost entries backed by a .workitem marker so
+	// builtin shortcuts (de, ex, ...) behave the same as custom workflow
+	// shortcuts whose path goes through completeInRelativePath.
+	if relPath := string(cat); relPath != "" {
+		candidates = applyRecentFirstOrder(filepath.Join(jumpResult.Path, relPath), candidates)
```
obey-agent, line 237: `string(cat)` here treats the underlying string of `nav.Category` as a directory path under the campaign root. This is a leaky abstraction: it works only because today every `Category` constant happens to equal its on-disk directory name relative to campaign root.

If `nav.Category` ever gains a value that does not map 1:1 to a directory (or starts using a prefix), this site will silently apply recent-first ordering against the wrong tree and the bug will look like "completion order looks weird sometimes".

Either:

1. **Add an explicit accessor** on `nav.Category`:
   ```go
   // In nav/category.go
   func (c Category) DirPath() string { return string(c) }
   ```
   Then the contract is documented and a future change has to update one place.
2. **Pass the resolved path through `nav.DirectJump`** instead of recomputing from `string(cat)`. `jumpResult.Path` is already the campaign root; the category-to-subpath mapping should live in `nav`, not be inferred at call sites.

At minimum, add a comment cross-referencing where the `Category → directory` contract is defined so the next maintainer knows what to update if categories ever diverge from directory names.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview

This is a substantial, well-scoped sequence: the doctor/sync pair gives the workflow surface real observability, the dotted-domain finding codes are a stable contract worth defending, and the recent-first refactor cleanly unifies builtin and custom completion paths. The package uses the internal `camperrors` framework consistently and threads context through every I/O call. The two things that block approval are a real exit-code bug (and a comment that lies about it in two files) and a JSON-contract bug in `shortcut add`. Both are small fixes; both will bite consumers if shipped.

## Key Findings

1. **Doctor does NOT exit 2 on error-severity findings as the PR description claims.** `errDoctorIssuesFound` is a `*ValidationError`, which `cmd/camp/main.go` exits with code 1 (only `*CommandError` carries an `ExitCode`). The integration test at `workflow_test.go:250` uses `require.Error` which masks the actual exit code. The same false claim is repeated in `show.go:21` for `errWorkflowNotFound`. (inline on `doctor.go:55` and `show.go:22`)

2. **`shortcut_add.go:119` always reports `applied: true` in JSON output**, even on the no-change path. Consumers that dispatch on `applied` cannot tell whether anything was written. (inline on `shortcut_add.go:119`)

3. **`applySyncActions` is 128 lines** vs the 50-line standard, with five distinct action kinds inlined. Extracting per-kind apply functions makes each independently testable. Also: the in-place slice filter `concepts[:0]` at line 130 mutates the cfg backing array; defensive `make` is safer. (inline on `sync.go:99`)

4. **`syncDeduplicateShortcut` silently picks the lexicographically-first key** and discards the others under `--apply`, with no indication in the human output of which key won. Surface the choice and prefer the already-normalized variant when present. (inline on `sync.go:190`)

5. **`applyCreatePlan` has no rollback for partial failures.** Doctor + sync can recover the partial state, so this is defensible, but document the recovery path and invalidate the nav cache on error too so completion does not serve stale data after a partial write. (inline on `create.go:297`)

6. **`create.go` 513 lines, `complete.go` 515 lines** both just over the 500-line standard. Worth a split now while the surface is still being shaped. (inline on `create.go:29`)

7. **`string(nav.Category)` to derive a campaign-relative directory** is a leaky abstraction at `complete.go:238`. Works today by accident of how Category constants are spelled; add an explicit accessor on the type so a future Category change does not break completion ordering silently. (inline on `complete.go:238`)

## What is Done Well

- Internal `camperrors` framework used consistently. No `fmt.Errorf` regressions in the new code.
- Context properly threaded through every I/O call (`config.LoadCampaignConfigFromCwd(ctx)`, `SaveJumpsConfig(ctx, ...)`, etc.).
- Dotted-domain finding codes (`workflow.shortcut.missing-target`, `workflow.cache.stale`, ...) are a genuinely good stable contract for tooling.
- Dry-run is the default for `sync`; `--apply` is the explicit opt-in. Correct conservative default.
- `--json` output across every subcommand with `schema_version` and `generated_at`. The schemas are versioned at `JSONSchemaVersion = "workflow/v1"` which gives consumers a real migration story.
- Recent-first ordering factored into `internal/complete/order.go` and shared by both builtin (`completeInCategory`) and custom (`completeInRelativePath`) paths. Benchmark documented at 22ms / 5000 entries, well under the 200ms shell-completion budget.
- `DESIGN.md` (570 lines) including §8 open questions is exemplary scope discipline; future reviewers can see what was deferred and why.
- Snapshot test (`testdata/create_research_human.golden`) and container integration tests cover both the golden-path human output and the doctor + sync round-trip.
- Builtin types (`intent`, `design`, `explore`, `festival`, plus the `code_reviews`/`pipelines` auto-source defaults) are correctly excluded from user-workflow enumeration with a clear comment explaining why.
- `applyRecentFirstOrder` returns input unchanged when no entries have `.workitem` markers, preserving existing ordering for directories that have not yet been touched by workitem creation.

## Staff Standard

Would I submit this as my best work at a staff level? Mostly yes. The architectural moves (factored ordering, dotted finding codes, dry-run-default sync, versioned JSON) are the right shape, and the test discipline is strong. The exit-code lie is what would catch me in a Meta interview: it is documented behavior that does not match the code, and the integration test was not strong enough to surface the gap. That, plus the JSON `applied: true` bug, are the kind of contract surprises that look like deliberate misrepresentation in a postmortem even when they are just oversights. Address the 8 inline comments, especially #1 and #2, and this is a clean approve.

obey-agent (approved):
## Verdict: Approve

## Overview

Every blocker from the previous round is addressed cleanly: the exit-code routing uses `camperrors.NewCommand`, the `Applied` JSON contract is honest, `applySyncActions` decomposed into a `syncState` with per-kind methods, the dedupe heuristic now prefers normalized keys and surfaces the choice, `applyCreatePlan` documents and invalidates the cache on every exit, the file size violations are split clean, and `string(nav.Category)` is now a typed `Category.Dir()` accessor.

## Findings Resolved

1. **Doctor + show exit code now actually 2.** `errDoctorIssuesFound` and `errWorkflowNotFound` are `camperrors.NewCommand(..., 2, ...)` instead of `NewValidation`. `cmd/camp/main.go` already maps `CommandError.ExitCode` correctly, so the contract is now end-to-end honest. The `NewCommand` helper at `internal/errors/errors.go:296` is the right primitive.

2. **`shortcut_add.go:119 Applied: !noChange`** — single-line fix, JSON contract no longer lies on the no-change path.

3. **`applySyncActions` decomposed into `syncState` with per-kind apply methods** (`removeShortcut`, `removeConcept`, `addConcept`, `deduplicateShortcut`). Each is small and testable; the outer dispatch is a clean switch. The in-place slice mutation (`concepts[:0]`) is also gone — `make([]config.ConceptEntry, 0, len(concepts))` everywhere.

4. **`deduplicateShortcut` prefers the already-normalized key** when present (`keptEntry, hasNormalized := s.jumps.Shortcuts[normalized]`), falls back to lexicographic only when no normalized variant exists. Records `Kept` and `Removed` on the action; `syncActionDetail` surfaces the choice in human output (`fixed workflow.shortcut.duplicate (plan; kept plan; removed Plan)`). Both the heuristic improvement AND the operator-visibility ask landed.

5. **`applyCreatePlan` uses `defer invalidateNavigationCache(cmd, campaignRoot)`** so cache invalidation runs on every exit (success or failure). The function comment now documents the non-transactional contract explicitly: "if a later write fails, earlier filesystem/config writes remain". Doctor + sync will repair the partial state.

6. **`create.go` (513 → 133 lines)** split into `create.go` + `create_plan.go` (161) + `create_apply.go` (166) + `create_emit.go` (86). Each file matches a single concern. `complete.go` (515 → 286 lines) split into `complete.go` + `complete_path.go` (239). All files now under 500.

7. **`Category.Dir()` is the typed accessor.** Implementation at `internal/nav/shortcuts.go:97-99` and the call site in `complete.go:236` use it. No more `string(cat)` for path derivation.

- The `syncState` refactor is the right abstraction, not just a function split — `apply` orchestrates, `applyAction` dispatches, `save` and `invalidateCacheIfNeeded` finalize. Each per-kind method (`removeShortcut`, `removeConcept`, etc.) returns `bool` for "did this change state" which makes the dirty-flag tracking clean.
- The dedupe fix went beyond what the comment asked for: not just surfacing the choice, but actually changing the heuristic to prefer the normalized form. Smaller surprise surface.
- The `defer invalidateNavigationCache` pattern in `applyCreatePlan` is the right shape — it guarantees cache freshness even on partial failure, which makes the documented "doctor + sync repair" contract actually load-bearing.
- File splits respect natural concern boundaries (plan / apply / emit, complete / complete_path) rather than arbitrary chops.

## Verification

- `just test unit` — 2411/2411 passed
- `just test integration-run TestIntegration_Workflow` — 7/7 passed (`StatusDirs`, `Idempotent`, `List`, `Show`, `ShortcutAdd`, `DoctorAndSync`, `CreateDryRun`)

## Staff Standard

Would I submit this as my best work at a staff level? Yes. The exit-code routing now uses the right primitive end-to-end, the sync refactor is well-shaped, the file splits track real concern boundaries, and the test suite is clean. Ready to merge.

obey-agent (commented):
## Verdict: Approve (already approved; current open-state check)

## Overview
I rechecked the current open state of the base workflow-surface PR. It remains the integrated trunk for the stacked CW0003 follow-up PRs and already has an `obey-agent` approval after the earlier blockers were resolved.

## Key Findings
- No new findings from this pass.
- The stacked follow-up PRs (#308-#313) carry the active review feedback for the remaining hardening work.

## What's Done Well
- The workflow surface is well documented, tested, and already went through a request-changes → approve cycle.

## Staff Standard
Still comfortable with the existing approval for this base PR.

internal/commands/workitem/COMMIT_DESIGN.md
```diff
@@ -0,0 +1,251 @@
+# camp workitem commit — Staging Plan Design
```
obey-agent: Please remove this design document from the implementation package before merge. The repo can keep durable CLI behavior documentation under docs/, but internal/commands/workitem should contain code, tests, and necessary testdata rather than festival planning contracts. Otherwise the PR adds a second source of truth for commit behavior that will drift from the implementation.

internal/commands/workflow/DESIGN.md
```diff
@@ -0,0 +1,570 @@
+# `camp workflow` command surface — design
```
obey-agent: This internal design contract should not ship inside the implementation package. Please remove it from the PR, and move any durable user/operator-facing material into the public docs surface instead. Leaving large planning/audit docs under internal/ makes the production source tree carry festival scratch context and creates stale secondary contracts that future maintainers may treat as authoritative.

internal/workitem/links/SCHEMA.md
```diff
@@ -0,0 +1,375 @@
+# Workitem link registry — schema
```
obey-agent: This schema contract should not live as a standalone Markdown artifact under internal/. Please remove it from the PR and migrate any durable schema information that users or maintainers need into docs/ or code-level tests/comments. Keeping this file here makes the implementation package carry a static contract document that can diverge from the Go types, validators, and test fixtures.

obey-agent (changes requested):
Verdict: Request Changes

The PR still includes internal design/contract Markdown artifacts under implementation packages. Those documents look like festival planning and static contracts, not production source files, and they should not ship in internal/ alongside the code.

Required cleanup:
- Remove internal/commands/workflow/DESIGN.md.
- Remove internal/commands/workitem/COMMIT_DESIGN.md.
- Remove internal/workitem/links/SCHEMA.md.
- Move any durable user/operator-facing material that should survive into the docs/ surface, or encode implementation contracts in Go types, tests, and narrow code comments.

Staff standard: I would not merge this aggregate PR while those planning documents remain in the implementation tree, because they create stale parallel contracts and add non-production festival context to the source package layout.

internal/commands/workflow/create.go
```diff
@@ -2,25 +2,32 @@ package workflow
...
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
-	"github.com/Obedience-Corp/camp/internal/nav"
-	navindex "github.com/Obedience-Corp/camp/internal/nav/index"
 	"github.com/Obedience-Corp/camp/internal/pathsafe"
 )
 
+// JSONSchemaVersion is the contract version for `camp workflow` JSON output.
+const JSONSchemaVersion = "workflow/v1"
+
+// statusDirs are the status sub-directories scaffolded inside every workflow
```
obey-agent: Blocking: this scaffold intentionally mirrors `.campaign/intents/`, but that is not the intended model for custom workflows here. The current `workflow/design` and `workflow/explore` layouts are lightweight and use the standardized dungeon format; adding top-level `inbox`, `active`, and `ready` buckets dramatically expands the scope and changes the workflow model. Please do not introduce intent-style buckets for custom workflows in this PR.

internal/commands/workflow/workflow.go
```diff
@@ -14,5 +14,10 @@ navigation config and workitem type support.`,
 	}
 
 	cmd.AddCommand(newCreateCommand())
```
obey-agent, line 16: Blocking: registering `camp workflow create` makes camp a second custom-workflow authoring surface. That is beyond the CW0003 scope as I understand it. The requested scope was workitem linking and commit tracking, similar to how `fest link` associates a festival with a working directory. Custom workflow definitions are already a Festival/fest concern via `WORKFLOW.md`; adding a camp-side custom workflow creator in the bundled product creates two competing ways for users to create custom workflows. Please remove or defer this creation surface unless the product direction is explicitly changed.

internal/commands/workflow/create_plan.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	if _, err := os.Stat(absPath); err != nil {
+		if !os.IsNotExist(err) {
+			return nil, camperrors.Wrap(err, "stat workflow directory")
+		}
+		plan.WorkflowDirCreate = true
+	}
+
+	for _, sub := range statusDirs {
```
obey-agent: Blocking: this loop materializes the expanded status-bucket model on disk for every custom workflow. That is the concrete behavior that turns a light workitem-linking/commit-tracking festival into a new workflow-management system. Please remove this scaffolded status directory creation from CW0003; if custom workflow creation is needed later, it should be reconciled with the existing Festival/fest `WORKFLOW.md` model first.

obey-agent (changes requested):
Verdict: Request Changes

Additional blocker: the workflow creation surface appears to have expanded beyond the intended CW0003 scope. The festival goal was workitem linking and workitem-aware commit tracking, similar in spirit to linking a festival to a working directory; it was not to introduce a second custom-workflow authoring system in camp.

Specific concerns are inline:
- `camp workflow create` makes camp a custom workflow creation surface even though Festival/fest already owns custom workflows through `WORKFLOW.md`.
- The scaffold mirrors `.campaign/intents/` with `inbox`, `active`, and `ready` buckets, which does not match the current lightweight `workflow/design` and `workflow/explore` shape.
- Materializing those buckets on disk turns this into a broader workflow-management change and should be removed or deferred until the camp/fest product boundary is explicitly reconciled.

This should block PR #302 in addition to the internal Markdown cleanup.

internal/commands/workflow/create.go
obey-agent: More precise requirement: do not mirror `.campaign/intents/` here. For camp workflow collections, remove the top-level `inbox`, `active`, and `ready` scaffold. Keep the lightweight collection root plus the standard dungeon subtree used by existing `workflow/design` and `workflow/explore` (`dungeon/completed`, `dungeon/archived`, `dungeon/someday`). Top-level `workflow/<type>/<item>` entries should remain valid and ergonomic.

internal/commands/workflow/workflow.go
obey-agent: Correction after product-boundary clarification: the existence of `camp workflow create` is not itself the blocker. Camp does need this light workflow-collection creation path for navigation/custom workitem types, separate from Fest custom `WORKFLOW.md` definitions. The blocker is the expanded status model: this camp command should mirror `workflow/design` and `workflow/explore`, where top-level items remain ergonomic active/ready work and only the standard `dungeon/{completed,archived,someday}` subtree is scaffolded.

internal/commands/workflow/create_plan.go
obey-agent: Narrowing this blocker: the loop should not create `inbox`, `active`, or `ready`. It should create only the standard dungeon directories for a camp workflow collection. Tests, golden output, JSON fields, doctor/show/list scaffold checks, and docs in this PR should be updated to stop treating the intent-style active buckets as the contract.

obey-agent (changes requested):
Verdict: Request Changes

Correction/narrowing of the previous workflow-surface blocker: `camp workflow create` itself is valid. Camp needs this lightweight workflow-collection creation path for navigation and custom workitem types, separate from Fest custom `WORKFLOW.md` workflows.

The blocker is the scaffold model in PR #302. Camp workflow collections should mirror the existing `workflow/design` and `workflow/explore` ergonomics: top-level `workflow/<type>/<item>` entries are valid active/ready work, and the command should only create the standard dungeon subtree (`dungeon/completed`, `dungeon/archived`, `dungeon/someday`). It should not create intent-style `inbox`, `active`, or `ready` top-level buckets.

Please update the code, tests, golden output, JSON shape/checks, and any docs/design artifacts in this PR to reflect that narrower contract. The internal Markdown artifacts still need to be removed or moved as previously noted.

internal/commands/workflow/create.go
```diff
@@ -2,44 +2,56 @@ package workflow
...
+// collection. They mirror the workitem-collection layout used by
+// `.campaign/intents/` and are documented in DESIGN.md §3.1.
+var statusDirs = []string{
```
obey-agent: Blocking from the latest PR #302 review is still present on the current head: this still mirrors the intent-style scaffold by creating top-level `inbox`, `active`, and `ready` buckets. The narrowed product requirement was that camp workflow collections mirror `workflow/design` / `workflow/explore`: top-level `workflow/<type>/<item>` entries remain valid, and creation only scaffolds the standard dungeon subtree (`dungeon/completed`, `dungeon/archived`, `dungeon/someday`). Please remove these three buckets and update `create_plan`, dry-run/human/JSON output, tests, goldens, and docs accordingly.

internal/commands/workflow/DESIGN.md
obey-agent: Still blocking on the current head: this internal design/audit Markdown artifact remains in an implementation package. Please remove it from `internal/`; durable user/operator-facing material belongs under `docs/`, and implementation contracts should live in Go types, tests, and narrow code comments.

internal/commands/workitem/COMMIT_DESIGN.md
obey-agent: Still blocking on the current head: this design contract remains under `internal/commands/workitem`. Please remove it from the implementation package and move any durable reference material into `docs/` or encode it in tests/comments close to the code.

internal/workitem/links/SCHEMA.md
obey-agent: Still blocking on the current head: this standalone schema contract remains under `internal/workitem/links`. The source package should not carry a parallel Markdown schema that can drift from the Go structs, validators, and tests; move durable schema/reference content to `docs/` or keep the contract in code/tests.

internal/commands/workflow/create_emit.go
```diff
@@ -0,0 +1,155 @@
...
+		Title:         plan.Title,
+		WorkflowDir:   plan.WorkflowRel,
+		StatusDirs:    statusDirsForOutput(),
+		OBEYWritten:   plan.OBEYWrite && !opts.DryRun,
+		Shortcut:      plan.Shortcut,
+		Concept:       plan.Concept,
+		Replaced:      append([]string(nil), plan.Replaced...),
+		NoChanges:     plan.NoChanges,
+		DryRun:        opts.DryRun,
+		Applied:       !opts.DryRun,
```
obey-agent: `runCreate` skips `applyCreatePlan` whenever `plan.NoChanges` is true, but this JSON contract reports `applied: true` for an idempotent no-op rerun because it only checks `!opts.DryRun`. That is the same class of automation-contract bug we fixed for `workflow shortcut add`: consumers cannot tell whether the invocation actually wrote anything. Please make this reflect real mutation, e.g. `Applied: !opts.DryRun && !plan.NoChanges`, and update `TestRunCreateJSONNoChanges` plus docs that currently disagree about this field.

internal/workitem/links/validate.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	if link.CreatedAt.IsZero() {
+		addErr("created_at", "required")
+	} else {
+		skew := now.Sub(link.CreatedAt)
+		if skew < 0 {
+			skew = -skew
+		}
+		if skew > maxClockSkewReject {
```
obey-agent: This turns `created_at` into a one-day TTL. Because `skew` is absolute, every otherwise-valid persisted link becomes invalid more than 24h after it was created, and `camp workitem doctor` will start reporting `workitem.schema.violation` on healthy registries. Historical link timestamps should remain valid; only zero values and probably future timestamps beyond a tolerance should be rejected. Please remove the past-age rejection and add a regression test with an old-but-valid `created_at`.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview
I re-reviewed the current aggregate head after the stacked PRs were merged into PR #302, and I started from the latest `obey-agent` request-changes comments because they identify the active blockers. The workitem-linking, commit-routing, JSON-envelope, and hardening slices are much stronger now; `go test ./...`, the focused integration sweep, and `just build-camp` all pass. The PR still should not merge because the latest workflow-surface blockers remain in the current head, and the aggregate pass found two additional contract issues.

## Key Findings
- The narrowed scaffold blocker is still present: `camp workflow create` still scaffolds top-level `inbox`, `active`, and `ready` buckets. The current contract should mirror `workflow/design` and `workflow/explore`: top-level `workflow/<type>/<item>` workitems, plus only the standard `dungeon/{completed,archived,someday}` subtree. Inline: `internal/commands/workflow/create.go`.
- The internal Markdown artifacts called out in the latest review are still present under implementation packages: `internal/commands/workflow/DESIGN.md`, `internal/commands/workitem/COMMIT_DESIGN.md`, and `internal/workitem/links/SCHEMA.md`. Inline on each file.
- `camp workflow create --json` reports `applied: true` on idempotent no-op reruns, even though `runCreate` skips `applyCreatePlan` when `plan.NoChanges` is true. Inline: `internal/commands/workflow/create_emit.go`.
- Link validation treats `created_at` as expiring after 24h because it rejects absolute clock skew in either direction. That makes healthy persisted links age into `workitem.schema.violation`. Inline: `internal/workitem/links/validate.go`.

## What’s Done Well
- The previously flagged PR #313 `--json=false` bug is fixed in `jsoncontract.Requested`, with unit and integration coverage.
- The commit-path safety fixes are present: dirty-index refusal, error wrapping, and SHA warning behavior are integrated.
- The links registry now uses locked load/mutate/save and ID collision retry; the aggregate integration sweep covers the key workitem and workflow flows.
- I kept durable local review notes at `workflow/code_reviews/camp-pr-302-rereview-2026-05-26/notes.md` in the campaign repo.

## Verification
- `go test ./internal/commands/workflow ./internal/jsoncontract ./internal/commands/workitem`
- `go test ./...`
- `just test integration-run 'TestIntegration_(Workflow|Workitem|CommitTags|JSON|Repair)'`
- `just build-camp`

## Staff Standard
Not ready to merge. The implementation is close, but the latest product-scope blockers still exist in code/tests/docs, and the `created_at` aging bug would turn valid user data into doctor errors over time. Fix those and this aggregate PR should be ready for another focused pass.

obey-agent (changes requested):
## Verdict: Request Changes

Do not approve PR #302 yet. The aggregate branch now contains the full CW0003 stack, but the current head still ships product-boundary drift and stale review/planning artifacts that would become confusing public behavior if merged to `main`.

Required before approval:

- Narrow `camp workflow create` to the intended camp workflow-collection contract. The command itself is valid, but it should mirror `workflow/design` and `workflow/explore`: top-level `workflow/<type>/<item>` work remains normal active/ready work, and creation should scaffold only `dungeon/completed`, `dungeon/archived`, and `dungeon/someday`. It should not create intent-style `inbox`, `active`, or `ready` buckets. Update code, JSON output, human output, tests, goldens, generated docs, and `docs/workflow-reference.md`.
- Remove the internal Markdown artifacts from implementation packages: `internal/commands/workflow/DESIGN.md`, `internal/commands/workitem/COMMIT_DESIGN.md`, and `internal/workitem/links/SCHEMA.md`. Move durable material into public docs or encode it in Go types, tests, and narrow comments.
- Remove all public/help/code references to those artifact files. `camp workitem commit --help` currently points users at `internal/commands/workitem/COMMIT_DESIGN.md`, and generated CLI docs inherit that reference.
- Fix `camp workflow create --json` so `applied` reflects real mutation. Idempotent no-op reruns should not report `applied: true` when `runCreate` skipped `applyCreatePlan`.
- Fix link validation so old valid `created_at` timestamps do not age into schema violations after 24 hours. Reject zero timestamps and future timestamps beyond tolerance; do not reject historical links merely because they are old.
- Reconcile stale public docs. Several reference docs still present earlier CW0003 review findings as open even though the stacked PRs fixed them. The final docs should describe the aggregate branch behavior, not the intermediate review state.
- Rewrite the PR description. It still describes PR #302 as the original workflow slice, claims `inbox/active/ready` scaffolding as a feature, and points reviewers to internal design markdown. It should describe the final aggregate CW0003 branch and its camp/fest boundary clearly.

This review intentionally incorporates the prior `obey-agent` request-changes reviews, the current aggregate re-review, and Lance's product clarification: camp may create lightweight workflow collections for navigation/custom workitem types, but Fest/festival owns methodology `WORKFLOW.md` workflows and camp should not introduce a second intent-like lifecycle model.

A festival implementation phase is being added to track these fixes explicitly. Approval should wait until that phase is implemented, tested, pushed to `feat/workflow-surface-CW0003`, and the PR description is updated.

obey-agent:
## PR 302 blocker resolution note

This is a resolution summary for the fixed aggregate head `7e8c35f1f41445669ac5cf40ff3bb4c23cac19b9`. It is not an approval; it is meant to prepare the final obey-agent re-review.

### Workflow scaffold contract

The product-boundary blocker has been narrowed and implemented. `camp workflow create` remains in scope as a lightweight camp workflow-collection creator for navigation and custom workitem types, but it no longer creates intent-style live status queues. The scaffold now creates the workflow root and standard terminal dungeon subtree only. Code, JSON output, human output, goldens, integration tests, generated CLI docs, and `docs/workflow-reference.md` were updated to match that contract.

The earlier broader concern that `camp workflow create` itself was out of scope is superseded by Lance's clarification and the later review correction. The active blocker was the scaffold model, not the command's existence.

### Internal markdown artifact cleanup

The implementation-package markdown artifacts were removed from the PR: workflow design notes, workitem commit design notes, and the link schema document no longer live under `internal/`. Durable user/operator material was moved into public docs or encoded in Go types, tests, and narrow comments. Public/help/generated docs were reconciled so they no longer point users at those internal artifact paths.

### JSON `applied` fix

`camp workflow create --json` now reports mutation truthfully. Idempotent no-op reruns return `applied: false` instead of claiming a write occurred when `runCreate` skipped apply.

### Link `created_at` validation fix

Link validation no longer treats old valid `created_at` values as expired after 24 hours. It rejects missing/zero timestamps and far-future timestamps beyond tolerance, but persisted historical links no longer age into schema violations.

### Stale docs and PR body reconciliation

The stale public docs that described intermediate review findings as open were reconciled with the final aggregate behavior. The PR title/body were rewritten to describe the full CW0003 aggregate branch, including workflow collections, workitem linking/current/resolve/doctor, scoped commits, commit queries, JSON contracts, docs, tests, and the camp/fest boundary.

### Review-thread audit map

The local audit map is `festivals/active/camp-workitem-linking-and-commits-CW0003/005_FIX_PR_302_AGGREGATE/review/pr302_review_thread_map.md`.

Disposition summary:

- 8 older workflow-surface threads were already fixed before this cleanup phase.
- 11 active blocker threads were mapped to Phase 005 implementation tasks and fixed.
- 1 thread was superseded: the broader claim that `camp workflow create` itself was out of scope.
- 0 threads were waived.

### Tests and gates run

- `go test ./internal/commands/workflow ./internal/commands/workitem ./internal/complete`
- `golangci-lint run --new-from-merge-base origin/main ./... --max-issues-per-linter=0 --max-same-issues=0` (`0 issues`)
- `just lint` (`0 issues`; `lint-no-host-fs-tests` clean with no new violators)
- `just build-camp` (`21.19s`)
- `go test ./...`
- `just test integration` (`333/333` tests passed in `501.49s`)

### Remaining waivers

None. The known blockers from the latest request-changes review were either fixed, already fixed, or superseded as described above.

obey-agent (approved):
## Verdict: Approve

I re-reviewed PR #302 at head `7e8c35f1f41445669ac5cf40ff3bb4c23cac19b9`. The previously blocking aggregate issues are resolved on the branch: `camp workflow create` now uses the lightweight dungeon-only workflow-collection scaffold, internal implementation-package markdown artifacts are gone, public/help references no longer point at those artifacts, workflow-create JSON `applied` reflects actual mutation, and link `created_at` validation no longer expires historical links.

I also confirmed the updated PR description and the obey-agent resolution comment are present. The local closure map records zero waived blockers, with the broader "workflow create itself is out of scope" concern superseded by Lance's clarified camp/fest boundary.

Review checks run in this pass:

- `find internal/commands/workflow internal/commands/workitem internal/workitem/links -name *.md -print` returned no internal markdown artifacts.
- Artifact-reference grep found no `DESIGN.md`, `COMMIT_DESIGN.md`, or `SCHEMA.md` references in `internal`, `docs`, `pkg`, `cmd`, or `tests`.
- Code/docs spot checks covered `create.go`, `create_plan.go`, `create_emit.go`, `validate.go`, `docs/workflow-reference.md`, `docs/workitem-link-reference.md`, `docs/workitem-commit-reference.md`, and `docs/json-contracts.md`.
- `go test ./internal/commands/workflow ./internal/workitem/links ./internal/jsoncontract ./internal/commands/workitem` passed.

The earlier full gate run is also recorded on this head: `just lint`, `just build-camp`, `go test ./...`, and `just test integration` all passed after the lint-gate fix.

Staff standard: I would approve this as ready to merge to `main`. The remaining documented hardening items are follow-up planning inputs, not blockers for this PR.

internal/commands/workitem/link.go
```diff
@@ -0,0 +1,334 @@
...
+			WorkitemKey: workitemKeyForLink(wi),
+			Scope:       *scope,
+			Role:        role,
+			CreatedAt:   time.Now().UTC().Truncate(time.Second),
+			CreatedBy:   defaultCreatedBy(),
+		}
+		if err := registry.AddLink(link, opts.Replace); err != nil {
+			return err
+		}
+		knownIDs := workitemIDSetFromMaybeNil(wi, opts.AllowMissing)
```
obey-agent: Blocking: this validates the whole registry against a WorkitemIDs set containing only the workitem being added. If links.yaml already has a valid link for workitem A, adding a link for workitem B fails with `no workitem with id A found on disk` (I reproduced this with two design workitems linked to two projects). Build the full on-disk workitem ID set here, or skip existence validation for pre-existing links and let doctor own that check, before calling `links.Validate`; otherwise the registry cannot represent multiple workitems.

internal/commands/workitem/staging.go
```diff
@@ -0,0 +1,231 @@
...
+// ComputePlan resolves a workitem from the current context, branches on the
+// matrix row that applies, and returns a StagingPlan that the commit runner
+// can hand to commit.Workitem. Refusal modes (no workitem, empty plan with no
+// override) surface as typed errors so the CLI can map them to exit codes.
+func ComputePlan(ctx context.Context, campaignRoot string, opts PlanOptions) (*StagingPlan, error) {
+	if campaignRoot == "" {
+		return nil, camperrors.NewValidation("root", "campaign root required", nil)
+	}
+
+	res, err := resolver.Resolve(ctx, campaignRoot, resolver.Options{
```
obey-agent, line 99: Major: the festival staging row is unreachable from `camp workitem commit`. `ComputePlan` never accepts or passes a `FestivalID` into `resolver.Options`, and the CLI has no `--festival` flag, so `res.Source == SourceFestival` cannot happen here. I reproduced this by linking a workitem with `camp workitem link timeline --festival CT0001`, then running `camp workitem commit --dry-run --json` from `festivals/active/CT0001`; it returns no workitem context. Thread the festival id through the planner/CLI, or remove the documented festival matrix row. When this path is wired, also include the FE segment in the plan and actual commit tag; line 116 currently hard-codes the festival ref argument to empty.

internal/commands/workitem/doctor.go
```diff
@@ -0,0 +1,428 @@
...
+				Severity:    docSeverityWarning,
+				Target:      "workitem:" + rel,
+				Message:     "workitem at " + rel + " is missing the ref field added in v1alpha6",
+				FixHint:     "run camp workitem doctor --fix to backfill",
+				AutoFixable: true,
+			})
+		}
+	}
+
+	cur, err := links.LoadCurrent(ctx, root)
```
obey-agent, line 269: Major: this silently drops `LoadCurrent` errors, so a malformed or wrong-version `current.yaml` produces `doctor: 0 findings` while `camp workitem current` and the resolver fail on the same file. Surface this as a finding (and make `--fix` remove/quarantine it) to match the documented "Schema violations in links.yaml or current.yaml" doctor contract and give users a recovery path.

internal/commands/workitem/doctor.go
```diff
@@ -0,0 +1,428 @@
...
+	if err == nil && cur != nil {
+		if _, known := knownIDs[cur.WorkitemID]; !known {
```
obey-agent, line 280: Major: this uses a stable-ID-only set to validate `current.yaml`, but `camp workitem current` deliberately stores `wi.Key` when a resolved item has no stable ID. That makes key-based current selections for legacy design/explore workitems resolve successfully, then doctor reports `workitem.current.missing` and `--fix` removes them. Validate current selections through the same selector/resolver path used by `camp workitem current`/`resolve`, or include both stable IDs and keys in the known set.

obey-agent (commented):
**Verdict: Request Changes**

I did an independent re-review of PR #302 at `7e8c35f` against the prior CW0003 review notes. The earlier blockers around internal markdown artifacts, workflow scaffold shape, idempotent `workflow create --json`, and old `created_at` links are resolved, and the focused/full Go gates plus targeted integration suite pass.

**Key findings**
- Blocking: `camp workitem link` validates the registry against only the newly linked workitem ID, so a registry with links for workitem A cannot add a link for workitem B. See inline on `internal/commands/workitem/link.go`.
- Major: the documented festival row for `camp workitem commit` is unreachable because no festival id reaches the resolver, and the FE tag segment is hard-coded empty. See inline on `internal/commands/workitem/staging.go`.
- Major: `camp workitem doctor` drops malformed `current.yaml` errors, causing doctor to report clean while `current`/resolver fail. See inline on `internal/commands/workitem/doctor.go`.
- Major: doctor validates `current.yaml` against stable IDs only, so valid key-based current selections for legacy design/explore items are reported as missing and auto-fixable. See inline on `internal/commands/workitem/doctor.go`.

**What's done well**
The workflow create remediation is solid, link writes now use a load/mutate/save lock, JSON envelopes are broadly consistent, and the scoped commit path has good protections against dirty-index and silent-widen failures.

**Verification**
Passed: focused Go package tests, `go test ./...`, `just build-camp`, `just lint`, `just lint-no-host-fs-tests`, and targeted integration for workflow/workitem/link/commit/commit-query paths. I did not run the full `just test integration` suite.

**Staff standard**
I would not merge this yet. The multi-workitem link blocker breaks a central registry use case, and the current/festival gaps need either implementation or explicit deferral before the PR is production-ready.

obey-agent (changes requested):
Requesting changes based on the inline findings and full review summary I just posted. The multi-workitem link failure is a blocking registry correctness issue, and the festival commit plus `current.yaml` doctor gaps need implementation or explicit deferral before merge.

internal/commands/workitem/staging_branches.go
```diff
@@ -0,0 +1,311 @@
...
+	plan.Skip = append(plan.Skip, addSkip...)
+	plan.Stage = applyExcludes(added, opts.Excludes, &plan.Skip)
+	plan.Stage = dedupeSorted(plan.Stage)
+	return plan, nil
+}
+
+func computeFestivalPlan(ctx context.Context, root string, opts PlanOptions, plan *StagingPlan) (*StagingPlan, error) {
+	plan.RepoRoot = root
+	plan.Context = PlanContextFestival
+	scope := primaryFestivalScopePath(ctx, root, plan.Workitem)
```
obey-agent: Major: this re-queries the registry by workitem only and then stages the first primary festival link, not necessarily the festival link that `SourceFestival` just matched. I reproduced this with one workitem linked to both `festivals/active/AA0001` and `festivals/active/ZZ0002`: running `camp workitem commit --dry-run --json` from `ZZ0002` emitted `festival_ref: ZZ0002` / `FE-ZZ0002`, but staged `festivals/active/AA0001/FESTIVAL_GOAL.md`. Carry the matched festival scope from the resolver, or select the festival scope by the resolved festival id here, so the tag and staged paths cannot point at different festivals.

obey-agent (commented):
**Verdict: Request Changes**

I re-reviewed PR #302 at `8a7280c` after the latest fix commit. The previously filed blockers are fixed: multi-workitem link registries work, malformed `current.yaml` is reported/fixable by doctor, key-based current selections are no longer false positives, and single-festival `camp workitem commit` now reaches the festival matrix row and emits an `FE-` segment.

One remaining correctness issue should block merge: the festival staging path can diverge from the festival that was resolved. If the same workitem has primary links to two festival scopes, committing from the second festival emits the second festival ref in the tag but stages files from the first festival link. See the inline comment on `internal/commands/workitem/staging_branches.go`.

**Verification**
Passed: focused workitem/resolver/link/commit tests, `go test ./...`, `just build-camp`, `just lint`, and the targeted integration sweep for workflow/workitem/link/commit paths. Manual smoke checks also verified the four previous blockers and reproduced the remaining multi-festival staging mismatch.

**Staff standard**
Not ready to merge until the matched festival scope is carried through staging, or `computeFestivalPlan` selects the scope by the resolved festival id so the staged files and `FE-` tag always refer to the same festival.

obey-agent (changes requested):
Requesting changes based on the inline multi-festival staging finding. The previous blockers are fixed on `8a7280c`, but `camp workitem commit` can currently tag one festival while staging another when a workitem has multiple primary festival links.

lancekrogers:
Addressed the multi-festival staging mismatch at head `8f164e5`.

Root cause: `computeFestivalPlan` selected the first primary festival link for the workitem regardless of which festival the resolver matched, so committing from a second linked festival emitted that festival's `FE-` tag while staging the first festival's files.

Fix:
- Exported `resolver.FestivalScopeMatches` so resolution and staging share one matching rule.
- `computeFestivalPlan` / `primaryFestivalScopePath` now take the resolved festival id and select the festival scope via `FestivalScopeMatches`, so the `FE-` tag and the staged paths always refer to the same festival. When the id is set, only a matching link is staged (never a different festival's files).
- Extracted the pure `selectPrimaryFestivalScope` helper.

Tests:
- Unit `TestSelectPrimaryFestivalScope_*` — the "second festival" case is the direct regression guard (old code returned the first link).
- Integration `TestIntegration_WorkitemCommit_MultiFestivalStagesResolvedScope` reproduces the AA0001/ZZ0002 case: committing from ZZ0002 stages `festivals/active/ZZ0002/...` and never AA0001.

Gates: `go test ./...`, `just build-camp`, `just lint` (0 issues, no new host-fs violators), and the workitem-commit integration sweep all pass.

obey-agent (approved):
## Verdict: Approve

## Overview
I re-reviewed PR #302 at `8f164e5` with focus on the latest multi-festival staging blocker. The previous `8a7280c` fixes are still present, and the new patch correctly prevents `camp workitem commit` from emitting one `FE-` tag while staging a different festival's files.

## Key Findings
- No blocking findings on the current head.
- The latest change threads the resolved festival id into `computeFestivalPlan` and uses the same `FestivalScopeMatches` rule as the resolver, which closes the mismatch between resolver attribution and staged paths.
- The new unit coverage exercises selecting AA0001 vs ZZ0002 and the unmatched case. The new integration test directly covers the reproduced failure shape: one workitem linked to two festivals, commit from the second festival, and assert only the second festival is staged.

## What's Done Well
- Good integration over duplication: the resolver and staging code now share the same festival matching contract instead of maintaining parallel matching logic.
- The fix is narrowly scoped to the staging decision and keeps existing fallback behavior when no festival id is present.
- The test coverage is aligned with the regression rather than just asserting an implementation detail.

## Verification
- `go test ./internal/commands/workitem`
- `go test ./...`
- `just lint`

I attempted `just test integration-run 'TestIntegration_WorkitemCommit_FestivalContextFromCwd'`, but Docker/Colima was unavailable locally, so container integration did not start. The branch includes the targeted integration test for the multi-festival regression.

## Staff Standard
I would be comfortable merging this current head. The latest blocker is addressed in the right abstraction layer and I do not see a remaining production-readiness issue in this slice.

PR #303, opened 2026-05-25
feat(workitem): workitem link registry — links/current/resolve/doctor (CW0003 slice 2)

lancekrogers:
## Summary

Sequence 02 of the CW0003 festival. Ships the v1alpha1 workitem link registry and the user-facing command surface that consumes it. Branch-of-origin/main, independent of sequence 01 (PR #302).

### What lands

- **`internal/workitem/links/SCHEMA.md`** — frozen v1alpha1 contract. `links.yaml` and `current.yaml` schemas, scope kinds, roles, six-tier resolution priority, safety bounds, validation rules, link ID format, migration policy, and the quest cross-reference. Two example fixtures under `testdata/`.
- **`internal/workitem/links/`** — typed registry with `Load`, `Save`, `LoadCurrent`, `SaveCurrent`, `AddLink/RemoveLinkByID/FindByID/FindByScope/PrimaryForScope/Sort`, and `Validate`. YAML round-trip is byte-stable; saves are atomic under a cross-platform O_CREATE|O_EXCL file lock with a 5s timeout. Reuses `internal/quest/link.go::ValidateLinkPath` for campaign-root containment (no parallel implementation).
- **`internal/workitem/selector/`** — selector resolver. Order: stable `.workitem` id → key (`<type>:<path>`) → relative path → directory slug → optional fuzzy. Returns wrapped validation errors with candidate lists on ambiguity and near-match suggestions on not-found.
- **`internal/workitem/resolver/`** — deterministic six-tier context resolver (`explicit` → `ancestor` → `link` → `festival` → `current` → `none`) used by `camp workitem resolve` and the sequence-03 commit wrappers. Carries a per-tier `Trace[]` for `--json --explain` debugging and a `QuestID` field that stays empty until sequence 03 plumbs it.
- **CLI surface** under `internal/commands/workitem/`:
  - `camp workitem link <selector> [path]` with `--project`, `--festival`, `--worktree`, `--cwd`, `--role`, `--replace`, `--allow-missing`, `--json`
  - `camp workitem unlink` with `--id`, scope filters, and `--all`
  - `camp workitem links [selector]` with table + `--json`
  - `camp workitem current [selector] | --clear` with `--json`
  - `camp workitem resolve [--workitem] [--festival] [--json] [--explain]`
  - `camp workitem doctor [--json] [--fix]` reporting `workitem.link.broken`, `workitem.scope.broken`, `workitem.scope.out-of-bounds`, `workitem.link.duplicate-primary`, `workitem.schema.violation`, `workitem.current.missing`

## Test plan

- [x] `go test ./...` clean across the camp module
- [x] Unit coverage: `internal/workitem/links` 75.6%, `internal/commands/workitem` 39.3% (workitem command surface includes pre-existing TUI code not covered by these tests)
- [x] Round-trip and validation: every fixture loads, saves, reloads byte-identical; 100-iteration random round-trip; every SCHEMA.md §9 rule has positive + negative cases; 2-goroutine concurrent save preserves file integrity
- [x] Resolver: every tier exercised, ancestor-beats-link precedence verified, JSON shape asserted
- [x] Doctor: every finding code triggered, `--fix` round-trip verified
- [x] Containerized integration tests pass (`just test integration-run TestIntegration_LinkLifecycle | TestIntegration_Resolver | TestIntegration_Doctor`)

## Notes for reviewers

- `TestIntegration_ResolverQuestID` and the `stale-quest-id` doctor check are intentionally skipped/deferred. They depend on sequence 03 task 01 adding `quest_id` to the `.workitem` schema; the test files keep the contracts in place so they reappear once the dependency lands.
- `SCHEMA.md` §13 lists the open questions tracked out of this sequence (stable cross-machine `created_by` identity, time-based expiry, quest-link auto-mirror).
- `current.yaml` is meant to be gitignored. The gitignore policy is documented in §1.1; the actual `camp init` `.gitignore` template update is deferred to sequence 03 (it touches the same template).

### DIFF
--- a/internal/commands/workitem/doctor.go
+++ b/internal/commands/workitem/doctor.go
@@ -1,0 +1,310 @@
+package workitem
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/paths"
+	"github.com/Obedience-Corp/camp/internal/quest"
+	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
+	"github.com/Obedience-Corp/camp/internal/workitem/links"
+)
+
+// Doctor finding codes (dotted-domain form). Stable strings; consumers
+// dispatch on them.
+const (
+	codeBrokenLink       = "workitem.link.broken"
+	codeBrokenScope      = "workitem.scope.broken"
+	codeOutOfBounds      = "workitem.scope.out-of-bounds"
+	codeDuplicatePrimary = "workitem.link.duplicate-primary"
+	codeSchemaViolation  = "workitem.schema.violation"
+	codeCurrentMissing   = "workitem.current.missing"
+)
+
+const (
+	docSeverityError   = "error"
+	docSeverityWarning = "warning"
+	docSeverityInfo    = "info"
+)
+
+type docFinding struct {
+	Code        string `json:"code"`
+	Severity    string `json:"severity"`
+	Target      string `json:"target,omitempty"`
+	Message     string `json:"message"`
+	FixHint     string `json:"fix_hint,omitempty"`
+	AutoFixable bool   `json:"auto_fixable"`
+}
+
+// errDoctorIssues triggers a non-zero exit from cobra after we have already
+// emitted findings.
+var errDoctorIssues = camperrors.NewValidation("doctor", "doctor reported error-severity findings", nil)
+
+func newDoctorCommand() *cobra.Command {
+	var jsonOut, fix bool
+	cmd := &cobra.Command{
+		Use:   "doctor",
+		Short: "Report workitem link-registry health issues",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runDoctor(cmd.Context(), cmd, jsonOut, fix)
+		},
+		SilenceUsage:  true,
+		SilenceErrors: true,
+	}
+	cmd.Flags().BoolVar(&jsonOut, "json", false, "emit a structured JSON result")
+	cmd.Flags().BoolVar(&fix, "fix", false, "auto-repair findings tagged auto_fixable")
+	return cmd
+}
+
+func runDoctor(ctx context.Context, cmd *cobra.Command, jsonOut, fix bool) error {
+	_, root, err := config.LoadCampaignConfigFromCwd(ctx)
+	if err != nil {
+		return camperrors.Wrap(err, "not in a campaign directory")
+	}
+	registry, err := links.Load(ctx, root)
+	if err != nil {
+		return err
+	}
+
+	knownIDs, err := workitemIDsOnDisk(ctx, root)
+	if err != nil {
+		return err
+	}
+
+	findings := collectWorkitemFindings(ctx, root, registry, knownIDs)
+	if fix {
+		applied := autoFixWorkitemFindings(ctx, root, registry, findings)
+		if applied > 0 {
+			if err := links.Save(ctx, root, registry); err != nil {
+				return err
+			}
+			// Re-run findings after fixes for an accurate post-fix report.
+			knownIDs, _ = workitemIDsOnDisk(ctx, root)
+			findings = collectWorkitemFindings(ctx, root, registry, knownIDs)
+		}
+	}
+
+	if jsonOut {
+		if err := emitDocJSON(cmd.OutOrStdout(), findings); err != nil {
+			return err
+		}
+	} else {
+		emitDocHuman(cmd.OutOrStdout(), findings)
+	}
+
+	if hasErrorFinding(findings) {
+		return errDoctorIssues
+	}
+	return nil
+}
+
+func collectWorkitemFindings(ctx context.Context, root string, registry *links.Links, knownIDs map[string]struct{}) []docFinding {
+	var findings []docFinding
+
+	// Schema-level validation.
+	for _, v := range links.Validate(ctx, registry, links.ValidateOptions{
+		CampaignRoot: root,
+		WorkitemIDs:  nil, // existence is checked below as a separate finding
+	}) {
+		findings = append(findings, docFinding{
+			Code:     codeSchemaViolation,
+			Severity: docSeverityError,
+			Target:   targetForLinkID(v.LinkID),
+			Message:  v.Field + ": " + v.Message,
+		})
+	}
+
+	primarySeen := make(map[string]string)
+	for _, link := range registry.Links {
+		if _, known := knownIDs[link.WorkitemID]; !known {
+			findings = append(findings, docFinding{
+				Code:        codeBrokenLink,
+				Severity:    docSeverityError,
+				Target:      "link:" + link.ID,
+				Message:     "workitem_id " + link.WorkitemID + " is not present on disk",
+				FixHint:     "auto-fix removes the link",
+				AutoFixable: true,
+			})
+		}
+		if !scopeTargetExists(root, link.Scope.Path) {
+			findings = append(findings, docFinding{
+				Code:        codeBrokenScope,
+				Severity:    docSeverityError,
+				Target:      "link:" + link.ID,
+				Message:     "scope path " + link.Scope.Path + " does not exist",
+				FixHint:     "remove the link or restore the directory; auto-fix removes the link",
+				AutoFixable: true,
+			})
+		}
+		if err := quest.ValidateLinkPath(root, link.Scope.Path); err != nil && errors.Is(err, camperrors.ErrInvalidInput) {
+			findings = append(findings, docFinding{
+				Code:     codeOutOfBounds,
+				Severity: docSeverityError,
+				Target:   "link:" + link.ID,
+				Message:  "scope path " + link.Scope.Path + " escapes the campaign root",
+			})
+		}
+		if link.Role == links.RolePrimary {
+			key := string(link.Scope.Kind) + "::" + link.Scope.Path
+			if other, dup := primarySeen[key]; dup {
+				findings = append(findings, docFinding{
+					Code:     codeDuplicatePrimary,
+					Severity: docSeverityError,
+					Target:   "scope:" + key,
+					Message:  "primary links " + other + " and " + link.ID + " collide on the same scope",
+				})
+			} else {
+				primarySeen[key] = link.ID
+			}
+		}
+	}
+
+	cur, err := links.LoadCurrent(ctx, root)
+	if err == nil && cur != nil {
+		if _, known := knownIDs[cur.WorkitemID]; !known {
+			findings = append(findings, docFinding{
+				Code:        codeCurrentMissing,
+				Severity:    docSeverityWarning,
+				Target:      "current:" + cur.WorkitemID,
+				Message:     "current.yaml points at workitem " + cur.WorkitemID + " which is not present on disk",
+				FixHint:     "auto-fix removes current.yaml",
+				AutoFixable: true,
+			})
+		}
+	}
+
+	sort.SliceStable(findings, func(i, j int) bool {
+		if findings[i].Code != findings[j].Code {
+			return findings[i].Code < findings[j].Code
+		}
+		return findings[i].Target < findings[j].Target
+	})
+	return findings
+}
+
+func autoFixWorkitemFindings(ctx context.Context, root string, registry *links.Links, findings []docFinding) int {
+	applied := 0
+	for _, f := range findings {
+		if !f.AutoFixable {
+			continue
+		}
+		switch f.Code {
+		case codeBrokenLink, codeBrokenScope:
+			id := strings.TrimPrefix(f.Target, "link:")
+			if registry.RemoveLinkByID(id) {
+				applied++
+			}
+		case codeCurrentMissing:
+			if err := links.SaveCurrent(ctx, root, nil); err == nil {
+				applied++
+			}
+		}
+	}
+	return applied
+}
+
+func targetForLinkID(linkID string) string {
+	if linkID == "" {
+		return "registry"
+	}
+	return "link:" + linkID
+}
+
+func scopeTargetExists(root, scopePath string) bool {
+	if scopePath == "" {
+		return false
+	}
+	abs := filepath.Join(root, filepath.FromSlash(scopePath))
+	_, err := os.Stat(abs)
+	if err == nil {
+		return true
+	}
+	return !errors.Is(err, fs.ErrNotExist)
+}
+
+func workitemIDsOnDisk(ctx context.Context, root string) (map[string]struct{}, error) {
+	cfg, err := config.LoadCampaignConfig(ctx, root)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "load campaign config")
+	}
+	resolver := paths.NewResolverFromConfig(root, cfg)
+	items, err := wkitem.Discover(ctx, root, resolver)
+	if err != nil {
+		return nil, err
+	}
+	set := make(map[string]struct{}, len(items))
+	for _, item := range items {
+		if item.StableID != "" {
+			set[item.StableID] = struct{}{}
+		}
+	}
+	return set, nil
+}
+
+func hasErrorFinding(findings []docFinding) bool {
+	for _, f := range findings {
+		if f.Severity == docSeverityError {
+			return true
+		}
+	}
+	return false
+}
+
+func emitDocHuman(w io.Writer, findings []docFinding) {
+	if len(findings) == 0 {
+		fmt.Fprintln(w, "doctor: 0 findings")
+		return
+	}
+	fmt.Fprintf(w, "doctor: %d finding(s)\n", len(findings))
+	for _, f := range findings {
+		fmt.Fprintf(w, "  [%s] %s %s — %s\n", f.Severity, f.Code, f.Target, f.Message)
+		if f.FixHint != "" {
+			fmt.Fprintln(w, "    hint: "+f.FixHint)
+		}
+	}
+}
+
+func emitDocJSON(w io.Writer, findings []docFinding) error {
+	if findings == nil {
+		findings = []docFinding{}
+	}
+	out := struct {
+		SchemaVersion string       `json:"schema_version"`
+		GeneratedAt   time.Time    `json:"generated_at"`
+		Findings      []docFinding `json:"findings"`
+		ErrorCount    int          `json:"error_count"`
+		WarningCount  int          `json:"warning_count"`
+		InfoCount     int          `json:"info_count"`
+	}{
+		SchemaVersion: "workitem-doctor/v1alpha1",
+		GeneratedAt:   time.Now().UTC(),
+		Findings:      findings,
+	}
+	for _, f := range findings {
+		switch f.Severity {
+		case docSeverityError:
+			out.ErrorCount++
+		case docSeverityWarning:
+			out.WarningCount++
+		case docSeverityInfo:
+			out.InfoCount++
+		}
+	}
+	enc := json.NewEncoder(w)
+	enc.SetIndent("", "  ")
+	return enc.Encode(out)
+}
--- a/internal/commands/workitem/doctor.go
+++ b/internal/commands/workitem/doctor.go
@@ -26,12 +26,13 @@ import (
 // Doctor finding codes (dotted-domain form). Stable strings; consumers
 // dispatch on them.
 const (
-	codeBrokenLink       = "workitem.link.broken"
-	codeBrokenScope      = "workitem.scope.broken"
-	codeOutOfBounds      = "workitem.scope.out-of-bounds"
-	codeDuplicatePrimary = "workitem.link.duplicate-primary"
-	codeSchemaViolation  = "workitem.schema.violation"
-	codeCurrentMissing   = "workitem.current.missing"
+	codeBrokenLink         = "workitem.link.broken"
+	codeBrokenScope        = "workitem.scope.broken"
+	codeOutOfBounds        = "workitem.scope.out-of-bounds"
+	codeScopeUnvalidatable = "workitem.scope.unvalidatable"
+	codeDuplicatePrimary   = "workitem.link.duplicate-primary"
+	codeSchemaViolation    = "workitem.schema.violation"
+	codeCurrentMissing     = "workitem.current.missing"
 )
 
 const (
@@ -140,7 +141,8 @@ func collectWorkitemFindings(ctx context.Context, root string, registry *links.L
 				AutoFixable: true,
 			})
 		}
-		if !scopeTargetExists(root, link.Scope.Path) {
+		scopeMissing := !scopeTargetExists(root, link.Scope.Path)
+		if scopeMissing {
 			findings = append(findings, docFinding{
 				Code:        codeBrokenScope,
 				Severity:    docSeverityError,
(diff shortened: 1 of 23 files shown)


internal/workitem/links/save.go
```diff
@@ -0,0 +1,150 @@
...
+					return
+				}
+				released = true
+				_ = os.Remove(lockPath)
+			}, nil
+		}
+		if !errors.Is(err, fs.ErrExist) {
+			return nil, camperrors.Wrap(err, "acquire lock")
+		}
+		if time.Now().After(deadline) {
```
obey-agent: **Stale lock recovery is missing.** If a camp invocation crashes or is SIGKILLed between `os.OpenFile(O_CREATE|O_EXCL)` and the deferred `release()`, the lock file orphans. Every subsequent invocation hits the 5s timeout and fails with the misleading "another camp invocation holds it?" message — but no camp is actually running.

The error text on line 142 even hints at the failure mode, so the author was aware. Fix one of two ways:

1. **Stale by age**: when `O_EXCL` returns `ErrExist`, `os.Stat` the lock file and treat it as stale if older than e.g. 30s. Then `os.Remove` it and retry.
2. **PID liveness**: write the PID to the lock file on acquire; on `ErrExist`, read the PID, check `kill(pid, 0)` (or platform equivalent), remove if the process is gone.

Option 1 is simpler and cross-platform. Option 2 is more correct but adds complexity. Either way the current code makes any process death a 5-second wait + hard failure for every later camp invocation. For a CLI run from shell loops, that is fatal to UX.

Also worth doing: change the line 142 error from `wrap(ErrInvalidInput, "...")` to a more specific error like `wrap(ErrLockTimeout, ...)` so callers can distinguish lock contention from genuine validation failures.

internal/workitem/resolver/resolver.go
```diff
@@ -0,0 +1,304 @@
...
+			if err == nil {
+				return wi, TraceStep{Tier: SourceAncestor, Result: "match", Detail: filepath.ToSlash(rel)}
+			}
+		}
+		if dir == root || dir == filepath.Dir(dir) {
+			break
+		}
+		dir = filepath.Dir(dir)
+		// Guard: don't walk above the campaign root.
+		if !strings.HasPrefix(dir, root) {
```
obey-agent: **`strings.HasPrefix(dir, root)` for path containment is a known macOS bug.** On macOS, `/var` is a symlink to `/private/var` and `/tmp` is a symlink to `/private/tmp`. If `root` is resolved through one and `cwd` is resolved through the other, this prefix check breaks early and the resolver silently skips the ancestor tier. Same trap fires for any campaign mounted under a symlinked path.

Use `filepath.Rel` (which already handles `..` semantics) and check that:

```go
if rel, err := filepath.Rel(root, dir); err != nil || strings.HasPrefix(rel, "..") {
    break
}
```

For full safety against symlink mismatches, call `filepath.EvalSymlinks` on both `root` and `cwd` once at the top of `Resolve` (or `resolveAncestor`) before the walk. Project memory: "macOS has `/var` → `/private/var` symlink, causing path comparison issues. Tests must use `filepath.EvalSymlinks()` to resolve paths before comparison."

Without this fix the ancestor tier silently fails for users who happened to symlink their campaign or run camp from inside `/tmp/some-test-campaign`, and the resolver quietly falls through to the next tier — exactly the silent-wrong-answer pattern this PR is trying to eliminate.

internal/workitem/resolver/resolver.go
```diff
@@ -0,0 +1,304 @@
...
+	wi, err := selector.Resolve(ctx, root, best.WorkitemID, selector.ResolveOptions{})
+	if err != nil {
+		// Workitem on disk vanished; surface as a tier miss with a note —
+		// doctor will report the same condition as broken-link.
+		return nil, TraceStep{
+			Tier:   SourceLink,
+			Result: "error",
+			Detail: "primary link " + best.ID + " points to missing workitem " + best.WorkitemID,
+		}
+	}
```
obey-agent, line 231: **Silent fall-through when a primary link points to a missing workitem.** When `selector.Resolve` fails for a primary path link, this returns `nil` and an "error" trace step, then the orchestrator (line 100-109) proceeds to the next tier. The user gets the festival or current answer instead of being told that the link they explicitly set up is broken.

This is precisely the failure mode that links + commit routing must NOT have. A user who runs `camp workitem link` then weeks later deletes the workitem will start silently routing commits to the wrong workitem the next time the resolver runs. The trace contains the info but `--explain` is opt-in; the default `camp workitem resolve` output buries it.

Two options:

1. **Hard-fail at the resolver level**: if a primary link covers cwd but its workitem is missing, return an error (or a Resolution with Source=`SourceLink` and a `Broken` flag), and let the caller decide. This is the strictest, surfaces the issue immediately.
2. **Emit a stderr warning on demote**: keep falling through, but print a one-line warning like `warning: primary link lnk_xxx points to missing workitem; ignoring`. Cheap and the user sees it without `--explain`.

Doctor will eventually catch this as `workitem.link.broken`, but doctor is opt-in. The resolver runs on every commit (per the sequence-03 plan). Catching it silently here is the exact path to "my commits landed on the wrong workitem for 3 weeks."

If option 1 is too strict, at minimum elevate the `Result: "error"` to a real Resolution-level signal so the human-readable resolve output surfaces it without `--explain`.

internal/workitem/selector/selector.go
```diff
@@ -0,0 +1,140 @@
...
+//     priority that produced any match;
+//   - a wrapped ErrAmbiguous when multiple workitems tie at that priority;
+//   - a wrapped ErrNotFound when nothing matches.
+//
+// Resolution order:
+//  1. exact stable .workitem id
+//  2. exact key (`<type>:<path>`)
+//  3. exact RelativePath
+//  4. exact directory-base slug
+//  5. fuzzy substring on key or title (only when opts.AllowFuzzy)
```
obey-agent, line 42: **This doc comment promises sentinel errors that do not exist.** Says "wrapped ErrAmbiguous when multiple workitems tie" and "wrapped ErrNotFound when nothing matches", but the actual returns at lines 73 (via `ambiguous`) and 94 (via `notFound`) both call `camperrors.NewValidation("selector", ...)`. Neither path attaches a sentinel error consumers can dispatch on with `errors.Is`.

This bites the resolver: at `resolver.go:213-218`, `resolveLink` treats every selector error as "broken link". If `selector.Resolve` returned a wrapped `ErrAmbiguous`, the resolver could distinguish "workitem deleted" (real broken link) from "selector matched two things" (config bug). Today both produce the same trace step.

Either:

1. **Add the sentinels** (preferred since the doc already promises them):
   ```go
   var (
       ErrSelectorAmbiguous = camperrors.NewValidation("selector", "ambiguous", nil)
       ErrSelectorNotFound  = camperrors.NewValidation("selector", "not found", nil)
   )

   func ambiguous(...) error {
       return camperrors.Wrap(ErrSelectorAmbiguous, "... details ...")
   }
   ```
   Then consumers can `errors.Is(err, selector.ErrSelectorAmbiguous)`.
2. **Or fix the doc** to remove the wrapped-error promise.

Option 1 is the right move because the broken-link / silent-fall-through finding in `resolver.go:218` becomes much easier to fix once selector errors are categorized.

internal/commands/workitem/doctor.go
```diff
@@ -0,0 +1,310 @@
...
+			findings = append(findings, docFinding{
+				Code:        codeBrokenScope,
+				Severity:    docSeverityError,
+				Target:      "link:" + link.ID,
+				Message:     "scope path " + link.Scope.Path + " does not exist",
+				FixHint:     "remove the link or restore the directory; auto-fix removes the link",
+				AutoFixable: true,
+			})
+		}
+		if err := quest.ValidateLinkPath(root, link.Scope.Path); err != nil && errors.Is(err, camperrors.ErrInvalidInput) {
```
obey-agent: **This silently swallows every error from `quest.ValidateLinkPath` except `ErrInvalidInput`.** The condition `err != nil && errors.Is(err, camperrors.ErrInvalidInput)` only reports when the error is exactly that sentinel. If `ValidateLinkPath` returns a wrapped I/O error (filesystem stat failure, permission denied, etc.), the finding is never emitted and doctor reports "0 findings" while the registry is in fact unhealthy.

Fix:

```go
if err := quest.ValidateLinkPath(root, link.Scope.Path); err != nil {
    severity := docSeverityError
    msg := "scope path " + link.Scope.Path + " escapes the campaign root"
    if !errors.Is(err, camperrors.ErrInvalidInput) {
        // Non-ErrInvalidInput is a probe failure, not an out-of-bounds.
        // Report as a separate finding (or wrap with the actual cause) so
        // the operator can see what went wrong rather than getting silence.
        msg = "scope path " + link.Scope.Path + " could not be validated: " + err.Error()
    }
    findings = append(findings, docFinding{
        Code:     codeOutOfBounds,
        Severity: severity,
        Target:   "link:" + link.ID,
        Message:  msg,
    })
}
```

Or add a separate finding code like `workitem.scope.unvalidatable` so consumers can dispatch. The current behavior — silent pass when a probe fails — is the opposite of what doctor exists to do.

internal/commands/workitem/links.go
```diff
@@ -0,0 +1,99 @@
...
+		var matched []links.Link
+		for _, link := range registry.Links {
+			if link.WorkitemID == wi.StableID {
+				matched = append(matched, link)
+			}
+		}
+		filtered = matched
+	}
+
+	registry.Sort()
```
obey-agent: **`registry.Sort()` runs after the filter, so filtered output is not sorted.** When `selectorArg` is empty (line 47), `filtered := registry.Links` aliases the underlying slice, and the subsequent `Sort()` (line 62) sorts both. But in the filter branch (lines 49-60), `matched` is a fresh slice built in encounter order, and the subsequent `Sort()` only orders `registry.Links` — leaving `filtered` in whatever order the registry happened to be iterated.

Two-line fix: hoist the sort:

```go
registry.Sort()

filtered := registry.Links
if selectorArg != "" {
    // ... build matched in sorted order since registry.Links is now sorted
    filtered = matched
}
```

Without this fix, `camp workitem links <selector> --json` is non-deterministic across runs, which breaks any consumer that diffs the output.

internal/workitem/links/validate.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Now is the reference time for created_at skew checks. Defaults to
+	// time.Now() if zero.
+	Now time.Time
+}
+
+var linkIDRegex = regexp.MustCompile(LinkIDPattern)
+
+const (
+	maxClockSkewWarn   = 5 * time.Minute
```
obey-agent: **`maxClockSkewWarn = 5 * time.Minute` is declared but never used.** The validator at line 162-168 only enforces `maxClockSkewReject` (24h). The warn threshold is dead code.

Either:

1. **Implement the warning channel.** Today `Validate` returns `[]ValidationError`. Promote `ValidationError` to carry a `Severity` field (`error|warning`), or return `(errs, warnings)` separately, then surface warnings as `info`-severity doctor findings.
2. **Delete the constant** and update SCHEMA.md if the spec promised warn-level skew reporting.

Partial implementations like this rot fast — a future reader sees the constant, assumes it is wired up, builds on the assumption, and ships a regression.

internal/commands/workitem/link.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	link := links.Link{
+		ID:          generateLinkID(),
+		WorkitemID:  workitemIDForLink(wi, opts),
+		WorkitemKey: workitemKeyForLink(wi),
+		Scope:       *scope,
+		Role:        role,
+		CreatedAt:   time.Now().UTC().Truncate(time.Second),
+		CreatedBy:   "camp_workitem_link",
```
obey-agent: **`CreatedBy: "camp_workitem_link"` is misleading.** Every link from every user on every machine gets the literal string `camp_workitem_link` written into the registry. That is the name of a command, not an identity. SCHEMA.md §13 acknowledges this as a deferred open question ("stable cross-machine `created_by` identity"), but the current placeholder is actively wrong semantically: a reviewer or doctor looking at `links.yaml` cannot tell who actually created the link.

Until §13 is resolved, the OS username is a better placeholder and matches what every other git-adjacent tool does:

```go
import "os/user"

func defaultCreatedBy() string {
    if u, err := user.Current(); err == nil && u.Username != "" {
        // sanitize for createdByRegex: ^[A-Za-z0-9_-]{1,64}$
        sanitized := createdBySanitizer.ReplaceAllString(u.Username, "-")
        if len(sanitized) > 64 {
            sanitized = sanitized[:64]
        }
        return sanitized
    }
    return "camp_workitem_link" // last-resort fallback
}
```

This avoids the contract lie ("created_by" should identify who) without committing to the §13 cross-machine identity story.

internal/workitem/resolver/resolver.go
```diff
@@ -0,0 +1,304 @@
...
+	if wi, step := resolveCurrent(ctx, root); wi != nil {
+		result.Workitem = wi
+		result.Source = SourceCurrent
+		result.Reason = step.Detail
+		result.QuestID = questIDOf(wi)
+		result.Trace = append(result.Trace, step)
+		return result, nil
+	} else {
+		result.Trace = append(result.Trace, step)
+	}
```
obey-agent, line 131: **Six tier blocks below are copy-pasted with the same structure.** Lines 78-131 repeat the pattern (`if wi, step := resolveX(...); wi != nil { ... } else { result.Trace = append(...) }`) six times. The only varying parts are the function call, the `Source` constant, and the `Reason` text. Easily factored:

```go
tiers := []struct {
    src  Source
    fn   func() (*workitem.WorkItem, TraceStep)
    base string
}{
    {SourceExplicit, func() (*workitem.WorkItem, TraceStep) { return resolveExplicit(ctx, root, opts) }, "explicit --workitem flag"},
    {SourceAncestor, func() (*workitem.WorkItem, TraceStep) { return resolveAncestor(ctx, root, cwd) }, "nearest ancestor .workitem"},
    {SourceLink,     func() (*workitem.WorkItem, TraceStep) { return resolveLink(ctx, root, cwd) }, ""},
    {SourceFestival, func() (*workitem.WorkItem, TraceStep) { return resolveFestival(ctx, root, opts.FestivalID) }, ""},
    {SourceCurrent,  func() (*workitem.WorkItem, TraceStep) { return resolveCurrent(ctx, root) }, ""},
}
for _, t := range tiers {
    wi, step := t.fn()
    result.Trace = append(result.Trace, step)
    if wi != nil {
        result.Workitem = wi
        result.Source = t.src
        result.Reason = firstNonEmpty(t.base, step.Detail)
        result.QuestID = questIDOf(wi)
        return result, nil
    }
}
```

Adding a tier becomes one new struct entry instead of 12 lines of copy-paste. More importantly: future bug fixes to the per-tier orchestration get applied once, not six times. This matters because the silent-fall-through fix proposed in the `resolveLink` comment will need to touch this structure too.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview

The link-registry design is genuinely impressive: the v1alpha1 schema is comprehensive, persistence is atomic with file locking, the six-tier resolver is well-shaped, and doctor + sync wire-up matches the dotted-domain-finding-code pattern already established in PR #302. SCHEMA.md §13 makes the open questions explicit. The package uses `camperrors` correctly and threads `context.Context` through every I/O call. What blocks approval are two correctness bugs (stale-lock recovery, macOS symlink path comparison), one semantic safety issue (silent fall-through on broken primary link), and a handful of contract gaps (selector sentinels, doctor swallowing non-`ErrInvalidInput`, links sort-after-filter). Each is a small fix; together they would catch a careful reviewer at staff level.

## Key Findings

1. **`acquireFileLock` has no stale-lock recovery.** Any SIGKILL or crash between `O_CREATE|O_EXCL` and `release()` orphans the lock file; every subsequent camp invocation hits the 5s timeout. The error string already hints at this scenario, so the author was aware. Add age-based or PID-based stale detection. (inline `save.go:140`)

2. **`resolveAncestor` uses `strings.HasPrefix(dir, root)` for containment.** On macOS this breaks across `/var` and `/private/var` symlinks; the ancestor tier silently fails for any campaign mounted under a symlinked path. Use `filepath.Rel` (and `filepath.EvalSymlinks` on `root`/`cwd` at entry). Captured in repo memory after a prior incident. (inline `resolver.go:167`)

3. **Silent fall-through when a primary link points to a missing workitem.** The resolver demotes to the next tier and the trace step is buried unless the user passed `--explain`. For a feature whose explicit purpose is correct commit routing, this is the exact silent-wrong-answer pattern to avoid. Either hard-fail or surface a stderr warning on demote. (inline `resolver.go:218`)

4. **`selector.Resolve` doc lies about sentinel errors.** Promises `ErrAmbiguous` and `ErrNotFound` but returns generic `camperrors.NewValidation`. Consumers (including the resolver, see finding #3) cannot dispatch programmatically. Either add the sentinels or fix the doc. The sentinels are the right move because they unlock the broken-link fix. (inline `selector.go:37`)

5. **Doctor silently swallows non-`ErrInvalidInput` errors from `quest.ValidateLinkPath`.** Any I/O or permission failure passes silently, so doctor reports "0 findings" while the registry is in fact unhealthy. (inline `doctor.go:153`)

6. **`links.go` calls `Sort()` after building the filtered slice.** When a selector filter is applied, the output is in match-encounter order, breaking consumers that diff JSON output across runs. (inline `links.go:62`)

7. **`maxClockSkewWarn = 5 * time.Minute` is dead code.** Either wire it through the validator as a warning channel or delete it; partial implementations rot. (inline `validate.go:53`)

8. **`CreatedBy: "camp_workitem_link"` is misleading.** That is a command name, not an identity. SCHEMA.md §13 acknowledges this is deferred, but the placeholder should at least be the OS username so the field carries some signal until cross-machine identity lands. (inline `link.go:120`)

9. **Six tier blocks in `resolver.Resolve` are copy-pasted.** Easily factored to a tier-dispatch slice; tightens the code path that finding #3 needs to fix. (inline `resolver.go:131`)

## What is Done Well

- `camperrors` used consistently (`Wrap`, `Wrapf`, `NewValidation`); no `fmt.Errorf` regressions.
- `context.Context` threaded through every I/O call, with `ctx.Err()` checks at function entry.
- SCHEMA.md is 375 lines including §9 validation rules with positive + negative test cases and §13 explicit open-questions section. This is exemplary contract discipline.
- `Empty()` factory ensures YAML marshals as `links: []` not `links: null`. Tiny detail; huge for consumer reliability.
- `Sort()` enforces stable on-disk order across runs, minimizing diff noise in `links.yaml`.
- Atomic write via `fsutil.WriteFileAtomically` plus the (incomplete) file lock is the right shape for concurrent camp invocations.
- Schema versioning is real (`workitem-links/v1alpha1`, `workitem-current/v1alpha1`, `workitem-doctor/v1alpha1`), with version-mismatch validation that returns a clear error rather than silently accepting future formats.
- `quest.ValidateLinkPath` reuse over reimplementation, matching the no-parallel-implementations rule.
- Resolver carries per-tier `Trace[]` so `--json --explain` is mechanical; sequence-03 commit wrappers will get this debugging surface for free.
- `--json` output across every subcommand with `schema_version` + `generated_at`, mirroring the workflow PR pattern.
- Test files are thorough: 457-line `links_test`, 398-line `link_test`, 324-line `resolve_test`, plus 253-line container integration coverage.
- Dotted-domain finding codes (`workitem.link.broken`, `workitem.scope.out-of-bounds`, etc.) consistent with PR #302 vocabulary — building a real cross-feature contract.

## Staff Standard

Would I submit this as my best work at a staff level? The architectural moves are right and the test discipline is strong, but the two correctness bugs (stale lock, macOS symlinks) and the silent-fall-through pattern are the kind of findings a staff reviewer surfaces before the next sequence starts depending on this code. Sequence 03 will plumb commit routing through `resolver.Resolve`; shipping with the silent broken-link demote means the first user with a stale primary link gets their commits silently misrouted. Address the 9 inline comments, especially #1, #2, #3, and #5, and this becomes a clean approve.

obey-agent (approved):
## Verdict: Approve

## Overview

Every blocker from the previous round is addressed cleanly. The lock-recovery primitive is lifted into `internal/fsutil/lock.go` with age-based stale detection, the resolver canonicalizes paths through `EvalSymlinks` and uses a tier-dispatch slice, `resolveLink` now surfaces broken-link as a real error rather than silently demoting, the selector exports `ErrSelectorAmbiguous`/`ErrSelectorNotFound` sentinels that the resolver dispatches on, doctor adds the `workitem.scope.unvalidatable` finding code, `links` sorts before filtering, `CreatedBy` derives from `os/user`, and the dead `maxClockSkewWarn` is gone.

## Findings Resolved

1. **Stale lock recovery via `internal/fsutil/lock.go`.** Age-based detection (`staleLockAfter = 30s`), retries through `removeStaleLock`, and the timeout error wraps `ErrTimeout` instead of the generic `ErrInvalidInput`. Lifted out of links/ so future writers can reuse it. (`internal/workitem/links/save.go` now imports `fsutil.AcquireFileLock`.)

2. **`resolveAncestor` and `resolveLink` canonicalize paths.** `canonicalPath` calls `filepath.Abs` + `filepath.EvalSymlinks` at Resolve entry; `pathOutside` uses `filepath.Rel`-based check instead of `strings.HasPrefix`. The macOS `/var → /private/var` and `/tmp → /private/tmp` cases now resolve identically regardless of which path the campaign or cwd was reached through.

3. **`resolveLink` returns a real error on broken primary link.** Lines 219-231: distinguishes `selector.ErrSelectorNotFound` (workitem deleted, surfaced via `workitem_link` validation error) from other selector failures. The tier orchestrator at line 105-107 propagates this error, so the user sees the broken-link signal up-front rather than getting a silently-demoted festival/current answer.

4. **Selector sentinels are real.** `ErrSelectorAmbiguous` and `ErrSelectorNotFound` exported at `selector.go:27-28`; `ambiguous()` and `notFound()` wrap them via `camperrors.Wrap`. The resolver uses `errors.Is(err, selector.ErrSelectorNotFound)` to dispatch.

5. **Doctor adds `codeScopeUnvalidatable` (`workitem.scope.unvalidatable`).** Switch statement at `doctor.go:155-170` distinguishes ErrInvalidInput (out-of-bounds) from other errors (unvalidatable). The new finding fires only when the scope is not already reported missing, avoiding duplicate noise. Severity `error`, no auto-fix.

6. **`links.go` calls `Sort()` before the filter loop** (line 47). Filtered output is now deterministic across runs.

7. **`maxClockSkewWarn` constant deleted.** No partial implementation rot.

8. **`CreatedBy` derived from `user.Current()`** with sanitization (`createdBySanitizer = [^A-Za-z0-9_-]+`). Falls back to `"camp_workitem_link"` only when no user identity is available. SCHEMA.md §13 still tracks the cross-machine identity open question.

9. **Tier orchestrator uses a `resolveTier` slice** with uniform iteration. Adding a future tier is one struct entry; bug fixes touch one shared loop. Pairs with the broken-link fix from #3 — the error-propagation change touched one place instead of six.

- `removeStaleLock` is a self-contained helper that other writers can compose with: `tryAcquireFileLock` either succeeds, returns `false` after clearing stale state (so the caller retries), or returns a real error. Clean separation.
- The `canonicalPath` fallback (line 132-136): if `EvalSymlinks` fails (path does not yet exist), fall back to the cleaned abs path. The right defensive shape.
- `resolveLink` returns BOTH a trace step AND an error — the trace remains complete for `--explain` even when the error short-circuits the resolver. Future operators debugging "why did this fail at the link tier" see the full reasoning.
- The doctor change does not blindly fire `codeScopeUnvalidatable` on top of `codeBrokenScope`; the `case !scopeMissing` guard avoids duplicate findings for the same root cause. Good signal hygiene.
- `defaultCreatedBy` regex sanitization matches the validator's `createdByRegex` exactly (`[A-Za-z0-9_-]{1,64}`), so the username field always round-trips cleanly.

## Verification

- `just test unit` — 2433/2433 passed
- `just test integration-run TestIntegration_LinkLifecycle` — PASS
- `just test integration-run TestIntegration_Doctor` — PASS (`BrokenLinks`, `Fix`)
- `just test integration-run TestIntegration_Resolver` — PASS (`Tiers` passes; `QuestID` skipped pending sequence 03 task 01, which is the documented contract)

## Staff Standard

Would I submit this as my best work at a staff level? Yes. The shared lock primitive, the resolver canonicalization, and the broken-link error propagation are all the kind of architectural decisions that would pass at staff review the first time. Ready to merge.

PR #304, opened 2026-05-25
feat(workitem): ref + quest_id schema, doctor backfill, tag composition (CW0003 slice 3 part 1)

lancekrogers:
## Summary

Sequence 03 of the CW0003 festival, **tasks 01–03**. Stacked on top of PR #303 (seq 02). Adds the `.workitem` schema bump that the rest of the sequence relies on, the doctor backfill for legacy workitems, and the full consolidated-tag composer in `internal/git`/`pkg/commitkit`.

### Tasks landed

**Task 01 — `.workitem` schema → v1alpha6.** New optional fields `ref` (deterministic `WI-<6 hex>` derived from id via `internal/workitem/ref.go::Derive`/`DeriveUnique`) and `quest_id` (captured at create/adopt time via `quest.ResolveContext`). v1alpha4/v1alpha5 markers still load without error. `camp workitem create` and `camp workitem adopt` both gain a `--quest` flag and surface a `ref:` line (plus optional `quest:` line) in their output.

**Task 02 — doctor backfills missing refs.** New finding code `workitem.ref.missing` (severity `warning`, `auto_fixable: true`). `camp workitem doctor --fix` rewrites each refless workitem in place via `yaml.Node` so all other fields and the existing key order are preserved; the `ref` lands immediately after `id` for minimal diffs. Backfill uses `DeriveUnique` against the currently-on-disk ref set (rescanned per fix, not stale-snapshotted) for collision avoidance. `ApplyMetadata` extended to surface `Ref`/`QuestID` into `WorkItem.SourceMetadata` so post-fix re-checks see the new values.

**Task 03 — consolidated tag composer.** `git.FormatContextTagsFull(campaignID, questID, festRef, workitemRef)` composes `[OBEY-CAMPAIGN-{cid}[-{qst}][-FE-{ref}][-WI-WI-{ref}]]`. Component order is fixed; absent components are omitted entirely. `FormatCampaignTag`/`FormatContextTags` now delegate so existing callers keep byte-identical output. `git.ParseTag(subject)` returns `TagComponents{CampaignID, QuestID, FestRef, WorkitemRef}` via a two-stage parser (RE2 strips the bracket; a small state machine peels the inner string on the known prefixes — clean disambiguation of `FE-<ref>-WI-<ref>` from `FE-<hyphenated ref>` without lookahead). `commitkit.PrependContextTagsFull` and `commitkit.ParseTag` re-export the helpers for both camp and fest consumers.

## Stacking note

This PR depends on PR #303 (sequence 02). Merging order:
1. PR #302 (sequence 01 — workflow surface)
2. PR #303 (sequence 02 — link registry)
3. PR #304 (this one — sequence 03 schema + doctor + tag)
4. Sequence 03 task 04+ (wiring the new tag into camp/fest commit wrappers) ships as a follow-up

## Test plan

- [x] `go test ./...` clean across the camp module
- [x] `internal/workitem/ref.go` — 5 unit tests (determinism, uniqueness, collision retry, helper)
- [x] `internal/commands/workitem/ref_quest_test.go` — create writes ref + omits quest_id when none active, three sequential creates produce unique refs, adopt writes ref, legacy v1alpha5 still loads
- [x] `internal/commands/workitem/doctor_ref_test.go` — backfill on legacy workitem clears the finding, preserves all other fields, handles multiple workitems without ref collision
- [x] `internal/git/campaign_tag_full_test.go` — every combo of present/absent components, parser on each known shape, backward-compat for `FormatCampaignTag` one-arg and two-arg forms, 100-iteration property round-trip
- [x] `pkg/commitkit/commitkit_tags_test.go` — `PrependContextTagsFull` matrix + parse-round-trip smoke
- [x] `internal/workitem/testdata/workitem_full.yaml` fixture bumped to v1alpha6 to match the new schema constant; `TestLoadMetadata_FullFixture` updated accordingly

## Notes for reviewers

- The two-stage tag parser in `internal/git/campaign_tag.go::ParseTag` is the riskiest piece. The property test exercises 100 random combinations; if a future component is added (e.g. `BR-<branch>`), the state machine needs a new arm.
- `ApplyMetadata` (`internal/workitem/merge.go`) now writes to `SourceMetadata` when `Ref` or `QuestID` is present. This is additive — existing readers that don't look at those keys are unaffected — but it does change the `SourceMetadata` shape for v1alpha6 workitems.
- The `workitem.ref.missing` doctor check runs on every invocation. On a fresh `camp` install with no v1alpha5 workitems the check is a no-op; on a campaign migrating from v1alpha5, expect one warning per legacy workitem until `--fix` is run.

### DIFF
--- a/internal/commands/workitem/doctor.go
+++ b/internal/commands/workitem/doctor.go
@@ -34,6 +34,7 @@ const (
 	codeSchemaViolation    = "workitem.schema.violation"
 	codeCurrentMissing     = "workitem.current.missing"
 	codeMissingRefField    = "workitem.ref.missing"
+	codeWorkitemScanFailed = "workitem.scan.failed"
 )
 
 const (
@@ -89,7 +90,7 @@ func runDoctor(ctx context.Context, cmd *cobra.Command, jsonOut, fix bool) error
 
 	findings := collectWorkitemFindings(ctx, root, registry, knownIDs)
 	if fix {
-		applied := autoFixWorkitemFindings(ctx, root, registry, findings)
+		applied := autoFixWorkitemFindings(ctx, root, registry, findings, cmd.ErrOrStderr())
 		if applied > 0 {
 			if err := links.Save(ctx, root, registry); err != nil {
 				return err
@@ -190,7 +191,14 @@ func collectWorkitemFindings(ctx context.Context, root string, registry *links.L
 	// the order in which DeriveUnique fills collisions during --fix is
 	// deterministic.
 	missingRefPaths, err := workitemPathsMissingRef(ctx, root)
-	if err == nil {
+	if err != nil {
+		findings = append(findings, docFinding{
+			Code:     codeWorkitemScanFailed,
+			Severity: docSeverityError,
+			Target:   "workitem-tree",
+			Message:  "could not scan workitems for ref backfill: " + err.Error(),
+		})
+	} else {
 		for _, rel := range missingRefPaths {
 			findings = append(findings, docFinding{
 				Code:        codeMissingRefField,
@@ -226,8 +234,9 @@ func collectWorkitemFindings(ctx context.Context, root string, registry *links.L
 	return findings
 }
 
-func autoFixWorkitemFindings(ctx context.Context, root string, registry *links.Links, findings []docFinding) int {
+func autoFixWorkitemFindings(ctx context.Context, root string, registry *links.Links, findings []docFinding, errw io.Writer) int {
 	applied := 0
+	needsRefBackfill := false
 	for _, f := range findings {
 		if !f.AutoFixable {
 			continue
@@ -243,10 +252,15 @@ func autoFixWorkitemFindings(ctx context.Context, root string, registry *links.L
 				applied++
 			}
 		case codeMissingRefField:
-			rel := strings.TrimPrefix(f.Target, "workitem:")
-			if backfillRef(ctx, root, rel) == nil {
-				applied++
-			}
+			needsRefBackfill = true
+		}
+	}
+	if needsRefBackfill {
+		n, err := backfillMissingRefs(ctx, root)
+		applied += n
+		if err != nil {
+			fmt.Fprintf(errw, "warning: backfill refs: %v\n", err)
+			return applied
 		}
 	}
 	return applied
--- a/internal/commands/workitem/doctor.go
+++ b/internal/commands/workitem/doctor.go
@@ -33,6 +33,7 @@ const (
 	codeDuplicatePrimary   = "workitem.link.duplicate-primary"
 	codeSchemaViolation    = "workitem.schema.violation"
 	codeCurrentMissing     = "workitem.current.missing"
+	codeMissingRefField    = "workitem.ref.missing"
 )
 
 const (
@@ -185,6 +186,23 @@ func collectWorkitemFindings(ctx context.Context, root string, registry *links.L
 		}
 	}
 
+	// Workitems missing the ref field added in v1alpha6. Sorted by path so
+	// the order in which DeriveUnique fills collisions during --fix is
+	// deterministic.
+	missingRefPaths, err := workitemPathsMissingRef(ctx, root)
+	if err == nil {
+		for _, rel := range missingRefPaths {
+			findings = append(findings, docFinding{
+				Code:        codeMissingRefField,
+				Severity:    docSeverityWarning,
+				Target:      "workitem:" + rel,
+				Message:     "workitem at " + rel + " is missing the ref field added in v1alpha6",
+				FixHint:     "run camp workitem doctor --fix to backfill",
+				AutoFixable: true,
+			})
+		}
+	}
+
 	cur, err := links.LoadCurrent(ctx, root)
 	if err == nil && cur != nil {
 		if _, known := knownIDs[cur.WorkitemID]; !known {
@@ -222,6 +240,11 @@ func autoFixWorkitemFindings(ctx context.Context, root string, registry *links.L
 			}
 		case codeCurrentMissing:
 			if err := links.SaveCurrent(ctx, root, nil); err == nil {
+				applied++
+			}
+		case codeMissingRefField:
+			rel := strings.TrimPrefix(f.Target, "workitem:")
+			if backfillRef(ctx, root, rel) == nil {
 				applied++
 			}
 		}
--- a/internal/commands/workitem/doctor_ref_backfill.go
+++ b/internal/commands/workitem/doctor_ref_backfill.go
@@ -1,0 +1,167 @@
+package workitem
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"sort"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/fsutil"
+	"github.com/Obedience-Corp/camp/internal/paths"
+	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
+)
+
+// workitemPathsMissingRef discovers every workitem on disk and returns the
+// campaign-relative paths to those whose .workitem marker has no ref field.
+// Paths are sorted lexicographically so DeriveUnique's collision retry has
+// deterministic input ordering during a doctor --fix pass.
+func workitemPathsMissingRef(ctx context.Context, root string) ([]string, error) {
+	cfg, err := config.LoadCampaignConfig(ctx, root)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "load campaign config")
+	}
+	resolver := paths.NewResolverFromConfig(root, cfg)
+	items, err := wkitem.Discover(ctx, root, resolver)
+	if err != nil {
+		return nil, err
+	}
+	var missing []string
+	for _, item := range items {
+		ref, _ := item.SourceMetadata["ref"].(string)
+		if ref != "" {
+			continue
+		}
+		missing = append(missing, item.RelativePath)
+	}
+	sort.Strings(missing)
+	return missing, nil
+}
+
+// backfillRef rewrites the .workitem at relPath to include a derived ref
+// field, computed via DeriveUnique against every other workitem's existing
+// ref (rescanned fresh to avoid stale snapshots in long-running doctor
+// passes). All other fields and YAML key order are preserved by mutating
+// only the parsed mapping in place.
+func backfillRef(ctx context.Context, root, relPath string) error {
+	abs := filepath.Join(root, filepath.FromSlash(relPath), wkitem.MetadataFilename)
+	raw, err := os.ReadFile(abs)
+	if err != nil {
+		return camperrors.Wrapf(err, "read %s", abs)
+	}
+
+	var doc yaml.Node
+	if err := yaml.Unmarshal(raw, &doc); err != nil {
+		return camperrors.Wrapf(err, "parse %s", abs)
+	}
+
+	id, ok := lookupScalar(&doc, "id")
+	if !ok {
+		return camperrors.NewValidation("id", "missing id in "+abs, nil)
+	}
+	if existing, ok := lookupScalar(&doc, "ref"); ok && existing != "" {
+		// Another concurrent process already filled it in. No-op.
+		return nil
+	}
+
+	existingRefs, err := refsExcludingPath(ctx, root, relPath)
+	if err != nil {
+		return err
+	}
+	ref := wkitem.DeriveUnique(id, existingRefs)
+
+	if err := insertScalarAfter(&doc, "id", "ref", ref); err != nil {
+		return err
+	}
+
+	out, err := yaml.Marshal(&doc)
+	if err != nil {
+		return camperrors.Wrap(err, "marshal updated workitem")
+	}
+	return fsutil.WriteFileAtomically(abs, out, 0o644)
+}
+
+func refsExcludingPath(ctx context.Context, root, skipRel string) (map[string]bool, error) {
+	cfg, err := config.LoadCampaignConfig(ctx, root)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "load campaign config")
+	}
+	resolver := paths.NewResolverFromConfig(root, cfg)
+	items, err := wkitem.Discover(ctx, root, resolver)
+	if err != nil {
+		return nil, err
+	}
+	out := make(map[string]bool, len(items))
+	for _, item := range items {
+		if item.RelativePath == skipRel {
+			continue
+		}
+		ref, _ := item.SourceMetadata["ref"].(string)
+		if ref != "" {
+			out[ref] = true
+		}
+	}
+	return out, nil
+}
+
+// lookupScalar finds the scalar value of `key` in a top-level mapping
+// document. Returns (value, true) on hit; ("", false) when the document is
+// not a mapping or the key is absent.
+func lookupScalar(doc *yaml.Node, key string) (string, bool) {
+	root := mappingRoot(doc)
+	if root == nil {
+		return "", false
+	}
+	for i := 0; i+1 < len(root.Content); i += 2 {
+		k := root.Content[i]
+		v := root.Content[i+1]
+		if k.Kind == yaml.ScalarNode && k.Value == key && v.Kind == yaml.ScalarNode {
+			return v.Value, true
+		}
+	}
+	return "", false
+}
+
+// insertScalarAfter inserts {key: value} immediately after the mapping pair
+// whose key equals `after`. If `after` is missing, the new pair is appended
+// at the end. This keeps the on-disk diff minimal for backfills.
+func insertScalarAfter(doc *yaml.Node, after, key, value string) error {
+	root := mappingRoot(doc)
+	if root == nil {
+		return camperrors.NewValidation("doc", "top-level YAML is not a mapping", nil)
+	}
+	insertAt := len(root.Content)
+	for i := 0; i+1 < len(root.Content); i += 2 {
+		k := root.Content[i]
+		if k.Kind == yaml.ScalarNode && k.Value == after {
+			insertAt = i + 2
+			break
+		}
+	}
+	keyNode := &yaml.Node{Kind: yaml.ScalarNode, Tag: "!!str", Value: key}
+	valueNode := &yaml.Node{Kind: yaml.ScalarNode, Tag: "!!str", Value: value}
+
+	updated := make([]*yaml.Node, 0, len(root.Content)+2)
+	updated = append(updated, root.Content[:insertAt]...)
+	updated = append(updated, keyNode, valueNode)
+	updated = append(updated, root.Content[insertAt:]...)
+	root.Content = updated
+	return nil
+}
+
+func mappingRoot(doc *yaml.Node) *yaml.Node {
+	if doc == nil {
+		return nil
+	}
+	root := doc
+	if root.Kind == yaml.DocumentNode && len(root.Content) > 0 {
+		root = root.Content[0]
+	}
+	if root.Kind != yaml.MappingNode {
+		return nil
+	}
+	return root
+}
--- a/internal/commands/workitem/doctor_ref_backfill.go
+++ b/internal/commands/workitem/doctor_ref_backfill.go
@@ -42,12 +42,76 @@ func workitemPathsMissingRef(ctx context.Context, root string) ([]string, error)
 }
 
 // backfillRef rewrites the .workitem at relPath to include a derived ref
-// field, computed via DeriveUnique against every other workitem's existing
-// ref (rescanned fresh to avoid stale snapshots in long-running doctor
-// passes). All other fields and YAML key order are preserved by mutating
-// only the parsed mapping in place.
+// field. All other fields and YAML key order are preserved by mutating only
+// the parsed mapping in place.
 func backfillRef(ctx context.Context, root, relPath string) error {
+	existingRefs, err := refsExcludingPath(ctx, root, relPath)
+	if err != nil {
+		return err
+	}
+	id, err := workitemIDAtPath(root, relPath)
+	if err != nil {
+		return err
+	}
+	ref, err := wkitem.DeriveUnique(ctx, id, existingRefs)
+	if err != nil {
+		return err
+	}
+	return backfillRefWithRef(ctx, root, relPath, ref)
+}
+
+func backfillMissingRefs(ctx context.Context, root string) (int, error) {
+	cfg, err := config.LoadCampaignConfig(ctx, root)
+	if err != nil {
+		return 0, camperrors.Wrap(err, "load campaign config")
+	}
+	resolver := paths.NewResolverFromConfig(root, cfg)
+	items, err := wkitem.Discover(ctx, root, resolver)
+	if err != nil {
+		return 0, err
+	}
+
+	existingRefs := make(map[string]bool, len(items))
+	var pending []wkitem.WorkItem
+	for _, item := range items {
+		ref, _ := item.SourceMetadata["ref"].(string)
+		if ref != "" {
+			existingRefs[ref] = true
+			continue
+		}
+		pending = append(pending, item)
+	}
+	sort.Slice(pending, func(i, j int) bool {
+		return pending[i].RelativePath < pending[j].RelativePath
+	})
+
+	applied := 0
+	for _, item := range pending {
+		if err := ctx.Err(); err != nil {
+			return applied, err
+		}
+		ref, err := wkitem.DeriveUnique(ctx, item.StableID, existingRefs)
+		if err != nil {
+			return applied, err
+		}
+		existingRefs[ref] = true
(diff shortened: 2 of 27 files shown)


internal/workitem/metadata.go
```diff
@@ -14,13 +14,17 @@ import (
...
 
 const MetadataKind = "workitem"
 
+// acceptedWorkitemVersions are loadable .workitem schema versions. v1alpha4
+// and v1alpha5 are accepted for backward compatibility; v1alpha6 is the
+// current shape and gains the Ref and QuestID fields.
 var acceptedWorkitemVersions = map[string]bool{
 	"v1alpha4": true,
 	"v1alpha5": true,
+	"v1alpha6": true,
```
obey-agent, line 28: Adding `v1alpha6` to this map without updating the error message in `validateMetadata` (line 93, just below this hunk) leaves a stale list. The user-facing error still says `supported: v1alpha4, v1alpha5` and does not mention v1alpha6 even though the map now accepts it.

Either update the literal:

```go
return camperrors.NewValidation("version",
    "unsupported .workitem schema version (got "+m.Version+
        "; supported: v1alpha4, v1alpha5, v1alpha6); update .workitem `version:` to "+WorkitemSchemaVersion, nil)
```

Or, preferred, derive the list from `acceptedWorkitemVersions` so the message stays in sync the next time a version is added:

```go
versions := make([]string, 0, len(acceptedWorkitemVersions))
for v := range acceptedWorkitemVersions { versions = append(versions, v) }
sort.Strings(versions)
// ... use strings.Join(versions, ", ") in the error message
```

Hardcoded lists that drift from the source-of-truth map are exactly the kind of comment-rot a future schema bump will reintroduce silently. Staff-level: if you bump a constant, scan callsites that reference the constant by name OR by value.

internal/commands/workitem/doctor_ref_backfill.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if err := insertScalarAfter(&doc, "id", "ref", ref); err != nil {
+		return err
+	}
+
+	out, err := yaml.Marshal(&doc)
+	if err != nil {
+		return camperrors.Wrap(err, "marshal updated workitem")
+	}
+	return fsutil.WriteFileAtomically(abs, out, 0o644)
```
obey-agent, line 141: **No file lock around the rewrite. `os.ReadFile` (line 51) then `fsutil.WriteFileAtomically` (line 84) is a TOCTOU window.** If another camp process writes `.workitem` between the read and the write, that write is lost. The "concurrent process already filled it in" guard at lines 65-68 only checks the snapshot we parsed; it cannot see writes by other processes that landed after our parse.

Concrete scenario:
1. Doctor process A reads `.workitem` (no ref).
2. Doctor process B reads `.workitem` (no ref).
3. `camp workitem create --quest foo` modifies `.workitem` (e.g., adds quest_id).
4. A writes back with ref. (Step 3 changes lost.)
5. B writes back with ref + different unique value. (A also lost.)

The links package solved this in PR #303 via `acquireFileLock` (`internal/workitem/links/save.go:119`). Apply the same pattern here:

```go
release, err := acquireWorkitemLock(ctx, abs+".lock")
if err != nil { return err }
defer release()

raw, err := os.ReadFile(abs)
// ... existing logic ...
```

Either lift `acquireFileLock` from the links package into a shared `internal/fsutil` helper, or copy it (and lift later in a cleanup pass). Either way, the bug is real and the fix is well-established in this codebase.

The same defense applies in `backfillRef` callers running in parallel doctor invocations against shared workitems — without a lock, two `doctor --fix` runs on the same campaign can step on each other.

internal/commands/workitem/doctor_ref_backfill.go
```diff
@@ -0,0 +1,167 @@
...
+		if item.RelativePath == skipRel {
+			continue
+		}
+		ref, _ := item.SourceMetadata["ref"].(string)
+		if ref != "" {
+			out[ref] = true
+		}
+	}
+	return out, nil
+}
```
obey-agent, line 182: **O(N²) discover pattern: every `backfillRef` call re-walks the entire workitem tree.** `refsExcludingPath` calls `wkitem.Discover(ctx, root, resolver)` once per workitem being fixed. For a campaign migrating 50 legacy workitems from v1alpha5, this is 50 full filesystem walks (plus the initial one in `workitemPathsMissingRef`). For larger campaigns it gets noticeable.

The PR description justifies this as "rescanned per fix, not stale-snapshotted" for collision safety. But the same safety property is achievable with one discover + an in-memory set updated as you go:

```go
// In autoFixWorkitemFindings, or a new orchestrator:
items, err := wkitem.Discover(ctx, root, resolver)
if err != nil { return ... }

existingRefs := make(map[string]bool)
var pending []wkitem.WorkItem
for _, item := range items {
    if ref, _ := item.SourceMetadata["ref"].(string); ref != "" {
        existingRefs[ref] = true
    } else {
        pending = append(pending, item)
    }
}
sort.Slice(pending, func(i, j int) bool {
    return pending[i].RelativePath < pending[j].RelativePath
})

for _, item := range pending {
    ref := wkitem.DeriveUnique(item.StableID, existingRefs)
    existingRefs[ref] = true                 // add as we go
    if err := backfillRefWithRef(ctx, root, item.RelativePath, ref); err != nil {
        return err
    }
}
```

Combined with the file-lock fix from the other comment, this gives you both correctness (no stale snapshot, no concurrent overwrite) and good asymptotic behavior.

internal/commands/workitem/doctor.go
```diff
@@ -173,6 +174,23 @@ func collectWorkitemFindings(ctx context.Context, root string, registry *links.L
 		}
 	}
 
+	// Workitems missing the ref field added in v1alpha6. Sorted by path so
+	// the order in which DeriveUnique fills collisions during --fix is
+	// deterministic.
```
obey-agent, line 192: **Discover errors here are silently swallowed.** `if err == nil { ... }` means a scan failure (filesystem permission glitch, IO error, etc.) produces zero `workitem.ref.missing` findings — doctor reports a clean tree while in fact it could not check.

This is the same antipattern flagged on PR #303 doctor (`internal/commands/workitem/doctor.go:153` swallowing non-`ErrInvalidInput`). The fix here:

```go
missingRefPaths, err := workitemPathsMissingRef(ctx, root)
if err != nil {
    findings = append(findings, docFinding{
        Code:     "workitem.scan.failed",
        Severity: docSeverityError,
        Target:   "workitem-tree",
        Message:  "could not scan workitems for ref backfill: " + err.Error(),
    })
} else {
    for _, rel := range missingRefPaths {
        findings = append(findings, docFinding{ /* existing */ })
    }
}
```

Or, at minimum, write a one-line warning to stderr so operators see the silence. Today there is no signal at all.

internal/commands/workitem/doctor.go
```diff
@@ -212,6 +230,11 @@ func autoFixWorkitemFindings(ctx context.Context, root string, registry *links.L
 			if err := links.SaveCurrent(ctx, root, nil); err == nil {
 				applied++
 			}
+		case codeMissingRefField:
+			rel := strings.TrimPrefix(f.Target, "workitem:")
+			if backfillRef(ctx, root, rel) == nil {
+				applied++
```
obey-agent: **`backfillRef` errors are silently dropped here.** `if backfillRef(...) == nil { applied++ }` means a failed fix increments nothing — the user sees `sync: applied N / M findings` with no indication that backfill X failed (disk full, permission denied, YAML corruption, lock contention, etc.).

For doctor --fix this is acute: the user runs it expecting their tree to be clean, sees "applied" output, and only discovers via the next run that some workitems still have missing-ref findings — with zero indication of why.

Fix shape (matches the pattern this codebase already uses elsewhere):

```go
case codeMissingRefField:
    rel := strings.TrimPrefix(f.Target, "workitem:")
    if err := backfillRef(ctx, root, rel); err != nil {
        fmt.Fprintf(os.Stderr, "warning: backfill ref for %s: %v\n", rel, err)
        continue
    }
    applied++
```

Or accumulate errors and surface them in the doctor output as `info`-severity findings. Either way, silent drop is the wrong default.

Same antipattern from PR #303 carried forward; would be good to fix both at the same time so the pattern is consistent across the auto-fix dispatch.

internal/workitem/ref.go
```diff
@@ -0,0 +1,54 @@
...
+	candidate := Derive(id)
+	if !existing[candidate] {
+		return candidate
+	}
+	for n := 1; ; n++ {
+		candidate = Derive(fmt.Sprintf("%s#%d", id, n))
+		if !existing[candidate] {
+			return candidate
+		}
+	}
```
obey-agent, line 50: **Unbounded collision-retry loop with no ctx check.** With 24 bits of entropy and SHA-256 distribution, this loop will never run more than a handful of times in practice. But "never" is not "cannot": a pathological set of pre-existing refs that exhaust the namespace causes this to spin forever, blocking the calling goroutine with no cancellation.

Defensive shape:

```go
const maxRefCollisionRetries = 1 << 24 // entire 24-bit space, generous upper bound

func DeriveUnique(id string, existing map[string]bool) (string, error) {
    candidate := Derive(id)
    if !existing[candidate] {
        return candidate, nil
    }
    for n := 1; n <= maxRefCollisionRetries; n++ {
        candidate = Derive(fmt.Sprintf("%s#%d", id, n))
        if !existing[candidate] {
            return candidate, nil
        }
    }
    return "", fmt.Errorf("ref space exhausted for id %s", id)
}
```

The signature change ripples to two callers (`ref_quest.go`, `doctor_ref_backfill.go`); both already return errors. Cheap to thread.

Also worth a `ctx.Err()` check at the top of the loop so long-running doctor passes can be cancelled mid-backfill. This is a project-standard requirement ("Always check `ctx.Err()` before starting long operations") that the pure-CPU loop currently violates.

internal/git/campaign_tag.go
```diff
@@ -1,27 +1,53 @@
...
+	}
+	if festRef != "" {
+		parts = append(parts, "FE-"+festRef)
+	}
+	if workitemRef != "" {
+		// Workitem refs already carry the WI- prefix per
+		// internal/workitem/ref.go::Derive. Embed verbatim so the bracket
+		// reads `WI-WI-abcdef`, matching the documented tag grammar.
+		parts = append(parts, "WI-"+workitemRef)
+	}
```
obey-agent, line 40: **Contract ambiguity: this formatter accepts both `workitemRef="WI-abcdef"` and `workitemRef="abcdef"` and produces different round-trip results.**

- `FormatContextTagsFull("", "", "", "WI-abcdef")` → `[OBEY-CAMPAIGN--WI-WI-abcdef]`, parsed back as `WorkitemRef = "WI-abcdef"` ✓ (matches the docstring "includes the leading WI- prefix")
- `FormatContextTagsFull("", "", "", "abcdef")` → `[OBEY-CAMPAIGN--WI-abcdef]`, parsed back as `WorkitemRef = "abcdef"` ✗ (no leading prefix, contradicts docstring)

Today every internal caller goes through `workitem.Derive` which always returns `WI-<hex>`, so the second case does not fire. But the parser is permissive and the formatter accepts arbitrary input, so a fest commit wrapper passing the raw 6-hex (without prefix) would silently produce non-canonical output that round-trips inconsistently.

Two ways to lock the contract:

1. **Reject non-canonical input in the formatter:**
   ```go
   if workitemRef != "" {
       if !strings.HasPrefix(workitemRef, workitem.RefPrefix) {
           // Refuse to produce a tag with a non-canonical ref. Forces
           // callers to pass the `WI-` form documented in TagComponents.
           return ""
       }
       parts = append(parts, "WI-"+workitemRef)
   }
   ```
2. **Or normalize on the way in:**
   ```go
   if workitemRef != "" {
       if !strings.HasPrefix(workitemRef, workitem.RefPrefix) {
           workitemRef = workitem.RefPrefix + workitemRef
       }
       parts = append(parts, "WI-"+workitemRef)
   }
   ```

Option 1 is strictly safer (compile-time-ish guarantee callers obey the contract). Option 2 is friendlier. Either way, the current "accepts anything, produces ambiguous tags" shape is a footgun that will bite when sequence-03 task-04 wires this into commit wrappers.

The property test at `campaign_tag_full_test.go:158` always generates `"WI-" + randHex(t, 3)`, so this case is uncovered. Worth adding a negative test.

internal/commands/workitem/doctor_ref_backfill.go
```diff
@@ -0,0 +1,167 @@
...
+	keyNode := &yaml.Node{Kind: yaml.ScalarNode, Tag: "!!str", Value: key}
+	valueNode := &yaml.Node{Kind: yaml.ScalarNode, Tag: "!!str", Value: value}
+
+	updated := make([]*yaml.Node, 0, len(root.Content)+2)
+	updated = append(updated, root.Content[:insertAt]...)
+	updated = append(updated, keyNode, valueNode)
+	updated = append(updated, root.Content[insertAt:]...)
+	root.Content = updated
+	return nil
+}
```
obey-agent, line 233: **Edge case: existing `ref:` line with empty value bypasses the guard at line 65-68 and ends up with a duplicate `ref:` line.** The lookup returns `("", true)` for `ref: ""`; the caller only continues when `existing != ""`, so an empty-string ref triggers a second insert and the file ends up with two `ref:` keys. yaml-v3 silently accepts duplicate keys; the LAST one wins on next load, but the file is now non-canonical YAML.

How might an empty `ref:` exist? Partial write from an aborted earlier doctor --fix, a hand-edit, or an interop tool that wrote the field stub. Robust shape:

```go
// In insertScalarAfter, detect an existing key and update in place rather
// than insert.
for i := 0; i+1 < len(root.Content); i += 2 {
    k := root.Content[i]
    v := root.Content[i+1]
    if k.Kind == yaml.ScalarNode && k.Value == key {
        v.Value = value
        v.Tag = "!!str"
        return nil
    }
}
// ... existing insert-after logic
```

Separately, the `TestDoctor_FixBackfillsRefAndPreservesFields` test asserts only that the title string is present in the output (`strings.Contains(string(after), "title: Legacy legacy")`). It does NOT verify:
- Indent depth is preserved (yaml-v3 default emits 4-space; if `.workitem` files use 2-space, the rewrite changes indentation across every line, producing a noisy diff)
- Key order outside the `id → ref` insertion is preserved
- Trailing newline behavior matches the source

For a feature whose explicit goal is "minimal diff during backfill", these three properties deserve direct byte-level assertions. A test that writes a fixture with 2-space indent, runs backfill, and asserts the only delta is the inserted `ref:` line would close the gap.

internal/git/campaign_tag.go
```diff
@@ -44,3 +70,91 @@ func PrependContextTags(campaignID, questID, message string) string {
...
+			// Unknown trailing content — return what we have rather than guess.
+			rest = ""
+		}
+	}
+	return out
+}
+
+// indexOfNextPrefix returns the byte offset of the next known segment
+// prefix (`-FE-` or `-WI-`) inside s, or len(s) if none is found.
+func indexOfNextPrefix(s string) int {
```
obey-agent, line 151: **Parser assumes quest IDs never contain `-FE-` or `-WI-` substrings, but this assumption is implicit.**

`indexOfNextPrefix` only considers `"-FE-"` and `"-WI-"` (not `qst_`). The `case strings.HasPrefix(rest, "qst_")` branch at line 118 reads until the next `-FE-` or `-WI-`. This works today because `quest.GenerateID` produces `qst_YYYYMMDD_xxxxxx` with `xxxxxx` drawn from `[a-z0-9]` (see `internal/quest/slug.go:18`), so no quest ID can contain `-`.

If a future change ever loosens the quest alphabet to include `-`, or if a quest ID arrives from an external source with embedded hyphens (e.g. a synthetic test ID like `qst_FE-something`), this parser silently corrupts the result. The 100-iteration property test at `campaign_tag_full_test.go:144` only generates hex-style quest IDs, so this case is uncovered.

Two small mitigations, pick one:

1. **Document the invariant** explicitly above `ParseTag`:
   ```go
   // ParseTag assumes quest IDs match `qst_[0-9]+_[a-z0-9]+` per
   // internal/quest/slug.go. If the quest alphabet is ever extended to
   // include `-`, update indexOfNextPrefix and add adversarial test cases.
   ```
2. **Add an adversarial test case**: a quest ID with `-FE-` in it, assert that ParseTag rejects or returns an empty struct rather than mis-parsing. This pins the invariant in CI.

Without one of these, the next person who edits `questAlphabet` ships a silent parser bug.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview

Three tasks in one PR and the scope is held tightly: deterministic ref derivation, idempotent doctor backfill, and a composable tag grammar with a two-stage parser. The schema bump is conservative (v1alpha4/v1alpha5/v1alpha6 all load) and the test surface is real (5 unit tests on ref derivation, a 100-iteration property test on tag round-trip, multi-workitem collision coverage in doctor --fix). `camperrors` and `context.Context` discipline carry through. What blocks approval are two correctness gaps in the backfill path (no file lock, silent error drops carried over from PR #303), one contract ambiguity in the tag formatter that will land in commit wrappers in the next sequence, and a handful of robustness gaps that are cheap to close now.

## Key Findings

1. **`backfillRef` has no file lock.** `os.ReadFile` → `WriteFileAtomically` is a TOCTOU window; concurrent doctor + create can clobber each other. Lift `acquireFileLock` from `internal/workitem/links/save.go` (already established in PR #303) into a shared helper. (inline `doctor_ref_backfill.go:84`)

2. **`workitemPathsMissingRef` errors silently dropped in doctor finding collection.** Scan failure produces zero findings; doctor reports a clean tree while in fact it could not check. Same antipattern as PR #303 doctor. (inline `doctor.go:179`)

3. **`backfillRef` errors silently dropped in autoFix dispatch.** `if backfillRef(...) == nil { applied++ }` swallows the cause. User sees "applied N/M" with no clue why M-N failed; next run will surface the same findings again. (inline `doctor.go:236`)

4. **Formatter accepts both prefixed and unprefixed workitem refs and produces different round-trip semantics for each.** Today every internal caller passes the canonical `WI-<hex>` form, so the bug does not fire — but sequence-03 task-04 wires this into commit wrappers where the contract gap becomes a real footgun. Reject non-canonical input or normalize on the way in. (inline `campaign_tag.go:37`)

5. **`DeriveUnique` has an unbounded loop with no `ctx.Err()` check.** Per project standard "Always check `ctx.Err()` before starting long operations". Bounded iteration with an error return is cheap. (inline `ref.go:39`)

6. **O(N²) `wkitem.Discover` pattern in backfill.** `refsExcludingPath` re-walks the entire workitem tree once per workitem being fixed. A migration of 50 v1alpha5 workitems triggers 50 full discovers. One discover + an in-memory set with file locking gives both correctness and good asymptotic behavior. (inline `doctor_ref_backfill.go:108`)

7. **Schema-version error message went stale.** Bumping the constant to v1alpha6 and adding it to `acceptedWorkitemVersions` did not update the literal in `validateMetadata` that says `supported: v1alpha4, v1alpha5`. Derive the list from the map so the next bump cannot drift. (inline `metadata.go:27`)

8. **`insertScalarAfter` does not handle an existing empty `ref:` line and produces duplicate keys.** Test `TestDoctor_FixBackfillsRefAndPreservesFields` asserts the title is present but does not lock down indent or byte-level diff minimality — yaml-v3 default indent is 4-space and could silently change every line in `.workitem` files written with 2-space. (inline `doctor_ref_backfill.go:153`)

9. **`ParseTag` carries an undocumented invariant that quest IDs never contain `-FE-` or `-WI-`.** Holds today because of the quest alphabet, but the property test never exercises adversarial quest IDs. Either document the constraint above `ParseTag` or add a negative test case so the assumption is pinned in CI. (inline `campaign_tag.go:144`)

## What is Done Well

- `wkitem.Derive` is deterministic, well-documented, and explicitly safe to recompute (the comment calls out this property and the tests pin it down).
- Backward compat is honest: `acceptedWorkitemVersions` keeps v1alpha4/v1alpha5 readable, and the test `TestLoadMetadata_LegacyV1Alpha5StillLoads` proves it.
- `ApplyMetadata` keeps the additive surface tight — existing readers that ignore `SourceMetadata["ref"]`/`["quest_id"]` are completely unaffected.
- `yaml.Node`-based rewrite is the right tool for the diff-preservation goal. Insert-after-`id` keeps the change to one line in practice.
- Quest resolution failure is non-fatal in create/adopt — a quest-unaware user is not blocked. The stderr warning is the correct UX.
- Two-stage tag parser is genuinely clever: regex peels the bracket; the inner state machine disambiguates `FE-<ref>-WI-<ref>` from `FE-<hyphenated ref>` without RE2 lookahead. Worth the careful design.
- 100-iteration round-trip property test on the tag composer + parser is the right level of confidence for a public commit-message contract.
- `commitkit.PrependContextTagsFull` and `commitkit.ParseTag` are clean re-exports, keeping fest et al. on the stable public API.
- Doctor finding code `workitem.ref.missing` slots cleanly into the dotted-domain vocabulary already established in PR #302/#303.

## Staff Standard

Would I submit this as my best work at a staff level? The deterministic-ref design and the tag parser would pass; the backfill plumbing would not. The two silent-error-drop sites repeat the exact antipattern flagged on PR #303, and the missing file lock around `.workitem` rewrites is the kind of correctness gap that produces "my workitem field disappeared" tickets months after merge. Address the 9 inline comments, especially #1 (file lock), #2-#3 (error visibility), and #4 (tag formatter contract), and this becomes a clean approve.

obey-agent (approved):
## Verdict: Approve

## Overview

Every blocker from the previous round is addressed substantively. The TOCTOU window on backfills is closed via the shared `fsutil.AcquireFileLock` lifted in PR #303, the O(N²) discover pattern is replaced with a single-pass `backfillMissingRefs` that maintains an in-memory ref set, errors no longer get swallowed at either the scan or the apply step, `DeriveUnique` is bounded with ctx checks, the tag formatter normalizes non-canonical workitem refs, and `insertScalarAfter` no longer produces duplicate keys.

## Findings Resolved

1. **Schema-version error message derives from the source map.** `supportedWorkitemVersions()` reads `acceptedWorkitemVersions`, sorts, and joins. The next schema bump cannot drift the literal out of sync.

2. **`backfillRefWithRef` acquires a file lock via `fsutil.AcquireFileLock`** (the shared helper from PR #303). TOCTOU window closed; concurrent doctor + create races on the same `.workitem` are serialized.

3. **`backfillMissingRefs` does ONE Discover** and maintains an in-memory `existingRefs` set updated as it goes. The pending list is sorted by `RelativePath` for deterministic collision-retry ordering. O(N²) → O(N).

4. **`workitemPathsMissingRef` scan failures now emit `workitem.scan.failed`** (severity `error`) instead of silently passing. Operators see a real signal when the registry could not be checked.

5. **`autoFixWorkitemFindings` collects pending backfills**, calls the batched `backfillMissingRefs`, and on error emits `fmt.Fprintf(errw, "warning: backfill refs: %v\n", err)` to stderr. Partial counts still propagate via `applied`.

6. **`DeriveUnique` accepts `ctx`** and checks `ctx.Err()` at entry and inside the retry loop. Bounded by `maxRefCollisionRetries = 1 << 24`; returns a validation error when exhausted instead of spinning forever.

7. **`FormatContextTagsFull` normalizes `workitemRef`** by prepending `WI-` when missing (line 33-35). Callers passing `"abcdef"` and `"WI-abcdef"` now produce identical tags. The doctype comment kept its explanation of why the bracket reads `WI-WI-abcdef`.

8. **`insertScalarAfter` detects existing key and updates in place.** Lines 211-219 walk the mapping once: if the key exists, set `v.Value = value` and return; otherwise track the insertion point. No more duplicate-key risk on partially-written `.workitem` files.

9. **`ParseTag` documents the quest-id invariant.** Comment now explicitly states the parser assumes `qst_[0-9]+_[a-z0-9]+` per `internal/quest/slug.go`, with a hint to update `indexOfNextPrefix` if the alphabet ever grows. Future maintainers cannot miss this assumption.

- `backfillMissingRefs` is the right architectural shape: one Discover, sort, then linear pass with the set updated in lockstep with the writes. Composes correctly with the file-lock from finding #2 — each individual write is serialized, but the batch does not re-scan after every fix.
- `insertScalarAfter` is now idempotent. Re-running `doctor --fix` after a partial write does not corrupt the YAML — the second invocation hits the existing-key branch and either no-ops (same value) or updates in place.
- Error paths are honest: `autoFixWorkitemFindings` does not swallow on backfill failure, but it does return the partial count so the operator sees how many fixes did land before the failure.
- The tag normalization at `FormatContextTagsFull:33-35` keeps the property test green AND closes the contract gap, without changing the existing wire format for callers passing canonical input.
- `ref_quest.go:25` calls `DeriveUnique(ctx, ...)` with the new signature — the ctx threading is end-to-end. Cancellation propagates from CLI through to the inner hash retries.

## Verification

- `just test unit` — 2454/2454 passed
- `just test integration-run TestIntegration_CommitTags` — 4/4 applicable passed (`CampCommit`, `CampPCommit`, `NoContext`, `ExplicitOverride`); `FestCommit` and `QuestAndWorkitem` are SKIP-with-reason pending fest-side wiring, documented in the test files

## Staff Standard

Would I submit this as my best work at a staff level? Yes. The lock + single-pass + bounded-loop + honest-error shape is exactly the kind of correctness/observability discipline that holds up at staff review. Ready to merge.

Issue #105, opened 2026-02-28
Bug: camp clone reports false success for submodules due to parallel init race


davidlovas:
## Summary

`camp clone` can report submodules as successfully initialized when they actually aren't. This is caused by a race condition in the parallel submodule initialization logic.

## Reproduction

```
camp clone git@github.com:Obedience-Corp/Obey-campaign.git
```

Output shows `20/22 initialized` during Phase 3, but validation correctly detects **5** uninitialized submodules — a discrepancy of 3 false successes.

## Root Cause

Two interacting issues in `initSubmoduleGraceful` (`internal/clone/git.go`):

### 1. `git submodule init` error silently discarded

```go
// Line 367-368
cmd := exec.CommandContext(ctx, "git", "-C", repoDir, "submodule", "init", subPath)
_ = cmd.Run() // Ignore error, may already be initialized
```

Under parallel execution (default 4 goroutines), concurrent `git submodule init` calls compete for the `.git/config` file lock. Some fail due to lock contention, but the error is silently swallowed.

### 2. `git submodule update` (without `--init`) returns exit 0 for unregistered submodules

```
$ git submodule update projects/agent-simulator
Submodule path 'projects/agent-simulator' not initialized
Maybe you want to use 'update --init'?
$ echo $?
0
```

Git prints a warning but returns **exit code 0**, so the code treats it as success.

### Combined effect

1. `git submodule init <path>` fails due to config lock contention → error ignored
2. `git submodule update <path>` returns exit 0 (does nothing) → counted as success
3. Phase 3 reports success, validation later correctly detects the submodule is not initialized
4. User sees contradictory output: "20/22 initialized" but 5 validation failures

## Suggested Fix

Replace the separate `init` + `update` with a single atomic `git submodule update --init`:

```go
// Before (race-prone):
cmd := exec.CommandContext(ctx, "git", "-C", repoDir, "submodule", "init", subPath)
_ = cmd.Run()
cmd = exec.CommandContext(ctx, "git", "-C", repoDir, "submodule", "update", subPath)

// After (safe):
cmd := exec.CommandContext(ctx, "git", "-C", repoDir, "submodule", "update", "--init", subPath)
```

`--init` makes the update command atomically register and clone the submodule, eliminating the lock contention window. All existing tests pass with this change.

## Additional Note

Separately from this bug, the Obey-campaign repo has 3 submodules pointing to repos that no longer exist:

| Submodule | URL | Status |
|-----------|-----|--------|
[...]

PR #305, opened 2026-05-25
feat(workitem): camp workitem commit + commits (CW0003 seq-04)

lancekrogers:
## Summary

Sequence 04 of the CW0003 festival: camp workitems become a first-class execution context with scoped commits and a cross-repo query.

**Stacked on:** #304 (seq-03). Re-base onto main after #303 and #304 land.

## What landed

- `commit.Workitem` in `internal/git/commit/` — thin mirror of `commit.Quest` that wraps `doCommit` and surfaces `WorkitemRef` into the consolidated campaign tag.
- `camp workitem commit [selector]` — scoped commit command. Resolves the active workitem (positional, `--workitem`, or cwd), computes a staging plan, prints it to stderr, and hands the file scope to `commit.Workitem` with `SelectiveOnly` so empty plans never widen to `git add .` at the campaign root.
- `camp workitem commits [selector]` — cross-repo query. Fans out `git log --grep=-WI-<ref>` across the campaign root and every linked project/repo/worktree with a bounded worker pool, then filters via `commitkit.ParseTag` to reject incidental string matches.
- Staging matrix design contract at `internal/commands/workitem/COMMIT_DESIGN.md` plus sample plan fixtures under `internal/commands/workitem/testdata/plans/`.
- Six containerized integration tests in `tests/integration/workitem_commit_test.go` — the load-bearing one is `TestIntegration_WorkitemCommit_RefusesSilentWiden` (any path outside the workitem dir leaking into HEAD fails loudly).

## Routing logic

The planner branches by (in order):
1. `--staged` short-circuits — commit the current index.
2. `--project <name>` — force project-repo context.
3. cwd-first: if cwd is inside a sub-git-repo, stage from there (covers the natural `cd projects/<name>; camp workitem commit <selector>` flow).
4. Resolver tier: `SourceLink` → linked-project plan, `SourceFestival` → festival plan, default → workitem-dir plan scoped at campaign root.

`--include` / `--exclude` apply after the matrix.

## Test plan

- [x] `go build ./...`
- [x] `just test unit` — 2458/2458
- [x] `just test integration-run TestIntegration_WorkitemCommit` — 6/6

### DIFF
--- a/internal/commands/workitem/commit.go
+++ b/internal/commands/workitem/commit.go
@@ -1,0 +1,215 @@
+package workitem
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/git/commit"
+)
+
+// noContextHint is the multi-line refusal message printed when the resolver
+// returns no workitem. Mirrors COMMIT_DESIGN.md §7.
+const noContextHint = `no workitem context resolved from cwd
+
+Try one of:
+  --workitem <selector>            explicit workitem to scope this commit
+  cd workflow/<type>/<slug> && ...  run from inside a workitem directory
+  camp workitem current <selector>  set a session-wide default`
+
+func newCommitCommand() *cobra.Command {
+	var (
+		flagMessage                 string
+		flagWorkitem                string
+		flagProject                 string
+		flagIncludes                []string
+		flagExcludes                []string
+		flagStaged                  bool
+		flagIncludeSubmodulePointer bool
+		flagDryRun                  bool
+		flagJSON                    bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "commit",
+		Short: "Commit changes scoped to the resolved workitem",
+		Long: `Stage and commit changes belonging to a resolved workitem.
+
+The staging plan is computed from the resolver context (cwd-aware, with
+explicit --workitem or --project overrides) and printed to stderr before
+the commit runs. The plan never silently widens to "git add ." at the
+campaign root.
+
+See internal/commands/workitem/COMMIT_DESIGN.md for the full staging matrix
+and flag precedence.`,
+		Annotations: map[string]string{
+			"agent_allowed": "true",
+			"agent_reason":  "Scoped commit command; honors --json and --dry-run for automation",
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			return runCommit(ctx, cmd, commitFlags{
+				Message:                 flagMessage,
+				Workitem:                flagWorkitem,
+				Project:                 flagProject,
+				Includes:                flagIncludes,
+				Excludes:                flagExcludes,
+				Staged:                  flagStaged,
+				IncludeSubmodulePointer: flagIncludeSubmodulePointer,
+				DryRun:                  flagDryRun,
+				JSON:                    flagJSON,
+			})
+		},
+	}
+	cmd.Flags().StringVarP(&flagMessage, "message", "m", "", "commit message (required unless --dry-run)")
+	cmd.Flags().StringVar(&flagWorkitem, "workitem", "", "explicit workitem selector (overrides cwd-based resolution)")
+	cmd.Flags().StringVar(&flagProject, "project", "", "force project-repo context by name (skips resolver)")
+	cmd.Flags().StringArrayVar(&flagIncludes, "include", nil, "additional path to stage (repeatable; relative to repo root)")
+	cmd.Flags().StringArrayVar(&flagExcludes, "exclude", nil, "path to remove from the staging plan (repeatable)")
+	cmd.Flags().BoolVar(&flagStaged, "staged", false, "commit whatever is already in the git index")
+	cmd.Flags().BoolVar(&flagIncludeSubmodulePointer, "include-submodule-pointer", false, "include dirty project submodule pointers in the plan")
+	cmd.Flags().BoolVar(&flagDryRun, "dry-run", false, "print the staging plan and exit without committing")
+	cmd.Flags().BoolVar(&flagJSON, "json", false, "emit the staging plan and commit result as JSON on stdout")
+	return cmd
+}
+
+type commitFlags struct {
+	Message                 string
+	Workitem                string
+	Project                 string
+	Includes                []string
+	Excludes                []string
+	Staged                  bool
+	IncludeSubmodulePointer bool
+	DryRun                  bool
+	JSON                    bool
+}
+
+func runCommit(ctx context.Context, cmd *cobra.Command, flags commitFlags) error {
+	if flags.Message == "" && !flags.DryRun {
+		return camperrors.NewValidation("message", "commit message required (use -m \"<msg>\")", nil)
+	}
+
+	cfg, campaignRoot, err := config.LoadCampaignConfigFromCwd(ctx)
+	if err != nil {
+		return camperrors.Wrap(err, "not in a campaign directory")
+	}
+
+	plan, err := ComputePlan(ctx, campaignRoot, PlanOptions{
+		Explicit:                flags.Workitem,
+		Project:                 flags.Project,
+		Includes:                flags.Includes,
+		Excludes:                flags.Excludes,
+		StagedOnly:              flags.Staged,
+		IncludeSubmodulePointer: flags.IncludeSubmodulePointer,
+		CampaignID:              cfg.ID,
+	})
+	if err != nil {
+		if errors.Is(err, ErrNoWorkitemContext) {
+			fmt.Fprintln(cmd.ErrOrStderr(), noContextHint)
+			os.Exit(2)
+		}
+		return err
+	}
+
+	if !flags.JSON {
+		if perr := PrintPlan(cmd.ErrOrStderr(), plan); perr != nil {
+			return perr
+		}
+	}
+
+	if len(plan.Stage) == 0 && len(plan.PreStaged) == 0 {
+		return camperrors.NewValidation("plan",
+			"empty staging plan; pass --include <path> or run from inside a changed workitem", nil)
+	}
+
+	if flags.DryRun {
+		if flags.JSON {
+			return emitJSON(cmd.OutOrStdout(), plan, "")
+		}
+		return nil
+	}
+
+	res := commit.Workitem(ctx, commit.WorkitemOptions{
+		Options: commit.Options{
+			CampaignRoot:  plan.RepoRoot,
+			CampaignID:    cfg.ID,
+			Files:         plan.Stage,
+			PreStaged:     plan.PreStaged,
+			SelectiveOnly: true,
+		},
+		Action:      commit.WorkitemScope,
+		WorkitemID:  plan.Workitem.StableID,
+		WorkitemRef: plan.WorkitemRef,
+		QuestID:     plan.QuestID,
+		Title:       flags.Message,
+	})
+	if res.Err != nil {
+		return res.Err
+	}
+	if res.NoChanges {
+		if flags.JSON {
+			return emitJSON(cmd.OutOrStdout(), plan, "")
+		}
+		fmt.Fprintln(cmd.OutOrStdout(), res.Message)
+		return nil
+	}
+
+	sha, _ := lastCommitSHA(ctx, plan.RepoRoot)
+	if flags.JSON {
+		return emitJSON(cmd.OutOrStdout(), plan, sha)
+	}
+	fmt.Fprintf(cmd.OutOrStdout(), "committed %s\n", sha)
+	return nil
+}
+
+func emitJSON(w writeFlusher, plan *StagingPlan, sha string) error {
+	payload := struct {
+		Workitem    string         `json:"workitem"`
+		Ref         string         `json:"workitem_ref,omitempty"`
+		QuestID     string         `json:"quest_id,omitempty"`
+		Tag         string         `json:"tag"`
+		Context     PlanContext    `json:"context"`
+		ContextNote string         `json:"context_note,omitempty"`
+		RepoRoot    string         `json:"repo_root"`
+		Stage       []string       `json:"stage"`
+		PreStaged   []string       `json:"pre_staged,omitempty"`
+		Skip        []SkippedEntry `json:"skip,omitempty"`
+		SHA         string         `json:"sha,omitempty"`
+	}{
+		Workitem:    plan.Workitem.StableID,
+		Ref:         plan.WorkitemRef,
+		QuestID:     plan.QuestID,
+		Tag:         plan.Tag,
+		Context:     plan.Context,
+		ContextNote: plan.ContextNote,
+		RepoRoot:    plan.RepoRoot,
+		Stage:       plan.Stage,
+		PreStaged:   plan.PreStaged,
+		Skip:        plan.Skip,
+		SHA:         sha,
+	}
+	enc := json.NewEncoder(w)
+	enc.SetIndent("", "  ")
+	return enc.Encode(payload)
+}
+
+type writeFlusher interface {
+	Write([]byte) (int, error)
+}
+
+func lastCommitSHA(ctx context.Context, repoRoot string) (string, error) {
+	out, err := exec.CommandContext(ctx, "git", "-C", repoRoot, "rev-parse", "HEAD").Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(out)), nil
+}
--- a/internal/commands/workitem/commit.go
+++ b/internal/commands/workitem/commit.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"os"
 	"os/exec"
 	"strings"
 
@@ -120,7 +119,7 @@ func runCommit(ctx context.Context, cmd *cobra.Command, flags commitFlags) error
 	if err != nil {
 		if errors.Is(err, ErrNoWorkitemContext) {
 			fmt.Fprintln(cmd.ErrOrStderr(), noContextHint)
-			os.Exit(2)
+			return camperrors.NewCommand("camp workitem commit", 2, "", err)
 		}
 		return err
 	}
--- a/internal/commands/workitem/commit.go
+++ b/internal/commands/workitem/commit.go
@@ -39,26 +39,31 @@ func newCommitCommand() *cobra.Command {
 	)
 
 	cmd := &cobra.Command{
-		Use:   "commit",
+		Use:   "commit [selector]",
 		Short: "Commit changes scoped to the resolved workitem",
 		Long: `Stage and commit changes belonging to a resolved workitem.
 
 The staging plan is computed from the resolver context (cwd-aware, with
-explicit --workitem or --project overrides) and printed to stderr before
-the commit runs. The plan never silently widens to "git add ." at the
+explicit positional <selector> or --project overrides) and printed to stderr
+before the commit runs. The plan never silently widens to "git add ." at the
 campaign root.
 
 See internal/commands/workitem/COMMIT_DESIGN.md for the full staging matrix
 and flag precedence.`,
+		Args: cobra.MaximumNArgs(1),
 		Annotations: map[string]string{
 			"agent_allowed": "true",
 			"agent_reason":  "Scoped commit command; honors --json and --dry-run for automation",
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
+			selector := flagWorkitem
+			if selector == "" && len(args) == 1 {
+				selector = args[0]
+			}
 			return runCommit(ctx, cmd, commitFlags{
 				Message:                 flagMessage,
-				Workitem:                flagWorkitem,
+				Workitem:                selector,
 				Project:                 flagProject,
 				Includes:                flagIncludes,
 				Excludes:                flagExcludes,
--- a/internal/commands/workitem/commits.go
+++ b/internal/commands/workitem/commits.go
@@ -1,0 +1,318 @@
+package workitem
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"sort"
+	"strings"
+	"sync"
+	"text/tabwriter"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/workitem/links"
+	"github.com/Obedience-Corp/camp/internal/workitem/resolver"
+	"github.com/Obedience-Corp/camp/pkg/commitkit"
+)
+
+const (
+	commitsDefaultLimit = 100
+	commitsLogFormat    = "%H%x09%an%x09%aI%x09%s"
+)
+
+// CommitRecord is the per-row payload emitted by `camp workitem commits`.
+// Repo is the campaign-relative path of the git repo the commit was found
+// in (".") for the campaign root.
+type CommitRecord struct {
+	SHA      string                  `json:"sha"`
+	Author   string                  `json:"author"`
+	Date     time.Time               `json:"date"`
+	Subject  string                  `json:"subject"`
+	Repo     string                  `json:"repo"`
+	TagParts commitkit.TagComponents `json:"tag,omitempty"`
+}
+
+// commitsQueryError records a per-repo failure surfaced under the --json
+// `errors` key. Default table output drops these silently to keep the
+// happy-path output readable.
+type commitsQueryError struct {
+	Repo string `json:"repo"`
+	Err  string `json:"error"`
+}
+
+func newCommitsCommand() *cobra.Command {
+	var (
+		flagRef      string
+		flagJSON     bool
+		flagLimit    int
+		flagOffset   int
+		flagWorkitem string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "commits [selector]",
+		Short: "List commits referencing a workitem across linked repos",
+		Long: `Search the campaign root and every linked project/repo/worktree/festival
+repo for commits whose campaign tag references this workitem's ref.
+
+Default sort: most recent first across all repos. Use --json for structured
+output. Repos that are not git checkouts or that fail their git log invocation
+are skipped silently in table mode and reported under "errors" in JSON mode.`,
+		Args: cobra.MaximumNArgs(1),
+		Annotations: map[string]string{
+			"agent_allowed": "true",
+			"agent_reason":  "Read-only query with --json output for automation",
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			selector := ""
+			if len(args) == 1 {
+				selector = args[0]
+			}
+			if flagWorkitem != "" {
+				selector = flagWorkitem
+			}
+			return runCommitsQuery(ctx, cmd, commitsFlags{
+				Selector: selector,
+				Ref:      flagRef,
+				JSON:     flagJSON,
+				Limit:    flagLimit,
+				Offset:   flagOffset,
+			})
+		},
+	}
+	cmd.Flags().StringVar(&flagRef, "ref", "", "query by workitem ref directly (e.g. WI-abc123) — skips resolver")
+	cmd.Flags().StringVar(&flagWorkitem, "workitem", "", "alias for the positional <selector>")
+	cmd.Flags().BoolVar(&flagJSON, "json", false, "emit JSON instead of the default table")
+	cmd.Flags().IntVar(&flagLimit, "limit", commitsDefaultLimit, "maximum commits to return")
+	cmd.Flags().IntVar(&flagOffset, "offset", 0, "number of commits to skip (after sorting)")
+	return cmd
+}
+
+type commitsFlags struct {
+	Selector string
+	Ref      string
+	JSON     bool
+	Limit    int
+	Offset   int
+}
+
+func runCommitsQuery(ctx context.Context, cmd *cobra.Command, flags commitsFlags) error {
+	cfg, campaignRoot, err := config.LoadCampaignConfigFromCwd(ctx)
+	if err != nil {
+		return camperrors.Wrap(err, "not in a campaign directory")
+	}
+	_ = cfg
+
+	ref := flags.Ref
+	if ref == "" {
+		res, rerr := resolver.Resolve(ctx, campaignRoot, resolver.Options{Explicit: flags.Selector})
+		if rerr != nil {
+			return camperrors.Wrap(rerr, "resolve workitem")
+		}
+		if res == nil || res.Workitem == nil {
+			return camperrors.NewValidation("workitem",
+				"no workitem context resolved; pass <selector> or --ref WI-...", nil)
+		}
+		ref = refOf(res.Workitem)
+		if ref == "" {
+			return camperrors.NewValidation("workitem",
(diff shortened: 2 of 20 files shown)


internal/commands/workitem/commit.go
```diff
@@ -0,0 +1,220 @@
...
+		Includes:                flags.Includes,
+		Excludes:                flags.Excludes,
+		StagedOnly:              flags.Staged,
+		IncludeSubmodulePointer: flags.IncludeSubmodulePointer,
+		CampaignID:              cfg.ID,
+	})
+	if err != nil {
+		if errors.Is(err, ErrNoWorkitemContext) {
+			fmt.Fprintln(cmd.ErrOrStderr(), noContextHint)
+			os.Exit(2)
```
obey-agent: **`os.Exit(2)` from within a cobra `RunE` is wrong.** Three problems:

1. **Skips deferred cleanup.** Any `defer cancel()` / `defer file.Close()` between `runCommit` start and this point is bypassed.
2. **Bypasses `cmd/camp/main.go` exit-code routing.** The codebase already has the right mechanism: `camperrors.CommandError{Err: ..., ExitCode: ...}` propagated through `RunE → cobra.Execute() → main.go`, where line 14-16 maps it to `os.Exit(cmdErr.ExitCode)`. This is the same gap I flagged on PR #302 (`doctor.go:55`) and PR #303 (`show.go:21`). The fix is established in the codebase; please use it.
3. **Untested.** Integration tests cannot distinguish exit 1 from exit 2 today (`require.Error` does not assert the code), so this slipped through.

Fix:

```go
if errors.Is(err, ErrNoWorkitemContext) {
    fmt.Fprintln(cmd.ErrOrStderr(), noContextHint)
    return &camperrors.CommandError{Err: err, ExitCode: 2}
}
```

The PR description specifically calls out exit 2 as the contract (`# Routing logic` section implies it via the resolver). Make sure an integration test captures the actual exit code via `tc.ExecCommand` returning `(out, exitCode, err)` and asserts `2`.

internal/commands/workitem/staging.go
```diff
@@ -0,0 +1,571 @@
...
+		if len(line) < 4 {
+			continue
+		}
+		path := strings.TrimSpace(line[3:])
+		// Rename entries look like "old -> new"; take the new path.
+		if i := strings.Index(path, " -> "); i >= 0 {
+			path = path[i+4:]
+		}
+		files = append(files, path)
+	}
```
obey-agent: **`git status --porcelain` without `-z` is not safe for arbitrary filenames.** Porcelain v1 quotes paths containing spaces, tabs, newlines, or non-printable bytes (controlled by `core.quotePath`). The parser at lines 423-436 does not unquote, so:

- A file like `path with"quote"` arrives as `"path with\"quote\""` and gets staged verbatim — the resulting `git add` either fails or stages the wrong path.
- A renamed file `old name -> new name` (where the names contain spaces) confuses the `-> ` split at line 432.
- A file containing a literal newline gets split across two lines and treated as two paths, both broken.

Robust shape:

```go
args := []string{"-C", repoRoot, "status", "--porcelain", "-z", "--untracked-files=all"}
if prefix != "" {
    args = append(args, "--", prefix)
}
out, err := exec.CommandContext(ctx, "git", args...).Output()
// ... split on \0; for renames, the entry is XY space old\0new\0
```

With `-z`, paths are NUL-terminated and never quoted. Renames are emitted as two NUL-separated entries (old then new). The parsing flips from "fragile string parsing" to "split on \0".

Same fix applies to `pathIsDirty` (line 455) and `listDirtySubmodulePointers` (line 463). The latter is especially exposed because submodule paths are user-controlled.

`TestIntegration_WorkitemCommit_RefusesSilentWiden` is described as load-bearing, but I would bet it does not exercise a filename with a space. Worth adding `workflow/feature/test slug with spaces/` to the fixture before this lands.

internal/commands/workitem/commits.go
```diff
@@ -0,0 +1,317 @@
...
+		"-C", repo,
+		"log", "--all",
+		"--pretty=format:"+commitsLogFormat,
+		"--grep="+grep,
+	)
+	output, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+	rel := filepath.Base(repo)
```
obey-agent: **The `Repo` field doc says one thing and the implementation does another.**

Doc (line 34-41):
> Repo is the campaign-relative path of the git repo the commit was found in (".") for the campaign root.

Implementation (line 255):
```go
rel := filepath.Base(repo)
```

Concrete collision: a campaign with both `projects/foo` and `projects/worktrees/foo` will emit two `CommitRecord.Repo = "foo"` rows that the user cannot distinguish. The output table renders `foo` for both; the JSON consumer cannot route commits back to the right tree.

Fix: thread the campaign root through `queryRepo` and compute the actual relative path:

```go
func queryRepo(ctx context.Context, campaignRoot, repo, ref string) ([]CommitRecord, error) {
    // ...
    rel, err := filepath.Rel(campaignRoot, repo)
    if err != nil || rel == "" {
        rel = filepath.Base(repo) // fallback for non-campaign roots
    }
    if rel == "." {
        rel = "." // explicit "campaign root" marker per the doc
    }
    // ...
}
```

Either fix the implementation or fix the doc — but the current state is a silent contract lie that bites the first consumer with name-colliding repos.

internal/commands/workitem/staging.go
```diff
@@ -0,0 +1,571 @@
...
+	plan.ContextNote = workitemDirNote(plan.Workitem, src)
+
+	stage, err := listChangedFilesUnder(ctx, root, plan.Workitem.RelativePath)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "list workitem changes")
+	}
+
+	if dirty, err := pathIsDirty(ctx, root, linkRegistryRelPath); err == nil && dirty {
+		stage = append(stage, linkRegistryRelPath)
+	}
```
obey-agent: **Silent auto-stage of `linkRegistryRelPath` contradicts the PR's "never silently widens" framing.** Lines 215-217 here (and the matching line in `computeFestivalPlan` at 276-278) silently bundle `.campaign/workitems/links.yaml` into the commit whenever it's dirty. A user running `camp workitem commit -m "fix bug"` whose link registry happens to be dirty from a previous `camp workitem link` call will see two unrelated changes in HEAD with no warning.

This is the opposite of the contract the PR description sells:

> The plan never silently widens to "git add ." at the campaign root.

Three ways to make it honest:

1. **Drop the auto-stage** — require `--include .campaign/workitems/links.yaml` for link-registry changes.
2. **Surface it in the plan output** — `PrintPlan` already shows what's being staged; add a line like `  +  .campaign/workitems/links.yaml (link registry auto-included; --exclude to drop)`.
3. **Gate behind a flag** — `--include-link-registry` default-on or default-off, but explicit.

(2) is the minimum: even keeping the current behavior, the user sees the inclusion in the stage list before the commit fires. Today the stage list shows the path but no annotation that it came from auto-detection rather than the workitem scope. Add a `SkippedEntry`-style annotation.

Also: `TestIntegration_WorkitemCommit_RefusesSilentWiden` should cover this case explicitly — if the rule is "never silently widens", a test that proves a dirty link registry does NOT silently appear in HEAD (or DOES with explicit annotation) needs to exist.

internal/commands/workitem/commits.go
```diff
@@ -0,0 +1,317 @@
...
+		records = nil
+	}
+	if flags.Limit > 0 && flags.Limit < len(records) {
+		records = records[:flags.Limit]
+	}
+
+	if flags.JSON {
+		return emitCommitsJSON(cmd.OutOrStdout(), records, queryErrs)
+	}
+	return emitCommitsTable(cmd.OutOrStdout(), records)
```
obey-agent: **Table mode silently drops per-repo failures.** `runCommitsQuery` returns both `records` and `queryErrs`. The JSON path emits errors under `errors`; the table path (line 151) does not. A user running `camp workitem commits foo` whose query failed on 3 of 12 linked repos sees fewer commits with zero indication that anything went wrong.

Same pattern: if the load-bearing finding here is "commits referencing this workitem", the consumer needs to know when the search was incomplete.

Minimum fix in `emitCommitsTable`:

```go
func emitCommitsTable(w io.Writer, records []CommitRecord, errs []commitsQueryError) error {
    if len(records) == 0 && len(errs) == 0 {
        fmt.Fprintln(w, "no commits found")
        return nil
    }
    // ... existing rendering ...
    if len(errs) > 0 {
        fmt.Fprintf(w, "\nwarning: %d repo(s) failed; re-run with --json for details\n", len(errs))
    }
    return tw.Flush()
}
```

Also: the warning message should be on `cmd.ErrOrStderr()` not stdout so consumers piping to `grep` / `wc` don't see it. Right now `emitCommitsTable` writes everything to the same writer — either keep that pattern but route the warning to stderr explicitly, or have the runner emit the warning around the table call.

Side note: the `isGitRepo` skip (line 283-286) silently swallows ALL errors as "not a git repo" — permission denied, missing git binary, etc. all collapse to "no rows". Worth distinguishing: log the unexpected ones to stderr or include them in `queryErrs` so they surface in JSON mode.

internal/commands/workitem/commits.go
```diff
@@ -0,0 +1,317 @@
...
+		return nil, nil
+	}
+	grep := "-WI-" + ref
+	cmd := exec.CommandContext(ctx, "git",
+		"-C", repo,
+		"log", "--all",
+		"--pretty=format:"+commitsLogFormat,
+		"--grep="+grep,
+	)
+	output, err := cmd.Output()
```
obey-agent, line 279: **No per-repo timeout in the fan-out.** Each `queryRepo` runs `git log --all --grep=...` through `exec.CommandContext(ctx, ...)`, but `ctx` is the parent context with no per-call deadline. A hanging git invocation (network filesystem stall, a misconfigured submodule, a `git log` against a 500k-commit history with a slow grep) blocks the entire query until the user Ctrl-C's the whole command.

The fan-out is the wrong layer to cap; the inner call needs a deadline.

Fix:

```go
const perRepoTimeout = 30 * time.Second

func queryRepo(ctx context.Context, repo, ref string) ([]CommitRecord, error) {
    if !isGitRepo(repo) {
        return nil, nil
    }
    cctx, cancel := context.WithTimeout(ctx, perRepoTimeout)
    defer cancel()
    cmd := exec.CommandContext(cctx, "git", /* ... */)
    output, err := cmd.Output()
    if cctx.Err() == context.DeadlineExceeded {
        return nil, fmt.Errorf("git log timeout after %s", perRepoTimeout)
    }
    // ... rest
}
```

This pairs with the table-mode error surfacing comment: a timed-out repo becomes a `commitsQueryError`, the user gets `warning: 1 repo failed` plus the timeout reason via `--json`. Without this, the whole command hangs on one bad repo and the user has no signal which one.

Separately: the worker pool uses `runtime.NumCPU()` (line 187) as the worker cap. Git log is IO-bound, not CPU-bound; on a 32-core box this spawns 32 git invocations even when most repos are small. Cap at `min(NumCPU, 8)` to keep the fan-out from thrashing IO.

internal/commands/workitem/staging.go
```diff
@@ -0,0 +1,571 @@
...
+	return out
+}
+
+// ErrNoWorkitemContext is returned when ComputePlan cannot resolve any
+// workitem context. The CLI maps this to exit code 2 with a help hint.
+var ErrNoWorkitemContext = camperrors.NewValidation(
+	"workitem",
+	"no workitem context resolved",
+	nil,
+)
```
obey-agent, line 195: **`internal/commands/workitem/staging.go` is 571 lines, over the 500-line standard.** This is exactly the kind of file that splits cleanly along natural concern boundaries:

- **`staging.go`** — `ComputePlan`, types (`PlanContext`, `PlanOptions`, `SkippedEntry`, `StagingPlan`), `PrintPlan`.
- **`staging_branches.go`** — the per-branch `compute{StagedOnly,Project,Workitem,Link,Festival}Plan` functions plus `cwdSubGitRepo` and `pickPrimaryProjectScopePath`.
- **`staging_paths.go`** — `applyIncludes`, `applyExcludes`, `relativeToRepo`, `dedupeSorted`.
- **`staging_git.go`** — `listChangedFilesUnder`, `listStagedFiles`, `pathIsDirty`, `listDirtySubmodulePointers`, `listSubmodulePointerSkips`.

Same staff-level argument that drove the manager.go split on obey #105: a reviewer holding the staging matrix in head needs the branches and the git helpers to live in separate files so a future change to one does not silently leak assumptions into the other.

Separately: `applyExcludes` at line 521 mutates the input array in place via `kept := stage[:0]`. Same antipattern from PR #303 sync.go. Use `kept := make([]string, 0, len(stage))` — the defensive allocation is cheap, the future debugging cost of an aliased slice mutation is not.

internal/commands/workitem/commits.go
```diff
@@ -0,0 +1,317 @@
...
+// deduplicated by canonical path. Always includes the campaign root.
+func enumerateQueryRepos(ctx context.Context, campaignRoot string) ([]string, error) {
+	seen := map[string]bool{}
+	out := []string{campaignRoot}
+	seen[campaignRoot] = true
+
+	registry, err := links.Load(ctx, campaignRoot)
+	if err != nil {
+		return out, nil // links.yaml may not exist yet; campaign root is enough
+	}
```
obey-agent, line 173: `enumerateQueryRepos` silently swallows `links.Load` errors at line 162-164:

```go
if err != nil {
    return out, nil // links.yaml may not exist yet; campaign root is enough
}
```

The comment's "may not exist yet" case is correct only when the error is `fs.ErrNotExist`. Any other failure (malformed YAML, schema-version mismatch, permission glitch) is silently swallowed, and the user's cross-repo query falls back to "campaign root only" with zero indication that their link registry is broken — they see "no commits found" and conclude the search did its job.

Discriminate:

```go
registry, err := links.Load(ctx, campaignRoot)
if err != nil {
    if errors.Is(err, fs.ErrNotExist) {
        return out, nil
    }
    // Surface non-fs errors so the user knows their registry needs attention.
    return out, camperrors.Wrap(err, "load link registry")
}
```

Even better: don't fail the whole query, but include the load error in the `queryErrs` slice so `--json` consumers see it.

Pairs with the table-mode-drops-errs comment: the user's feedback loop here is broken end-to-end if `links.Load` quietly returns nothing.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview

The split between `commit.Workitem` (thin tag wrapper, ~34 lines) and the planner (`staging.go`) is the right shape, the staging-matrix design contract in `COMMIT_DESIGN.md` is the kind of doc that ages well, and the 6 containerized integration tests cover the actual fail-loud rule the feature exists to enforce. What blocks approval is one bug that breaks established codebase conventions (`os.Exit` from RunE), one robustness gap that bites on any non-trivial filename (`git status --porcelain` without `-z`), one contract lie (`CommitRecord.Repo` doc vs implementation), and one silent inclusion that contradicts the PR-description framing ("never silently widens"). The smaller findings are observability gaps you would want to close before this becomes the default commit path for camp workitems.

## Key Findings

1. **`os.Exit(2)` from within `runCommit` (`commit.go:123`)** bypasses defers and the `CommandError.ExitCode` routing that `cmd/camp/main.go` already implements for exactly this purpose. Same gap I flagged on PR #302 (`doctor.go:55`) and PR #303 (`show.go:21`) — the fix mechanism exists in the codebase, please use it. Integration tests cannot distinguish exit 1 from exit 2 today; this slipped through. (inline `commit.go:123`)

2. **`git status --porcelain` without `-z`** in `listChangedFilesUnder`, `pathIsDirty`, and `listDirtySubmodulePointers`. Filenames with spaces, tabs, newlines, or non-printable chars get quoted by git's porcelain v1 and the parser does not unquote. Rename entries (`old -> new`) break on names containing spaces. Switch to `--porcelain -z` and NUL-split. (inline `staging.go:436`)

3. **`CommitRecord.Repo` field doc lies about the value.** Doc says "campaign-relative path... '.' for the campaign root"; implementation at `commits.go:255` is `filepath.Base(repo)`. Two repos with the same basename (e.g., `projects/foo` and `projects/worktrees/foo`) collide in output and consumers cannot route commits back. (inline `commits.go:255`)

4. **Silent auto-stage of `.campaign/workitems/links.yaml` in `computeWorkitemDirPlan` and `computeFestivalPlan`** contradicts the PR's "never silently widens" framing. A user committing workitem changes gets the link registry bundled with zero annotation in the plan. Either drop the auto-stage, surface it explicitly in `PrintPlan`, or gate behind a flag. (inline `staging.go:217`)

5. **Table mode silently drops per-repo failures** (`commits.go:151`). JSON path emits errors under `errors`, table path does not. A user whose query failed on 3 of 12 linked repos sees fewer commits with no signal. (inline `commits.go:151`)

6. **No per-repo timeout in the cross-repo fan-out** (`commits.go:251`). One hanging git log blocks the whole query until Ctrl-C. Wrap each `exec.CommandContext` with a 30s deadline; cap workers at `min(NumCPU, 8)` since git log is IO-bound. (inline `commits.go:251`)

7. **`enumerateQueryRepos` silently swallows `links.Load` errors** (`commits.go:164`). Only `fs.ErrNotExist` should be silent; malformed YAML or permission glitches should surface. (inline `commits.go:164`)

8. **`staging.go` is 571 lines, over the 500 standard.** Splits cleanly along branch/path/git seams. `applyExcludes` also mutates the input slice in place — same antipattern from PR #303 sync.go. (inline `staging.go:571`)

## What is Done Well

- `commit.Workitem` is a clean 34-line wrapper that mirrors `commit.Quest` exactly — no parallel implementation, just threading `WorkitemRef` and `QuestID` into `doCommit` via the existing `Options` struct.
- `COMMIT_DESIGN.md` (251 lines) documents the routing logic, staging matrix, flag precedence, and refusal modes. Future maintainers can hold the model in head without re-deriving it from code.
- Six sample plan fixtures under `testdata/plans/` lock down the stable lines the integration tests can grep against.
- `TestIntegration_WorkitemCommit_RefusesSilentWiden` is exactly the load-bearing test for this feature — proving that any path outside the workitem dir leaking into HEAD fails loudly is the security/correctness invariant. Adding adversarial cases (filename with spaces, dirty link registry, conflicting workitem dirs) before this lands would close the gap.
- `commitkit.ParseTag` filter at `commits.go:267-269` correctly rejects incidental string matches — `--grep=-WI-WI-abc123` could match a literal commit body containing `WI-abc123`; filtering through `ParseTag` ensures only properly-tagged commits land in the result set.
- `enumerateQueryRepos` deduplicates by canonical path (line 157, 170-173) so a project linked via both `ScopeProject` and `ScopeRepo` is queried once.
- Worker pool fan-out for cross-repo query is the right shape — `git log` per repo is independent work, blocking the main thread on serial calls would dominate latency on campaigns with many linked repos.
- `Annotations: agent_allowed` on both `commit` and `commits` plus `--json` / `--dry-run` flags means these surface cleanly for automation, matching the agent-friendly contract from earlier sequences.

## Staff Standard

Would I submit this as my best work at a staff level? The architectural decisions (`commit.Workitem` mirror, DESIGN.md, plan-then-commit split, fan-out query) are sound. The four blockers (#1-#4) are correctness gaps that any careful reviewer would catch and that the existing test surface lets slip through. Address them, plus #5-#7 for observability and #8 for maintainability, and this is a clean approve. The bones are right; the polish is missing.

internal/commands/workitem/commits.go
```diff
@@ -0,0 +1,380 @@
...
+		"--grep="+grep,
+	)
+	output, err := cmd.Output()
+	if errors.Is(cctx.Err(), context.DeadlineExceeded) {
+		return nil, fmt.Errorf("git log timeout after %s", commitsPerRepoTimeout)
+	}
+	if err != nil {
+		return nil, err
+	}
+	rel := filepath.Base(repo)
```
obey-agent: **Still using `filepath.Base(repo)`** at line 286. The doc on `CommitRecord.Repo` (line 38-40) still says "campaign-relative path of the git repo, ('.') for the campaign root", but the implementation:

- For the campaign root itself, returns the campaign directory name (e.g. `"my-campaign"`), not `"."`.
- For `projects/foo` and `projects/worktrees/foo`, both basename to `"foo"` — colliding rows in output that consumers cannot route.

The other fixes in this commit thread `ctx` through helpers and add timeouts; threading `campaignRoot` through `queryRepo` is the same shape:

```go
func queryRepo(ctx context.Context, campaignRoot, repo, ref string) ([]CommitRecord, error) {
    // ...
    rel, err := filepath.Rel(campaignRoot, repo)
    if err != nil || strings.HasPrefix(rel, "..") {
        rel = filepath.Base(repo)
    }
    // ...
    Repo: rel
}
```

`searchRepos` already has campaignRoot at its call site (`runCommitsQuery:136-138`); pass it through the job struct or capture it in the worker closure. Add a unit test with two repos that share a basename to lock the contract.

internal/commands/workitem/staging.go
```diff
@@ -0,0 +1,600 @@
...
+	}
+	plan.ContextNote = workitemDirNote(plan.Workitem, src)
+
+	stage, err := listChangedFilesUnder(ctx, root, plan.Workitem.RelativePath)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "list workitem changes")
+	}
+
+	if dirty, err := pathIsDirty(ctx, root, linkRegistryRelPath); err == nil && dirty {
+		stage = append(stage, linkRegistryRelPath)
```
obey-agent: **Link registry still silently auto-staged** at lines 216-217 (and the matching line in `computeFestivalPlan`). No annotation in `PrintPlan`, no flag to gate it, no `Skip` entry explaining the inclusion. The user sees `A .campaign/workitems/links.yaml` in the stage list with no indication it was auto-added rather than part of the workitem scope.

This still contradicts the PR description's "never silently widens" framing, and it's the kind of bundling that surprises operators when they look at the resulting commit.

Three options, pick one. The minimum I'd ask is option 1:

1. **Mark the auto-inclusion in `PrintPlan`** by carrying an annotation through the plan:
   ```go
   type StageEntry struct {
       Path   string
       Source string // "workitem-dir" | "link-registry" | "submodule-pointer" | "include-flag"
   }
   ```
   Then render as `  A  .campaign/workitems/links.yaml (link registry auto-included)`.
2. **Gate behind `--include-link-registry`** (default-on for one release, then default-off).
3. **Drop the auto-stage entirely** and require `--include .campaign/workitems/links.yaml` when the user wants it bundled.

Whichever you pick, `TestIntegration_WorkitemCommit_RefusesSilentWiden` should grow a sub-case that proves either (a) the link registry is annotated visibly or (b) it is not auto-included. Today the test doesn't cover this path.

internal/commands/workitem/staging.go
```diff
@@ -0,0 +1,600 @@
...
+	return out
+}
+
+// ErrNoWorkitemContext is returned when ComputePlan cannot resolve any
+// workitem context. The CLI maps this to exit code 2 with a help hint.
+var ErrNoWorkitemContext = camperrors.NewValidation(
+	"workitem",
+	"no workitem context resolved",
+	nil,
+)
```
obey-agent, line 195: **`staging.go` is now 600 lines** (up from 571 in the previous round), still over the 500-line standard. The porcelain-z refactor added the `gitStatusPorcelainZ`, `parseGitStatusPorcelainZ`, `parseNULPathList`, `splitNULFields`, and `gitStatusEntry` helpers — all good additions — but the file did not get split.

The same split shape I suggested last round still applies:

- **`staging.go`** — `ComputePlan`, types, `PrintPlan`
- **`staging_branches.go`** — the per-branch `compute{StagedOnly,Project,Workitem,Link,Festival}Plan` functions plus `cwdSubGitRepo` and `pickPrimaryProjectScopePath`
- **`staging_paths.go`** — `applyIncludes`, `applyExcludes`, `relativeToRepo`, `dedupeSorted`
- **`staging_git.go`** — `listChangedFilesUnder`, `listStagedFiles`, `pathIsDirty`, `listDirtySubmodulePointers`, `listSubmodulePointerSkips`, **plus the new `gitStatusPorcelainZ`/`parseGitStatusPorcelainZ`/`parseNULPathList`/`splitNULFields` helpers** — the porcelain-z code wants to live with the other git-shell wrappers anyway.

The split made PR #302 cleaner end-to-end (manager.go on obey #105 followed the same playbook). Two-file moves and an import update.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview

Five of eight blockers from the previous round are addressed cleanly: `os.Exit(2)` replaced with `camperrors.NewCommand(..., 2, ...)`, all `git status --porcelain` callers now use `-z` with proper rename handling, per-repo `commitsPerRepoTimeout = 30 * time.Second` plus `commitsMaxWorkers = 8` cap, table mode now emits a stderr warning when repos fail, and `isGitRepo` discriminates real errors from "not a git repository". Three findings remain: `CommitRecord.Repo` still uses `filepath.Base` (collision risk + doc lie), `links.yaml` is still silently auto-staged with zero annotation, and `staging.go` grew from 571 to 600 lines without the split.

## Findings Resolved (5 of 8)

1. **`os.Exit(2)` → `camperrors.NewCommand("camp workitem commit", 2, "", err)`** at `commit.go:122`. Same primitive as PR #302/#303 fixes; main.go routes the exit code correctly.

2. **`git status --porcelain -z` everywhere** with `parseGitStatusPorcelainZ` handling rename entries (`R`/`C` codes skip the second NUL-delimited field). `TestParseGitStatusPorcelainZ_RenameKeepsNewPath` and `TestListChangedFilesUnder_PreservesPorcelainZPaths` lock the contract with a literal `"path with \"quote\".md"` fixture. Robust against the actual edge cases this time.

5. **Table mode warns on stderr.** `emitCommitsQueryWarnings` writes `warning: N repo(s) failed; re-run with --json for details\n` to `cmd.ErrOrStderr()`. `TestEmitCommitsQueryWarnings` pins the format exactly. Pipes through `grep`/`wc` still see only stdout, the operator still sees the warning.

6. **Per-repo timeout + worker cap.** `commitsPerRepoTimeout = 30 * time.Second` and `commitsMaxWorkers = 8` are real constants. `queryRepo` wraps with `context.WithTimeout`, distinguishes `context.DeadlineExceeded` for a clean error message. `commitsWorkerCount` covers the edge cases (0 repos, 1 repo, NumCPU > 8, etc.) with a unit test.

7. **`isGitRepo(ctx, path) (bool, error)`** stats first, errors on non-existent for non-`fs.ErrNotExist` cases, distinguishes "not a git repository" from genuine exec failures via stderr matching, propagates `ctx.Err()`. The legitimate "skip silently" case still works; the previously-swallowed real errors now surface.

## Findings Still Open

3. **`CommitRecord.Repo` is still `filepath.Base(repo)`** at `commits.go:286`. Doc on line 38-40 still says `"campaign-relative path of the git repo, ('.') for the campaign root"`, but the implementation returns the basename. Two repos with the same basename still collide; the campaign root itself emits its directory name rather than `"."`. (inline `commits.go:286`)

4. **`linkRegistryRelPath` is still silently auto-staged** by `computeWorkitemDirPlan` (`staging.go:216-217`) and `computeFestivalPlan` (`:277-278`). No annotation in `PrintPlan`, no flag gate, no `Skip` entry explaining the inclusion. The PR description's "never silently widens" framing still does not match the behavior. (inline `staging.go:217`)

8. **`staging.go` is now 600 lines** (up from 571). The porcelain-z helpers are good additions, but the file still wants the four-way split I suggested last round (`staging.go` / `staging_branches.go` / `staging_paths.go` / `staging_git.go`). (inline `staging.go:600`)

## Verification

- `just test unit` — 2472/2472 passed
- `just test integration-run TestIntegration_WorkitemCommit` — 6/6 passed (`WorkitemDirContext`, `LinkedProject`, `RefusesSilentWiden`, `IncludeExclude`, `CrossRepo`, `FailureModes`)

## What is Done Well

- The porcelain-z parser handles rename codes correctly (`R`/`C` consume the next NUL field). Test fixture uses an actual quoted-path filename, not a synthetic one.
- The exit-code fix uses `camperrors.NewCommand` consistently with the same pattern PR #302/#303 adopted — the exit-2 contract is now load-bearing across the workflow/workitem command surface.
- `commitsWorkerCount` is a tiny pure function with table-driven test coverage. The kind of small-but-load-bearing helper that earns its keep at staff review.
- `isGitRepo` discrimination via `CombinedOutput` + stderr substring is pragmatic — git's exit codes for "not a repo" overlap with other failures, so the textual check is the right pragmatism.

## Staff Standard

Five fixes are well-shaped. Three open: a contract lie (`Repo` doc vs impl), a silent inclusion contradicting the "never silently widens" PR framing, and a file that drifted further over the size standard. The architectural choices are sound; closing the remaining three is mostly mechanical work plus one test.

obey-agent (approved):
## Verdict: Approve

## Overview

All three remaining findings from the previous round are addressed substantively. `repoDisplayPath` returns the campaign-relative path matching the documented contract (with `"."` for the campaign root and basename fallback only for genuine edge cases), `StageAnnotations` carries per-path notes that surface visibly in `PrintPlan`, and `staging.go` is split into four focused files that match the natural concern boundaries.

## Findings Resolved

3. **`CommitRecord.Repo` now reflects the documented contract.** `repoDisplayPath(campaignRoot, repo)` uses `filepath.Rel` + `filepath.ToSlash` to produce campaign-relative paths; `"."` for the root; basename fallback only when `Rel` fails or the result escapes. Threaded through `searchRepos` → `queryRepo`. The previously-colliding case of `projects/foo` vs `projects/worktrees/foo` now renders as distinct paths.

4. **Link registry auto-stage is annotated.** New `StageAnnotations map[string]string` on `StagingPlan`, populated via `addStageNote(linkRegistryRelPath, stageAnnotationLinkRegistry)` in both `computeWorkitemDirPlan` and `computeFestivalPlan`. `PrintPlan` renders these as `A  .campaign/workitems/links.yaml (link registry auto-included)`. The "never silently widens" framing is now honest end-to-end. `TestComputePlan_AutoIncludedLinkRegistryIsAnnotated` locks the contract.

8. **`staging.go` split into four files** matching the boundaries from my last review:
   - `staging.go` (195 lines) — `ComputePlan`, types, `PrintPlan`, `addStageNote`
   - `staging_branches.go` (248) — `cwdSubGitRepo`, the per-branch `compute{Workitem,Link,Festival,Project,WorkitemDir}Plan` helpers, `pickPrimaryProjectScopePath`
   - `staging_git.go` (123) — `listChangedFilesUnder`, `listStagedFiles`, `pathIsDirty`, `listDirtySubmodulePointers`, `gitStatusPorcelainZ`, `parseGitStatusPorcelainZ`, `parseNULPathList`, `splitNULFields`
   - `staging_paths.go` (78) — `applyIncludes`, `applyExcludes`, `relativeToRepo`, `dedupeSorted`
   
   Every file now under 500. The porcelain-z helpers landed in `staging_git.go` exactly where they belong.

## Bonus Fix

The `enumerateQueryRepos` error-swallow case I flagged in the first review round (and partially noted in round 2) is now fully closed: `links.Load` errors propagate via `camperrors.Wrap(err, "load link registry")` instead of returning the campaign-root-only fallback. `links.Load` returns `Empty(), nil` for non-existent files, so this only fires on genuine corruption — which is exactly the signal operators need.

## Verification

- `just test unit` — 2476/2476 passed
- `just test integration-run TestIntegration_WorkitemCommit` — 6/6 passed (`WorkitemDirContext`, `LinkedProject`, `RefusesSilentWiden`, `IncludeExclude`, `CrossRepo`, `FailureModes`)

## What Is Done Particularly Well

- `repoDisplayPath` handles every documented case (`.` for root, campaign-relative for sub-repos) AND degrades safely (basename) only when `Rel` produces something unusable. The fallback path is reachable only for truly weird inputs, not the happy path.
- `StageAnnotations` as a `map[string]string` keyed by path is the right shape — generic enough to carry future annotations (submodule-pointer, force-include, etc.) without a new type per concern.
- The four-way file split matches concern boundaries exactly: orchestrator vs branch routing vs git-shell vs path-math. A future contributor touching one section does not touch the others.
- `staging_git.go` is the right home for the porcelain-z helpers that landed in round 2 — keeping git-shell wrappers in one file keeps the `bytes`/`os/exec` imports localized.

## Staff Standard

Yes. The shape of the planner is now clean: types and orchestration in `staging.go`, branch routing in `staging_branches.go`, git plumbing in `staging_git.go`, path math in `staging_paths.go`. The link-registry annotation closes the "silent widening" gap that contradicted the PR's framing, and `repoDisplayPath` honors the documented `CommitRecord.Repo` contract. Ready to merge after #302/#303/#304 land in sequence.

PR #308, opened 2026-05-26
fix(CW0003 seq-01): atomic config writes + links RMW lock + workflow doctor stale walk

lancekrogers:
Closes 4 merge-blocking findings from CW0003 Phase 002 (REVIEW). First sequence of Phase 003 (FIX_BLOCKERS).

## Findings closed

- **`CW0003-workflow-03`** — Multi-file config writes non-atomic. `SaveJumpsConfig` / `SaveCampaignConfig` now go through `fsutil.WriteFileAtomically`.
- **`CW0003-workflow-13` (proposed)** — Raw `os.WriteFile` on `jumps.yaml` / `campaign.yaml`. Closed by the same swap.
- **`CW0003-links-06`** — Last-writer-wins RMW window on `links.yaml`. New `links.WithLock(ctx, root, fn)` holds the lock across the full Load -> Mutate -> Save transaction. 3 RMW call sites migrated (link, unlink, doctor --fix).
- **`CW0003-workflow-02`** — `workflow.cache.stale` mtime check missed most real staleness. Replaced single-dir `os.Stat` with bounded recursive walk (`maxWorkflowWalkEntries = 10_000`). `collectFindings` and the helper now take `ctx` (incidental chip at `CW0003-workflow-04`).

## Test evidence

- New unit tests in `internal/commands/workflow/doctor_cache_test.go` cover fresh / touched-deep / missing-workflow-tree paths.
- New containerized integration tests:
  - `TestIntegration_WorkflowConfigSave_NoTempLeftover` — asserts atomic-write codepath runs cleanly with no `.tmp-*` leftovers.
  - `TestIntegration_LinkConcurrentNoLoss` — 10-way concurrent `camp workitem link` race, asserts all 10 links land in the registry.
- All existing tests pass: `internal/config/...`, `internal/workitem/links/...`, `internal/commands/workitem/...`, `internal/commands/workflow/...`.

## Target

Targets `feat/workflow-surface-CW0003` (the integrated trunk), NOT `main`. Trunk merges to main after Phase 003 closes all 5 sequences and `COMPLETENESS_MEMO.md` reissues GO.

## Festival tracking

Festival: `camp-workitem-linking-and-commits-CW0003` / Phase 003 / sequence 01 (`atomic_config_writes`). Commit body cites each closed finding ID.

### DIFF
--- a/docs/OBEY.md
+++ b/docs/OBEY.md
@@ -25,3 +25,31 @@ docs/
 - Use Markdown for all documentation
 - Keep documentation close to the code it describes
 - Update docs when code changes
+
+## Narrative References
+
+Hand-authored references that explain concepts beyond the per-command
+auto-generated CLI reference.
+
+- [campaign-directory-reference.md](campaign-directory-reference.md)
+- [campaign-settings-files.md](campaign-settings-files.md)
+- [workflow-reference.md](workflow-reference.md) — custom workflow surface
+  (`camp workflow create/list/show/doctor/sync` and `shortcut add`).
+- [workitem-link-reference.md](workitem-link-reference.md) — link registry,
+  `lnk_*` IDs, resolver tiers, broken-link recovery.
+- [workitem-commit-reference.md](workitem-commit-reference.md) — scoped
+  workitem commits, staging matrix, refusal mode, cross-repo query, tag
+  grammar.
+- [SHORTCUTS.md](SHORTCUTS.md)
+- [shell-integration.md](shell-integration.md)
+- [leverage-score.md](leverage-score.md)
+
+## Migrations
+
+Operator-facing rollout notes for behavior changes that affect existing
+campaigns.
+
+- [migrations/dungeon-behavior-upgrade-2026-03.md](migrations/dungeon-behavior-upgrade-2026-03.md)
+- [migrations/intent-path-targeting-upgrade-2026-03.md](migrations/intent-path-targeting-upgrade-2026-03.md)
+- [migrations/workitem-v1alpha6-2026-05.md](migrations/workitem-v1alpha6-2026-05.md) —
+  `.workitem` schema bump (added `ref` and `quest_id`).
--- a/docs/campaign-directory-reference.md
+++ b/docs/campaign-directory-reference.md
@@ -33,6 +33,9 @@ After `camp init`, the directory usually looks like this:
 │   ├── allowlist.json
 │   ├── fresh.yaml
 │   └── jumps.yaml
+├── workitems/              # link registry + current selection
+│   ├── links.yaml
+│   └── current.yaml
 ├── skills/
 │   ├── camp-navigation/
 │   ├── camp-projects/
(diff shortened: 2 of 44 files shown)


obey-agent (commented):
## Verdict: Approve

## Overview

Tight, focused fix bundle that closes 4 merge-blocking findings from CW0003 Phase 002. The refactor preserves the existing public `Save` / `Load` API while adding a real concurrency primitive (`links.WithLock`), routes config writes through `fsutil.WriteFileAtomically` with a 2-line swap, and replaces the broken mtime-on-root staleness check with a bounded recursive walk. All changes lean on existing helpers (camperrors, fsutil) rather than introducing parallel implementations.

## Findings closed

- `CW0003-workflow-03` — `SaveJumpsConfig` / `SaveCampaignConfig` now atomic
- `CW0003-workflow-13` (proposed) — same swap covers the second escalated finding
- `CW0003-links-06` — RMW window closed by `links.WithLock` migration across 3 call sites
- `CW0003-workflow-02` — `workflow.cache.stale` now walks the tree recursively with a 10k-entry cap; `collectFindings` gains ctx (incidental chip at `CW0003-workflow-04`)

## What's done well

- **`Save` / `saveLocked` split is the right shape.** Public `Save` keeps single-shot lock acquisition for compatibility; new `saveLocked` skips re-acquisition so `WithLock` can hold the lock across the full Load -> Mutate -> Save window. Same shape for `Load` / `loadLocked`.
- **`ErrSkipSave` sentinel is exactly the API a doctor-style "inspect then maybe mutate" flow wants.** The `errors.Is` check inside `WithLock` is correctly defensive.
- **Context propagation through `collectFindings` -> `isNavCacheStaleForWorkflow`** with explicit `ctx.Err()` inside the walk is staff-Go correct.
- **The 10-way concurrent link race test** (`TestIntegration_LinkConcurrentNoLoss`) is the right shape of evidence for the lock fix: shell-orchestrated parallel invocations against the real binary in the container, asserting every link lands in `links.yaml`.
- **Atomic-write evidence via the `NoTempLeftover` assertion** is honest: it demonstrates the codepath runs without leftover temp files. The PR description acknowledges the fail-injection variant wasn't workable in the shared container (which runs as root), so we lean on fsutil's package-level test coverage for the rename-rollback contract.
- **Bounded walk with named constant (`maxWorkflowWalkEntries = 10_000`)** beats an unbounded scan; explicit `filepath.SkipAll` exit is clean.

- `runLink` generates `link.ID` BEFORE `WithLock` acquires the lock. Two concurrent invocations could in theory generate the same ID. `CW0003-links-01` (seq-04 of Phase 003) introduces the 32-attempt retry loop that closes this. Calling out so the seq-04 reviewer is aware the fix expects this code shape.
- `doctor.go runDoctor` computes `knownIDs` outside `WithLock`. A workitem created between `workitemIDsOnDisk` and `collectWorkitemFindings` inside the closure would be a registry-only entry. The pre-refactor code had the same race window, so this PR is net-neutral. Worth tracking if `workitem create` ever lands on the doctor's hot path.

## Staff standard

Would I submit this as my own work? Yes. Refactor minimal, the public API contract preserved, real tests for the real bugs, named constants, explicit error wrapping, ctx propagation. The two pre-existing concerns above are tracked and have explicit fixes in flight (seq-04). Clean approve.

obey-agent (approved):
## Verdict: Approve

## Overview
This is a clean hardening slice. Atomic config writes move to the shared fs utility, links registry mutations now hold the RMW lock across load/mutate/save, and the stale cache check is meaningfully broader while still bounded and context-aware.

## Key Findings
- No blocking findings in this PR.

## What's Done Well
- `links.WithLock` is the right abstraction for the registry transaction window.
- The new containerized concurrent-link integration test targets the actual race being fixed.
- The recursive workflow cache check observes context cancellation and avoids unbounded walks.

## Staff Standard
Yes, I would be comfortable merging this slice.

obey-agent (approved):
Verdict: Approve

Overview:
The new push adds the missing stress coverage rather than changing the production path, and it strengthens the earlier hardening story. The concurrent registry test now verifies no links are lost under `WithLock`, and the atomic write integration test exercises failed rename cleanup without requiring privileged container mounts.

Key Findings:
- No blocking findings in the pushed delta.

What's Done Well:
- The unit test now asserts every concurrent writer survives, not just that the YAML remains parseable.
- The integration test checks both destination preservation and temporary-file cleanup on write failure.

Verification:
- `just test unit` passed: 74 packages, 2501 tests.
- `just test integration-run 'TestIntegration_WorkflowConfigSaveAtomic_ENOSPC|TestIntegration_LinkConcurrentNoLoss'` passed.

Staff Standard:
Yes. I would be comfortable merging this slice.

PR #309, opened 2026-05-26
fix(CW0003 seq-02): doctor + tag grammar + workflow dry-run (10 findings)

lancekrogers:
Closes 10 merge-blocking findings from CW0003 Phase 002 (REVIEW). Second sequence of Phase 003.

## Findings closed

|---|---|---|
| `CW0003-format-01` | format | Doctor false-positives on intents/festivals (52 in real repos) — marker-existence filter |
| `CW0003-format-02` | format | ref / quest_id shape unvalidated — regex validation at load time |
| `CW0003-format-03` | format | `tagShellRegex` unanchored — quote-guard rejects Revert-subject false positives |
| `CW0003-format-04` | format | ParseTag silently drops unknown trailing segments — unknown branch now seeks next prefix |
| `CW0003-format-06` | format | `backfillMissingRefs` aborted on first per-item error — collects failures and continues |
| `CW0003-format-13` | format | Silent WI-less tag for v1alpha5 workitems — stderr warning via `StagingPlan.Warnings` |
| `CW0003-links-12` (proposed) | links | Doctor wedged on malformed `links.yaml` — new `QuarantineBroken` + `workitem.registry.parse-error` finding |
| `CW0003-links-13` (proposed) | links | Resolver aborted on tier-internal errors — `result="error"` trace + continue to next tier |
| `CW0003-workflow-01` | workflow | `workflow create --dry-run` showed one header line — new `planActionLines` emits per-action lines per DESIGN.md §4.1 |
| `CW0003-workflow-10` | workflow | Dedup `fix_hint` misrepresented actual behavior — hint now matches `deduplicateShortcut` policy |

## Test evidence

- `internal/workitem/metadata_test.go` — 3 new table cases (invalid ref shape, ref wrong length, invalid quest_id shape)
- `internal/git/campaign_tag_full_test.go` — 3 new adversarial cases (Revert subject, quoted backport subject, malformed middle segment with recoverable WI- suffix)
- `internal/commands/workitem/resolve_test.go` — `TestResolve_BrokenPrimaryLinkDoesNotFallThrough` inverted: now asserts fall-through with an `error` trace step instead of asserting hard error
- All existing tests pass: `internal/workitem/...`, `internal/git/...`, `internal/commands/workitem/...`, `internal/commands/workflow/...`

## Coordination notes

- This PR coordinates with seq-04 (`links-registry-hardening`) on the resolver fall-through pattern. Seq-04 closes `CW0003-links-03` using the same `result="error"` shape this PR established for `links-13`. Whichever lands first sets the contract; the other rebases mechanically.
- `CW0003-format-13` policy: chose option (b) warn-and-skip instead of option (a) auto-backfill-mid-commit. Rationale: auto-backfill at commit time would couple the commit success to file-lock acquisition; warn-and-skip preserves commit success while making the gap visible to the user.

## Target

Targets `feat/workflow-surface-CW0003` (the integrated trunk), NOT `main`.

## Festival tracking

Festival `camp-workitem-linking-and-commits-CW0003` / Phase 003 / sequence 02 (`doctor_and_tag_grammar`).

### DIFF
--- a/internal/commands/workitem/doctor.go
+++ b/internal/commands/workitem/doctor.go
@@ -91,7 +91,10 @@ func runDoctor(ctx context.Context, cmd *cobra.Command, jsonOut, fix bool) error
 				AutoFixable: true,
 			}
 			if jsonOut {
-				return emitDocJSON(cmd.OutOrStdout(), []docFinding{parseFinding})
+				if jerr := emitDocJSON(cmd.OutOrStdout(), []docFinding{parseFinding}); jerr != nil {
+					return jerr
+				}
+				return errDoctorIssues
 			}
 			emitDocHuman(cmd.OutOrStdout(), []docFinding{parseFinding})
 			return camperrors.Wrap(err, "load links registry")
--- a/internal/commands/workitem/doctor.go
+++ b/internal/commands/workitem/doctor.go
@@ -35,6 +35,7 @@ const (
 	codeCurrentMissing     = "workitem.current.missing"
 	codeMissingRefField    = "workitem.ref.missing"
 	codeWorkitemScanFailed = "workitem.scan.failed"
+	codeRegistryParseError = "workitem.registry.parse-error"
 )
 
 const (
@@ -80,7 +81,34 @@ func runDoctor(ctx context.Context, cmd *cobra.Command, jsonOut, fix bool) error
 	}
 	registry, err := links.Load(ctx, root)
 	if err != nil {
-		return err
+		if !fix {
+			parseFinding := docFinding{
+				Code:        codeRegistryParseError,
+				Severity:    docSeverityError,
+				Target:      "registry:links.yaml",
+				Message:     "links.yaml cannot be parsed: " + err.Error(),
+				FixHint:     "run `camp workitem doctor --fix` to quarantine the broken file and bootstrap an empty registry",
+				AutoFixable: true,
+			}
+			if jsonOut {
+				return emitDocJSON(cmd.OutOrStdout(), []docFinding{parseFinding})
+			}
+			emitDocHuman(cmd.OutOrStdout(), []docFinding{parseFinding})
+			return camperrors.Wrap(err, "load links registry")
+		}
+		quarantined, qerr := links.QuarantineBroken(ctx, root)
+		if qerr != nil {
+			return camperrors.Wrap(qerr, "quarantine broken registry")
+		}
+		if quarantined != "" {
+			fmt.Fprintf(cmd.ErrOrStderr(),
+				"quarantined broken links.yaml to %s; bootstrapped empty registry\n",
+				quarantined)
+		}
+		registry, err = links.Load(ctx, root)
+		if err != nil {
+			return camperrors.Wrap(err, "reload after quarantine")
+		}
 	}
 
 	knownIDs, err := workitemIDsOnDisk(ctx, root)
@@ -256,8 +284,11 @@ func autoFixWorkitemFindings(ctx context.Context, root string, registry *links.L
 		}
 	}
 	if needsRefBackfill {
-		n, err := backfillMissingRefs(ctx, root)
+		n, failures, err := backfillMissingRefs(ctx, root)
 		applied += n
+		for _, f := range failures {
+			fmt.Fprintf(errw, "warning: backfill ref for %s: %v\n", f.RelativePath, f.Err)
+		}
 		if err != nil {
 			fmt.Fprintf(errw, "warning: backfill refs: %v\n", err)
 			return applied
--- a/internal/workitem/resolver/resolver.go
+++ b/internal/workitem/resolver/resolver.go
@@ -101,17 +101,19 @@ func Resolve(ctx context.Context, root string, opts Options) (*Resolution, error
 	}
 	for _, tier := range tiers {
 		wi, step, err := tier.fn()
-		if err != nil {
-			step.Result = "error"
-			if step.Detail == "" {
-				step.Detail = err.Error()
-			} else {
-				step.Detail = step.Detail + ": " + err.Error()
-			}
-			result.Trace = append(result.Trace, step)
-			continue
-		}
 		result.Trace = append(result.Trace, step)
+		if err != nil {
+			// Operational failures (cancellation, registry parse, unexpected
+			// I/O, malformed selector input) bubble up so callers don't get a
+			// silently-misattributed lower-priority workitem. The only
+			// recoverable case is the stale-link condition where a primary
+			// path link points to a workitem that no longer exists; tier
+			// helpers signal that by encoding the failure in step.Result
+			// without returning a non-nil err.
+			result.Source = tier.source
+			result.Reason = firstNonEmpty(step.Detail, err.Error())
+			return result, err
+		}
 		if wi != nil {
 			result.Workitem = wi
 			result.Source = tier.source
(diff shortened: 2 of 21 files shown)


obey-agent (commented):
## Verdict: Approve

## Overview

Big bundle (10 findings) but each fix is small, localized, and well-bounded. The hardest call was on `CW0003-format-13` (silent WI-less tag) where I picked warn-and-skip (option b) over auto-backfill-mid-commit (option a) so a backfill failure cannot derail a commit. Trade-off recorded in the PR body. Parser fixes are particularly conservative: the unanchored regex got a quote-guard (`(?:^|[^"])`) rather than full anchoring, which preserves the prepend-style tag format that `PrependContextTagsFull` round-trips. Unknown-segment recovery via `indexOfNextPrefix` preserves trailing WI- refs that the old silent-drop branch was losing.

## What's done well

- **Marker-existence filter for doctor false-positives.** `hasWorkitemMarker(root, relPath)` is the right invariant: only paths with a `.workitem` on disk are valid backfill targets. Intents (.md) and festivals (.fest.yaml dirs) drop out by construction, not by source-id heuristics.
- **`backfillMissingRefs` failure accumulation.** Returning `(applied, []backfillFailure, err)` lets the doctor report per-item failures without aborting the whole queue. The caller in `autoFixWorkitemFindings` prints them individually to stderr. Honest reporting, not silent recovery.
- **`QuarantineBroken` + `codeRegistryParseError`.** The dedicated finding code lets `doctor --json` consumers detect the wedge condition specifically. The quarantined filename includes a unix-nano timestamp so repeated wedges don't collide.
- **Resolver `result="error"` trace propagation.** The error is preserved as a trace step rather than swallowed, so `camp workitem resolve --json` consumers can see why a higher tier failed even though resolution continued.
- **Adversarial test cases** for the Revert-subject, backport-subject, and malformed-middle-segment cases are concrete and check the exact behaviors from the audit doc.
- **`StagingPlan.Warnings`** as a plan-level concept keeps the planner pure (no stderr writes) while letting the caller render them however it wants.

- The links package's `Save` and the new `QuarantineBroken` both acquire `links.yaml.lock` independently. There's no RMW-style helper here yet (that's seq-01's `WithLock` which targets the trunk via PR #308). When both PRs merge, callers should converge on `WithLock` for any new mutation flow; `QuarantineBroken` already does its own lock acquisition because it's a one-shot replace, not RMW.
- The `runDoctor` ctx-bound `findings` slice mutation is the same shape as before this PR; ctx threading through `collectWorkitemFindings` is not changed here (seq-03's commit-safety PR may touch related ctx plumbing).
- Test file `resolve_test.go` was inverted to assert new behavior; the test's `chdir + os.Chdir` plumbing is `CW0003-tests-02` follow-up territory and not in scope here.

## Staff standard

Would I submit this as my own work? Yes. The parser fixes are particularly satisfying because they preserve the existing public API (no signature changes to ParseTag, no breaking change to the tag format) while closing the silent-drop and false-positive bugs at the regex layer. The warn-and-skip choice for v1alpha5 is the more conservative option and the rationale is documented in the PR body. Clean approve.

internal/commands/workitem/doctor.go
```diff
@@ -80,7 +81,34 @@ func runDoctor(ctx context.Context, cmd *cobra.Command, jsonOut, fix bool) error
...
+			parseFinding := docFinding{
+				Code:        codeRegistryParseError,
+				Severity:    docSeverityError,
+				Target:      "registry:links.yaml",
+				Message:     "links.yaml cannot be parsed: " + err.Error(),
+				FixHint:     "run `camp workitem doctor --fix` to quarantine the broken file and bootstrap an empty registry",
+				AutoFixable: true,
+			}
+			if jsonOut {
+				return emitDocJSON(cmd.OutOrStdout(), []docFinding{parseFinding})
```
obey-agent: The JSON path exits successfully for a malformed `links.yaml` even though the finding is `severity: error`. The normal path returns an error after emitting the finding, and the later common path also exits non-zero when error findings exist; this early return bypasses both. Agent callers using `--json` will treat a corrupt registry as healthy. Please emit the JSON and then return the same non-zero doctor error/command error contract.

internal/workitem/resolver/resolver.go
```diff
@@ -101,10 +101,17 @@ func Resolve(ctx context.Context, root string, opts Options) (*Resolution, error
...
 		if err != nil {
-			return nil, err
+			step.Result = "error"
+			if step.Detail == "" {
+				step.Detail = err.Error()
+			} else {
+				step.Detail = step.Detail + ": " + err.Error()
+			}
+			result.Trace = append(result.Trace, step)
+			continue
```
obey-agent: Same blocker as the links-hardening slice: this downgrades all resolver tier errors to trace entries. It should not fall through on cancellation, unreadable/malformed registries, or unexpected selector errors, because that can silently pick a lower-priority workitem and tag/commit against the wrong context. Narrow the fallback to the stale-link condition and keep operational failures as returned errors.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview
This closes several real format and doctor gaps, but two correctness issues remain. `workitem doctor --json` can report a corrupt registry as a successful process, and the resolver fallback masks too many failure classes.

## Key Findings
- Inline: malformed `links.yaml` under `--json` must still exit non-zero after emitting JSON.
- Inline: resolver fall-through needs to be limited to recoverable stale-link cases, not all tier errors.

## What's Done Well
- Ref/quest ID validation and deterministic backfill ordering are solid.
- The dry-run plan output improvement makes workflow create more inspectable.

## Staff Standard
The direction is good, but I would not ship this while automation can see a corrupt registry as success or silently resolve the wrong workitem after operational failure.

internal/commands/workitem/doctor.go
obey-agent: This still exits 0 for a corrupt registry under `--json`. The payload contains an `error`-severity `workitem.registry.parse-error` and `error_count: 1`, but returning only `emitDocJSON(...)` makes automation treat the repository as healthy. I reproduced this on the current head with malformed `links.yaml`: `camp workitem doctor --json` printed the error finding and exited `0`. Please emit the JSON and then return `errDoctorIssues` (or equivalent) when the parse-error finding is severity `error`, and update the integration test to assert the command error is non-nil instead of ignoring it.

internal/workitem/resolver/resolver.go
obey-agent: The resolver fallback is still too broad. This `continue` treats every tier failure as recoverable, including malformed registries, selector/read failures, and unexpected festival/current errors; a later lower-priority tier can then produce a workitem and hide the operational failure. The trace is useful for diagnostics, but correctness should only fall through for the specific recoverable stale-link case this workflow wanted to tolerate. Please classify recoverable tier errors explicitly and return the rest; keep cancellation and parse/read errors as hard failures.

obey-agent (changes requested):
Verdict: Request Changes

Overview:
The new push adds useful coverage around tag parsing, ref backfill, and doctor fixtures, but the two prior correctness blockers are still present. Automation can still see a corrupt workitem registry as a successful doctor run, and resolver fallback still masks operational failures too broadly.

Key Findings:
- Inline: `workitem doctor --json` still exits 0 when `links.yaml` is malformed even though the JSON contains `error_count: 1`; I reproduced this on the current head.
- Inline: resolver tier fallback still converts all tier errors into trace entries and continues, instead of limiting fall-through to the specific recoverable stale-link case.

What's Done Well:
- The intent/festival false-positive tests are valuable and keep the ref backfill scan scoped to real `.workitem` markers.
- The tag parser is now anchored and warnings-aware, which is a stronger contract than the previous regex behavior.
- The ref backfill helper preserves commit flow while making legacy markers visible to the user.

Verification:
- `go test ./internal/commands/workitem ./internal/workitem/resolver ./internal/git ./pkg/commitkit` passed.
- Manual malformed `links.yaml` reproduction confirmed `camp workitem doctor --json` exits 0 on this head.

Staff Standard:
Not yet. I would not merge while an error-severity doctor finding exits successfully and resolver errors can be silently overridden by lower-priority tiers.

PR #310, opened 2026-05-26
fix(CW0003 seq-03): commit-path safety + JSON schema_version (6 findings)

lancekrogers:
Closes 6 merge-blocking findings from CW0003 Phase 002. Third sequence of Phase 003.

## Findings closed

|---|---|---|
| `CW0003-commit-01` | commit | `fmt.Errorf` in `commits.go:281,351` removed; both now route through `camperrors.New(fmt.Sprintf(...))` |
| `CW0003-commit-02` | commit | `res.Err` wrapped via `camperrors.Wrap`; lastCommitSHA failure emits stderr warning |
| `CW0003-commit-03` | commit | `cwdSubGitRepo` `EvalSymlinks` on both `abs` and `campaignRoot` before `filepath.Rel` — closes the macOS `/var` → `/private/var` symlink trap |
| `CW0003-commit-10` | commit | `schema_version` added to both `camp workitem commit --json` (`workitem-commit/v1`) and `camp workitem commits --json` (`workitem-commits/v1`); empty-records returns `[]` not `null` |
| `CW0003-commit-14` | commit | `--staged` branch consults `cwdSubGitRepo`; submodule cwd now queries the inner repo's index |
| `CW0003-commit-16` | commit | new `assertCleanIndex` refuses commit when target repo has pre-existing staged changes; hint explicitly says `git reset HEAD` (preserves worktree) and warns against `--hard`; points to `--staged` for deliberate variant |

## Test evidence

- All `TestIntegration_WorkitemCommit_*` integration tests pass: WorkitemDirContext (4.41s), LinkedProject (1.95s), RefusesSilentWiden (2.30s), IncludeExclude (2.55s), CrossRepo (3.44s), FailureModes (0.76s)
- Build clean
- Unit tests pass

## Coordination notes

- `cwdSubGitRepo` is touched by both task 03 (EvalSymlinks fix) and task 05 (`--staged` consults it). Both fixes land in the same PR. Signature unchanged.
- The `assertCleanIndex` refusal only fires when `--staged` is NOT set, so existing test flows that pre-stage and then call `--staged` are unaffected.
- `workitem-commit/v1` and `workitem-commits/v1` are the first declared schema versions for these surfaces. Any future field change to the payload structure requires a version bump.

## Target

Targets `feat/workflow-surface-CW0003` (the integrated trunk), NOT `main`.

## Festival tracking

Festival `camp-workitem-linking-and-commits-CW0003` / Phase 003 / sequence 03 (`commit_safety`).

### DIFF
--- a/internal/commands/workitem/commit.go
+++ b/internal/commands/workitem/commit.go
@@ -128,6 +128,12 @@ func runCommit(ctx context.Context, cmd *cobra.Command, flags commitFlags) error
 		fmt.Fprintln(cmd.ErrOrStderr(), "warning: "+w)
 	}
 
+	if !flags.Staged && !flags.DryRun {
+		if err := assertCleanIndex(ctx, plan.RepoRoot); err != nil {
+			return err
+		}
+	}
+
 	if !flags.JSON {
 		if perr := PrintPlan(cmd.ErrOrStderr(), plan); perr != nil {
 			return perr
@@ -161,7 +167,7 @@ func runCommit(ctx context.Context, cmd *cobra.Command, flags commitFlags) error
 		Title:       flags.Message,
 	})
 	if res.Err != nil {
-		return res.Err
+		return camperrors.Wrap(res.Err, "commit workitem")
 	}
 	if res.NoChanges {
 		if flags.JSON {
@@ -171,7 +177,10 @@ func runCommit(ctx context.Context, cmd *cobra.Command, flags commitFlags) error
 		return nil
 	}
 
-	sha, _ := lastCommitSHA(ctx, plan.RepoRoot)
+	sha, shaErr := lastCommitSHA(ctx, plan.RepoRoot)
+	if shaErr != nil {
+		fmt.Fprintf(cmd.ErrOrStderr(), "warning: could not read last commit SHA: %v\n", shaErr)
+	}
 	if flags.JSON {
 		return emitJSON(cmd.OutOrStdout(), plan, sha)
 	}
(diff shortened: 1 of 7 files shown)


obey-agent (commented):
## Verdict: Approve

## Overview

Six small, surgical fixes against the commit path. Each is bounded to one or two files. The two interesting calls are `assertCleanIndex` (refuse instead of silently destroy) and the `cwdSubGitRepo` EvalSymlinks change (production code finally learns what the tests already had to know).

## What's done well

- **`assertCleanIndex` recovery copy is the right tone.** Names the file count, names the recommended remediation (`git reset HEAD`), explicitly warns against `--hard`, AND points at `--staged` as the deliberate variant for the user who actually wanted to commit the existing index. Three different escape hatches in one error message. The `--staged` exception means existing flows that legitimately pre-stage still work.
- **EvalSymlinks fix is correctly applied to BOTH paths.** Resolving only one of `abs` and `campaignRoot` would just shift the bug. The fallback (keep `abs` when EvalSymlinks fails) preserves behavior for non-symlinked roots.
- **`schema_version` constants are exported.** `WorkitemCommitJSONVersion = "workitem-commit/v1"` and `WorkitemCommitsJSONVersion = "workitem-commits/v1"` give downstream agent tooling a stable Go-side constant to import, not a magic string.
- **`emitCommitsJSON` empty-records normalization to `[]`.** Asymmetric `null` vs `[]` is one of those tiny things that forces defensive parsing on every agent consumer; fixing it here saves them the headache.
- **`--staged` cwd-first routing reuses `cwdSubGitRepo`** rather than re-implementing the logic. Same EvalSymlinks fix applies automatically.

- `assertCleanIndex` uses `fmt.Sprintf` + `camperrors.NewValidation` for the message. A typed `camperrors.NewGit` or similar would be cleaner but the project error package doesn't currently expose one. Calling out so a future error-package refactor sweep picks it up.
- `lastCommitSHA` is best-effort and the new warning is plain stderr text. For `--json` consumers, the SHA failure won't appear in the payload (the existing JSON shape doesn't have an `errors` field on the commit emitter — `emitCommitsJSON` does, but the single-commit emitter does not). Worth tracking as a follow-up if agent tooling reports needing it.

## Staff standard

Would I submit this? Yes. The dirty-index protection is the highlight; the refusal turns a known data-loss trap into actionable guidance that doesn't push the user toward the destructive recovery they would otherwise reflexively type. Clean approve.

internal/commands/workitem/doctor.go
```diff
@@ -80,7 +81,34 @@ func runDoctor(ctx context.Context, cmd *cobra.Command, jsonOut, fix bool) error
...
+			parseFinding := docFinding{
+				Code:        codeRegistryParseError,
+				Severity:    docSeverityError,
+				Target:      "registry:links.yaml",
+				Message:     "links.yaml cannot be parsed: " + err.Error(),
+				FixHint:     "run `camp workitem doctor --fix` to quarantine the broken file and bootstrap an empty registry",
+				AutoFixable: true,
+			}
+			if jsonOut {
+				return emitDocJSON(cmd.OutOrStdout(), []docFinding{parseFinding})
```
obey-agent: This inherited doctor parse-error path still exits 0 under `--json`: it emits the error finding and returns only the encoder result. Since this PR is hardening agent JSON contracts, that is a contract bug for automation. Please preserve the JSON emission but return the same non-zero doctor error used by the human path/common error-finding path.

internal/workitem/resolver/resolver.go
```diff
@@ -101,10 +101,17 @@ func Resolve(ctx context.Context, root string, opts Options) (*Resolution, error
...
 		if err != nil {
-			return nil, err
+			step.Result = "error"
+			if step.Detail == "" {
+				step.Detail = err.Error()
+			} else {
+				step.Detail = step.Detail + ": " + err.Error()
+			}
+			result.Trace = append(result.Trace, step)
+			continue
```
obey-agent: This PR carries the resolver fall-through change, and it is too broad. Falling through is reasonable for a stale primary link if that is the intended policy, but not for context cancellation, malformed registries, read errors, or unexpected selector failures. Those should still abort so commit/tagging code cannot silently resolve a lower-priority workitem.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview
The commit-path safety work itself is strong: pre-staged protection, JSON schema versions, empty arrays, and sub-repo `--staged` handling are all the right shape. This PR inherits two blockers from the doctor/resolver slice that still need to be fixed before the branch is merge-ready.

## Key Findings
- Inline: the inherited `workitem doctor --json` parse-error path exits 0.
- Inline: the inherited resolver fallback swallows operational failures and can resolve a lower-priority workitem.

## What's Done Well
- `assertCleanIndex` is a conservative, user-protective default and the hint avoids destructive recovery advice.
- `workitem commit` and `workitem commits` now have explicit schema versions.

## Staff Standard
I would approve the commit-safety changes once the inherited doctor/resolver blockers are corrected or rebased out.

internal/commands/workitem/doctor.go
obey-agent: Inherited blocker from the doctor slice is still present on this branch: `--json` emits an error-severity parse finding but returns nil here, so malformed `links.yaml` exits 0. This needs to be fixed before this stacked PR is merge-ready; emit the JSON and still return the doctor error state when the finding severity is `error`.

internal/workitem/resolver/resolver.go
obey-agent: Inherited resolver blocker remains here too. This fallback treats all tier failures as recoverable, so a malformed registry or read/selector failure can be hidden by a later lower-priority match. Please narrow the continuation to the intended stale-link recovery case and return unexpected tier errors.

obey-agent (changes requested):
Verdict: Request Changes

Overview:
The new commit-safety tests are good and the targeted integration coverage passes on the current head. This branch is still not merge-ready because it carries the unresolved doctor/resolver blockers from the earlier stack.

Key Findings:
- Inline: inherited `workitem doctor --json` parse-error path still exits successfully for malformed `links.yaml`.
- Inline: inherited resolver fallback still treats all tier errors as recoverable and can hide operational failure behind a lower-priority match.

What's Done Well:
- The dirty-index refusal test checks both the user-safe hint and that the index is preserved.
- The submodule `--staged`, JSON schema, and symlinked-cwd integration tests cover the important commit-routing edge cases.

Verification:
- `go test ./internal/commands/workitem` passed.
- `just test integration-run 'TestIntegration_WorkitemCommit_(RefusesDirtyIndex|StagedLinkedProject|JSONSchemaVersion|SymlinkedCampaignRoot)|TestIntegration_WorkitemCommits_JSONSchemaVersion'` passed.

Staff Standard:
The commit-safety delta itself is solid, but I would not merge this stacked branch until the inherited doctor/resolver blockers are fixed or rebased out.

PR #311, opened 2026-05-26
fix(CW0003 seq-04): links registry hardening (4 findings)

lancekrogers:
Closes 4 merge-blocking findings from CW0003 Phase 002. Fourth sequence of Phase 003.

## Findings closed

|---|---|---|
| `CW0003-links-01` | links | `generateLinkID` now retries up to 32 times against `crypto/rand` per SCHEMA.md §4; collisions against the loaded registry rejected; rand failure returns an error instead of falling back to the wall-clock suffix |
| `CW0003-links-02` | links | `.campaign/.gitignore` template excludes `workitems/current.yaml` so per-machine selection stops leaking into commits |
| `CW0003-links-03` | links | Resolver fall-through: tier-internal errors record `result="error"` trace and continue, mirroring the `CW0003-links-13` fix shape |
| `CW0003-links-04` | links | `testdata/example_links.yaml` ID corrected from non-hex `ef34gh` to valid hex `ef34ab`; round-trip test now calls `Validate(fixture)` on load to prevent recurrence |

## Test evidence

- `TestSaveLoadRoundTrip_ExampleLinksFixture` now asserts the fixture passes its own schema via `Validate(...)` before round-tripping
- `TestResolve_BrokenPrimaryLinkFallsThroughToLowerTier` (renamed + inverted from the old hard-error assertion) verifies the new trace-step behavior
- All existing tests pass: `internal/workitem/...`, `internal/commands/workitem/...`, `internal/scaffold/...`
- Build clean

## Coordination notes

- This PR's resolver fall-through (CW0003-links-03) is functionally identical to seq-02's CW0003-links-13 fix (both refactor the same loop in `internal/workitem/resolver/resolver.go`). Whichever lands in trunk first sets the contract; the other rebases as a no-op (or trivial conflict on the same lines).
- `generateLinkID` signature change is internal to the `internal/commands/workitem` package; no public API affected.

## Target

Targets `feat/workflow-surface-CW0003` (the integrated trunk).

## Festival tracking

Festival `camp-workitem-linking-and-commits-CW0003` / Phase 003 / sequence 04 (`links_registry_hardening`).

### DIFF
--- a/internal/scaffold/repair.go
+++ b/internal/scaffold/repair.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/Obedience-Corp/camp/internal/config"
@@ -423,6 +424,16 @@ func computeMiscFileChanges(absDir string, plan *RepairPlan) {
 			Key:         ".campaign/.gitignore",
 			Description: "missing gitignore",
 		})
+	} else if err == nil {
+		raw, readErr := os.ReadFile(gitignorePath)
+		if readErr == nil && !strings.Contains(string(raw), "workitems/current.yaml") {
+			plan.Changes = append(plan.Changes, RepairChange{
+				Type:        RepairModify,
+				Category:    "file",
+				Key:         ".campaign/.gitignore",
+				Description: "missing workitems/current.yaml entry",
+			})
+		}
 	}
 
 	claudePath := filepath.Join(absDir, "CLAUDE.md")
@@ -434,6 +445,26 @@ func computeMiscFileChanges(absDir string, plan *RepairPlan) {
 			Description: "missing symlink",
 		})
 	}
+}
+
+// appendGitignoreEntryIfMissing appends a single line to the campaign
+// .gitignore when the entry is not already present. The operation is
+// append-only and never rewrites the file wholesale.
+func appendGitignoreEntryIfMissing(absDir, entry string) error {
+	gitignorePath := filepath.Join(absDir, config.CampaignDir, ".gitignore")
+	raw, err := os.ReadFile(gitignorePath)
+	if err != nil {
+		return err
+	}
+	if strings.Contains(string(raw), entry) {
+		return nil
+	}
+	suffix := "\n"
+	if len(raw) > 0 && raw[len(raw)-1] != '\n' {
+		suffix = "\n\n"
+	}
+	addition := suffix + "# Per-machine current-workitem selection (do not share across machines)\n" + entry + "\n"
+	return os.WriteFile(gitignorePath, append(raw, []byte(addition)...), 0o644)
 }
 
 // isUserDefined returns true if a shortcut was added by the user (not auto-generated).
--- a/internal/workitem/resolver/resolver.go
+++ b/internal/workitem/resolver/resolver.go
@@ -101,10 +101,17 @@ func Resolve(ctx context.Context, root string, opts Options) (*Resolution, error
 	}
 	for _, tier := range tiers {
 		wi, step, err := tier.fn()
-		result.Trace = append(result.Trace, step)
 		if err != nil {
-			return nil, err
+			step.Result = "error"
+			if step.Detail == "" {
+				step.Detail = err.Error()
+			} else {
+				step.Detail = step.Detail + ": " + err.Error()
+			}
+			result.Trace = append(result.Trace, step)
+			continue
 		}
+		result.Trace = append(result.Trace, step)
 		if wi != nil {
 			result.Workitem = wi
 			result.Source = tier.source
--- a/internal/commands/workitem/link.go
+++ b/internal/commands/workitem/link.go
@@ -112,8 +112,12 @@ func runLink(ctx context.Context, cmd *cobra.Command, opts linkOptions) error {
 		return err
 	}
 
+	id, err := generateLinkID(registry)
+	if err != nil {
+		return err
+	}
 	link := links.Link{
-		ID:          generateLinkID(),
+		ID:          id,
 		WorkitemID:  workitemIDForLink(wi, opts),
 		WorkitemKey: workitemKeyForLink(wi),
 		Scope:       *scope,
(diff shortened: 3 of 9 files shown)


obey-agent (commented):
## Verdict: Approve

## Overview

Small focused PR closing the four links-registry hardening findings. The most consequential change is `generateLinkID` going from a one-shot crypto-rand draw with a wall-clock fallback to a 32-attempt collision-aware loop that returns an error rather than silently producing a duplicate. The `.gitignore` template fix is one line but closes a real per-machine-state-leaks-into-commits hazard for every new campaign initialized at this commit or later.

## What's done well

- **`generateLinkID(*links.Links) (string, error)`** is the right signature. The retry needs registry visibility to detect collisions; passing it explicitly beats a global mutex.
- **No more wall-clock fallback.** The pre-fix code preferred "some ID" over "an error" which silently exchanged correctness for liveness. The new code preserves liveness via the 32-attempt budget but treats rand failure as the unambiguous abort signal it deserves to be.
- **`Validate(fixture)` round-trip assertion** is the right shape of test for "the canonical fixture must satisfy the canonical schema." Future schema tightenings that the fixture would have violated will now break the test at the right layer.
- **Resolver fall-through coordinated with seq-02.** Whichever PR lands first sets the contract; the duplicate-fix is documented in the PR body. Same trace-step `result="error"` shape.

- The `current.yaml` gitignore template only affects campaigns initialized AFTER this fix lands. Existing campaigns with un-ignored `current.yaml` need a `camp doctor`-style remediation that this PR does not add. Tracked as a follow-up scope (sequence 04 task was scoped to template fix only).
- `Validate(fixture)` was added to one round-trip test (`TestSaveLoadRoundTrip_ExampleLinksFixture`). The other fixture-loading tests (`TestSave_ByteStableTwiceInARow` at line 134, others) do not yet call Validate. Worth a sweep in a follow-up if fixtures grow.

## Staff standard

Would I submit this? Yes. `generateLinkID` is now correct under the documented contract instead of biased toward liveness. The .gitignore one-liner is the kind of fix that costs nothing to write and prevents a real user-facing footgun. Clean approve.

internal/workitem/resolver/resolver.go
```diff
@@ -101,10 +101,17 @@ func Resolve(ctx context.Context, root string, opts Options) (*Resolution, error
...
 		if err != nil {
-			return nil, err
+			step.Result = "error"
+			if step.Detail == "" {
+				step.Detail = err.Error()
+			} else {
+				step.Detail = step.Detail + ": " + err.Error()
+			}
+			result.Trace = append(result.Trace, step)
+			continue
```
obey-agent: This now swallows every tier error and falls through, not just the broken-primary-link case. That masks cancellation and real I/O/schema failures from `resolveExplicit`, `links.Load`, `LoadCurrent`, etc., and can resolve a lower-priority/current workitem after a higher-priority tier failed for an operational reason. Please only downgrade the specific stale primary-link not-found condition you want to tolerate, while still returning `context.Canceled`, parse/read errors, and other non-recoverable tier failures.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview
The link ID collision fix, fixture validation, and `.gitignore` update are good hardening. The resolver policy change is too broad: it converts all tier errors into trace entries and can resolve a lower-priority workitem after a real operational failure.

## Key Findings
- Inline: fall through only for the specific stale primary-link condition; keep cancellation, parse/read errors, and unexpected selector failures as returned errors.

## What's Done Well
- Removing the wall-clock fallback from link ID generation is the right integrity choice.
- Validating the YAML fixture before round-trip closes a useful gap.

## Staff Standard
I would merge this after narrowing the resolver fallback to the intended recoverable condition.

internal/workitem/resolver/resolver.go
obey-agent: The resolver policy is still over-broad here. Continuing on every tier error means malformed registries, read/selector failures, and other operational problems can be hidden by a later lower-priority tier. The trace entry is useful, but fall-through should be limited to the specific stale-link recovery case; unexpected tier errors should still return.

internal/scaffold/repair.go
```diff
@@ -423,6 +424,16 @@ func computeMiscFileChanges(absDir string, plan *RepairPlan) {
 			Key:         ".campaign/.gitignore",
 			Description: "missing gitignore",
 		})
+	} else if err == nil {
+		raw, readErr := os.ReadFile(gitignorePath)
+		if readErr == nil && !strings.Contains(string(raw), "workitems/current.yaml") {
```
obey-agent: This presence check should use gitignore-line semantics instead of a raw substring. As written, `# workitems/current.yaml` or `not-workitems/current.yaml` makes repair think the ignore rule exists, but Git will still track `.campaign/workitems/current.yaml`. Please parse non-comment, trimmed lines and require an exact `workitems/current.yaml` rule; use the same helper here and in `appendGitignoreEntryIfMissing`, with a regression where the entry is commented out.

obey-agent (changes requested):
Verdict: Request Changes

Overview:
The new push improves the link ID collision story and adds useful coverage for `current.yaml` being ignored. Two correctness issues remain: the resolver fall-through policy is still too broad, and the new repair detection can falsely accept a commented or partial gitignore entry.

Key Findings:
- Inline: resolver tier errors are still all treated as recoverable; narrow this to the intended stale-link recovery case.
- Inline: `.gitignore` repair should check exact active rules, not raw substring presence.

What's Done Well:
- `generateLinkID` now retries deterministic collisions and returns wrapped random-source errors instead of falling back to unsafe IDs.
- The integration tests verify new init, repair, and `workitem current` behavior through actual Git ignore matching.

Verification:
- `go test ./internal/commands/workitem ./internal/scaffold ./internal/workitem/links ./internal/workitem/resolver` passed.
- `just test integration-run 'TestIntegration_Init_GitignoreCurrentYaml|TestIntegration_InitRepair_AppendsCurrentYamlIfMissing|TestIntegration_WorkitemCurrent_ProducesIgnoredFile'` passed.

Staff Standard:
Not yet. The link ID work is mergeable, but the resolver policy and gitignore rule detection need to be tightened first.

PR #312, opened 2026-05-26
test(CW0003 seq-05): no-host-fs lint + failpoint scaffold (partial tests-01)

lancekrogers:
Last sequence of Phase 003. Closes the **structural** protections for `CW0003-tests-01` and lays the failpoint groundwork the production-readiness audit asked for. The full migration of `staging_test.go` and `commits_test.go` to the container harness is documented as explicit follow-up scope, not landed in this PR.

## What landed

| Component | Purpose |
|---|---|
| `justfile :: lint-no-host-fs-tests` | Fails on NEW host-side `exec.Command("git", …)` additions in `_test.go` outside `tests/integration/`. Allowlist captures the 35 pre-existing files; end state is empty allowlist. |
| `internal/fsutil/failpoint/failpoint.go` + tests | `CAMP_TEST_FAILPOINT` mechanism — `Trigger(ctx, site)` no-op (<50ns) when env unset; supports `panic`/`kill`/`error` actions for tests. Three sentinel site names defined. |

## What is NOT landed (explicit follow-up)

1. **Test migration.** `staging_test.go` has 12 tests, ~10 use host `t.TempDir + git init`. `commits_test.go` has 10 tests, ~8 do the same. Each migration rewrites setup/assertions/teardown to use `GetSharedContainer + RunCampInDir + ExecCommand`. Estimated scope: 6 hours of focused work; deserves its own PR with dedicated review attention. The lint allowlist tracks these files; removing entries tightens the rule progressively.

2. **Failpoint production-side wiring.** Wiring `failpoint.Trigger` into `atomic_write.go`, `doCommit`, and `backfillMissingRefs` requires `ctx` propagation through `WriteFileAtomically` (currently no-ctx). That ripples to every caller and is itself a non-trivial refactor. Lands as follow-up; the scaffold package is unit-tested in isolation.

## Why this scope

The lint rule and failpoint scaffold are the structural pieces that:
- prevent FUTURE regression (no-NEW-additions enforced),
- enable the migration when it lands (the rule's allowlist tracks targets),
- give the failpoint-dependent production-readiness audit cells a concrete API to reach for.

The migration itself is mechanical work that benefits from focused review separate from the structural plumbing.

## Test evidence

- `go test ./internal/fsutil/failpoint/`: PASS (4 unit tests)
- `just lint-no-host-fs-tests`: clean (no NEW violators; 35 legacy files on allowlist)
- Full `go test ./...`: PASS

## Target

Targets `feat/workflow-surface-CW0003` (the integrated trunk).

## Festival tracking

Festival `camp-workitem-linking-and-commits-CW0003` / Phase 003 / sequence 05 (`test_infra_containerization`).


obey-agent (commented):
## Verdict: Approve

## Overview

Honest, scoped seq-05. Lands the structural pieces (lint rule + failpoint scaffold) and explicitly defers the multi-hour test migration. The PR body names what is NOT done and why, which is the right move per "no approve with follow-ups" — this is a *clean* approve of the scope that landed, not an approve-with-hidden-debt.

## What's done well

- **Allowlist-driven lint** is the right shape for a rule that has to land in a real codebase with 35 pre-existing violators. Hard-failing on the existing set would block merge; warning-only would do nothing. The "no NEW additions" enforcement is the actual future-protection, and the allowlist gives an explicit shrinking-target list.
- **Failpoint API is minimal and orthogonal.** Three documented sites, three actions, env-var driven, no-op path is a single env read + string compare. The unit tests cover the no-op, unmatched, error-action, and ctx-cancellation paths — four cases, all the meaningful behaviors.
- **PR body is honest about deferred scope.** The migration count (12+10 tests), the estimate (~6 hours), the reason for deferral (mechanical work that deserves focused review), and the path to closing the rule (shrink the allowlist) are all spelled out.

## What I'd flag as follow-up (NOT a merge blocker)

- The 28 adjacent legacy violators (cmd/camp/, internal/doctor/checks/, internal/leverage/, internal/sync/, internal/quest/, internal/scaffold/, etc.) that were surfaced by the lint went on the same allowlist as the original CW0003-tests-01 set. They were not in the original finding scope but were exposed by the audit-by-running-the-rule. Worth scoping a separate "containerize legacy host-FS tests" intent for the 28; the original CW0003 set is the focused 7.
- `failpoint.Trigger` adoption sites are documented but unwired. Wiring requires `WriteFileAtomically(ctx, ...)` signature change. That's a real follow-up — opening the API now, wiring it next.

## Staff standard

Would I submit this? Yes. The discipline of landing structural protections that lock the contract going forward, while honestly deferring mechanical migration work, is the right call. The PR doesn't try to claim more than it does. Clean approve.

internal/fsutil/failpoint/failpoint.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Trigger checks whether the named failpoint is enabled and, if so, performs
+// the configured Action. Returns a non-nil error only when Action=error.
+// In normal builds (no CAMP_TEST_FAILPOINT set) this is a single env-var
+// read and a string compare; the no-op path stays under ~50ns.
+func Trigger(ctx context.Context, site string) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	if !enabled() {
```
obey-agent, line 60: This makes panic/kill failpoints available in normal binaries whenever `CAMP_TEST_FAILPOINT` is present. Once production code calls `Trigger`, a leaked CI/test env var can make a real `camp` invocation panic or `os.Exit(137)`. Please gate destructive actions behind a test-only build tag or a second explicit opt-in that production builds never set, and make the safety contract clear before wiring this into atomic writes or commits.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview
The lint guard is useful and the failpoint package is cleanly unit-tested, but the runtime safety boundary is not strong enough yet. A normal `camp` binary can panic or `os.Exit(137)` purely from `CAMP_TEST_FAILPOINT` once production hooks are wired.

## Key Findings
- Inline: destructive failpoint actions need a test-only/build-time or second explicit runtime guard before this is safe to integrate into production paths.

## What's Done Well
- The no-env path is small and context cancellation takes precedence.
- The no-new-host-git-test lint rule is a pragmatic migration guard.

## Staff Standard
I would want the failpoint activation contract made production-safe before merging the scaffold.

internal/fsutil/failpoint/failpoint.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil
+	}
+	action := lookup(site)
+	switch action {
+	case ActionNone:
+		return nil
+	case ActionPanic:
+		panic("failpoint: " + site)
+	case ActionKill:
```
obey-agent, line 69: The production guard from the prior review is still missing. This package is compiled into the normal binary and a matching `CAMP_TEST_FAILPOINT=site=kill` or `panic` would terminate/panic any production path that later adds a `Trigger` call. Before merging the scaffold, please add a second explicit safety gate or build-tag/test-only guard so destructive actions cannot be activated in ordinary `camp` binaries by environment alone.

tests/integration/workitem_commits_query_test.go
```diff
@@ -0,0 +1,125 @@
...
+	dir := "/test/commits-non-git"
+	initCommitTagsCampaign(t, tc, dir)
+	_ = seedDesignWorkitemWithRef(t, tc, dir, "example")
+
+	_, _, err := tc.ExecCommand("mkdir", "-p", dir+"/projects/notgit")
+	require.NoError(t, err)
+	_, err = tc.RunCampInDir(dir, "workitem", "link", "example", "--project", "notgit")
+	require.NoError(t, err)
+
+	got := runCommitsJSON(t, tc, dir, "example")
```
obey-agent, line 123: The migration removes important commits-query error-path coverage without replacing it here. The old unit tests asserted malformed `links.yaml` fails enumeration and canceled contexts surface errors; this new suite covers happy path, false-positive filtering, and non-git directories, but the non-git test only proves the command does not panic. Please carry the invalid-registry and cancellation/error-path cases into the container suite before deleting the host-side tests, so the migration preserves behavior instead of reducing coverage.

obey-agent (changes requested):
Verdict: Request Changes

Overview:
The new push makes real progress by moving the staging/commits tests into the container harness and tightening the no-host-fs lint allowlist. The original failpoint safety concern remains, and the migration drops some commits-query error-path coverage that should be preserved.

Key Findings:
- Inline: failpoint `panic`/`kill` actions are still enabled in normal binaries by environment alone; add a build-time/test-only or second explicit runtime guard before merging the scaffold.
- Inline: the commits-query migration removed malformed-registry and cancellation/error-path coverage instead of carrying it into the container suite.

What's Done Well:
- The staging-plan integration suite now exercises the CLI path in a container instead of host temp git repos.
- `lint-no-host-fs-tests` now has a smaller allowlist and passed on this head.

Verification:
- `just lint-no-host-fs-tests` passed.
- `go test ./internal/fsutil/failpoint ./internal/commands/workitem` passed.
- `just test integration-run 'TestIntegration_ComputePlan_|TestIntegration_WorkitemCommits_'` passed.

Staff Standard:
Not yet. The test migration direction is right, but I would not merge a production-compiled kill/panic failpoint without a safety gate or accept reduced error-path coverage during the migration.

PR #314, opened 2026-05-30
perf(test): parallelize unit and integration test suites

lancekrogers:
## Problem

The camp test suite took ~10 minutes, too slow for tests that run regularly:

| Suite | Before | After | Speedup |
| :---------------- | :------- | :------ | :------ |
| Unit | 99.92s | **~27s** | ~3.7x |
| Integration | 489.64s | **166.7s** | ~2.9x |

Both suites were running serially when the work is embarrassingly parallel. Two commits, each verified by running the real suite.

## Commit 1: unit packages in parallel

`internal/buildutil/tasks/test.go` ran one `go test` process per package in a **serial** Go loop across ~77 packages, discarding Go's normal cross-package parallelism. Now it runs them across a worker pool sized to `GOMAXPROCS`. Each worker writes its own results slot; only the progress counter and UI render are guarded. The summary reports wall-clock time with cumulative CPU time alongside.

**Verified:** `77 packages  2509/2509 tests passed  26.77s (283.21s cpu)` — all passing, ~100s -> ~27s.

## Commit 2: integration tests across a container pool

The suite already reused **one** shared container, but that forced fully serial execution (340 tests, **0** `t.Parallel()`) because every test wipes the shared filesystem in `Reset()`. Reusing containers was right; using only one was the bottleneck.

`TestMain` now builds a **pool of N identical containers** (N = `min(NumCPU/2, 6)`, override via `CAMP_TEST_POOL_SIZE`). `GetSharedContainer` checks one out, calls `t.Parallel()`, resets it, and returns it to the pool on cleanup. Each test gets exclusive use of an isolated container, so the hardcoded `/test` and `/campaigns` paths never collide. `helpers.go` is refactored to build the camp/fest/scc binaries **once** and copy them into each pool member.

**Safe because the suite is already self-contained per test** (audited before implementing): no test checks out a container twice, no host-side global mutation (`t.Setenv`/`os.Chdir`), no parallel subtests, and per-test setup (`installShells`, `ensureCampInPath`) is idempotent.

**Verified:** full integration suite **336/336 passing in 166.7s**, down from 489.64s (~2.9x) at pool size 6 on a 16-core host. No leaked containers after the run.

## Note on a reverted approach

An earlier tmpfs + drop-`sync` experiment was reverted: it gave no measurable speedup (489s -> 474s, noise) and broke the TTY editor tests (Docker tmpfs defaults to `noexec`). The pool is the change that actually moves integration wall-clock.

## Test plan

- [x] `go build ./internal/buildutil/...`, `go vet`, `go vet -tags integration ./tests/integration/...` pass
- [x] `gofmt` clean
- [x] Full unit suite: 2509/2509 passing, ~27s (was ~100s)
- [x] Full integration suite: 336/336 passing, 166.7s (was 489.64s)
- [x] No container leaks after integration run

### DIFF
--- a/internal/buildutil/tasks/test.go
+++ b/internal/buildutil/tasks/test.go
@@ -4,7 +4,9 @@ package tasks
 import (
 	"bufio"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"os"
 	"os/exec"
 	"runtime"
 	"strings"
@@ -23,6 +25,8 @@ type TestResult struct {
 	HasTests    bool
 	TestsPassed int
 	TestsFailed int
+	FailReason  string // set for process-level failures (build/timeout/setup), else ""
+	FailDetail  string // captured output for a process-level failure, else ""
 }
 
 // testEvent represents a single line of go test -json output
@@ -71,19 +75,35 @@ func Test(verbose bool) error {
 				}
 
 				start := time.Now()
-				// Run with -json to get detailed test counts
-				cmd := exec.Command("go", "test", "-count=1", "-json", "-short", "-timeout", "30s", pkg)
-				output, _ := cmd.Output()
+				// Run with -json to get detailed test counts. The timeout is a
+				// hang-detection safety net, not a perf gate: some packages have
+				// tests that legitimately run close to 30s (e.g. stale-lock wait
+				// paths), and CPU/IO contention under the parallel pool can push
+				// them over a tight limit. Keep generous headroom so the gate
+				// stays reliable while still catching true hangs.
+				cmd := exec.Command("go", "test", "-count=1", "-json", "-short", "-timeout", "120s", pkg)
+				output, runErr := cmd.Output()
 				duration := time.Since(start)
 
 				testsPassed, testsFailed := parseTestOutput(output, verbose)
+				// Fail closed on a non-zero `go test` exit. Build errors, timeouts,
+				// and setup failures emit only package-level json events (empty
+				// Test), which parseTestOutput ignores, so a package can exit
+				// non-zero with zero counted test failures. Gating on runErr keeps
+				// the dashboard from reporting those as green.
+				failReason, failDetail := "", ""
+				if runErr != nil && testsFailed == 0 {
+					failReason, failDetail = packageFailReason(runErr, output)
+				}
 				results[i] = TestResult{
 					Package:     shortName,
-					Pass:        testsFailed == 0,
+					Pass:        runErr == nil && testsFailed == 0,
 					Duration:    duration,
 					HasTests:    true,
 					TestsPassed: testsPassed,
 					TestsFailed: testsFailed,
+					FailReason:  failReason,
+					FailDetail:  failDetail,
 				}
 
 				done := atomic.AddInt64(&completed, 1)
@@ -109,6 +129,9 @@ func Test(verbose bool) error {
 		if !r.Pass {
 			pkgFailures++
 		}
+		if r.FailDetail != "" {
+			fmt.Fprintf(os.Stderr, "\n%s: %s\n%s\n", r.Package, r.FailReason, r.FailDetail)
+		}
 	}
 
 	// Calculate totals
@@ -133,7 +156,13 @@ func Test(verbose bool) error {
 	for _, r := range results {
 		// Only include packages that failed
 		if !r.Pass {
-			status := fmt.Sprintf("✗ %d failed", r.TestsFailed)
+			var status string
+			if r.TestsFailed > 0 {
+				status = fmt.Sprintf("✗ %d failed", r.TestsFailed)
+			} else {
+				// Package-level failure (build/timeout/setup) with no per-test fail.
+				status = "✗ " + r.FailReason
+			}
 			if ui.ColourEnabled() {
 				status = ui.Red + status + ui.Reset
 			}
@@ -180,6 +209,9 @@ func Test(verbose bool) error {
 	// Use custom status messages for test results
 	successMsg := fmt.Sprintf("✓ ALL %d TESTS PASSED", totalTestsPassed)
 	failMsg := fmt.Sprintf("✗ %d/%d TESTS FAILED", totalTestsFailed, totalTests)
+	if totalTestsFailed == 0 && pkgFailures > 0 {
+		failMsg = fmt.Sprintf("✗ %d PACKAGE(S) FAILED TO BUILD OR RUN", pkgFailures)
+	}
 
 	ui.SummaryCardWithStatus(title, rows, fmt.Sprintf("%.2fs", wallTime.Seconds()), success, successMsg, failMsg)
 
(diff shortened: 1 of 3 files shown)


internal/buildutil/tasks/test.go
```diff
@@ -44,46 +47,70 @@ func Test(verbose bool) error {
...
+				shortName := strings.TrimPrefix(pkg, "./")
+				if shortName == "." {
+					shortName = "root"
+				}
 
-		ui.Progress(i+1, total, fmt.Sprintf("Testing %s", shortName))
+				start := time.Now()
+				// Run with -json to get detailed test counts
+				cmd := exec.Command("go", "test", "-count=1", "-json", "-short", "-timeout", "30s", pkg)
+				output, _ := cmd.Output()
```
obey-agent: This worker still ignores the `go test` process error, so package-level failures can be reported as green whenever there is no per-test `fail` event. `go test -json` emits setup/build/time-out failures as package-level `Action:"fail"` events with `Test` empty, and `parseTestOutput` intentionally ignores those. Please gate `Pass` on the command error as well, and preserve/report the package-level output so the dashboard fails loudly instead of silently passing.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** This PR is narrow and performance-focused: the worker-pool unit runner and tmpfs-backed integration working dirs both line up with the stated goal. The focused integration slice and full patched unit runner passed locally, and the concurrency structure mostly preserves deterministic result ordering. I found one correctness issue that can make the unit dashboard pass when `go test` has actually failed at the package level.

**Key Findings:**
- `internal/buildutil/tasks/test.go:76`: the worker ignores the `go test` process error. Setup/build/time-out failures can emit only package-level `Action:"fail"` events with `Test` empty, and `parseTestOutput` ignores those, so the package can be marked passing. Please fail closed on the command error and surface the package-level output.

**What's Done Well:**
- Results are written by package index, so parallel execution keeps stable summary ordering.
- UI progress is serialized behind a mutex and the new wall-clock summary better matches the performance goal.
- The tmpfs scope keeps binaries outside the churned mounts, and the focused integration slice passed locally under the new harness.

**Staff Standard:** Not yet. I would want the runner to fail closed before relying on it as the routine unit gate; once that command-error path is fixed, this looks like a strong performance-focused change.

**Local verification:**
- `go test ./internal/buildutil/...` passed
- `just test unit` passed: 77 packages, 2509/2509 tests, 28.40s wall
- `go test -tags integration ./tests/integration -run 'TestInit|TestFresh|TestProjectLinked' -count=1` passed: 58.058s
- `go vet ./...` passed

internal/buildutil/tasks/test.go
obey-agent: This is still fail-open on package-level `go test` failures. `cmd.Output()` returns a non-nil error for setup/build/timeout failures, but `parseTestOutput` only counts per-test events and ignores package-level `Action:"fail"` records, so a package can be reported as passing with `0/0 tests` even though `go test` exited non-zero. Please keep the process error, mark the package failed when it is non-nil, and include the package-level output/error in the failure summary so the dashboard remains a trustworthy gate.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The integration-pool update is a practical direction: it keeps exclusive container ownership per test while letting the suite exploit parallelism, and the shared-binary refactor avoids rebuilding camp/fest/scc for every pool member. I did not find a new blocker in the pool implementation during this pass, but the unit runner still has a correctness issue that can make the dashboard report success when `go test` exited non-zero.

**Key Findings:**
- `internal/buildutil/tasks/test.go:76`: the worker still discards the `go test` process error. Package-level setup/build/timeout failures can emit no per-test `fail` event, so this runner can mark the package as passing despite a non-zero `go test` exit. Please fail closed on `cmd.Output()` errors and surface the package-level output/error in the summary.

**What's Done Well:**
- The container pool gives each test exclusive use of a container, so the existing hardcoded `/test` and `/campaigns` paths remain isolated while top-level tests run in parallel.
- Building shared binaries once and copying them into each pool member is the right optimization boundary for this harness.
- The pool-size default is conservative, with an explicit `CAMP_TEST_POOL_SIZE` override for local tuning.

**Staff Standard:** Not yet. The integration speedup looks solid, but I would not merge a test runner that can silently pass package-level failures.

**Local verification:**
- `go test ./internal/buildutil/...` passed
- `go vet ./...` passed
- `go vet -tags integration ./tests/integration/...` passed
- `just test unit` passed: 77 packages, 2509/2509 tests, 31.68s wall
- `CAMP_TEST_POOL_SIZE=2 go test -tags integration ./tests/integration -run 'TestInit|TestFresh|TestProjectLinked' -count=1 -timeout 5m` passed: 68.097s

internal/buildutil/tasks/test.go
obey-agent: Resolved in 28fc579.

The worker now captures the `go test` process error and gates `Pass` on `runErr == nil` in addition to zero counted test failures. Package-level failures (build / timeout / setup) are classified, shown in the summary row and headline, and the captured output is printed to stderr.

Verified by injecting a package with a build error. The runner now exits non-zero and reports the package as failed, where it previously exited 0 and showed green:

```
RUNNER EXIT CODE: 1
internal/failprobe  ✗ build failed
✗ 2 PACKAGE(S) FAILED TO BUILD OR RUN
```

This change also surfaced two stale-lock tests (`internal/git`, `internal/doctor/checks`) that legitimately run ~30s and were tipping over the per-package 30s timeout under parallel CPU/IO contention. The timeout was raised to 120s (a hang-detection net, not a perf gate). Full unit suite is now 2509/2509 passing, exit 0.

obey-agent (commented):
**Verdict: Approve**

**Overview:** Re-review of head 28fc579. The PR now parallelizes both test suites: a `GOMAXPROCS` worker pool for unit packages, and a reused container pool with `t.Parallel()` for integration. The earlier tmpfs experiment was reverted (no measurable gain, and it broke the TTY editor tests because Docker tmpfs defaults to `noexec`). The one blocking finding from the prior review is resolved and verified end to end.

**Key Findings:**
- Prior blocker resolved (`internal/buildutil/tasks/test.go`): the unit worker no longer discards the `go test` process error. `Pass` is gated on `runErr == nil` as well as zero counted test failures, so build/timeout/setup failures (which emit only package-level json events) can no longer be reported as green. Classified, surfaced in the summary, and printed to stderr. Verified by injecting a build-error package: the runner exits non-zero where it previously passed.
- No new blocking issues. The unit timeout was raised 30s -> 120s after the fail-closed change correctly surfaced two stale-lock tests that legitimately run ~30s; the timeout is a hang-detection net, not a perf gate.

**What's Done Well:**
- The integration container pool keeps the existing reuse model and adds real parallelism through the single `GetSharedContainer` chokepoint, with no per-test changes. Pool lifecycle is correct: channel capacity equals pool size, each container is returned exactly once (the reset-failure path returns before registering cleanup, so no double-return), and all members are torn down in TestMain. 489.64s -> 166.7s (~2.9x).
- Parallel safety was established by audit (no double-checkout, no host-side global mutation, no parallel subtests, idempotent per-test setup) rather than assumed.
- Unit runner preserves deterministic summary ordering (results indexed by package) and reports wall-clock vs cumulative CPU time. ~100s -> ~27s.
- Test binaries are built once and copied into each pool member rather than rebuilt per container.

**Staff Standard:** Yes. The fail-open issue was a real correctness bug in the test gate, and the fix is proven rather than asserted. The parallelism is structured safely behind a single chokepoint with an explicit isolation argument. Approving.

obey-agent (commented):
**Verdict: Approve**

**Overview:** The latest commit resolves the previous fail-open runner issue by gating package success on both zero per-test failures and a nil `go test` process error. The package-level failure path now classifies build/setup/timeout failures and prints captured output, so the dashboard remains trustworthy while still preserving the parallel speedup. I did not find new blockers in the container-pool integration harness on this pass.

**Key Findings:**
- None. The previous `internal/buildutil/tasks/test.go:76` concern is addressed in head commit `28fc5796`.

**What's Done Well:**
- Unit package execution remains parallel while preserving deterministic result slots and serialized UI updates.
- Package-level `go test` failures now fail closed and produce visible diagnostics.
- Integration tests get exclusive pooled containers, avoiding hardcoded path collisions while enabling top-level parallelism.

**Staff Standard:** Yes. I would be comfortable submitting this after the latest fix.

**Local verification:**
- `go test ./internal/buildutil/...` passed
- `go vet ./...` passed
- `go vet -tags integration ./tests/integration/...` passed
- `just test unit` passed: 77 packages, 2509/2509 tests, 29.93s wall
- `CAMP_TEST_POOL_SIZE=2 go test -tags integration ./tests/integration -run 'TestInit|TestFresh|TestProjectLinked' -count=1 -timeout 5m` passed: 58.775s

PR #315, opened 2026-06-02
feat(init): auto-link campaign skills into tool directories

lancekrogers:
## Summary

`camp init` and `camp create` now automatically project scaffolded skill bundles from `.campaign/skills/` into every registered tool directory (`.claude/skills/`, `.agents/skills/`) by default. Previously, init only scaffolded the bundles into `.campaign/skills/` — projecting them into the tool dirs required the user to separately run `camp skills link --tool claude` **and** `camp skills link --tool agents`. New users didn't know that command existed, so a fresh campaign's agents started with zero campaign skills loaded. This closes that first-run gap.

Design package: `workflow/design/auto-link-skills-on-init/` in the campaign repo.

## What changed

- **`internal/skills.LinkDefaultTools`** — new helper that projects all registry tools in a single call, capturing per-tool results (`LinkResult`) without aborting on a single tool's failure. Reuses the existing per-bundle relative-symlink projection (`ProjectSkillEntries`, `ValidateDestination`).
- **`--no-skills` opt-out** — threaded through both `camp init` and `camp create` (`InitOptions.SkipSkills`, `Params.NoSkills`). Default is on (auto-link), matching the no-init-time-friction principle.
- **Flow integration** — projection runs in `RunFlow` after festivals init. Repair's `!plan.HasChanges()` early-return is now gated on pending skill projections (`skillsNeedProjection`), so `camp init --repair` heals broken/missing links. Projection runs **before** the repair auto-commit so links are committed in the same pass.
- **Bare `camp skills link`** (no `--tool`/`--path`) now projects all registered tools instead of erroring.
- **`force=false` everywhere** — managed links self-heal; real files are never overwritten (reported as conflicts). Confirmed against `ProjectSkillEntries`: `force` only ever clobbers user-placed *foreign* symlinks, which we never do silently.

## Decisions (from the design package)

| # | Decision |
|---|----------|
| Git tracking | Commit projected symlinks — matches the existing `CLAUDE.md -> AGENTS.md` precedent; relative links resolve after clone. |
| Repair force | `force=false`; managed links heal without it, real files stay safe. |
| Bare `skills link` | Projects all registered tools. |
| Tool default | Full `ToolNames()` registry (`claude`, `agents`). |

## Notable bug caught during implementation

The acceptance test for repair healing exposed a real ordering bug: `RunFlow` early-returned at `!plan.HasChanges()` *before* the projection step, so a campaign with only a broken skill link (no missing scaffold files) would never heal. Fixed by detecting pending skill projections before the early-return.

## Testing

- **Unit** (`internal/skills/link_test.go`): fresh projection, idempotency, broken-managed-link healing, real-file-conflict-not-overwritten, dry-run-touches-nothing, empty-skills-dir.
- **Integration** (`tests/integration/init_skills_test.go`): init auto-links both tools, `--no-skills` skips projection, `--repair` heals a deleted link and is idempotent, bare `skills link` projects all tools.
- Existing skills lifecycle tests updated: `setupSkillsCampaign` resets `.claude/skills`/`.agents/skills` post-init (with `sync` for Colima overlayfs) so they still start from an unlinked slate.
- `golangci-lint`: 0 issues. host-fs lint: clean. Full Init+Skills integration suite green across 4 non-cached runs.

### DIFF
--- a/cmd/camp/init/command.go
+++ b/cmd/camp/init/command.go
@@ -271,13 +271,16 @@ func RunFlow(ctx context.Context, p Params, w Writers, isInteractive bool) error
 	// before the repair auto-commit so healed links are committed in the same pass.
 	var skillsLinked []string
 	var skillsWarnings []string
+	var skillsProjectedPaths []string
 	if !p.NoSkills && !p.DryRun {
-		skillsLinked, skillsWarnings = projectCampaignSkills(result.CampaignRoot)
+		skillsLinked, skillsWarnings, skillsProjectedPaths = projectCampaignSkills(result.CampaignRoot)
 	}
 
-	// Auto-commit after repair (scaffold creates + migrations + skills).
+	// Auto-commit after repair (scaffold creates + migrations + skills). The
+	// repair commit stages a selective file list, so newly projected skill links
+	// must be passed in explicitly or they would be left untracked.
 	if p.Repair && !p.DryRun {
-		commitRepairChanges(ctx, result, opts.RepairPlan, migrationCount, w)
+		commitRepairChanges(ctx, result, opts.RepairPlan, migrationCount, skillsProjectedPaths, w)
 	}
 
 	// Initialize Festival Methodology (unless dry-run).
@@ -374,16 +377,18 @@ func skillsNeedProjection(root string) bool {
 // projectCampaignSkills links campaign skill bundles into every registered tool
 // directory under root. It is best-effort: failures are returned as human-readable
 // warnings rather than aborting the init flow. It returns the tools that were
-// fully linked and any warnings to surface.
-func projectCampaignSkills(root string) (linked []string, warnings []string) {
+// fully linked, any warnings to surface, and the campaign-root-relative paths of
+// the skill links that were created or replaced (so the caller can stage exactly
+// those paths in a selective repair commit).
+func projectCampaignSkills(root string) (linked []string, warnings []string, projectedPaths []string) {
 	skillsDir := filepath.Join(root, campaign.CampaignDir, intskills.SkillsSubdir)
 	if _, err := os.Stat(skillsDir); err != nil {
-		return nil, nil
+		return nil, nil, nil
 	}
 
 	results, err := intskills.LinkDefaultTools(root, skillsDir, false, false, os.Stderr)
 	if err != nil {
-		return nil, []string{fmt.Sprintf("skills not linked: %v", err)}
+		return nil, []string{fmt.Sprintf("skills not linked: %v", err)}, nil
 	}
 
 	for _, res := range results {
@@ -395,7 +400,16 @@ func projectCampaignSkills(root string) (linked []string, warnings []string) {
 			warnings = append(warnings, fmt.Sprintf("%s: %d conflicting path(s) left untouched (%s); run 'camp skills status'",
 				res.Tool, res.Summary.Conflicts, strings.Join(res.Summary.ConflictNames, ", ")))
 		}
+		relPath, err := intskills.ResolveToolPath(res.Tool)
+		if err == nil {
+			for _, slug := range res.Summary.CreatedNames {
+				projectedPaths = append(projectedPaths, filepath.Join(relPath, slug))
+			}
+			for _, slug := range res.Summary.ReplacedNames {
+				projectedPaths = append(projectedPaths, filepath.Join(relPath, slug))
+			}
+		}
 		linked = append(linked, res.Tool)
 	}
-	return linked, warnings
+	return linked, warnings, projectedPaths
 }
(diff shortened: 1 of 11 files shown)


cmd/camp/init/command.go
```diff
@@ -254,7 +267,15 @@ func RunFlow(ctx context.Context, p Params, w Writers, isInteractive bool) error
 		migrationCount = moved
 	}
 
-	// Auto-commit after repair (scaffold creates + migrations).
+	// Project campaign skills into tool directories (.claude/skills, .agents/skills)
+	// before the repair auto-commit so healed links are committed in the same pass.
+	var skillsLinked []string
+	var skillsWarnings []string
+	if !p.NoSkills && !p.DryRun {
+		skillsLinked, skillsWarnings = projectCampaignSkills(result.CampaignRoot)
```
obey-agent: This links the tool skill directories before the repair auto-commit, but the commit path never receives those new paths. `commitRepairChanges` builds a selective file list from `initResult` and the repair plan, then calls `commit.Repair` with `SelectiveOnly: true`; `projectCampaignSkills` does not add `.claude/skills/...` or `.agents/skills/...` to that scope. I reproduced an older campaign with no projected tool skills: `camp init --repair --yes` printed "Committed changes to git" but left `.agents/` and `.claude/` untracked. Please have the projection step return the created/replaced paths and include them in the repair commit scope, and add a regression test that verifies repair commits newly projected skill links for an existing campaign.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** This is a well-scoped improvement to the init first-run path and it reuses the existing skill projection primitives instead of adding a parallel implementation. The core init/link behavior is covered, but the repair auto-commit path still misses the newly projected tool skill links.

**Key Findings:**
- Inline comment on `cmd/camp/init/command.go`: `camp init --repair --yes` can create `.agents/` and `.claude/` skill projections while the repair commit selectively stages only scaffold/migration paths. I reproduced this with an older campaign shape; the command printed a successful commit message while leaving both tool directories untracked.

**What's Done Well:** The `LinkDefaultTools` helper keeps per-tool failures isolated, the projection code preserves user files, and the tests cover fresh projection, idempotency, broken-link healing, conflicts, dry-run behavior, and bare `camp skills link`.

**Validation:** `go test ./internal/skills ./cmd/camp/init` and `go test -tags integration ./tests/integration -run 'TestInit_AutoLinksSkills|TestInit_NoSkillsFlag|TestInit_RepairHealsBrokenSkillLink|TestSkills_LinkNoFlagsProjectsAllTools'` both pass locally. GitHub reports no checks for the branch.

**Staff Standard:** Not yet. I would be comfortable after the repair commit scope includes the projected skill paths and a regression test proves older campaigns get those links committed.

lancekrogers:
Thanks for the catch — confirmed and fixed in 7afde4c.

You were right: `commitRepairChanges` builds a selective file list (`SelectiveOnly: true`) from scaffold/migration paths only, so the links `projectCampaignSkills` created during `--repair` were never staged. Fresh `init` was fine (initial commit adds everything); only the selective repair commit missed them.

**Fix:**
- `ProjectionSummary` now reports `CreatedNames`/`ReplacedNames` per bundle.
- `projectCampaignSkills` returns the campaign-root-relative projected paths.
- `commitRepairChanges` / `buildRepairCommitFiles` / `buildRepairCommitMessage` stage and document those paths; `hasChanges` now also triggers for a skills-only repair (no scaffold changes).

**Regression test (as requested):** `TestInit_RepairCommitsProjectedSkillLinks` reproduces an older campaign — removes `.claude/skills`/`.agents/skills` and commits that removal so the tree starts clean — then runs `camp init --repair --yes` and asserts each projected link is (a) restored and (b) git-tracked (`git ls-files --error-unmatch`), with a clean `git status --porcelain` afterward. Added two unit tests for the commit file/message builders too.

**Validation:** `go test ./internal/skills ./cmd/camp/init ./cmd/camp` and the full `-tags integration` Init+Skills suite pass; `golangci-lint` 0 issues, host-fs lint clean.

obey-agent (commented):
**Verdict: Approve**

**Overview:** Re-reviewing the latest head, the previous repair auto-commit issue is addressed. The PR now carries the created/replaced skill projection paths out of the projection step and passes them into the selective repair commit, which closes the untracked `.claude/` / `.agents/` regression I reproduced earlier.

**Key Findings:**
- No new blocking findings.
- Previous finding addressed: projected skill links are included in `buildRepairCommitFiles`, the repair commit message records them, and `TestInit_RepairCommitsProjectedSkillLinks` verifies older campaigns get the restored links tracked with a clean working tree afterward.

**What's Done Well:** The fix stays integrated with the existing projection summary instead of adding a separate filesystem crawl. The new regression test targets the exact failure mode from the prior review, and the projection path tracking is limited to created/replaced links, which keeps the selective commit scope tight.

**Validation:** `go test ./internal/skills ./cmd/camp/init` passed. `go test -tags integration ./tests/integration -run 'TestInit_AutoLinksSkills|TestInit_NoSkillsFlag|TestInit_RepairHealsBrokenSkillLink|TestInit_RepairCommitsProjectedSkillLinks|TestSkills_LinkNoFlagsProjectsAllTools'` passed locally. GitHub reports no checks for this branch.

**Staff Standard:** Yes. I would be comfortable submitting this after the current revisions.

PR #318, opened 2026-06-05
feat(workitem): add 'camp workitem priority' CLI command

lancekrogers:
## Summary

Adds `camp workitem priority <selector> <high|medium|low|clear>` so manual priority can be set non-interactively.

Until now, manual priority (high/medium/low) was assignable **only** from the interactive workitem TUI (press `P`). The TUI writes `.campaign/settings/workitems.json` through the `internal/workitem/priority` package. That meant any non-interactive caller — agents, scripts, and the festival-app workitems viewer — had no way to set priority without re-implementing the on-disk store format, which would be a parallel implementation of camp's own logic.

This command is a thin wrapper over the same `priority` package the TUI already uses, so the CLI and the dashboard stay byte-for-byte compatible on the store.

## Behavior

```
camp workitem priority <selector> <high|medium|low|clear> [--json]
```

- **selector** accepts the same forms as `camp workitem current`: a stable `.workitem` id, the workitem key (`<type>:<path>`), a relative path, or a directory slug (resolved via `selector.Resolve`).
- **level** is `high`, `medium`, `low`, or `clear` (`none` is accepted as an alias for `clear`; `med` for `medium`; case-insensitive).
- Writes through `priority.Set` / `priority.Clear` + `SaveOrDelete`, so clearing removes the entry and deletes the store file when it becomes empty — identical to the TUI's semantics.
- `--json` emits a structured result (schema `workitem-priority/v1alpha1`) for automation; the command carries the `agent_allowed` annotation.
- Validates the level **before** touching the store; unknown levels and unresolvable selectors error out without writing anything.

## Verification

End-to-end against a scratch campaign:

```
$ camp workitem priority design:workflow/design/example high --json
{
  "schema_version": "workitem-priority/v1alpha1",
  "key": "design:workflow/design/example",
  "priority": "high",
  "cleared": false
}
$ camp workitem --json | jq '.items[] | [.key, .manual_priority]'
["design:workflow/design/example", "high"]      # set priority surfaces in list output
$ camp workitem priority example clear
cleared priority: design:workflow/design/example
# .campaign/settings/workitems.json is deleted once empty
$ camp workitem priority example bogus
Error: validation error on priority: unknown priority "bogus" (valid: high, medium, low, clear)   # exit 1
```

Gates:
- `go build ./...`: clean
- `just lint`: 0 issues (`lint-no-host-fs-tests`: no new violators)
- `go test ./internal/workitem/... ./internal/commands/workitem/...`: all pass

Tests added (`priority_test.go`): set-by-id/key/slug, level aliases + case-insensitivity, clear + empty-store deletion, `--json` set and cleared shapes, invalid level (no write), unknown selector, and context cancellation (no store mutation on cancel).

## Follow-up

This unblocks wiring a priority picker into the festival-app workitems viewer (it will shell out to this command rather than writing the store directly). That UI change ships separately.

### DIFF
--- a/internal/commands/workitem/json_contract.go
+++ b/internal/commands/workitem/json_contract.go
@@ -5,6 +5,8 @@ const (
 	WorkitemCreateJSONVersion = "workitem-create/v1alpha1"
 	// WorkitemResolveJSONVersion is the schema version of camp workitem resolve --json.
 	WorkitemResolveJSONVersion = "workitem-resolve/v1alpha1"
+	// WorkitemPriorityJSONVersion is the schema version of camp workitem priority --json.
+	WorkitemPriorityJSONVersion = "workitem-priority/v1alpha1"
 	// WorkitemCommitJSONVersion is the schema version of camp workitem commit --json.
 	WorkitemCommitJSONVersion = "workitem-commit/v1alpha1"
 	// WorkitemCommitsJSONVersion is the schema version of camp workitem commits --json.
--- a/internal/commands/workitem/priority.go
+++ b/internal/commands/workitem/priority.go
@@ -1,0 +1,127 @@
+package workitem
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/jsoncontract"
+	"github.com/Obedience-Corp/camp/internal/workitem/priority"
+	"github.com/Obedience-Corp/camp/internal/workitem/selector"
+)
+
+func newPriorityCommand() *cobra.Command {
+	var jsonOut bool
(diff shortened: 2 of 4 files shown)


obey-agent (commented):
**Verdict: Approve**

**Overview:** This is a focused CLI wrapper around the existing `internal/workitem/priority` store, and it avoids creating a parallel on-disk implementation. The command validates the requested level before resolving or loading the store, uses the same selector machinery as adjacent workitem commands, and preserves the TUI's `Set`/`Clear` + `SaveOrDelete` semantics.

**Key Findings:** No blocking findings. I did not leave inline comments because I did not find a correctness, maintainability, or production-readiness issue worth calling out at line level.

**What's Done Well:** The tests cover selector forms, aliases/case handling, clear semantics including empty-store deletion, JSON output, invalid levels without writes, unknown selectors, and cancellation before mutation. The JSON contract wrapping and `agent_allowed` annotation match nearby workitem command patterns.

**Verification:**
- `go test ./internal/workitem/... ./internal/commands/workitem/...`
- `go build ./...`
- `just lint`

**Staff Standard:** Yes, I would be comfortable submitting this as my own work.

PR #319, opened 2026-06-10
feat(intent): notes category, tags, rename, and workflows as sub-concepts

lancekrogers:
Implements the **intent capture UX** feature across three threads. All work is on `projects/camp`; the festival and design package live at the campaign root and are out of scope for this PR.

## Thread 1 — Notes category + tags
- A note is a separate **category** (not a `Type`), stored in `.campaign/intents/notes/` outside the inbox/ready/active lifecycle (flat + `archived`). Notes are excluded from intent `list`/`gather`/index.
- Fast capture: `camp intent note "..."` and a `cnote` shell shortcut; a TUI note quick-add (text only, no type/concept).
- Explorer **notes view** (`N`) and **note→intent convert** via `camp intent convert` and a TUI action (`c`).
- Frontmatter **tags** via a reusable shortcut overlay (`ctrl+t` in capture, `T` in the explorer), `--tag` flags on `note`/`add`/`edit`, and a configurable `intents.tags` list seeded at init. A bare note needs zero tag interaction.

## Thread 2 — Rename intent
- `id:` frontmatter is made **authoritative**: perf-safe resolution (fast path `<id>.md` + lazy `id→path` index, no per-lookup directory scans), so a renamed filename slug never breaks lookups or cross-references.
- Single shared `SlugFromTitle` (merge's duplicate removed).
- `IntentService.Rename` regenerates `<new-slug>-<stable-suffix>.md`, preserves the id, handles filename collisions, and adds `audit.EventRename` / `commit.IntentRename`.
- Exposed via `camp intent rename <id> "<title>"` and the explorer (`R`, prefilled overlay, list refresh + reselect). One service method backs both.

## Thread 3 — Workflows as sub-concepts
- Recursive `ConceptEntry.Children`; `DefaultConcepts()` emits the nested `projects / workflow{festivals,design,explore,code_reviews,pipelines} / docs` shape; `camp init` seeds it.
- Concept service resolves a parent's children (config children merged with on-disk subdirs under the parent path); the `intent` workflow is filtered from the picker.
- The existing two-step picker renders the parent → sub-concept cascade (no new step needed).
- `camp workflow create` registers the collection as a **child** of `workflow`; `doctor`/`sync`/`list` descend into the nested tree.
- `camp init --repair` migrates a flat `concepts:` block into the nested shape, idempotently and non-destructively (custom/unknown concepts preserved).
- `camp concepts` renders the tree with a customization hint; upgrade guide added under `docs/migrations/`.

## Testing
- New unit/TUI tests for every thread (notes routing/exclusion/convert, id resolution, rename + collision, tag overlay + persistence, nested config round-trip, child resolution, picker cascade, workflow-create nesting, and migration: flat→nested, idempotency, custom survival).
- `just lint` is clean; the host-fs-tests guard reports no new violators.
- The only full-suite failures are the **pre-existing** `cmd/camp` manifest tests for `workitem priority` (#318) — they fail identically on the base of this branch and are unrelated to this work.

### DIFF
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -213,7 +213,7 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 			Concept: saved.Concept,
 			Body:    saved.Body,
 			Author:  saved.Author,
-			Tags:    saved.Tags,
+			Tags:    mergeTags(tagsFlag, saved.Tags),
 		}
 		if createNote {
 			if err := runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, savedOpts); err != nil {
@@ -243,7 +243,7 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 		Concept: result.Concept,
 		Body:    result.Body,
 		Author:  result.Author,
-		Tags:    result.Tags,
+		Tags:    mergeTags(tagsFlag, result.Tags),
 	}
 
 	if createNote {
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -85,6 +85,7 @@ func init() {
 	flags.String("body-file", "", "Read intent body from file (- for stdin, 10 MiB cap)")
 	flags.String("concept", "", "Set the concept field (e.g., projects/camp)")
 	flags.String("author", "", "Override the default author attribution")
+	flags.StringArray("tag", nil, "Add a tag (repeatable)")
 }
 
 func runIntentAdd(cmd *cobra.Command, args []string) error {
@@ -99,6 +100,7 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 	targetCampaign, args = normalizeIntentAddCampaignArgs(args, targetCampaign)
 	conceptFlag, _ := cmd.Flags().GetString("concept")
 	authorFlag, _ := cmd.Flags().GetString("author")
+	tagsFlag, _ := cmd.Flags().GetStringArray("tag")
 
 	// Resolve body from --body / --body-file (mutual exclusivity checked inside)
 	body, bodySet, err := resolveBody(cmd)
@@ -143,6 +145,7 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 			Author:  author,
 			Body:    body,
 			Concept: conceptFlag,
+			Tags:    tagsFlag,
 		}
 
 		// Deep capture overrides ultra-fast; body flags pre-fill the template
@@ -176,11 +179,12 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 
 	// Run BubbleTea TUI
 	model, err := runIntentAddTUI(ctx, conceptSvc, tui.AddOptions{
-		DefaultType:  intentType,
-		FullMode:     fullMode,
-		Author:       author,
-		CampaignRoot: campaignRoot,
-		Shortcuts:    shortcuts,
+		DefaultType:   intentType,
+		FullMode:      fullMode,
+		Author:        author,
+		CampaignRoot:  campaignRoot,
+		Shortcuts:     shortcuts,
+		AvailableTags: cfg.IntentTags(),
 	})
 	if err != nil {
 		return err
@@ -194,6 +198,7 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 			Concept: saved.Concept,
 			Body:    saved.Body,
 			Author:  saved.Author,
+			Tags:    saved.Tags,
 		}
 		if err := runFastCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, savedOpts); err != nil {
 			return err
@@ -217,6 +222,7 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 		Concept: result.Concept,
 		Body:    result.Body,
 		Author:  result.Author,
+		Tags:    result.Tags,
 	}
 
 	// Deep capture if requested
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -48,6 +48,7 @@ PROGRAMMATIC (agent) FLAGS:
   --body              Set intent body from a literal string
   --body-file         Read intent body from a file (- for stdin)
   --concept           Set the concept field (e.g., "projects/camp")
+  --note              Create a note instead of a lifecycle intent
   --author            Override the default author attribution
 
   --body and --body-file are mutually exclusive.
@@ -60,6 +61,8 @@ Examples:
   camp intent add                        Fast TUI (3-step form)
   camp intent add --campaign             Pick a target campaign interactively
   camp intent add --full                 Full TUI (includes body)
+  camp intent add --note                 Note TUI (title + body, no type/concept)
+  camp intent add --note "Meeting note" --body "Follow up next week"
   camp intent add -e "Complex feature"   Deep capture with editor
   camp intent add -t feature "New API"   Set type explicitly
   camp intent add "Fix login" --body "The login page returns 500"
@@ -84,6 +87,7 @@ func init() {
 	flags.String("body", "", "Set intent body as a literal string")
 	flags.String("body-file", "", "Read intent body from file (- for stdin, 10 MiB cap)")
 	flags.String("concept", "", "Set the concept field (e.g., projects/camp)")
+	flags.Bool("note", false, "Create a note instead of a lifecycle intent")
 	flags.String("author", "", "Override the default author attribution")
 	flags.StringArray("tag", nil, "Add a tag (repeatable)")
 }
@@ -97,10 +101,18 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 	fullMode, _ := cmd.Flags().GetBool("full")
 	targetCampaign, _ := cmd.Flags().GetString("campaign")
 	noCommit, _ := cmd.Flags().GetBool("no-commit")
+	createNote, _ := cmd.Flags().GetBool("note")
 	targetCampaign, args = normalizeIntentAddCampaignArgs(args, targetCampaign)
 	conceptFlag, _ := cmd.Flags().GetString("concept")
 	authorFlag, _ := cmd.Flags().GetString("author")
 	tagsFlag, _ := cmd.Flags().GetStringArray("tag")
+
+	if createNote && cmd.Flags().Changed("concept") {
+		return camperrors.Wrap(camperrors.ErrInvalidInput, "--note cannot be combined with --concept")
+	}
+	if createNote && cmd.Flags().Changed("type") {
+		return camperrors.Wrap(camperrors.ErrInvalidInput, "--note cannot be combined with --type")
+	}
 
 	// Resolve body from --body / --body-file (mutual exclusivity checked inside)
 	body, bodySet, err := resolveBody(cmd)
@@ -148,6 +160,13 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 			Tags:    tagsFlag,
 		}
 
+		if createNote {
+			if useEditor {
+				return runDeepNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
+			}
+			return runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
+		}
+
 		// Deep capture overrides ultra-fast; body flags pre-fill the template
 		if useEditor {
 			return runDeepCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
@@ -170,17 +189,13 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 	}
 
 	// Build navigation shortcuts map (key -> path) for @ completion
-	shortcuts := make(map[string]string)
-	for key, sc := range cfg.Shortcuts() {
-		if sc.HasPath() {
-			shortcuts[key] = sc.Path
-		}
-	}
+	shortcuts := navigationShortcuts(cfg)
 
 	// Run BubbleTea TUI
 	model, err := runIntentAddTUI(ctx, conceptSvc, tui.AddOptions{
 		DefaultType:   intentType,
 		FullMode:      fullMode,
+		NoteMode:      createNote,
 		Author:        author,
 		CampaignRoot:  campaignRoot,
 		Shortcuts:     shortcuts,
@@ -199,6 +214,12 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 			Body:    saved.Body,
 			Author:  saved.Author,
 			Tags:    saved.Tags,
+		}
+		if createNote {
+			if err := runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, savedOpts); err != nil {
+				return err
+			}
+			continue
 		}
 		if err := runFastCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, savedOpts); err != nil {
 			return err
@@ -225,12 +246,26 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 		Tags:    result.Tags,
 	}
 
+	if createNote {
+		return runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
+	}
+
 	// Deep capture if requested
 	if useEditor {
 		return runDeepCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
 	}
 
 	return runFastCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
+}
+
+func navigationShortcuts(cfg *config.CampaignConfig) map[string]string {
+	shortcuts := make(map[string]string)
+	for key, sc := range cfg.Shortcuts() {
+		if sc.HasPath() {
+			shortcuts[key] = sc.Path
+		}
+	}
+	return shortcuts
 }
 
 func validateIntentAddArgs(cmd *cobra.Command, args []string) error {
@@ -372,6 +407,26 @@ func runDeepCapture(ctx context.Context, svc *intent.IntentService, intentsDir s
 	}
 
 	return finalizeCreatedIntent(ctx, result, intentsDir, cfg, campaignRoot, noCommit)
+}
+
+// runDeepNoteCapture opens the note template in $EDITOR and saves it to notes/.
+func runDeepNoteCapture(ctx context.Context, svc *intent.IntentService, intentsDir string, cfg *config.CampaignConfig, campaignRoot string, noCommit bool, opts intent.CreateOptions) error {
+	editorFn := func(ctx context.Context, path string) error {
+		return editor.Edit(ctx, path)
+	}
+	opts.Category = intent.CategoryNote
+	opts.Type = ""
+	opts.Concept = ""
+
+	result, err := svc.CreateWithEditor(ctx, opts, editorFn)
+	if err != nil {
+		if errors.Is(err, camperrors.ErrCancelled) {
+			return intent.ErrCancelled
+		}
+		return camperrors.Wrap(err, "failed to create note")
+	}
+
+	return finalizeCreatedNote(ctx, result, intentsDir, cfg, campaignRoot, noCommit)
 }
 
 func finalizeCreatedIntent(ctx context.Context, result *intent.Intent, intentsDir string, cfg *config.CampaignConfig, campaignRoot string, noCommit bool) error {
--- a/cmd/camp/intent/note.go
+++ b/cmd/camp/intent/note.go
@@ -1,0 +1,182 @@
+package intent
+
+import (
+	"context"
+	"fmt"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/concept"
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/git"
+	"github.com/Obedience-Corp/camp/internal/git/commit"
+	"github.com/Obedience-Corp/camp/internal/intent"
+	"github.com/Obedience-Corp/camp/internal/intent/audit"
+	"github.com/Obedience-Corp/camp/internal/intent/tui"
+	navtui "github.com/Obedience-Corp/camp/internal/nav/tui"
+	"github.com/Obedience-Corp/camp/internal/paths"
+)
+
+var intentNoteCmd = &cobra.Command{
+	Use:   "note [text]",
+	Short: "Capture a quick note",
+	Long: `Capture a freeform note. Notes are a separate category from intents: they
+are stored in .campaign/intents/notes/ and do not flow through the
+inbox → ready → active lifecycle. A note carries no type or concept; tags
+organize them.
+
+Fast capture skips the type wheel and concept picker entirely.
+
+Examples:
+  camp intent note "check the daemon socket path"   Capture a note immediately
+  camp intent note "follow up" --body "details..."  Note with a longer body
+  echo "body" | camp intent note "idea" --body-file -
+  camp intent note                                  Quick-add note (TUI)`,
+	Args: cobra.MaximumNArgs(1),
+	RunE: runIntentNote,
+}
+
+func init() {
+	Cmd.AddCommand(intentNoteCmd)
+
+	flags := intentNoteCmd.Flags()
+	flags.Bool("no-commit", false, "Don't create a git commit")
+	flags.String("body", "", "Set note body as a literal string")
+	flags.String("body-file", "", "Read note body from file (- for stdin, 10 MiB cap)")
+	flags.String("author", "", "Override the default author attribution")
+}
+
+func runIntentNote(cmd *cobra.Command, args []string) error {
+	ctx := cmd.Context()
+
+	noCommit, _ := cmd.Flags().GetBool("no-commit")
+	authorFlag, _ := cmd.Flags().GetString("author")
+
+	body, _, err := resolveBody(cmd)
+	if err != nil {
+		return err
+	}
+
+	cfg, campaignRoot, err := config.LoadCampaignConfigFromCwd(ctx)
+	if err != nil {
+		return err
+	}
+
+	resolver := paths.NewResolverFromConfig(campaignRoot, cfg)
+	svc := intent.NewIntentService(campaignRoot, resolver.Intents())
+	if err := svc.EnsureDirectories(ctx); err != nil {
+		return camperrors.Wrap(err, "ensuring intent directories")
+	}
+
+	// Fast path: note text provided as an argument
+	if len(args) > 0 {
+		author := "agent"
+		if cmd.Flags().Changed("author") {
+			author = authorFlag
+		}
+		opts := intent.CreateOptions{Title: args[0], Author: author, Body: body}
+		return runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
+	}
+
+	// No argument: non-TTY requires note text (can't launch TUI)
+	if !navtui.IsTerminal() {
+		return camperrors.Wrap(camperrors.ErrInvalidInput, "note text required in non-interactive mode\n       Usage: camp intent note <text> [flags]")
+	}
+
+	author := git.GetUserName(ctx)
+	if cmd.Flags().Changed("author") {
+		author = authorFlag
+	}
+
+	conceptSvc := concept.NewService(campaignRoot, cfg.Concepts())
+	model, err := runIntentNoteTUI(ctx, conceptSvc, tui.AddOptions{
+		NoteMode:     true,
+		Author:       author,
+		CampaignRoot: campaignRoot,
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, saved := range model.SavedResults() {
+		if err := runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, intent.CreateOptions{
+			Title:  saved.Title,
+			Author: saved.Author,
+		}); err != nil {
+			return err
+		}
+	}
+
+	result := model.Result()
+	if result == nil {
+		if len(model.SavedResults()) > 0 {
+			return nil
+		}
+		return intent.ErrCancelled
+	}
+
+	return runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, intent.CreateOptions{
+		Title:  result.Title,
+		Author: result.Author,
+	})
+}
+
+func runIntentNoteTUI(ctx context.Context, conceptSvc concept.Service, opts tui.AddOptions) (*tui.IntentAddModel, error) {
+	model := tui.NewIntentAddModel(ctx, conceptSvc, opts)
+
+	p := tea.NewProgram(model, tea.WithAltScreen())
+	finalModel, err := p.Run()
+	if err != nil {
+		return nil, camperrors.Wrap(err, "TUI error")
+	}
+
+	m, ok := finalModel.(tui.IntentAddModel)
+	if !ok {
+		return nil, camperrors.Wrapf(camperrors.ErrInvalidInput, "unexpected model type: %T", finalModel)
+	}
+
+	return &m, nil
+}
+
+func runNoteCapture(ctx context.Context, svc *intent.IntentService, intentsDir string, cfg *config.CampaignConfig, campaignRoot string, noCommit bool, opts intent.CreateOptions) error {
+	result, err := svc.CreateNote(ctx, opts)
+	if err != nil {
+		return camperrors.Wrap(err, "failed to create note")
+	}
+
+	return finalizeCreatedNote(ctx, result, intentsDir, cfg, campaignRoot, noCommit)
+}
+
+func finalizeCreatedNote(ctx context.Context, result *intent.Intent, intentsDir string, cfg *config.CampaignConfig, campaignRoot string, noCommit bool) error {
+	if err := appendIntentAuditEvent(ctx, intentsDir, audit.Event{
+		Type:  audit.EventCreate,
+		ID:    result.ID,
+		Title: result.Title,
+		To:    string(result.Status),
+	}); err != nil {
+		return err
+	}
+
+	fmt.Printf("✓ Note created: %s\n", result.Path)
+
+	if !noCommit {
+		files := commit.NormalizeFiles(campaignRoot, result.Path, audit.FilePath(intentsDir))
+		commitResult := commit.Intent(ctx, commit.IntentOptions{
+			Options: commit.Options{
+				CampaignRoot:  campaignRoot,
+				CampaignID:    cfg.ID,
+				Files:         files,
+				SelectiveOnly: true,
+			},
+			Action:      commit.IntentCreate,
+			IntentTitle: result.Title,
+		})
+		if commitResult.Message != "" {
+			fmt.Printf("  %s\n", commitResult.Message)
+		}
+	}
+
+	return nil
+}
--- a/cmd/camp/intent/note.go
+++ b/cmd/camp/intent/note.go
@@ -27,13 +27,14 @@ are stored in .campaign/intents/notes/ and do not flow through the
 inbox → ready → active lifecycle. A note carries no type or concept; tags
 organize them.
 
-Fast capture skips the type wheel and concept picker entirely.
+Fast capture skips the TUI. Interactive capture uses the same title/body/tag
+flow as intent add, but skips the type wheel and concept picker.
 
 Examples:
   camp intent note "check the daemon socket path"   Capture a note immediately
   camp intent note "follow up" --body "details..."  Note with a longer body
   echo "body" | camp intent note "idea" --body-file -
-  camp intent note                                  Quick-add note (TUI)`,
+  camp intent note                                  Note TUI (title + body)`,
 	Args: cobra.MaximumNArgs(1),
 	RunE: runIntentNote,
 }
@@ -97,6 +98,7 @@ func runIntentNote(cmd *cobra.Command, args []string) error {
 		NoteMode:      true,
 		Author:        author,
 		CampaignRoot:  campaignRoot,
+		Shortcuts:     navigationShortcuts(cfg),
 		AvailableTags: cfg.IntentTags(),
 	})
 	if err != nil {
@@ -107,6 +109,7 @@ func runIntentNote(cmd *cobra.Command, args []string) error {
 		if err := runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, intent.CreateOptions{
 			Title:  saved.Title,
 			Author: saved.Author,
+			Body:   saved.Body,
 			Tags:   mergeTags(tags, saved.Tags),
 		}); err != nil {
 			return err
@@ -124,6 +127,7 @@ func runIntentNote(cmd *cobra.Command, args []string) error {
 	return runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, intent.CreateOptions{
 		Title:  result.Title,
 		Author: result.Author,
+		Body:   result.Body,
 		Tags:   mergeTags(tags, result.Tags),
 	})
 }
--- a/cmd/camp/intent/note.go
+++ b/cmd/camp/intent/note.go
@@ -46,6 +46,7 @@ func init() {
 	flags.String("body", "", "Set note body as a literal string")
 	flags.String("body-file", "", "Read note body from file (- for stdin, 10 MiB cap)")
 	flags.String("author", "", "Override the default author attribution")
+	flags.StringArrayP("tag", "t", nil, "Add a tag (repeatable)")
 }
 
 func runIntentNote(cmd *cobra.Command, args []string) error {
@@ -53,6 +54,7 @@ func runIntentNote(cmd *cobra.Command, args []string) error {
 
 	noCommit, _ := cmd.Flags().GetBool("no-commit")
 	authorFlag, _ := cmd.Flags().GetString("author")
+	tags, _ := cmd.Flags().GetStringArray("tag")
 
 	body, _, err := resolveBody(cmd)
 	if err != nil {
@@ -76,7 +78,7 @@ func runIntentNote(cmd *cobra.Command, args []string) error {
 		if cmd.Flags().Changed("author") {
 			author = authorFlag
 		}
-		opts := intent.CreateOptions{Title: args[0], Author: author, Body: body}
+		opts := intent.CreateOptions{Title: args[0], Author: author, Body: body, Tags: tags}
 		return runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, opts)
 	}
 
@@ -92,9 +94,10 @@ func runIntentNote(cmd *cobra.Command, args []string) error {
 
 	conceptSvc := concept.NewService(campaignRoot, cfg.Concepts())
 	model, err := runIntentNoteTUI(ctx, conceptSvc, tui.AddOptions{
-		NoteMode:     true,
-		Author:       author,
-		CampaignRoot: campaignRoot,
+		NoteMode:      true,
+		Author:        author,
+		CampaignRoot:  campaignRoot,
+		AvailableTags: cfg.IntentTags(),
 	})
 	if err != nil {
 		return err
@@ -104,6 +107,7 @@ func runIntentNote(cmd *cobra.Command, args []string) error {
 		if err := runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, intent.CreateOptions{
 			Title:  saved.Title,
 			Author: saved.Author,
+			Tags:   mergeTags(tags, saved.Tags),
 		}); err != nil {
 			return err
 		}
@@ -120,6 +124,7 @@ func runIntentNote(cmd *cobra.Command, args []string) error {
 	return runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, intent.CreateOptions{
 		Title:  result.Title,
 		Author: result.Author,
+		Tags:   mergeTags(tags, result.Tags),
 	})
 }
 
@@ -138,6 +143,23 @@ func runIntentNoteTUI(ctx context.Context, conceptSvc concept.Service, opts tui.
 	}
 
 	return &m, nil
+}
+
+// mergeTags unions flag tags and overlay tags, preserving order and dropping
+// duplicates and empties.
+func mergeTags(a, b []string) []string {
+	seen := make(map[string]bool)
+	var out []string
+	for _, list := range [][]string{a, b} {
+		for _, t := range list {
+			if t == "" || seen[t] {
+				continue
+			}
+			seen[t] = true
+			out = append(out, t)
+		}
+	}
+	return out
 }
 
 func runNoteCapture(ctx context.Context, svc *intent.IntentService, intentsDir string, cfg *config.CampaignConfig, campaignRoot string, noCommit bool, opts intent.CreateOptions) error {
--- a/internal/commands/workflow/create_plan.go
+++ b/internal/commands/workflow/create_plan.go
@@ -152,6 +152,18 @@ func planConcept(cfg *config.CampaignConfig, plan *createPlan, opts createOption
 			return nil
 		}
 	}
+
+	// A legacy flat top-level workflow concept counts as an existing entry: apply
+	// folds it under the parent rather than adding a duplicate child.
+	if idx := flatWorkflowConceptIndex(concepts, opts.Type); idx != -1 {
+		existing := concepts[idx]
+		if existing.Path != plan.WorkflowRel && !opts.Replace {
+			return camperrors.NewValidation("type",
+				"concept "+opts.Type+" already points to "+existing.Path+"; use --replace to update it", nil)
+		}
+		plan.Concept.Replaced = true
+		plan.Concept.Existing = existing.Path
+	}
 	return nil
 }
 
--- a/internal/commands/workflow/create_plan.go
+++ b/internal/commands/workflow/create_plan.go
@@ -123,27 +123,34 @@ func planShortcut(cfg *config.CampaignConfig, plan *createPlan, opts createOptio
 func planConcept(cfg *config.CampaignConfig, plan *createPlan, opts createOptions) error {
 	plan.Concept.Name = opts.Type
 	plan.Concept.Path = plan.WorkflowRel
+	plan.Concept.Parent = workflowParentName
 
 	concepts := cfg.ConceptList
 	if len(concepts) == 0 {
 		concepts = cfg.Concepts()
 	}
 
-	for _, concept := range concepts {
-		if !strings.EqualFold(concept.Name, opts.Type) {
+	// New collections nest under the workflow parent; look for an existing child.
+	for _, parent := range concepts {
+		if !strings.EqualFold(parent.Name, workflowParentName) {
 			continue
 		}
-		if concept.Path == plan.WorkflowRel {
-			plan.Concept.NoChange = true
+		for _, child := range parent.Children {
+			if !strings.EqualFold(child.Name, opts.Type) {
+				continue
+			}
+			if child.Path == plan.WorkflowRel {
+				plan.Concept.NoChange = true
+				return nil
+			}
+			if !opts.Replace {
+				return camperrors.NewValidation("type",
+					"concept "+opts.Type+" already points to "+child.Path+"; use --replace to update it", nil)
+			}
+			plan.Concept.Replaced = true
+			plan.Concept.Existing = child.Path
 			return nil
 		}
-		if !opts.Replace {
-			return camperrors.NewValidation("type",
-				"concept "+opts.Type+" already points to "+concept.Path+"; use --replace to update it", nil)
-		}
-		plan.Concept.Replaced = true
-		plan.Concept.Existing = concept.Path
-		return nil
 	}
 	return nil
 }
--- a/internal/concept/children.go
+++ b/internal/concept/children.go
@@ -1,0 +1,72 @@
+package concept
+
+import "github.com/Obedience-Corp/camp/internal/config"
+
+// pickerHiddenChildren are workflow registry entries that should never appear in
+// the concept picker submenu even though they live in the config registry. The
+// intents workflow is config-managed but linking an intent to the intents
+// directory is noise (design Q3).
+var pickerHiddenChildren = map[string]bool{
+	"intent":  true,
+	"intents": true,
+}
+
+// isPickerHidden reports whether a child concept is filtered from the picker.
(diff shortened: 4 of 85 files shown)


internal/intent/notes.go
```diff
@@ -0,0 +1,176 @@
...
+	newPath := s.getIntentPath(StatusInbox, note.ID)
+	if err := os.MkdirAll(filepath.Dir(newPath), 0755); err != nil {
+		return nil, camperrors.Wrap(err, "creating directory")
+	}
+
+	data, err := SerializeIntent(note)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "serializing converted note")
+	}
+	if err := os.WriteFile(newPath, data, 0644); err != nil {
```
obey-agent, line 302: `Convert` is writing the converted note with `os.WriteFile`, which truncates any existing `inbox/<note.ID>.md`. Notes and intents use the same ID generator but live in separate stores, so imported data or same-title captures in the same second can produce a note and an inbox intent with the same ID. Converting that note silently replaces the existing intent and then removes the note. Please check the target path before writing and return a conflict/`ErrFileExists`, or otherwise use a collision-safe move, and add a regression test for note/intent ID collisions. This is a data-loss path.

internal/intent/idindex_test.go
```diff
@@ -0,0 +1,92 @@
...
+func TestMove_ResolvesRenamedFileByID(t *testing.T) {
+	ctx := context.Background()
+	tmp := t.TempDir()
+	intentsDir := filepath.Join(tmp, "intents")
+	svc := NewIntentService(tmp, intentsDir)
+
+	id := "do-the-thing-20260101-000000"
+	writeRenamedIntent(t, intentsDir, "inbox", "renamed-thing-20260101-000000", id, "Renamed thing")
+
+	moved, err := svc.Move(ctx, id, StatusReady)
```
obey-agent, line 63: This test proves a renamed file can be found before/after `Move`, but it misses the rename contract. `Move` currently rebuilds the target as `getIntentPath(newStatus, intent.ID)`, so `renamed-thing-20260101-000000.md` becomes `do-the-thing-20260101-000000.md` after ordinary triage. `UpdateDirect`/editor status changes have the same pattern. Please preserve the current basename, or recompute from the current title using the same collision handling as rename, and assert the moved path keeps the renamed slug.

obey-agent (changes requested):
**Verdict: Request changes**

This is a substantial and useful intent-capture UX pass: notes as a category, tags, rename-by-stable-frontmatter ID, and nested workflow concepts all fit the direction of the campaign tooling. The main structure is coherent, and the tests cover a lot of the happy path. I am requesting changes for two data-integrity issues around moving files after this PR makes filenames mutable.

**Key findings**

- `Convert` can overwrite an existing `inbox/<id>.md` when a note and an intent share an ID. Since notes and intents are separate stores but use the same ID generator, conversion needs a target-exists check or collision-safe move before `os.WriteFile`.
- Renamed filenames are not durable across normal status moves. `Move`, `UpdateDirect`, and editor-driven status changes rebuild the target filename from `intent.ID`, so a renamed slug reverts to the old ID slug after triage. The new move test should assert basename preservation and the implementation should preserve or intentionally recompute the user-facing slug.

**What's done well**

- The frontmatter-ID resolver is a good direction: fast path plus lazy index avoids per-lookup directory scans while supporting renamed files.
- The note category is cleanly excluded from normal intent list/gather/index paths.
- The nested workflow concept migration is careful about preserving custom concepts and default additions.

**Verification**

- `just lint` passed, including `lint-no-host-fs-tests` with no new violators.
- `go test ./internal/intent ./internal/intent/gather ./internal/intent/tui ./internal/intent/tui/explorer ./internal/concept ./internal/config ./internal/scaffold ./internal/commands/workflow ./cmd/camp/intent` passed.
- `go build ./...` passed.
- `just test unit` fails only in `cmd/camp` with the two known manifest tests for `workitem priority`; I confirmed the same two failures reproduce on `origin/main`.

**Staff standard**

I would be comfortable with the overall design after the two file-move data integrity gaps are closed and covered by regression tests.

lancekrogers:
Thanks for the careful review. Both data-integrity findings are fixed in f037e9c:

**1. Convert overwriting an existing `inbox/<id>.md`**
`Convert` now stat-checks the target before writing and returns `ErrFileExists` instead of overwriting. Notes and intents are separate stores sharing one id generator, so a true id collision is now surfaced rather than silently clobbering the existing intent. New test `TestConvert_DoesNotOverwriteExistingIntent` constructs a note and an intent with a colliding id and asserts the convert is rejected and both files stay intact.

**2. Renamed slug reverting on status moves**
`Move`, `UpdateDirect`, and the editor-driven status change in `Edit` now preserve the file's current basename via a new `moveTargetPath` helper, instead of rebuilding `<id>.md` from `intent.ID`. Identity is still the `id:` frontmatter, so carrying the basename across directories is safe and the index resolver continues to find the file. New test `TestRename_SurvivesStatusMove` renames an intent, then moves it (via both `Move` and an audited `UpdateDirect` status change) and asserts the renamed basename is preserved and the id still resolves.

Verified end-to-end: `rename` then `move ... ready` keeps `correct-readable-title-<ts>.md` (no revert to the old slug). `just lint` is clean and the intent suite passes; the only full-suite failures remain the pre-existing `workitem priority` manifest tests.

internal/intent/service.go
```diff
@@ -380,10 +416,19 @@ func (s *IntentService) Count(ctx context.Context) ([]StatusCount, int, error) {
...
 func (s *IntentService) getIntentPath(status Status, id string) string {
 	return filepath.Join(s.intentsDir, string(status), id+".md")
 }
 
+// moveTargetPath returns the destination when moving an existing file to
+// newStatus, preserving its current basename so a renamed slug survives status
+// changes. Identity is the id: frontmatter, not the filename, so carrying the
+// basename across directories is safe.
+func (s *IntentService) moveTargetPath(newStatus Status, oldPath string) string {
+	return filepath.Join(s.intentsDir, string(newStatus), filepath.Base(oldPath))
```
obey-agent: Preserving the basename fixes the rename durability issue, but the target path is still not collision-safe. Rename uniqueness is only checked within the current status directory, so two different intents can legitimately have the same renamed basename in different statuses (for example same timestamp suffix, same new title, different original IDs). Moving one into the other status now writes to that basename and overwrites the existing file. Please make `moveTargetPath`/its callers check whether the target exists and belongs to a different frontmatter `id`, then return a conflict or choose a disambiguated basename before any `os.WriteFile`/`Save` runs. This should cover `Move`, `Edit` status changes, and `UpdateDirect`.

obey-agent (changes requested):
**Verdict: Request changes**

Thanks for the update. The two prior blockers are mostly addressed: note conversion now refuses an occupied `inbox/<id>.md`, and normal moves/updates preserve renamed basenames instead of reverting to the original ID slug.

One data-loss issue remains in the basename-preserving move path. Because rename uniqueness is scoped to the current status directory, two different intents can share the same renamed basename in different statuses. `Move`, `Edit`, and `UpdateDirect` can now write that basename into the target status without checking whether it already belongs to a different frontmatter `id`, which can overwrite another intent. I left the inline comment on `moveTargetPath`.

**Verification**

- `just lint` passed (`lint-no-host-fs-tests` clean; no new violators).
- `go test ./internal/intent ./internal/intent/gather ./internal/intent/tui ./internal/intent/tui/explorer ./internal/concept ./internal/config ./internal/scaffold ./internal/commands/workflow ./cmd/camp/intent` passed.
- `go build ./...` passed.
- `just test unit` still fails only in `cmd/camp` with the known manifest-test bucket; 2564/2566 tests passed.

**Staff standard**

I would approve the intent changes after the cross-status basename collision is handled with a conflict or disambiguated destination and a regression test.

lancekrogers:
Addressed the remaining cross-status basename collision in `1f087d7`.

**The bug.** Preserving the renamed basename on a status change was correct, but `rename` only enforces filename uniqueness *within a single status directory*. Two distinct intents (same timestamp suffix, same renamed title, different `id:`) can therefore legitimately share a basename across statuses. `Move`, `Edit`, and `UpdateDirect` carried the basename into the target status unconditionally, so moving one onto the other's basename overwrote an unrelated intent.

**The fix.** `moveTargetPath` is now id-aware. It accepts a destination only when the path is free or already held by the *same* frontmatter `id`; when a *different* `id` occupies it, it disambiguates with a `-2`, `-3`, ... suffix (the same scheme `rename` already uses via `uniqueRenamePath`). The check runs before any `os.WriteFile`/`Save`, and all three callers (`Move`, `Edit` status change, `UpdateDirect`) route through it.

I chose disambiguation over a hard conflict error so routine triage never fails because of an unrelated intent's basename. No data is lost either way, and it stays consistent with rename's existing collision behavior.

**Regression tests** (`internal/intent/rename_test.go`): `TestMove_`, `TestUpdateDirect_`, and `TestEdit_DoesNotOverwriteCrossStatusBasename`. Each builds the two-intents-sharing-a-basename precondition, then asserts the moved intent is disambiguated, the other survives untouched (same path, id, title), and both resolve by `id` with two distinct files in `ready`. I confirmed all three **fail** without the production change (`a collided onto b's basename ...`) and pass with it.

**Verification.**
- `go test ./internal/intent/...` green; `go build ./...` clean.
- `just lint`: `0 issues`, `lint-no-host-fs-tests` clean (no new violators).
- End-to-end CLI smoke: seeded two intents sharing `shared-title-20200101-120000.md` across `inbox`/`ready`, then `camp intent move` of the inbox one into `ready` produced `shared-title-20200101-120000-2.md` and left the existing file untouched. `camp intent list` shows both, and `camp intent show <id>` resolves each to its own distinct id.

The two known `cmd/camp` `workitem priority` manifest tests still fail identically on `origin/main`, unrelated to this PR.

obey-agent (approved):
**Verdict: Approve**

The new `1f087d7` change closes the remaining data-loss issue from the prior review. `moveTargetPath` is now ID-aware, preserves renamed basenames when safe, and disambiguates cross-status basename collisions before any write can overwrite a different frontmatter `id`.

**Key findings**

No blocking findings in this re-review. The new regression coverage exercises the three relevant mutation paths: `Move`, `UpdateDirect`, and editor-driven status changes through `Edit`.

**What's done well**

- The fix keeps the user-facing rename behavior while preventing cross-status clobbering.
- The tests build the exact two-intents/same-basename precondition and verify both IDs remain resolvable afterward.
- Error and path handling stay consistent with the existing intent service style.

**Verification**

- `go test ./internal/intent/...` passed.
- `go build ./...` passed.
- `just lint` passed (`lint-no-host-fs-tests` clean; no new violators).

**Staff standard**

Yes. With the collision-safe move path in place, I would be comfortable submitting this intent-file behavior as my own work.

internal/intent/rename.go
```diff
@@ -0,0 +1,114 @@
...
+	if err := ctx.Err(); err != nil {
+		return nil, camperrors.Wrap(err, "context cancelled")
+	}
+
+	newTitle = strings.TrimSpace(newTitle)
+	if newTitle == "" {
+		return nil, camperrors.Wrap(camperrors.ErrInvalidInput, "rename: new title is empty")
+	}
+
+	oldPath, err := s.resolveByID(id)
```
obey-agent, line 28: **Renaming a note fails with `not found` — confirmed by running it.** `Rename` resolves through `resolveByID`, which scans only `AllStatuses()` (and `buildIDIndexLocked` does the same), so a note id never resolves. The explorer binds `R` in the notes view (update_keys.go:102 says "intent or note"), so pressing R on a note always surfaces "Rename failed: ... not found".

Repro: `svc.CreateNote(...)` then `svc.Rename(ctx, note.ID, "new title")` returns `ErrNotFound`.

Fix one of two ways: make `resolveByID`/the id index also cover `NoteStatuses()` (and add a rename test for a note), or gate the `R` binding on `!m.notesMode` and drop the "or note" claim. Given notes are tag-organized freeform text, rename support seems worth having.

internal/intent/templates/intent.md.tmpl
```diff
@@ -3,9 +3,10 @@ id: {{ .ID }}
 title: {{ .Title }}
 type: {{ .Type }}
 concept: {{ .Concept }}
-status: inbox
+status: {{ .Status }}
 created_at: {{ .CreatedAt }}
 author: {{ .Author }}
+tags: [{{ range $i, $t := .Tags }}{{ if $i }}, {{ end }}{{ $t }}{{ end }}]
```
obey-agent, line 9: Tags are interpolated **unescaped** into a YAML flow sequence here (same in note.md.tmpl:7). Custom tags are free text (tag overlay input, `--tag` flags), so a tag containing `,` silently splits into two tags, and `:`, `]`, or `#` corrupts the frontmatter so `loadIntent` fails and the item disappears from listings. `SerializeIntent` is safe (yaml.Marshal quotes), so only the create path is exposed — meaning the file is born corrupt.

Either validate tags to a slug-safe charset at every entry point (overlay's `addCustomTag`, the `--tag` flags), or quote each tag in the template. Validation is the better fix: it also keeps `tags: [a b c]` greppable and consistent.

cmd/camp/intent/add.go
```diff
@@ -176,11 +179,12 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
...
-		FullMode:     fullMode,
-		Author:       author,
-		CampaignRoot: campaignRoot,
-		Shortcuts:    shortcuts,
+		DefaultType:   intentType,
+		FullMode:      fullMode,
+		Author:        author,
+		CampaignRoot:  campaignRoot,
+		Shortcuts:     shortcuts,
+		AvailableTags: cfg.IntentTags(),
```
obey-agent, line 202: `camp intent add --tag x` is silently dropped on the interactive path: `tagsFlag` is read but the TUI results below use `saved.Tags`/`result.Tags` alone. `note.go` handles this exact case with `mergeTags(tags, saved.Tags)` — do the same here (or pass the flag tags into `AddOptions` as initial selections so the overlay shows them pre-toggled).

cmd/camp/intent/note.go
```diff
@@ -0,0 +1,204 @@
...
+		CampaignRoot:  campaignRoot,
+		AvailableTags: cfg.IntentTags(),
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, saved := range model.SavedResults() {
+		if err := runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, intent.CreateOptions{
+			Title:  saved.Title,
```
obey-agent, line 110: `camp intent note --body "details"` with no positional text resolves `body` at the top of the command but never passes it into the TUI-path `CreateOptions` (here and in the `result` branch below) — the body is silently lost. Either include `Body: body` in both TUI-path CreateOptions, or reject `--body`/`--body-file` without positional text the way the non-TTY branch rejects missing text.

obey-agent (commented):
## Staff Review — PR #319 (intent capture UX: notes, tags, rename, workflow sub-concepts)

**Verdict: Request Changes**

### Overview
This is a well-architected PR across all three threads. The id-as-identity design (frontmatter `id:` authoritative, fast-path stat + lazy index, basename-preserving moves with collision-safe disambiguation) is the right call, and the cross-status clobber tests show real care about failure modes. The nested-concept migration is idempotent and preserves custom entries. Test coverage is genuinely strong — error paths first, collision preconditions asserted, TUI flows exercised end-to-end.

### Key Findings (see inline comments)
1. **Note rename is broken** (`internal/intent/rename.go:28`) — `resolveByID` and the id index only scan `AllStatuses()`, so `Rename` on a note returns `not found`. The explorer offers `R` in the notes view ("intent or note"), so the action always fails. Confirmed by running it against the branch.
2. **Tag YAML injection at create time** (`intent.md.tmpl:9`, `note.md.tmpl:7`) — tags are interpolated unescaped into a YAML flow sequence; a free-text custom tag containing `,`/`:`/`]`/`#` splits or corrupts the frontmatter at birth, making the item unparseable. Validate tags to a slug-safe charset at the entry points.
3. **`camp intent add --tag` dropped on the interactive path** (`cmd/camp/intent/add.go:187`) — `note.go` merges flag tags into TUI results; `add.go` doesn't.
4. **`camp intent note --body` dropped on the TUI path** (`cmd/camp/intent/note.go:108`) — body is resolved but never passed into the TUI-path `CreateOptions`.

Minor, no inline comment: the `camp intent rename` help example uses a short slug (`add-dark`) but resolution is exact-id only — the example as written would fail.

### What's Done Well
- `TestConvert_DoesNotOverwriteExistingIntent` and the `sharedBasenameAcrossStatuses` suite are exactly the kind of adversarial tests this storage model needs.
- Single `SlugFromTitle` replacing the duplicate in merge.go — correct consolidation instead of a parallel implementation.
- `moveTargetPath`/`collisionSafeMovePath` semantics (reuse own orphan, never clobber another id) are clearly documented and tested.
- Context checks and `camperrors` wrapping are consistent throughout; the id index is properly mutex-guarded and invalidated on every mutation.

### Staff Standard
The architecture and test discipline here are work to be proud of. But the notes feature ships with a user-facing action that fails 100% of the time (rename in notes view) and a create path that can write unparseable files from ordinary tag input — those cross the line from polish items to correctness gaps. Fix the four inline items and this is a clean approve.

internal/intent/rename.go
obey-agent: Re-verified at 11c270f: still reproduces. `svc.CreateNote(...)` then `svc.Rename(ctx, note.ID, ...)` returns `intent not found` — `resolveByID` and `buildIDIndexLocked` still scan only `AllStatuses()`, and the explorer still binds `R` in the notes view.

internal/intent/templates/intent.md.tmpl
obey-agent: Re-verified at 11c270f: still reproduces, with a new wrinkle. A tag containing `:` (e.g. `x: y`) now fails at create time with `parsing rendered template: yaml: unmarshal errors` because the render→parse round-trip chokes on the corrupt flow sequence — so the command errors instead of writing a corrupt file. A tag containing `,` (e.g. `a, b`) still silently splits into two tags. Tag validation at the entry points (overlay `addCustomTag`, `--tag` flags) is still needed.

cmd/camp/intent/add.go
obey-agent: Re-verified at 11c270f: still unaddressed. `tagsFlag` is read at line 108 but only reaches the fast path (line 160); the TUI saved/result paths still pass `saved.Tags`/`result.Tags` with no merge. `note.go`'s `mergeTags(tags, saved.Tags)` remains the pattern to mirror.

cmd/camp/intent/note.go
obey-agent: Verified fixed at 11c270f: the note TUI now collects a body (title → body editor), both TUI-path `CreateOptions` include `Body`, and `camp intent add --note` covers the editor path via `CreateWithEditor` (which routes by the parsed frontmatter status, so deep-captured notes land in notes/ correctly). Covered by `TestIntentAddModel_NoteModeCollectsBody` and `TestIntentAdd_NoteFlagRoutesThroughAddCommand`.

obey-agent (commented):
## Staff Re-Review — PR #319 at 11c270f

**Verdict: Request Changes**

### Overview
The two new commits since my last review fix one of the four findings and add a sensible timestamp-precision improvement. I re-verified every finding empirically against 11c270f rather than trusting the commit messages. Three of the four findings still reproduce, and none of the three open threads carry a reply or a code change.

### Re-Verification Results (see thread replies)
- **Fixed: note body flow** (`cmd/camp/intent/note.go`, `add_model.go`). The note TUI now collects a body via the editor step, both TUI-path CreateOptions carry Body, and the new `camp intent add --note` editor path routes correctly to notes/ because CreateWithEditor derives the destination from the parsed frontmatter status. Good test coverage for both.
- **Still broken: note rename** (`internal/intent/rename.go`). Re-ran the probe: `CreateNote` then `Rename(note.ID, ...)` still returns `intent not found`. `resolveByID` and the id index still scan only `AllStatuses()`, and the explorer still binds `R` in the notes view.
- **Still broken: hostile tag input** (`intent.md.tmpl` / `note.md.tmpl`). Re-ran the probe: a tag containing `:` now fails creation with `parsing rendered template: yaml: unmarshal errors`, and a tag containing `,` still silently splits into two tags. The render-then-parse round trip turns the corruption into a hard create failure for some inputs, which is better than writing a corrupt file, but ordinary free-text tags still break the command. Validate tags at the entry points.
- **Still unaddressed: `camp intent add --tag` on the interactive path** (`cmd/camp/intent/add.go:108`). The flag still reaches only the fast path; the TUI saved/result branches pass `saved.Tags`/`result.Tags` with no merge.

### Notes on the New Work
- The `--note` flag on `add` with conflict guards against `--type`/`--concept`, plus the shared `navigationShortcuts` extraction, is clean and well tested, including a real command-level routing test asserting notes land in notes/ and never inbox/.
- `FormatCreatedAt` moving to RFC3339 is a behavior change smuggled in by commit message ("preserve fresh timestamps"), but it is correct: I verified legacy date-only `created_at:` frontmatter still parses, and the round-trip test pins the new format.

### Staff Standard
The work that landed is good work. But a re-review request with three of four findings untouched and no thread replies is not ready to flip the verdict. The note rename failure remains a user-facing action that fails 100% of the time it is offered. Fix the three open items and I expect this to be a clean approve.

internal/intent/notes.go
```diff
@@ -0,0 +1,258 @@
...
+		return nil, camperrors.Wrap(err, "context cancelled")
+	}
+
+	if _, err := s.GetNote(ctx, id); err == nil {
+		return nil, camperrors.Wrap(ErrFileExists, id)
+	} else if !errors.Is(err, ErrNotFound) {
+		return nil, err
+	}
+
+	it, err := s.Find(ctx, id)
```
obey-agent: **Destructive move on a fuzzy match — confirmed by running it.** `Find` falls back to substring matching when exact id resolution fails, so `MoveIntentToNote(ctx, "widget")` resolved and moved `fix-the-widget-renderer-20260609-100000` into notes/. A service method that relocates a file and strips its lifecycle metadata should never act on a fuzzy guess; a caller passing a truncated or mistyped id silently converts whichever intent happens to contain that substring.

The opposite direction already gets this right: `MoveNoteToStatus` resolves via exact `GetNote`. Use `resolveByID` + `loadIntent` (or `Get`) here instead of `Find`, and add a test asserting a partial id returns `ErrNotFound`.

internal/intent/tui/add_model.go
```diff
@@ -445,6 +467,13 @@ func (m IntentAddModel) updateBody(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		return m, m.openExternalEditor()
 	}
 
+	// Handle ctrl+t for tags from the editor step.
+	if msg.String() == "ctrl+t" {
```
obey-agent, line 477: Moving ctrl+t into the body editor (with a test pinning that the title step does not open it) leaves the default fast TUI with no tag entry point: `camp intent add` without `--full` runs title, type, concept and never reaches a body step, so ctrl+t is unreachable in the most common interactive flow. Combined with the still-open finding that `--tag` flags are dropped on the interactive path, fast interactive capture now has zero way to attach a tag.

If pulling ctrl+t out of the title step was to avoid a key conflict, consider offering it from the type/concept steps too, or at minimum merge the `--tag` flag values into the TUI results so flags remain the escape hatch.

obey-agent (commented):
## Staff Re-Review — PR #319 at b861122

**Verdict: Request Changes**

### Overview
The three new commits are a meaningful feature push: notes become first-class in the explorer (a combined default view with a Notes group, bidirectional moves between notes and the lifecycle via MoveIntentToNote/MoveNoteToStatus, notes in search and status filters, and per-category action-menu gating). The work is well tested and the service-level convert path got safer. However, none of the three findings open since my first review were addressed, and the new move API introduces a fresh correctness bug I confirmed by running it.

### Key Findings (see inline comments and thread replies)
1. **New: destructive move on a fuzzy match** (`internal/intent/notes.go:140`). `MoveIntentToNote` resolves through `Find`, which substring-matches when exact resolution fails. Probe: `MoveIntentToNote(ctx, "widget")` moved `fix-the-widget-renderer-20260609-100000` into notes/ and stripped its lifecycle metadata. The reverse direction (`MoveNoteToStatus`) correctly uses exact `GetNote`. Resolve exactly.
2. **Still open: note rename fails 100% of the time** (`internal/intent/rename.go:28`). Re-ran the probe at this head: `Rename` on a note still returns `intent not found`. This delta made notes more prominent (they now appear in the default explorer view where `R` is offered on every row), which widens the surface of this bug, not narrows it.
3. **Still open: hostile tag input** (`intent.md.tmpl:9`). Templates unchanged; tags with `:` still fail creation with a YAML parse error and tags with `,` still silently split.
4. **Still open: `--tag` dropped on the interactive add path** (`cmd/camp/intent/add.go:108`), now compounded: ctrl+t moved into the body editor, so the default fast TUI (title/type/concept, no body step) has no tag entry point at all (`add_model.go:471`).

### What's Done Well
- `MoveNoteToStatus` replacing the per-path `os.Stat` guard with an id-level `resolveByID` collision check is a real improvement over the code I reviewed last round.
- `MoveIntentToNote` clears every lifecycle field, reuses `moveTargetPath` for collision-safe placement, and the test asserts the frontmatter actually dropped `type:`/`concept:`.
- The explorer integration is disciplined: gather/promote/archive/delete are explicitly gated for notes with honest status messages instead of half-working actions, and the TUI flow tests (intent→note, note→ready with type picker) exercise the real service.
- In-memory fuzzy search over loaded items keeps notes searchable in the combined view without growing the search index contract.

### Staff Standard
The feature work here is good, and the test discipline continues to be a strength of this PR. But this is the second re-review where the original findings sat untouched while new surface area landed on top of them, and the new surface inherited one of the same root causes (resolution APIs that are too loose for destructive operations). Fix the exact-resolution bug, the three open findings, and this approves cleanly. I would hold my own work to the same bar: no new features on top of known-broken paths.

lancekrogers:
Addressed all four findings in `c2212fb`. I verified each against this head before fixing, and confirmed every new regression test fails without the production change.

**1. Destructive move on a fuzzy match (`notes.go`).** `MoveIntentToNote` now resolves with exact `Get` instead of `Find`, so a truncated or mistyped id returns `ErrNotFound` rather than converting whichever intent happens to contain that substring and stripping its lifecycle metadata. Test `TestMoveIntentToNote_RejectsPartialID` asserts a partial id is rejected and the intent stays in inbox.

**2. Note rename failing 100% of the time (`notes.go`, `rename.go`).** Note resolution is now id-authoritative: `GetNote` resolves through a new `resolveNoteByID` (filename fast path plus a small note-dir scan), and `Rename` falls back to it when lifecycle resolution misses. A note now renames in place in `notes/`, keeps its id and note status, and stays resolvable afterward (including renamed notes, which the old filename-based `GetNote` could not find). The CLI `camp intent rename` also falls back to the note store, and its help example now shows a full id since resolution is exact. Test `TestRename_WorksForNotes`; verified end-to-end (`camp intent rename <note-id> "..."` renamed the note within `notes/` and kept its id).

**3. Hostile tag input (`tags.go`, wired into the service create/update paths).** Added `validateAndNormalizeTags` at the single choke point every create/update funnels through. It trims, drops empties, de-duplicates, and rejects any tag whose characters would break the flow-style YAML the templates render (comma, colon, brackets, and so on) with a clear `ErrInvalidInput` before any write. A comma no longer silently splits a tag and a colon no longer corrupts frontmatter. Tests `TestValidateAndNormalizeTags`, `TestCreateDirect_RejectsHostileTags`, `TestCreateDirect_NormalizesTags`; verified end-to-end (`--tag "a,b"` and `--tag "key: value"` are rejected with a readable error, clean tags accepted).

**4. `--tag` dropped on the interactive add path, and fast-TUI tag entry (`add.go`, `add_model.go`).**
- The real bug: `add.go` built the TUI `CreateOptions` from `saved.Tags`/`result.Tags` without merging the `--tag` flag (`note.go` already merged it). Fixed: both the save-and-new loop and the final result now use `mergeTags(tagsFlag, ...)`. Test `TestMergeTags`.
- On "never reaches a body step": I could not reproduce that. The existing `TestIntentAddModel_BodyCtrlS` drives the default (non-`--full`) flow title -> type -> concept -> body, so ctrl+t at the body step was already reachable in the fast flow (`fullMode` turns out to be unused; the body step is always entered). That said, making users walk to the last step just to tag is poor, so I took the other suggestion and made ctrl+t open the overlay on the type and concept steps too, advertised in their help lines. Tests `TestIntentAddModel_TypeStepOpensTags` and `TestIntentAddModel_ConceptStepOpensTags` (the title step still does not open it, per the existing test).

**Verification.**
- `just lint` clean (`lint-no-host-fs-tests` clean, no new violators); `go build ./...` clean; `go test ./internal/intent/... ./cmd/camp/intent/...` green.
- Genuineness check: for findings 1, 2, 3, and 4b I reverted the production changes with the new tests in place and confirmed each fails (fuzzy move resolves `widget`; note rename returns `intent not found`; `a,b` creates silently while `key: value` throws a YAML unmarshal error; type/concept steps do not open the overlay), then restored.

The two known `cmd/camp` `workitem priority` manifest tests still fail identically on `origin/main`, unrelated to this PR.

internal/concept/children.go
```diff
@@ -0,0 +1,72 @@
...
+
+	for _, ch := range parent.Children {
+		if isPickerHidden(ch.Name) {
+			continue
+		}
+		items = append(items, Item{
+			Name:     ch.Name,
+			Path:     ch.Path,
+			IsDir:    true,
+			Children: countFn(ch.Path),
```
obey-agent: This makes configured child concepts drillable based on `countFn(ch.Path)`, but the picker still drills by appending `item.Name` to the parent concept subpath and then calls `ListItems(parent.Name, subpath)`. For the new default `workflow > festivals` child, `item.Path` is `festivals/` while the parent is `workflow/`; pressing `l`/right therefore loads `ListItems("workflow", "festivals")`, which reads `workflow/festivals` instead of `festivals/` and usually shows an empty/wrong submenu. Because the top-level festivals concept was removed from the defaults, this means users can select only `festivals/` but cannot browse to `festivals/active/<festival>` through the picker. Please carry the child concept/path into subsequent drill calls (or switch the selected concept to the child for traversal) and add a regression for right-drilling into the configured `festivals/` child.

obey-agent (commented):
**Verdict: Request Changes**

**Overview**
The latest head addresses the earlier note rename, tag validation, interactive tag, and exact move concerns. The remaining issue is in the nested workflow concept picker: the new default tree removes the top-level festivals concept, but drilling from `workflow > festivals` now resolves under the wrong parent path.

**Key Findings**
- Inline on `internal/concept/children.go`: configured child concepts are marked drillable, but the picker continues traversal as `ListItems("workflow", "festivals")`. For the default `workflow > festivals` child, that reads `workflow/festivals` instead of `festivals/`, so users cannot browse to `festivals/active/<festival>` through the default picker.

**What's Done Well**
- The prior data-integrity fixes around note conversion, renamed basename preservation, and cross-status basename collisions remain solid.
- The note/intent move APIs now resolve destructive operations exactly and have useful regression coverage.
- No Codex/co-author advertising was found in the PR body, changed files, or commit messages.

**Verification**
- `go test ./internal/intent/... ./cmd/camp/intent/... ./internal/concept/... ./internal/config/... ./internal/commands/workflow/... ./internal/scaffold/... ./internal/shell/...` passed.
- `just lint` passed.
- `go build ./...` passed.
- `just test unit` failed only in `cmd/camp`: `TestManifestCommand_AllRestrictedCommandsPresent` and `TestManifestCommand_AllCommandsHaveAnnotations` for the known `workitem priority` manifest bucket.

**Staff Standard**
Not yet. I would approve the PR after the picker carries the configured child path/concept into drill traversal and adds a regression for drilling into `workflow > festivals`.

obey-agent (changes requested):
Formal review state: request changes for the inline `workflow > festivals` drill regression. See the detailed review summary and inline comment posted by obey-agent.

internal/intent/rename.go
```diff
@@ -0,0 +1,121 @@
...
+		oldPath = notePath
+	}
+	it, err := s.loadIntent(oldPath)
+	if err != nil {
+		return nil, camperrors.Wrap(ErrNotFound, id)
+	}
+
+	newPath := s.uniqueRenamePath(filepath.Dir(oldPath), renameBase(it.ID, newTitle), oldPath)
+
+	it.Title = newTitle
```
obey-agent, line 46: Rename only rejects an empty new title before mutating and serializing the intent. That bypasses the same validation used by `UpdateDirect`, so `camp intent rename <id> x` can persist a title that `Intent.Validate()` rejects (`ErrTitleTooShort`) and still return success. Please validate the mutated intent before writing/moving the file, and add a regression test for an invalid rename title.

internal/config/defaults.go
```diff
@@ -23,42 +32,17 @@ func DefaultConcepts() []ConceptEntry {
...
-			Depth:       &depth1,
-		},
-		{
-			Name:        "explore",
-			Path:        "workflow/explore/",
-			Description: "Exploratory notes and discovery work",
-			Depth:       &depth1,
+			Description: "Workflows",
+			Children: []ConceptEntry{
+				{Name: "festivals", Path: "festivals/", Description: "Multi-step festival plans"},
```
obey-agent, line 40: This new `workflow -> festivals` child points at `festivals/`, which is not underneath the parent `workflow/` path. The picker renders that child correctly via `item.Path`, but drilling into it still builds `currentSubpath` from `item.Name`, so `workflow > festivals` loads `workflow/festivals` instead of `festivals/`. Either update the drill-down state to follow the child path or keep child paths relative to the parent, and cover this case in the concept picker tests.

internal/intent/tui/explorer/tags.go
```diff
@@ -0,0 +1,83 @@
...
+// persistTags writes the new tag set. Notes are saved directly (they live
+// outside the intent lookup); intents go through UpdateDirect for auditing.
+func (m *Model) persistTags(target *intent.Intent, tags []string) tea.Cmd {
+	return func() tea.Msg {
+		if target.Status.IsNote() {
+			note, err := m.service.GetNote(m.ctx, target.ID)
+			if err != nil {
+				return tagsUpdatedMsg{err: err}
+			}
+			note.Tags = tags
```
obey-agent: Note tag edits are saved directly through `Save`, so they bypass `validateAndNormalizeTags`, leave `updated_at` unchanged, and skip the audit/autocommit path used for lifecycle intents below. That makes notes behave differently from creates/edits and can persist an invalid tag set from the overlay. Please route note tag updates through a service method that normalizes/validates, refreshes `UpdatedAt`, and records the same audit/commit side effects.

internal/commands/workflow/create_plan.go
```diff
@@ -123,27 +123,34 @@ func planShortcut(cfg *config.CampaignConfig, plan *createPlan, opts createOptio
...
+	plan.Concept.Parent = workflowParentName
 
 	concepts := cfg.ConceptList
 	if len(concepts) == 0 {
 		concepts = cfg.Concepts()
 	}
 
-	for _, concept := range concepts {
-		if !strings.EqualFold(concept.Name, opts.Type) {
+	// New collections nest under the workflow parent; look for an existing child.
```
obey-agent, line 133: Existing campaigns can still have legacy flat workflow concepts such as `research -> workflow/research/` until they run repair. This scan only looks under the nested `workflow` parent, so `camp workflow create research ...` will append a second child concept with the same name/path instead of treating the existing top-level concept as a match; `--replace` also will not replace it. Please fold the legacy top-level entries into this check or run the same migration/normalization before planning.

obey-agent (commented):
Verdict: Request changes

This is a broad and useful feature set, but I found several correctness issues that should be fixed before merge:

- `rename` can persist titles that fail intent validation.
- `workflow -> festivals` exposes a concept-picker drill-down bug because the child path is outside the parent path.
- Note tag edits bypass tag validation, `updated_at`, audit, and autocommit behavior.
- `workflow create` can duplicate legacy flat workflow concepts in existing campaigns.

I also agree there is too much comment density in this patch. Keep comments that explain invariants and edge cases, like stable IDs across rename and zsh compinit ordering, but trim comments that restate function names or straight-line code. The service/TUI paths are already large enough that repeated narration makes the important safety comments harder to find.

Tests run:

- `go test ./internal/intent ./internal/intent/tui ./internal/intent/tui/explorer ./internal/concept ./internal/commands/workflow ./internal/scaffold ./cmd/camp/intent ./internal/shell`

Coordination note: this PR also touches the zsh shell template. If PR #320 lands first, rebase/resolve carefully so `cnote` lands on top of the deferred zsh completion registration instead of restoring unconditional `compdef` calls.

internal/concept/children.go
lancekrogers: Fixed in 1703b95. `ListItems` now remaps drill subpaths onto configured child concepts: when the first segment of the subpath names a visible child, traversal continues as the child concept with the remaining subpath, so `workflow > festivals` reads `festivals/` (and deeper drills read `festivals/active/...`). The remap also carries the child's own depth and ignore rules, which restores the behavior the old top-level festivals/design concepts had (depth-1 children mark their items DrillDisabled; child ignore lists apply at the child root). Depth-0 children are now DrillDisabled directly in the parent submenu, and disk-only dirs with no child mapping still fall through to the parent path unchanged.

Regression coverage: service-level tests for the child-path drill (with a `workflow/festivals` decoy dir on disk proving we no longer read it), nested drill, child depth, child ignore, depth-0 children, and disk-only fallthrough, plus an end-to-end picker test right-drilling `workflow > festivals > active` against a real FSService. All of them fail without the production change, reproducing your finding exactly:

```
--- FAIL: TestListItems_ParentChildren_DrillUsesChildPath (0.00s)
    children_test.go:132: drill items = [decoy], want [active planning] from festivals/, not workflow/festivals
--- FAIL: TestConceptPicker_DrillIntoConfiguredChild (0.00s)
    concept_picker_test.go:730: festivals drill items = [{Name:decoy Path:workflow/festivals/decoy ...}], want [active] from festivals/, not workflow/festivals
```

On the new head: `go test ./internal/concept/... ./internal/intent/tui/...`, `just lint`, and `go build ./...` pass; `go test ./...` fails only on the known pre-existing `workitem priority` manifest bucket in `cmd/camp` that you already noted.

obey-agent (commented):
## Verdict: Approve

### Overview
A large (about 5.6k line) three-thread feature (notes category + tags, id-authoritative rename, workflows as sub-concepts) that lands cleanly. The design choices are sound and the code is well-commented exactly where the "why" is non-obvious. I read the core logic in full rather than skimming, and verified the high-risk areas directly.

### Key Findings
No blocking issues. What I verified:
- Identity is genuinely id-authoritative. `resolveByID` uses an `<id>.md` fast path plus a lazily-built, mutation-invalidated id->path index, and every move / rename / convert path routes through collision-safe targets (`collisionSafeMovePath`, `uniqueRenamePath`, `pathAvailableForID`) that never overwrite a different intent's file. A renamed slug that drifts from its id still resolves.
- The `camp init --repair` concept migration (`migrateConceptsToNested` / `ensureDefaultConcepts`) is idempotent and non-destructive: custom and unknown concepts fall through to the preserve path, the merged list is written only when non-empty, and `init.go` otherwise preserves the existing config in repair mode.
- Tag validation rejects characters that would corrupt the flow-style YAML the templates render. I confirmed empirically that numeric and space-containing tags round-trip through yaml.v3 as strings (no parse error, no dropped note file), so the validator plus template are consistent.
- `rename` and `convert` wire `audit.EventRename` / `EventMove` and `commit.IntentRename` / `IntentMove` correctly at the command layer.
- Context propagation (ctx.Err checks) is present on every service entry point, the project error package is used throughout, and no `fmt.Errorf` was introduced in production code.

I also chased several potential index-out-of-range concerns in the explorer TUI (post-rename refresh, dungeon rebuild, notes-view toggle, action menu) and confirmed each is safe: callers are gated behind `SelectedIntent()` (which itself dereferences `groups[cursorGroup]`), `updateMove` clamps its index in-function against the same slice it indexes, `NewActionMenu` never builds an empty item list, and `toggleNotesMode` resets the cursor to a valid position.

### What's Done Well
Thorough test coverage for every thread: rename collision, note routing / exclusion / convert, nested-concept round-trip plus migration idempotency plus custom survival, and the picker parent->child cascade. Clean separation between the note category and the lifecycle (notes are excluded from `AllStatuses()` so intent queries never scan them). `go build` and `go vet` pass; the PR's note that the only full-suite failures are the pre-existing #318 workitem-priority manifest tests checks out.

### Staff Standard
Yes. This is careful, well-tested, well-reasoned work with the data-integrity edges (rename collisions, id drift, migration idempotency) actually thought through rather than hand-waved. I would be comfortable submitting it as my own.

obey-agent (changes requested):
## Verdict: Request Changes (correcting a premature Approve)

My earlier "Approve" summary on this PR was premature: I posted it without reconciling against the open findings from the prior Request Changes re-reviews. Retracting that verdict.

There is exactly one commit since the last Request Changes review (`c2212fb..1703b95`: "fix(concept): drill into configured child concepts by their own path"), and it touches only the four concept files. So three of the four findings from that review are necessarily still open. I re-verified each against `1703b95`.

### Fixed since last review
- **`workflow > festivals` picker drill** (`internal/concept/children.go`, `service.go`). `conceptItems` now remaps a drill subpath whose first segment names a configured child onto that child's own path/depth/ignore rules, so `workflow > festivals` reads `festivals/` (and deeper). Regression coverage added. Confirmed.

### Still open at 1703b95
1. **Rename persists titles that fail validation** (`internal/intent/rename.go:23-26`). `Rename` rejects only an empty title and never calls `Intent.Validate()` before serializing, so `camp intent rename <id> "x"` saves a title that `Validate()` rejects with `ErrTitleTooShort`. `UpdateDirect` validates after applying changes (`update.go:135`); `Rename` should do the same. Needs the validation plus a regression test. File untouched since the finding.
2. **Note tag edits bypass validation, `updated_at`, audit, and autocommit** (`internal/intent/tui/explorer/tags.go:52-62`). The note branch of `persistTags` does `note.Tags = tags; m.service.Save(...)`: it skips `validateAndNormalizeTags`, never refreshes `UpdatedAt`, and skips the `appendAuditEvent` + `autoCommitIntent` the intent branch runs immediately below. Route note tag updates through a service method that normalizes/validates, stamps `UpdatedAt`, and records the same audit/commit side effects. File untouched since the finding.
3. **`workflow create` duplicates legacy flat workflow concepts** (`internal/commands/workflow/create_plan.go:133-154`). `planConcept` only scans children under the `workflow` parent, so an existing campaign with a legacy top-level concept (for example `research -> workflow/research/`, pre-repair) is not matched; `camp workflow create research` then appends a second `research` child and `--replace` will not replace the legacy entry. Fold legacy top-level entries into the check (or normalize before planning) and cover it with a test. File untouched since the finding.

A style note from the prior review (trim comment density that restates code; keep the invariant and edge-case comments) also remains, but it is non-blocking.

### Verification
- Confirmed only `1703b95` landed since `c2212fb`, touching only `internal/concept/*` plus the concept picker test (`git diff --name-only c2212fb..1703b95`).
- Read `rename.go`, `explorer/tags.go`, `create_plan.go`, and `update.go` at the current head; all three findings reproduce by inspection as described, with `UpdateDirect` (`update.go:54,135,139`) being the validating/timestamping path the other two should mirror.

### Staff Standard
Not yet. Three findings from a formal Request Changes review remain untouched while only the picker-drill fix landed. Close the three open items with regression tests and this is a clean approve.

lancekrogers:
Addressed the three open findings from the last review in `b7df8ed`. The `8d97f2d` merge before it was only `main` (the #320 zsh fix), so those findings were untouched until this commit.

1. **Rename persists invalid titles** (`internal/intent/rename.go`) — `Rename` now runs `Intent.Validate()` on the mutated intent before any write/move, so a too-short title is rejected (`ErrTitleTooShort`) instead of saved. `TestRename_InvalidTitleRejectedWithoutMutating` asserts the error and that the original file is left untouched; it fails without the validation and passes with it.

2. **Note tag edits bypass validation / updated_at / audit / autocommit** (`internal/intent/tui/explorer/tags.go`) — added `IntentService.UpdateNoteTags`, the note-store equivalent of `UpdateDirect`'s tag path: it runs `validateAndNormalizeTags`, refreshes `UpdatedAt`, and returns a `FieldChange`. `persistTags` now routes notes through it, so the explorer emits the same `appendAuditEvent` + `autoCommitIntent` it does for lifecycle intents. Covered by `TestUpdateNoteTags_NormalizesStampsAndReturnsChange` and `TestUpdateNoteTags_RejectsInvalidTag`.

3. **`workflow create` duplicates legacy flat concepts** (`internal/commands/workflow/create_plan.go`, `create_apply.go`) — `planConcept` and `upsertConcept` now detect a legacy top-level workflow concept (for example `research -> workflow/research/` on a pre-repair campaign) via `flatWorkflowConceptIndex` and fold it under the `workflow` parent instead of appending a second child. Covered by `TestUpsertConceptFoldsLegacyFlatConcept` and the end-to-end `TestRunCreateFoldsLegacyFlatConcept`.

The earlier comment-density note: trimmed the new comments to keep only the invariant/why.

Gates on `b7df8ed`: `go build ./...`, `just lint` (0 issues, `lint-no-host-fs-tests` clean), and `go test ./internal/intent/... ./internal/commands/workflow/... ./internal/concept/... ./internal/config/... ./internal/scaffold/... ./cmd/camp/intent/...` all pass.

internal/intent/merge.go
```diff
@@ -60,48 +60,14 @@ func MergeIntents(sources []*Intent, opts MergeOptions) (*Intent, error) {
...
 }
 
-// generateMergedID creates an ID for the gathered intent.
+// generateMergedID creates an ID for the gathered intent. The slug uses the
+// shared SlugFromTitle so merged IDs slug identically to created ones.
 func generateMergedID(title string, t time.Time) string {
-	slug := generateSlugFromTitle(title)
+	slug := SlugFromTitle(title)
 	timestamp := t.Format("20060102-150405")
 	return fmt.Sprintf("%s-%s", timestamp, slug)
```
obey-agent: `MergeIntents` now says merged IDs share the same slug behavior as create/rename, but this still builds `timestamp-slug` (`20260612-120000-title`) instead of the repo-valid `slug-YYYYMMDD-HHMMSS` format produced by `GenerateID`. That means `MergeIntents` returns an `Intent` whose `Validate()` fails with `ErrInvalidIDFormat`; the current gather service happens to discard this ID by calling `CreateDirect`, but the internal merge API is now inconsistent with the ID contract and the PR removed the local slug test without adding a validity assertion.

Please generate merged IDs through `GenerateID(title, t)` or mirror its empty-slug handling (`slug == "" ? timestamp : slug + "-" + timestamp`), and add a regression like `MergeIntents(...).Validate()` / exact ID-shape coverage so gather metadata cannot drift from the canonical ID format again.

obey-agent (changes requested):
## Verdict: Request Changes

### Overview
The three open findings from the last re-review are addressed on `b7df8ed`: rename now validates before writing, note tag edits route through a validating/timestamping service method and keep audit/autocommit behavior, and `workflow create` folds legacy flat workflow concepts instead of duplicating them. The #320 zsh completion changes also survived the merge, including `cnote` on top of the deferred registration block.

### Key Findings
- `internal/intent/merge.go`: `MergeIntents` still returns gathered intents with IDs in `timestamp-slug` order while the repo ID contract and `GenerateID` use `slug-YYYYMMDD-HHMMSS`. That returned `Intent` fails `Validate()` with `ErrInvalidIDFormat`. I left an inline comment on `generateMergedID`; use `GenerateID(title, t)` or mirror its empty-slug handling and add a regression that validates the merged intent / exact ID shape.

### What's Done Well
- The prior data-integrity findings around exact note moves, note conversion collisions, renamed basename preservation, and cross-status basename collisions remain covered.
- The new `UpdateNoteTags` path mirrors lifecycle tag updates closely enough: normalization, `updated_at`, field changes, audit, and commit are all wired.
- The configured child concept drill fix is solid and the regression with a `workflow/festivals` decoy proves it follows the child path.

### Verification
- `go test ./internal/intent/... ./cmd/camp/intent/...`
- `go test ./internal/commands/workflow/... ./internal/concept/... ./internal/config/... ./internal/scaffold/... ./internal/shell/...`
- `go build ./...`
- `just lint`

### Staff Standard
Not yet. This is close, but the merge API should not produce an intent that violates the canonical ID format immediately after this PR centralizes slug generation and makes IDs frontmatter-authoritative.

internal/intent/merge.go
lancekrogers: Fixed in 3aa14f7. `MergeIntents` now routes through the shared `GenerateID` (deleting `generateMergedID` entirely), so merged IDs follow the canonical `slug-YYYYMMDD-HHMMSS` shape including the empty-slug fallback. `TestMergeIntents_Basic` now pins the contract three ways: slug-first prefix, `intentIDPattern` match, and `merged.Validate()` returning no errors. Verified load-bearing: with the old timestamp-slug order restored, all three assertions fail (`id does not match required format slug-YYYYMMDD-HHMMSS`).

Gates on 3aa14f7: `go test ./internal/intent/...` (incl. gather), `go build ./...`, `go vet`, and `just lint` all pass.

obey-agent (approved):
**Verdict: Approve**

Re-reviewed PR #319 at `3aa14f7` and specifically audited the older unresolved review threads. I confirmed the old findings are addressed in code/tests, then resolved the 15 stale/open review threads so the GitHub thread state now matches the implementation state.

**Older review items verified addressed**

- File safety: note conversion refuses ID collisions, renamed basenames survive status moves/updates, cross-status basename collisions are disambiguated, and `MoveIntentToNote` now resolves exact IDs only.
- Notes/tags UX: note rename works through the note store, invalid rename titles are rejected before mutation, tag create/update paths normalize and reject YAML-hostile tags, `camp intent add --tag` is merged into TUI results, note body flow is preserved, fast TUI tagging is available from type/concept/body steps, and note tag edits go through `UpdateNoteTags` with validation/audit/commit behavior.
- Workflow/concepts: `workflow > festivals` drill traversal follows the configured child path, and `camp workflow create` folds legacy flat workflow concepts instead of duplicating them.
- Merge IDs: `MergeIntents` now uses `GenerateID`, and the regression test validates the canonical `slug-YYYYMMDD-HHMMSS` shape via `Validate()`.

**Verification**

- `go test ./internal/intent/... ./cmd/camp/intent/...` passed.
- `go test ./internal/commands/workflow/... ./internal/concept/... ./internal/config/... ./internal/scaffold/... ./internal/shell/...` passed.
- `go build ./...` passed.
- `go vet ./...` passed.
- `just lint` passed (`lint-no-host-fs-tests` clean; no new violators).
- `go test ./...` still fails only in `cmd/camp` on the known `workitem priority` manifest bucket: `TestManifestCommand_AllRestrictedCommandsPresent` and `TestManifestCommand_AllCommandsHaveAnnotations`.

**Staff standard**

Yes. With the older review threads verified and resolved, and the latest merged-ID regression fixed, I would be comfortable submitting this as my own work.

internal/intent/notes.go
lancekrogers: Addressed at head 3aa14f7: `Convert` delegates to `MoveNoteToStatus`, which refuses an occupied id (`resolveByID` hit returns `ErrFileExists`) before any write, and the destination comes from collision-safe `moveTargetPath`. Regression: `TestConvert_DoesNotOverwriteExistingIntent`.

internal/intent/idindex_test.go
lancekrogers: Addressed: `Move`/`Edit`/`UpdateDirect` now route through `moveTargetPath`, which preserves the current (possibly renamed) basename. `TestRename_SurvivesStatusMove` pins the renamed slug across both `Move` and an `UpdateDirect` status change.

internal/intent/service.go
lancekrogers: Fixed in 1f087d7 (approved in the 06-10 re-review): `collisionSafeMovePath`/`pathAvailableForID` reuse a path only when it already holds the same frontmatter id and disambiguate with -2/-3 otherwise. Regressions: `TestMove_`/`TestUpdateDirect_`/`TestEdit_DoesNotOverwriteCrossStatusBasename`.

internal/intent/rename.go
lancekrogers: Fixed: `Rename` now falls back to `resolveNoteByID` when lifecycle resolution misses, so notes rename in place (the rename logic is directory-agnostic). Regressions: `TestRename_WorksForNotes` and the explorer-level `TestRenameAction_RenamesViaServiceAndReselects`.

internal/intent/templates/intent.md.tmpl
lancekrogers: Fixed via validation at the single entry point: `validateAndNormalizeTags` rejects any rune outside letters/numbers/space/`-_/.+` (so `,` `:` `]` `#` can never reach the template), and every create/update path funnels through it (`CreateDirect`, `CreateWithEditor`, `UpdateDirect`, `UpdateNoteTags`). Covered by `TestValidateAndNormalizeTags`; numeric and space-containing tags round-trip as strings.

cmd/camp/intent/add.go
lancekrogers: Fixed: both interactive paths now merge the flag tags — `Tags: mergeTags(tagsFlag, saved.Tags)` for ctrl+n saves and `mergeTags(tagsFlag, result.Tags)` for the final result (add.go:216,246), mirroring note.go. `mergeTags` is covered by `TestMergeTags`.

internal/intent/notes.go
lancekrogers: Fixed: `MoveIntentToNote` resolves via exact `Get` (`resolveByID`), never `Find`. Regression: `TestMoveIntentToNote_RejectsPartialID` asserts a partial id returns `ErrNotFound` and the intent does not move.

internal/intent/tui/add_model.go
lancekrogers: Fixed: ctrl+t now opens the tag overlay from the type and concept steps too, so the default fast TUI has tag entry points, and the `--tag` flag merge (other thread) restores flags as the escape hatch. Pinned by `TestIntentAddModel_TypeStepOpensTags` / `ConceptStepOpensTags` / `BodyStepOpensTags` (title step still excluded by design).

internal/intent/rename.go
lancekrogers: Fixed in b7df8ed: `Rename` validates the mutated intent before any write/move, so a too-short title is rejected (`ErrTitleTooShort`) and the original file is untouched. Regression: `TestRename_InvalidTitleRejectedWithoutMutating` (verified it fails without the fix).

internal/config/defaults.go
lancekrogers: Fixed in 1703b95 (same root cause as the children.go thread): `childForSubpath` remaps drill traversal onto the configured child's own path/depth/ignore rules, so `workflow > festivals` reads `festivals/`. Regressions: `TestConceptPicker_DrillIntoConfiguredChild` plus the children_test suite with a `workflow/festivals` decoy dir.

internal/intent/tui/explorer/tags.go
lancekrogers: Fixed in b7df8ed: note tag edits route through the new `IntentService.UpdateNoteTags` (normalizes/validates, refreshes `updated_at`, returns a `FieldChange`), and the explorer emits the same audit + autocommit as lifecycle intents. Regressions: `TestUpdateNoteTags_NormalizesStampsAndReturnsChange` / `_RejectsInvalidTag`.

internal/commands/workflow/create_plan.go
lancekrogers: Fixed in b7df8ed: `planConcept`/`upsertConcept` detect a legacy top-level workflow concept via `flatWorkflowConceptIndex` and fold it under the workflow parent (respecting `--replace`) instead of appending a duplicate child. Regressions: `TestUpsertConceptFoldsLegacyFlatConcept` and end-to-end `TestRunCreateFoldsLegacyFlatConcept`.

PR #320, opened 2026-06-12
fix(shell): make zsh completion resilient to compinit ordering

lancekrogers:
## Problem

User reported a shell bug affecting "some users." Root cause: the zsh init
script (`internal/shell/templates/zsh.sh.tmpl`) called `compdef` unconditionally
at source time. `compdef` only exists after `compinit` has run. Users who source
camp before `compinit`, or never run it (bare zsh, no framework), got three
`command not found: compdef` errors at every shell startup and no tab-completion.
Framework users (oh-my-zsh / prezto) run `compinit` early, so they never saw it,
hence "some users."

Reproduction on the old script (bare zsh, no compinit):

```
$ zsh -d -f -c 'eval "$(camp shell-init zsh)"; echo exit=$?'
(eval):171: command not found: compdef
(eval):200: command not found: compdef
(eval):279: command not found: compdef
exit=127
```

## Fix

Replace the three scattered `compdef` calls with a single registration block
resilient to `compinit` ordering:

- compdef available (compinit already ran) → register immediately. Unchanged
  behavior for oh-my-zsh / prezto users.
- compdef not yet available → defer registration to the first prompt via an
  `add-zsh-hook precmd` one-shot, so completion still loads once the user's own
  `compinit` runs.
- compinit never runs → guard silently skips. No errors, navigation still works.

## Cross-shell audit

We support bash, zsh, and fish, so all three were checked empirically:

- **bash**: not a bug, left unchanged. `return` inside `eval` only unwinds the
  eval (does not abort the host rc); completion registers on bash 3.2 and 5.x.
- **fish**: uses `complete`, which has no `compinit`-equivalent ordering
  dependency. Left unchanged.
- **zsh**: fixed as above.

## Tests

The existing harness always ran `compinit` before sourcing camp, which is
exactly why this bug was never caught. Added:

- `runZshScriptNoCompinit` helper (sources without compinit).
- `TestShellInit_ZshNoCompinitNoErrors`: asserts exit 0, no `command not found`,
  navigation still works. Fails on the old script (exit 127, three compdef
  errors).
- `TestShellInit_ZshDeferredCompletionRegistersAfterCompinit`: sources before
  compinit, runs compinit, fires precmd hooks, asserts completion registers.

Verification: `go test ./internal/shell/...`, the full zsh integration suite,
`just lint` (0 issues), and `just build` all pass. Manual repro of all three init
orderings against a freshly built binary confirmed clean.

### DIFF
--- a/internal/shell/templates/zsh.sh.tmpl
+++ b/internal/shell/templates/zsh.sh.tmpl
@@ -156,8 +156,6 @@ _cgo() {
     fi
   fi
 }
-compdef _cgo cgo
-
 # Run command from campaign root
 # Usage: cr <command> [args...]
 cr() {
@@ -185,8 +183,6 @@ _csw() {
 
   (( ${#campaigns} )) && compadd -a campaigns
 }
-compdef _csw csw
-
 # Quick intent capture
 # Usage: cint "my idea"
 cint() {
@@ -264,4 +260,31 @@ _camp() {
       ;;
   esac
 }
-compdef _camp camp
+# Register completions, resilient to compinit ordering.
+#
+# compdef only exists after `compinit` has run. Some users source camp before
+# compinit (or never run it), which previously printed three
+# "command not found: compdef" errors at every shell startup and skipped
+# completion entirely. Register immediately when compdef is available;
+# otherwise defer to the first prompt so completion still loads once the
+# user's own compinit runs, without erroring when it never does.
+_camp_register_completions() {
+  compdef _cgo cgo
+  compdef _csw csw
+  compdef _camp camp
+}
+
+if (( $+functions[compdef] )); then
+  _camp_register_completions
+  unfunction _camp_register_completions
+elif autoload -Uz add-zsh-hook 2>/dev/null && (( $+functions[add-zsh-hook] )); then
+  _camp_deferred_completions() {
+    if (( $+functions[compdef] )); then
+      _camp_register_completions
+      unfunction _camp_register_completions
+    fi
+    add-zsh-hook -d precmd _camp_deferred_completions
+    unfunction _camp_deferred_completions
+  }
+  add-zsh-hook precmd _camp_deferred_completions
+fi
--- a/internal/shell/templates/zsh.sh.tmpl
+++ b/internal/shell/templates/zsh.sh.tmpl
@@ -293,5 +293,12 @@ elif autoload -Uz add-zsh-hook 2>/dev/null && (( $+functions[add-zsh-hook] )); t
     add-zsh-hook -d precmd _camp_deferred_completions
     unfunction _camp_deferred_completions
   }
-  add-zsh-hook precmd _camp_deferred_completions
+  if ! add-zsh-hook precmd _camp_deferred_completions 2>/dev/null; then
+    # autoload only declared add-zsh-hook; its definition file is not on fpath
+    # (e.g. a stripped fpath), so the precmd registration just failed. Back out
+    # cleanly and stay silent: completion is skipped here, but sourcing the init
+    # must not emit a startup error.
+    unfunction _camp_deferred_completions 2>/dev/null
+    unfunction _camp_register_completions 2>/dev/null
+  fi
 fi
--- a/internal/shell/templates/zsh.sh.tmpl
+++ b/internal/shell/templates/zsh.sh.tmpl
@@ -275,11 +275,18 @@ _camp_register_completions() {
 }
 
 if (( $+functions[compdef] )); then
+  # Sourced again after compinit: drop any deferred hook a pre-compinit
+  # source left queued, so the first prompt stays silent.
+  if (( $+functions[_camp_deferred_completions] )); then
+    (( $+functions[add-zsh-hook] )) && add-zsh-hook -d precmd _camp_deferred_completions
+    precmd_functions=(${precmd_functions:#_camp_deferred_completions})
+    unfunction _camp_deferred_completions
+  fi
   _camp_register_completions
   unfunction _camp_register_completions
 elif autoload -Uz add-zsh-hook 2>/dev/null && (( $+functions[add-zsh-hook] )); then
   _camp_deferred_completions() {
-    if (( $+functions[compdef] )); then
+    if (( $+functions[compdef] )) && (( $+functions[_camp_register_completions] )); then
       _camp_register_completions
       unfunction _camp_register_completions
     fi
--- a/tests/integration/shell_init_helpers_test.go
+++ b/tests/integration/shell_init_helpers_test.go
@@ -195,6 +195,78 @@ func runFishScript(t *testing.T, tc *TestContainer, initScript, stubSetup, testC
 	return output, exitCode
 }
 
+// sourceStderrMarkerBegin and sourceStderrMarkerEnd delimit the captured stderr
+// of the init-sourcing step inside the combined shell output, so the Go side can
+// extract exactly what was written to stderr while sourcing the camp init.
+const (
+	sourceStderrMarkerBegin = "===CAMP_SRC_STDERR_BEGIN==="
+	sourceStderrMarkerEnd   = "===CAMP_SRC_STDERR_END==="
+)
+
+// runShellCleanStartup sources the camp init for a shell in a controlled
+// environment and reports whatever the init wrote to stderr while sourcing.
+//
+// This is the generalized guard against the whole class of "shell command breaks
+// at startup for some users" bugs: a clean shell startup must produce NOTHING on
+// stderr, regardless of the user's completion-system state or rc ordering. The
+// camp stub is installed before sourcing so the init's `command -v camp` guard
+// passes; `preamble` simulates the user's environment (e.g. running compinit, or
+// not). `verify` runs after sourcing and echoes markers to stdout.
+//
+// Returns (sourceStderr, combinedOutput, exitCode). A non-empty sourceStderr
+// means the init emitted errors/warnings at startup.
+func runShellCleanStartup(t *testing.T, tc *TestContainer, shellName, preamble, initScript, stub, verify string) (string, string, int) {
+	t.Helper()
+
+	initPath := "/test/clean-" + shellName + "-init"
+	srcStderrPath := "/test/clean-" + shellName + "-src.stderr"
+	scriptPath := "/test/clean-" + shellName + ".script"
+
+	if err := tc.WriteFile(initPath, initScript); err != nil {
+		t.Fatalf("write init script: %v", err)
+	}
+
+	fullScript := preamble + "\n" +
+		stub + "\n" +
+		"source " + initPath + " 2>" + srcStderrPath + "\n" +
+		"echo '" + sourceStderrMarkerBegin + "'\n" +
+		"cat " + srcStderrPath + "\n" +
+		"echo '" + sourceStderrMarkerEnd + "'\n" +
+		verify + "\n"
+	if err := tc.WriteFile(scriptPath, fullScript); err != nil {
+		t.Fatalf("write test script: %v", err)
+	}
+
+	var argv []string
+	switch shellName {
+	case "bash":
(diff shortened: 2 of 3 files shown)


obey-agent (commented):
**Verdict: Approve**

**Overview**
This PR fixes the zsh `compdef` ordering failure by registering completions immediately when `compinit` has already run and deferring registration through a one-shot `precmd` hook when camp is sourced first. The change is small, integrates with the existing template instead of adding a parallel path, and the new tests cover the previously-missing no-`compinit` and deferred-registration cases.

**Key Findings**
No blocking or non-blocking findings.

**What's Done Well**
- The fix preserves existing behavior for users whose framework already runs `compinit` early.
- The regression tests use the container-backed integration harness and exercise the bug-producing shell order directly.
- The shell template remains understandable: the new registration function centralizes the three `compdef` calls without changing the command wrappers.

**Verification**
- `go test ./internal/shell/...`
- `go test -count=1 -tags=integration ./tests/integration -run 'TestShellInit_Zsh(NoCompinitNoErrors|DeferredCompletionRegistersAfterCompinit|CompdefRegistered)$'`
- `go test -count=1 -tags=integration ./tests/integration -run 'TestShellInit_'`
- `just lint`
- `just build default-build`

**Staff Standard**

internal/shell/templates/zsh.sh.tmpl
```diff
@@ -264,4 +260,31 @@ _camp() {
...
+# user's own compinit runs, without erroring when it never does.
+_camp_register_completions() {
+  compdef _cgo cgo
+  compdef _csw csw
+  compdef _camp camp
+}
+
+if (( $+functions[compdef] )); then
+  _camp_register_completions
+  unfunction _camp_register_completions
```
obey-agent, line 286: This still has a startup-error edge case when the init is sourced more than once around `compinit`: first source before `compinit` registers `_camp_deferred_completions`, then a later source after `compinit` takes this immediate branch and `unfunction`s `_camp_register_completions`, but the old `precmd` hook is still pending. On the first prompt it then prints `_camp_deferred_completions: command not found: _camp_register_completions` plus an `unfunction` error.

I reproduced it with:

```zsh
source camp-init.zsh
autoload -Uz compinit && compinit -u
source camp-init.zsh
for f in $precmd_functions; do (( $+functions[$f] )) && $f; done
```

Please make this idempotent before unfunctioning the helper, e.g. remove any pending `_camp_deferred_completions` hook/function in the immediate path, or have the deferred hook tolerate an already-removed registrar. This shell init should stay silent even when users source it from multiple rc/plugin paths.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** This PR correctly identifies the zsh `compdef` ordering bug and the main fix direction is sound: register immediately after `compinit`, otherwise defer without breaking shell startup. The new integration coverage is strong and the cross-shell clean-startup matrix is a useful guardrail.

**Key Findings:**
- The deferred zsh hook is not idempotent when camp init is sourced before `compinit`, then sourced again after `compinit` before the first prompt. The immediate branch removes `_camp_register_completions` while the stale `precmd` hook is still queued, so the first prompt emits new startup errors. I left an inline comment on `internal/shell/templates/zsh.sh.tmpl` with a repro and suggested fixes.

**What's Done Well:** The tests exercise the exact originally reported no-`compinit` failure mode, preserve navigation behavior, and add a broader clean-startup assertion across bash, zsh, and fish.

**Staff Standard:** Not yet. I would be comfortable with this after the deferred hook is made idempotent for repeated sourcing, because shell init files need to stay silent even under messy rc/plugin ordering.

**Verification Run:** `go test ./internal/shell/...`, `just lint`, `just build-camp`, and `go test -tags integration ./tests/integration -run 'TestShellInit_'` all passed on the PR head.

internal/shell/templates/zsh.sh.tmpl
lancekrogers: Fixed in 6a7cd63. The immediate branch now deregisters and unfunctions any pending `_camp_deferred_completions` (via `add-zsh-hook -d` plus a direct `precmd_functions` filter) before registering, and the deferred hook guards on `_camp_register_completions` existing so a stale hook can never error. Added `TestShellInit_ZshRepeatedSourcingAroundCompinit` covering your repro plus three other source/compinit/prompt orderings; the repro case fails on the previous template and passes now. Full `TestShellInit_` integration suite, `just lint`, and `just build-camp` pass on the new head.

internal/shell/templates/zsh.sh.tmpl
```diff
@@ -264,4 +260,38 @@ _camp() {
...
+elif autoload -Uz add-zsh-hook 2>/dev/null && (( $+functions[add-zsh-hook] )); then
+  _camp_deferred_completions() {
+    if (( $+functions[compdef] )) && (( $+functions[_camp_register_completions] )); then
+      _camp_register_completions
+      unfunction _camp_register_completions
+    fi
+    add-zsh-hook -d precmd _camp_deferred_completions
+    unfunction _camp_deferred_completions
+  }
+  add-zsh-hook precmd _camp_deferred_completions
```
obey-agent: There is one remaining clean-startup edge case in this deferred path: `autoload -Uz add-zsh-hook` does not prove the function file can actually be loaded. In zsh, if a user has a stripped or broken `fpath`, this branch is still entered, but the actual `add-zsh-hook precmd _camp_deferred_completions` call prints `add-zsh-hook: function definition file not found` during shell startup.

Repro against this head:

```zsh
camp(){ :; }
fpath=()
eval "$(camp shell-init zsh)"
```

Please guard the actual `add-zsh-hook` invocation, not only the `autoload` declaration. A small fix would be to redirect/silence the registration call and clean up `_camp_deferred_completions` / `_camp_register_completions` when it fails, plus add a clean-startup case with `fpath=()` or an otherwise missing `add-zsh-hook`. The shell init contract here is that completion setup may be skipped, but sourcing the init must stay silent.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The previous repeated-sourcing/idempotency issue is fixed in `6a7cd63`, and I resolved that review thread. The implementation now handles the important source-before-`compinit`, source-after-`compinit`, and stale-hook paths, and the added integration coverage is strong.

**Key Findings:**
- One clean-startup edge remains in the deferred zsh path: `autoload -Uz add-zsh-hook` does not prove the function can actually be loaded. With a stripped/broken `fpath`, the final `add-zsh-hook precmd _camp_deferred_completions` call still prints `add-zsh-hook: function definition file not found` during shell startup. I left an inline comment on `internal/shell/templates/zsh.sh.tmpl` with a repro and a suggested guard.

**What's Done Well:**
- The original `compdef` ordering bug is fixed without changing the command wrappers.
- The repeated-sourcing regression is covered directly and the previous stale-hook failure no longer reproduces.
- The clean-startup matrix across bash, zsh, and fish is the right testing direction for shell init code.

**Verification Run:**
- `go test ./internal/shell/...`
- `go test -count=1 -tags=integration ./tests/integration -run 'TestShellInit_Zsh(NoCompinitNoErrors|DeferredCompletionRegistersAfterCompinit|RepeatedSourcingAroundCompinit|CompdefRegistered)$'`
- `go test -count=1 -tags=integration ./tests/integration -run 'TestShellInit_CleanStartupAcrossEnvironments$'`
- `go test -count=1 -tags=integration ./tests/integration -run 'TestShellInit_'`
- `just lint`
- `just build-camp`
- Manual zsh repro for the remaining `fpath=()` startup-noise case

**Staff Standard:** Not yet. This is close, but shell init runs on every terminal startup; if completion setup cannot be installed, it should degrade silently rather than emit another zsh startup error.

lancekrogers:
Fixed in `59012c9`. I reproduced the finding first against this head:

```
# no compinit (compdef absent) + stripped fpath
$ camp shell-init zsh > init.zsh
$ zsh -df -c 'fpath=(); source init.zsh; print rc=$?'
init.zsh:308: add-zsh-hook: function definition file not found
rc=1
```

**Root cause.** `autoload -Uz add-zsh-hook` only *declares* the autoload; it does not prove the function file is on `fpath`. So `(( $+functions[add-zsh-hook] ))` is true (the stub exists), the deferred branch is entered, and the `add-zsh-hook precmd _camp_deferred_completions` call is the first thing that actually tries to load the file — which fails on a stripped `fpath` and prints the error at startup.

**Fix.** Guard the registration call itself, not just the `autoload` declaration: silence it and, on failure, back out cleanly so nothing dangles.

```zsh
if ! add-zsh-hook precmd _camp_deferred_completions 2>/dev/null; then
  unfunction _camp_deferred_completions 2>/dev/null
  unfunction _camp_register_completions 2>/dev/null
fi
```

The stripped-`fpath` startup is now silent (`rc=0`, no stderr) and both helper functions are cleaned up. The healthy deferred path is unchanged: the hook still registers and fires after the user's own compinit.

**Test.** Added a `zsh_stripped_fpath_no_compinit` case (`preamble: emulate -R zsh` then `fpath=()`) to `TestShellInit_CleanStartupAcrossEnvironments`, which asserts the source writes nothing to stderr, exits 0, and still defines the `camp`/`cgo` wrappers. Before the fix this case fails on the `add-zsh-hook: function definition file not found` line; after, it passes.

**Verification.**
- `go test ./internal/shell/...` green.
- `go test -count=1 -tags=integration ./tests/integration -run 'TestShellInit_(CleanStartupAcrossEnvironments|ZshNoCompinitNoErrors|ZshDeferredCompletionRegistersAfterCompinit|ZshRepeatedSourcingAroundCompinit|ZshCompdefRegistered)$'` green, including the new `zsh_stripped_fpath_no_compinit` subtest and all prior zsh cases.
- `just lint` clean; `just build-camp` successful.
- Manual repro confirmed: `rc=1` with the stderr line before the change, `rc=0` and silent after.

obey-agent (commented):
Verdict: Approve

I did not find blocking issues. The zsh init change correctly defers Camp completion registration until after `compinit`/`add-zsh-hook` is available, and the focused coverage exercises no-compinit, post-compinit, and clean startup paths.

Tests run:

- `go test ./internal/shell ./tests/integration -run 'TestShellInit_Zsh|TestShellInit_CleanStartupAcrossEnvironments' -tags integration`

Coordination note: PR #319 also edits the zsh template. If it lands after this, preserve the deferred completion registration and add `cnote` there.

obey-agent (commented):
## Verdict: Approve

### Overview
A focused, correct fix for the `compdef`-before-`compinit` startup errors. The root-cause analysis is accurate (framework users run compinit early so they never saw it; bare or pre-compinit users got three `command not found: compdef` lines at every startup and no completion), and the fix handles every init ordering rather than only the reported one.

### Key Findings
No blocking issues. I traced all init orderings against the new registration block:
- compinit-first: registers immediately, unchanged for oh-my-zsh / prezto users.
- pre-compinit with add-zsh-hook autoloadable: defers to a one-shot precmd hook that registers once the user's compinit runs, then removes itself.
- stripped fpath (add-zsh-hook declared but its definition file unfindable): the `add-zsh-hook precmd` call fails, the cleanup branch unfunctions both helpers, and startup stays silent.
- repeated sourcing around compinit: the stale deferred hook is dropped (add-zsh-hook -d, the `precmd_functions` filter, and unfunction) before re-registering, so the first prompt does not error.

The one-shot precmd design (register-or-give-up at the first prompt) is the right tradeoff over a permanent retry hook that runs on every prompt forever.

### What's Done Well
The tests are the strongest part. `TestShellInit_CleanStartupAcrossEnvironments` generalizes the bug into a class ("a clean startup must write nothing to stderr") across bash, zsh, and fish, including the stripped-fpath case; the repeated-sourcing matrix captures the exact reported regression. Verified locally: `go vet -tags integration ./tests/integration/` and the shell build both pass.

### Staff Standard
Yes. Careful root-cause work, a fix that covers the whole class rather than the single report, and tests that would have caught the original bug. Worth keeping as a reference for shell-init robustness.

PR #322, opened 2026-06-12
test: isolate integration config state

lancekrogers:
## Summary
- clear the current Obey config home during integration container resets so pooled containers cannot leak registry or settings state
- update the camp settings TUI navigation test for the local settings menu entry
- align the outside-campaign project assertion with the empty-registry path

## Validation
- `CAMP_TEST_POOL_SIZE=1 just test integration-run 'Test(CampSettings_CampaignsDirEditAndClear|Project_NotInCampaign)'`
- `just test integration`
- `just test unit`
- `just lint`
- `just build-camp`

### DIFF
--- a/tests/integration/helpers_container.go
+++ b/tests/integration/helpers_container.go
@@ -19,10 +19,12 @@ import (
 // begins — required for consistency on macOS/Colima where Docker exec runs through
 // a virtualization layer (overlayfs in a Linux VM).
 func (tc *TestContainer) Reset() error {
-	// Remove all test artifacts and recreate clean directories
+	// Remove all test artifacts and recreate clean directories. Include both
+	// current and legacy config homes so registry/global settings never leak
+	// between tests that share a pooled container.
 	exitCode, _, err := tc.container.Exec(tc.ctx, []string{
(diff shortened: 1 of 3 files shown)


obey-agent (commented):
**Verdict:** Approve

**Overview:** This is a focused test-harness fix that addresses the integration failures at their shared isolation boundary. Clearing the current global config home in `Reset()` removes order-dependent registry/settings leakage, and the two test updates now match the current settings menu and resolver behavior.

**Key Findings:**
- No issues requiring changes.
- The reset change matches the active config and registry paths under `~/.obey/campaign`.
- The settings TUI navigation change is consistent with the current top-level menu order: Global, Local, Exit.
- The project assertion now covers the correct empty-registry branch after per-test isolation is restored.

**What's Done Well:** The patch keeps the behavioral surface narrow, centralizes cleanup in the shared container reset path, and removes a one-off cleanup from the individual settings test. The validation set is strong for this change: targeted integration, full integration, unit tests, lint, and build all passed.

**Staff Standard:** Yes. I would be comfortable submitting this as production-quality test infrastructure work.

PR #325, opened 2026-06-16
fix: stabilize camp gate follow-ups

lancekrogers:
## Summary

- update the `project add` integration assertion to match the current noninteractive campaign-selection contract
- add containerized integration coverage for `.githooks/pre-push` so hook-local Git environment variables do not leak into `just gate-push`
- make `pkg/commitkit` stale-lock tests deterministic by aging synthetic lock files past the production stale-lock age floor

## Why

The RC release flow exposed two local gate problems after the pre-push hook fix landed on `main`:

- `just test all` was failing `TestProject_NotInCampaign` because the test expected the empty-registry error even though it creates a registered campaign. With a registry present and no current campaign in noninteractive mode, the command correctly requires `--campaign`.
- The release-blocking hook environment issue needed reviewable regression coverage.

The hook regression now lives in `tests/integration` because it creates Git repos, writes an executable `just` fixture, and runs the real hook. The integration harness keeps that filesystem mutation inside an isolated container instead of the host filesystem.

While validating this branch, `gate-push` also exposed a flaky stale-lock fixture. The production lock cleaner intentionally waits for a minimum lock age before removing a lock, so tests that create a brand-new synthetic lock must age it before expecting cleanup.

## Validation

- `just test integration-run TestPrePushHook_ClearsGitHookEnvironment`
- `go test ./internal/buildutil ./pkg/commitkit -count=1`
- `go test ./pkg/commitkit -run 'TestStageAll/succeeds_with_stale_lock_file|TestCommit/succeeds_with_stale_lock_file' -count=1`
- `just test integration-run TestProject_NotInCampaign`
- `just test all`
- `just gate-push`
- push pre-push hook: `just gate-push` passed


obey-agent (commented):
Verdict: Approve

Overview:
I reviewed this as a standalone PR against `main`, not as the author. This is a narrow, test-only stabilization PR that addresses the observed RC gate failures without expanding production behavior.

Key Findings:
- No blocking findings.
- The hook regression is covered in the integration suite, which is the right boundary because it creates repos, writes an executable fixture, and runs the real pre-push hook inside an isolated container.
- The stale-lock fixture change is correct: production lock cleanup has a minimum age floor, so tests expecting stale cleanup must age synthetic lock files before invoking the code.
- The `project add` assertion now matches the current noninteractive campaign-selection contract for the case where registered campaigns exist, but no current or explicit campaign was selected.

What's Done Well:
- The test exercises the real `.githooks/pre-push` content instead of duplicating hook logic.
- The fake `just` fixture records args and env for Go-side assertions, keeping the test readable.
- Full local validation passed, including `just test all` and the actual pre-push `gate-push` during branch push.

Staff Standard:
I would be comfortable merging this. I would be proud to submit it as a follow-up because it tightens the gate without adding production risk, and it keeps filesystem mutation in the containerized integration harness.

obey-agent (approved):
Formal approval: no blocking findings. This is good enough to merge; see the review body I just posted for the detailed reasoning and residual risk read.

PR #326, opened 2026-06-16
fix(intent): make notes.go and rename.go writes atomic (C-2 follow-up)

obey-agent:
## What

Routes the 5 remaining non-atomic `os.WriteFile` sites in the intent package through `fsutil.WriteFileAtomically` (temp-file + fsync + rename), closing finding **C-2** from the camp #324 (camp-hardening) review.

- `internal/intent/rename.go` — in-place rewrite (slug unchanged) + slug-change move
- `internal/intent/notes.go` — `ArchiveNote`, `MoveIntentToNote`, `MoveNoteToStatus`

These landed after the camp-hardening festival's atomic-write sweep (R11/N-21) was planned (the #319/#321 intent-notes feature), so they were outside its scope and never converted. This restores the festival's "zero `os.WriteFile` on intent-state paths" invariant for the full current intent surface.

## Test

New `internal/intent/atomic_write_test.go` exercises both the rename and note-move paths and asserts the file round-trips and no `.tmp-*` litter remains after the atomic write. `just gate-fast` green (2684 tests, lint clean including the fmt.Errorf and host-fs-test ratchets).

## Scope note

A broader grep shows additional production `os.WriteFile` sites elsewhere in the package (`migration.go`, `migration_filesystem.go`, `promote/promote.go`, `feedback/tracker.go`, `feedback/builder.go`). Those were **not** part of C-2 and are intentionally left out of this PR to avoid over-scoping; capture as a separate intent if the same atomicity guarantee is wanted there.

---

Part of the intent-backlog cleanup. Base branch is the `intent-cleanup` umbrella. Closes intent `make-intent-notesgo-and-renamego-20260615-203000`. `[WI-ee98f2]`

### DIFF
--- a/internal/intent/atomic_write_test.go
+++ b/internal/intent/atomic_write_test.go
@@ -1,0 +1,91 @@
+package intent
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestAtomicWrites_RenameRoundTripsWithoutTempLitter(t *testing.T) {
+	ctx := context.Background()
+	tmp := t.TempDir()
+	intentsDir := filepath.Join(tmp, "intents")
+	svc := NewIntentService(tmp, intentsDir)
+
+	created, err := svc.CreateDirect(ctx, CreateOptions{
+		Title:     "atomic rename target",
+		Type:      TypeIdea,
+		Timestamp: time.Date(2026, 2, 1, 9, 0, 0, 0, time.UTC),
+	})
+	if err != nil {
+		t.Fatalf("CreateDirect: %v", err)
+	}
+
+	renamed, err := svc.Rename(ctx, created.ID, "atomic rename target updated")
+	if err != nil {
+		t.Fatalf("Rename: %v", err)
+	}
+	if _, err := os.Stat(renamed.Path); err != nil {
+		t.Errorf("renamed file missing: %v", err)
+	}
+
+	got, err := svc.Get(ctx, created.ID)
+	if err != nil {
+		t.Fatalf("Get after rename: %v", err)
+	}
+	if got.Title != "atomic rename target updated" {
+		t.Errorf("title = %q after rename", got.Title)
+	}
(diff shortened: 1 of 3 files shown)


obey-agent (commented):
**Verdict: Approve**

**Overview:** Focused, correct follow-up that swaps the three remaining `os.WriteFile` calls in `notes.go` and `rename.go` for `fsutil.WriteFileAtomically`, closing the partial-write window when an intent is renamed, archived, or promoted. The change is mechanical and a new test file guards the rename and note-promotion paths plus the absence of `.tmp-` litter.

**What's done well:**
- Right mechanism reused, not reinvented: same-directory temp + chmod + fsync + atomic rename, with existing-mode preservation, so no permission regression on the in-place rewrite path.
- The litter assertion (`strings.Contains(d.Name(), ".tmp-")`) matches the helper's real temp pattern (`base+".tmp-*"`), so it actually catches leaked temps rather than passing vacuously.
- `os.MkdirAll(filepath.Dir(newPath))` still runs before each atomic write, so the temp lands in an existing directory and the rename stays same-device (no EXDEV).

**Verified by trace:**
- All converted sites pass `(path, data, 0644)`, matching the helper signature. New-path writes hit a fresh file (defaultMode 0644); the in-place rewrite preserves the existing 0644. No behavior change beyond atomicity.
- New tests cover Rename (slug-change branch) and MoveNoteToStatus. ArchiveNote, MoveIntentToNote, and the in-place Rename branch lean on existing operation tests plus the shared helper, which is acceptable since the write mechanism is transparent to those tests.

**Out of scope, not a blocker on this PR:** the atomicity migration is not yet complete across `internal/intent`. Still on plain `os.WriteFile` in production paths: `feedback/builder.go` (2 sites), `feedback/tracker.go` (1), `promote/promote.go` (1), plus the one-shot writers `migration.go` / `migration_filesystem.go`. Worth a tracked workitem to finish the series so the whole package is uniformly crash-safe.

**Staff standard:** Yes. Tight scope, correct mechanism, and a test that fails for the right reason. I would submit this as my own work.

PR #334, opened 2026-06-16
fix(dungeon): scope and batch crawl link rewriting to stop the freeze

lancekrogers:
## Problem

`camp dungeon crawl` freezes on large workspaces and has to be killed. Reported while triaging `workflow/code_reviews` (24 items) and again on `workflow/bugs` (9 items).

The triage/inner crawl moves each item **and rewrites markdown links the moment "Move" is chosen** — so the hang is in the per-move link rewrite, not the final auto-commit. `mdlinks.RewriteForMove` called `collectMDFiles(campaignRoot)`, which walked the **entire campaign monorepo on every single move**:

- Measured **~6s per move** (0.9s walk + 5s read/regex of **27,434 `.md` files / 103 MB**), single-threaded, no progress output.
- A 24-item crawl spends **~2m22s** in blocked IO → reads as a freeze.

It also descended into all 30 `projects/*` git submodules and `worktrees/`. Those are independent repos, so the campaign-root auto-commit cannot stage their files (`git ls-files` at the campaign root never lists gitlink contents). Rewriting links inside them either errored on `git add` or left submodule working trees dirty — a correctness bug, not just perf.

## Fix

1. **Scope the walk** — `collectMDFiles` no longer descends into nested git repositories (submodules, worktrees). Link rewriting only touches files the campaign repo owns. Cuts the scan from ~27.4k to ~17.7k files (walk 914ms → 179ms).
2. **Batch the rewrite** — `RewriteForMove` is split into:
   - `RewriteMovedInternalLinks` — rewrites only the moved subtree's own links (cheap, per move).
   - `RewriteExternalLinksForMoves` — a single tree scan handling many moves at once, including links between two moved items.
   - `RewriteForMove` is kept as a composition, so single-move callers are unchanged.
3. **Defer external rewriting in the crawl** — `Service.BeginLinkBatch` records moves during the loop; a single `FlushLinkRewrites` runs after all moves, before the auto-commit. Matches the intended "decide → move → one commit" flow. Single-shot `camp dungeon move` / `promote` / `workitem` behavior is unchanged (guarded by `TestService_MoveToStatus_ExternalLinkRewriteFlowsIntoCommit`).
4. **O(files), not O(files × moves)** — the external rewrite runs the markdown regexes **once per file** and matches each link target against every move with cheap path arithmetic (a naive batch would re-scan every file once per move: ~9× slower for a 9-item crawl). The scan is parallelized across `runtime.NumCPU` workers over disjoint file chunks (no shared mutable state; results merged and sorted, so output is deterministic; verified under `go test -race`).

## Results (live campaign tree, ~17.7k md files)

| crawl | before | after |
|-------|--------|-------|
| 9 moves  | ~30–55s | **~1.0s** |
| 24 moves | ~2m22s  | **~1–2s** |

## Tests

- `mdlinks`: nested-git-repo skip, multi-move single scan, cross-moved-item links, internal-only pass leaves external referrers untouched.
- `dungeon`: `BeginLinkBatch` defers external rewrites until `FlushLinkRewrites` (referrers not rewritten mid-crawl; recorded only after flush).

## Verification

- `gofmt -l` clean, `go vet` clean
- `just build-camp` ✓
- `just lint` — 0 issues; no new host-fs-test violators; no `fmt.Errorf` regressions
- `go test ./...` passes; `go test -race ./internal/mdlinks/... ./internal/dungeon/...` clean

## Review follow-up (commit 3f5b740)

Addresses the "Request Changes" finding: once `BeginLinkBatch` was active, several non-abort error returns (a `RunTriageCrawl` error, the `svc.ListItems` call, a `RunCrawl` error) bailed out before `FlushLinkRewrites`, leaving already-moved items on disk with stale external links — a regression on error paths.

Fix: extracted `finalizeCrawl`, through which **every** exit path (success, abort, and each mid-crawl error) now routes, so the flush is unconditional by construction. A flush failure is surfaced (folded into the return when there is no prior error, else reported on stderr alongside it). Added `TestFinalizeCrawl_FlushesPendingRewritesOnSessionError`.

## Review follow-up (commit dfe4146)

Addresses the chained-move finding: `RewriteExternalLinksForMoves` returned on the first matching move, so a batch containing a chain (parent -> dungeon -> dated status) rewrote an external referrer to the intermediate location instead of the final destination. Now every matching move is applied in chronological order (one ordered pass = sequential-per-move equivalence); independent moves are unaffected. Regression tests added at both the mdlinks level (the exact repro) and the Service level (batched `MoveToDungeon` then `MoveToStatus`).

### DIFF
--- a/cmd/camp/dungeon/crawl.go
+++ b/cmd/camp/dungeon/crawl.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"sort"
 
 	intdungeon "github.com/Obedience-Corp/camp/internal/dungeon"
@@ -65,7 +66,9 @@ func runDungeonCrawl(cmd *cobra.Command, args []string) error {
 	svc := intdungeon.NewService(cmdCtx.CampaignRoot, cmdCtx.Dungeon.DungeonPath)
 	// Defer external markdown-link rewriting so the campaign tree is scanned
 	// once after all moves, not once per move (which is O(moves x workspace)
-	// and dominates wall-clock time on large workspaces).
+	// and dominates wall-clock time on large workspaces). finalizeCrawl flushes
+	// on every exit path, so a mid-crawl error never leaves moved items on disk
+	// with stale links.
 	svc.BeginLinkBatch()
 	relParent := RelFromRoot(cmdCtx.CampaignRoot, cmdCtx.Dungeon.ParentPath)
 	relDungeon := RelFromRoot(cmdCtx.CampaignRoot, cmdCtx.Dungeon.DungeonPath)
@@ -91,9 +94,9 @@ func runDungeonCrawl(cmd *cobra.Command, args []string) error {
 
 	// Run triage crawl if needed
 	if runTriage {
-		parentItems, err := svc.ListParentItems(ctx, cmdCtx.Dungeon.ParentPath)
-		if err != nil {
-			return camperrors.Wrap(err, "listing parent items")
+		parentItems, listErr := svc.ListParentItems(ctx, cmdCtx.Dungeon.ParentPath)
+		if listErr != nil {
+			return finalizeCrawl(ctx, cmdCtx, svc, triageSummary, innerSummary, aborted, camperrors.Wrap(listErr, "listing parent items"))
 		}
 		if len(parentItems) > 0 {
 			fmt.Printf(
@@ -108,7 +111,7 @@ func runDungeonCrawl(cmd *cobra.Command, args []string) error {
 				if errors.Is(err, intdungeon.ErrCrawlAborted) {
 					aborted = true
 				} else {
-					return camperrors.Wrap(err, "triage crawl failed")
+					return finalizeCrawl(ctx, cmdCtx, svc, triageSummary, innerSummary, aborted, camperrors.Wrap(err, "triage crawl failed"))
 				}
 			}
 		} else {
@@ -116,21 +119,11 @@ func runDungeonCrawl(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	// flushLinkRewrites applies the deferred external-link rewrites for every
-	// item moved so far in a single campaign-tree scan.
-	flushLinkRewrites := func() error {
-		if (triageSummary != nil && triageSummary.HasMoves()) ||
-			(innerSummary != nil && innerSummary.HasMoves()) {
-			fmt.Printf("%s Updating markdown cross-references...\n", ui.InfoIcon())
-		}
-		return svc.FlushLinkRewrites(ctx)
-	}
-
 	// Run inner crawl if the triage step did not abort.
 	if !aborted && runInner {
-		dungeonItems, err := svc.ListItems(ctx)
-		if err != nil {
-			return camperrors.Wrap(err, "listing dungeon items")
+		dungeonItems, listErr := svc.ListItems(ctx)
+		if listErr != nil {
+			return finalizeCrawl(ctx, cmdCtx, svc, triageSummary, innerSummary, aborted, camperrors.Wrap(listErr, "listing dungeon items"))
 		}
 		if len(dungeonItems) > 0 {
 			if runTriage {
@@ -147,7 +140,7 @@ func runDungeonCrawl(cmd *cobra.Command, args []string) error {
 				if errors.Is(err, intdungeon.ErrCrawlAborted) {
 					aborted = true
 				} else {
-					return camperrors.Wrap(err, "inner crawl failed")
+					return finalizeCrawl(ctx, cmdCtx, svc, triageSummary, innerSummary, aborted, camperrors.Wrap(err, "inner crawl failed"))
 				}
 			}
 		} else {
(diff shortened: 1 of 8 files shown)


cmd/camp/dungeon/crawl.go
```diff
@@ -63,6 +63,10 @@ func runDungeonCrawl(cmd *cobra.Command, args []string) error {
 	}
 
 	svc := intdungeon.NewService(cmdCtx.CampaignRoot, cmdCtx.Dungeon.DungeonPath)
+	// Defer external markdown-link rewriting so the campaign tree is scanned
+	// once after all moves, not once per move (which is O(moves x workspace)
+	// and dominates wall-clock time on large workspaces).
+	svc.BeginLinkBatch()
```
obey-agent, line 72: Starting batch mode here changes the command invariant: every path after a successful move needs to either flush the pending external rewrites or explicitly report that they were not applied. Today the normal and abort paths flush, but several later error exits return without flushing. For example, after a successful triage move, a later `RunTriageCrawl` non-abort error, the `svc.ListItems(ctx)` call before inner crawl, or a `RunCrawl` non-abort error returns from `runDungeonCrawl` before `FlushLinkRewrites` runs. That leaves moved items on disk while external markdown links still point at the old locations; before this PR each move rewrote those links immediately.

Please make flushing part of the error cleanup path as well, likely with a small helper/defer that flushes once when `triageSummary` or `innerSummary` has moves, and surface any flush failure alongside the original error. A regression test with one successful batched move followed by a forced later crawl/listing error would lock this down.

obey-agent (commented):
Verdict: Request Changes

Overview: This PR tackles the right problem and the core approach is solid: scope markdown scans away from nested repos and batch external rewrites so crawl no longer performs a campaign-wide scan per move. The mdlinks split is coherent, preserves single-move behavior, and the added tests cover the main batching and nested-repo cases.

Key Findings:
- Batched crawl rewrites are not flushed on several error exits after successful moves. I left an inline comment on `cmd/camp/dungeon/crawl.go`: once `BeginLinkBatch` is active, later non-abort errors can return before `FlushLinkRewrites`, leaving moved items on disk with stale external markdown links.

What's Done Well:
- `RewriteMovedInternalLinks` and `RewriteExternalLinksForMoves` separate the cheap per-move work from the expensive campaign scan cleanly.
- `collectMDFiles` now skips nested git repos, which matches campaign-root commit ownership and avoids dirtying submodules/worktrees.
- Existing single-shot move/promote/workitem behavior is preserved through `rewriteLinksAfterMove` and covered by tests.

Verification:
- `go test ./internal/mdlinks ./internal/dungeon` passed.
- `gofmt -l` on changed Go files was clean.
- `diff --check` was clean.
- `just build-camp` passed.
- `go test ./...` passed.
- `just lint` passed with 0 issues and no ratchet regressions.
- `just vet` passed.

Staff Standard: I would be proud of the design direction and the performance fix, but I would not merge until the command-level error cleanup flushes or reports pending external rewrites consistently.

internal/mdlinks/rewrite.go
```diff
@@ -187,27 +192,30 @@ func rewriteExternalLinksToMoved(content []byte, fileDir, oldPath, newPath strin
...
-			return target, false
+			dest := rel
+			if hadBrackets {
+				dest = "<" + rel + ">"
+			}
+			newTarget := dest + anchor
+			if newTarget == target {
+				return target, false
+			}
+			return newTarget, true
```
obey-agent: This returns after the first matching move, which regresses chained moves within one crawl. In the common auto-detect flow an item can move from the parent into `dungeon/` during triage and then from `dungeon/` into a dated status bucket during the inner crawl before the single batch flush runs. A referrer to the original parent path matches the first move and is rewritten to the intermediate `dungeon/item.md`, never to the final `dungeon/completed/<date>/item.md` path.

I confirmed this with a temporary repro test: moves `{bugs/issue.md -> dungeon/issue.md, dungeon/issue.md -> dungeon/completed/2026-06-16/issue.md}` rewrote `[issue](../bugs/issue.md)` to `[issue](../dungeon/issue.md)` instead of the final dated path. The previous implementation applied each move sequentially to the file content, so it handled this case. Please either collapse chained moves to their final destination before scanning, or keep applying move matches until the target no longer maps through the batch, and add a regression test for parent -> dungeon -> status in one batch.

obey-agent (commented):
Verdict: Request Changes

Overview: The new commit improves the intended performance shape by running markdown regexes once per file and parallelizing over disjoint chunks. The concurrency structure is straightforward and race tests for the touched packages pass, but the optimized rewrite path lost an important semantic from the earlier batch implementation.

Key Findings:
- New finding: chained moves in one crawl now rewrite external links to the intermediate path instead of the final destination. I left an inline comment on `internal/mdlinks/rewrite.go`; a parent -> dungeon -> dated status sequence rewrites `bugs/issue.md` links to `dungeon/issue.md` rather than `dungeon/completed/<date>/issue.md`.
- Previous finding still applies: batched crawl rewrites are not flushed on several non-abort error exits after successful moves. The earlier inline comment on `cmd/camp/dungeon/crawl.go` is still unresolved in this head.

What's Done Well:
- The O(files) direction is the right target for large campaign trees.
- The parallel worker setup uses disjoint file chunks and merges results deterministically.
- Existing focused tests, race tests, full tests, build, vet, and lint all pass on this head.

Verification:
- `go test ./internal/mdlinks ./internal/dungeon` passed.
- `go test -race ./internal/mdlinks ./internal/dungeon` passed.
- `gofmt -l` on changed files was clean.
- `git diff --check` was clean.
- `just build-camp` passed.
- `go test ./...` passed.
- `just lint` passed with 0 issues and no ratchet regressions.
- `just vet` passed.
- Temporary repro test failed as expected for the chained-move case: got `../dungeon/issue.md`, wanted `../dungeon/completed/2026-06-16/issue.md`.

Staff Standard: I would not merge yet. The performance work is close, but both the chained-move correctness regression and the pending-flush error path need to be fixed before this is production-ready.

internal/mdlinks/rewrite.go
obey-agent: This chained-move regression is still present on the current head. In auto-detect crawl, an item can move from the parent into `dungeon/` during triage and then from `dungeon/` into a dated status bucket during inner crawl before the single batch flush. Because this helper returns on the first matching move, a referrer to the original parent path is rewritten to the intermediate `dungeon/item.md` path and never reaches the final `dungeon/completed/<date>/item.md` destination.

I reran the repro against `3f5b740`: moves `{bugs/issue.md -> dungeon/issue.md, dungeon/issue.md -> dungeon/completed/2026-06-16/issue.md}` still rewrite `[issue](../bugs/issue.md)` to `[issue](../dungeon/issue.md)`. Please collapse chained moves before scanning, or keep resolving a matched target through the remaining moves until it reaches the final destination, with a regression test for parent -> dungeon -> status in one batch.

obey-agent (commented):
Verdict: Request Changes

Overview: The new `finalizeCrawl` change addresses the earlier pending-flush cleanup issue: the session error, abort, and success paths now all flush deferred external rewrites before returning or committing. The remaining blocker is the O(files) rewrite optimization’s chained-move semantics.

Key Findings:
- Still blocking: chained moves in one crawl rewrite external links to the intermediate path instead of the final destination. I left a current-head inline comment on `internal/mdlinks/rewrite.go`; parent -> dungeon -> dated status still produces links to `dungeon/item.md` rather than `dungeon/completed/<date>/item.md`.
- Resolved: the previous cleanup-path issue in `cmd/camp/dungeon/crawl.go` is fixed by `finalizeCrawl`, and `TestFinalizeCrawl_FlushesPendingRewritesOnSessionError` covers the session-error case.

What's Done Well:
- The cleanup refactor is clearer than scattered flush calls and makes the command-level invariant easier to audit.
- The focused and full gate suite is clean on this head.
- The performance direction remains sound once chained move mapping is made semantically equivalent to sequential rewrites.

Verification:
- `go test ./cmd/camp/dungeon ./internal/mdlinks ./internal/dungeon` passed.
- `go test -race ./cmd/camp/dungeon ./internal/mdlinks ./internal/dungeon` passed.
- `gofmt -l` on changed files was clean.
- `git diff --check` was clean.
- `just build-camp` passed.
- `go test ./...` passed.
- `just lint` passed with 0 issues and no ratchet regressions.
- `just vet` passed.
- Temporary chained-move repro still failed on `3f5b740`: got `../dungeon/issue.md`, wanted `../dungeon/completed/2026-06-16/issue.md`.

Staff Standard: I would not merge yet. The error-path regression is fixed, but the chained-move correctness bug still needs a fix and regression test before this is production-ready.

obey-agent (commented):
**Verdict: Approve**

Re-reviewed current head `dfe4146315e3c566216d070442d30bfb0b230da1`. The previous blockers are addressed: crawl finalization now flushes pending external rewrites on success, abort, and mid-crawl error paths, and batched external link rewriting now applies moves in chronological order so chained moves resolve to their final destination.

**Key Findings**

- No remaining blocking findings. The current mdlinks implementation preserves the intended O(files) scan shape while restoring sequential-per-move semantics for parent -> dungeon -> dated-status chains.
- The new regressions cover the earlier failures at the right levels: `TestFinalizeCrawl_FlushesPendingRewritesOnSessionError`, `TestRewriteExternalLinksForMoves_ChainedMovesResolveToFinalDestination`, and `TestService_BeginLinkBatch_ChainedMovesResolveToFinalDestination`.

**What's Done Well**

- The split between immediate moved-subtree rewrites and deferred external rewrites is clear and keeps single-shot move behavior intact.
- `collectMDFiles` now avoids nested git repositories, which matches campaign-root ownership and avoids dirtying submodules/worktrees.
- The concurrent external scan uses disjoint file chunks and deterministic result merging; race tests pass on the touched packages.

**Verification**

- `go test ./cmd/camp/dungeon ./internal/mdlinks ./internal/dungeon` passed.
- `go test -race ./cmd/camp/dungeon ./internal/mdlinks ./internal/dungeon` passed.
- `gofmt -l` on changed files was clean.
- `just build-camp` passed.
- `go test ./...` passed.
- `just lint` passed with 0 issues, no new host-fs-test violators, and no `fmt.Errorf` ratchet regressions.
- `just vet` passed.

**Staff Standard**

Yes. I would be comfortable submitting this: the performance fix is scoped to the right ownership boundary, the batching semantics now match the old sequential behavior, and the command-level cleanup invariant is covered.

PR #336, opened 2026-06-17
feat(camp): org/tags/status registry layer + camp festivals aggregator

lancekrogers:
## Summary

Adds an organizational layer to the camp campaign registry: **org**, **tags**, and **status** axes, the CLI to manage and filter on them, and `camp festivals` for a cross-campaign festival view. Filesystem + git only, no database. State lives in `~/.obey/campaign/registry.json` with atomic writes.

Implements design package `workflow/design/campaign-org-grouping-and-user-20260611-230319/` (v1 scope).

## The three axes

- **org** — single membership; every campaign is in exactly one org (default `default`).
- **tags** — a single global pool of labels; orthogonal to org.
- **status** — lifecycle, one of `active` / `inactive` / `reference`. Default `camp list` shows only `active`. The command is `camp lifecycle`, **not** `camp status` (the git-status wrapper is untouched).

A campaign at its defaults stores no extra keys, so existing registries are byte-identical until something is organized.

## Commands

- `camp org add/remove/rename/list/show` + bare `camp org` (single-membership; add reassigns; rename is atomic; renaming the fallback org moves the fallback).
- `camp tag add/rm/list` (set semantics over a global pool).
- `camp lifecycle set/list` (fixed status set, validated).
- `camp list --org/--tag/--status/--all/--group/--no-group` (active-only default, org grouping).
- `camp festivals --org/--tag` — composes `fest list --json` per campaign and aggregates; `fest list` is unchanged.

All commands support `--json`.

## Notable implementation decisions

- **Read-side normalization** maps zero values to defaults in memory and never rewrites the file on read; `SaveRegistry` denormalizes fallback-equal values so default campaigns stay key-free on disk (proven byte-identical by a golden test, #28).
- **Movable fallback**: renaming the `default` org stores a registry-level `default_org` and rewires normalization, so unset campaigns map to the new fallback.
- **`camp festivals` uses `fest list --json --progress`** to obtain `tasks.{completed,total}` for the `progress: {completed,total}` shape; the top-level `total` scalar in fest's output is skipped. Sequential one-invocation-per-campaign fan-out (no silent cap); campaigns without a `festivals/` workspace contribute nothing; a missing/failing `fest` yields a clear error naming the campaign.
- Shared `config.ValidateName` (org/tag name rules) and `config.ValidateStatus` (fixed status set) as single sources of truth.

## Migration

Forward-compatible by construction: no migration step, prompt, or version bump. Old registries normalize to defaults on load; an old binary that rewrites the file simply drops the new keys (documented, asserted).

## Testing

- Design test-plan cases #1-#51 implemented across the config, org, tag, lifecycle, list, and festivals packages (negative/invariant cases first).
- `#28` golden proves single-org all-active `camp list` output is byte-identical to pre-feature.
- `#46` holds by construction: only a camp command is added; `fest list` source is untouched.
- Full `go test -race ./...`: 89 packages pass. `just lint`: 0 issues; host-FS and fmt.Errorf ratchets clean. Stable + dev builds green.

## Docs

README "Organizing Campaigns" section + regenerated `docs/cli-reference`. Superseded intent `camp-registry-extension-20260428-231046` is archived and remains referenced from design doc 99.

Executed via the Festival methodology (`campaign-org-grouping-CO0001`); every commit carries the `FE-CO0001` tag.


cmd/camp/tag/tag.go
```diff
@@ -0,0 +1,243 @@
...
+		result.Tags = next
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return renderTagChange(cmd.OutOrStdout(), result, asJSON, true)
+}
+
+func runTagRm(cmd *cobra.Command, args []string) error {
```
obey-agent, line 99: `camp tag rm` accepts the tag arguments without running them through `config.ValidateName`, while `camp tag add` validates the same name grammar and the sequence goal calls for name validation on the tag command group. This means invalid labels such as `Bad-Tag` are silently treated as absent no-ops on remove, which weakens the CLI contract and hides user mistakes. Please validate every tag in `runTagRm` before entering `UpdateRegistry`, and add a negative no-write test mirroring `TestTagAdd_InvalidTag_NoWrite`.

cmd/camp/festivals/festivals.go
```diff
@@ -0,0 +1,256 @@
...
+	allCampaigns, _ := cmd.Flags().GetBool("all-campaigns")
+	asJSON, _ := cmd.Flags().GetBool("json")
+
+	reg, err := config.LoadRegistry(ctx)
+	if err != nil {
+		return camperrors.Wrap(err, "failed to load registry")
+	}
+	campaigns := selectCampaigns(reg, org, tags, allCampaigns)
+
+	festPath, err := fest.FindFestCLI()
```
obey-agent: This resolves the `fest` binary before checking whether there is any matching campaign workspace to query. As written, `camp festivals --org <empty-set>` or a registry where the selected campaigns have no `festivals/` directory can fail with “needs the fest CLI” even though no `fest list` invocation is required and the command should simply render an empty result. That makes the new camp -> fest coupling leak into no-op reads. Please filter to campaigns with a `festivals/` workspace first, return the empty JSON/human result when there is nothing to query, and only require `fest` when at least one `fest list` call will actually run.

obey-agent (commented):
**Verdict: Request Changes**

This PR implements the right feature shape and the overall architecture is mostly aligned with the design: registry normalization is read-side, writes go through `UpdateRegistry`, and the new command surfaces are covered by focused tests. I would not merge it yet because the local quality gate is not green and two changed-code paths need correction.

**Key Findings**
- `camp tag rm` does not validate tag names before mutation/no-op handling, so invalid labels are silently accepted as absent tags. I left an inline comment on `cmd/camp/tag/tag.go`.
- `camp festivals` requires the `fest` binary before checking whether any selected campaign actually has a `festivals/` workspace to query. That makes no-op reads fail unnecessarily when `fest` is unavailable. I left an inline comment on `cmd/camp/festivals/festivals.go`.
- Local gate evidence is blocking: `just build default-build` passed and `just lint` passed, but `just test all` failed in Docker integration. Unit tests passed `2754/2754`; integration reported `373/375` with `TestPrePushHook_ClearsGitHookEnvironment` failing because `.githooks/pre-push` is missing. A targeted rerun with `just test integration-run TestPrePushHook_ClearsGitHookEnvironment` reproduced the same error.

**What is done well**
The implementation extends the existing registry and command patterns instead of introducing a parallel persistence layer. The test coverage is broad across schema normalization, org/tag/lifecycle commands, list filtering, and festival aggregation, and the user-facing docs were regenerated with the new command surface.

**Staff Standard**
I would be proud of the direction and most of the implementation, but I would not submit or merge it as production-ready until the failed integration gate is fixed and the two inline issues are addressed.

internal/config/registry.go
lancekrogers, line 1: The new org/tags/status fields need to be preserved by existing registry update paths in this file. `Register` is still an add-or-update helper used by `camp register` / registry sync flows; when an ID already exists it replaces the whole `RegisteredCampaign` with only name/path/type/last_access. With this PR that means a routine re-register of an organized campaign silently clears `Org`, `Tags`, and `Status` back to fallback defaults on the next save/load. Please preserve existing metadata when updating an existing ID, and add a regression test that re-registering an organized campaign keeps org/tags/status.

obey-agent (commented):
**Verdict:** Request Changes

**Overview:** The org/tag/lifecycle model is well scoped and the new commands have good coverage around the direct mutation paths. The remaining production risk is that existing registry update flows can still rewrite entries without carrying the new metadata forward.

**Key Findings:**
- File-level inline: `internal/config/registry.go` / `Registry.Register` still replaces an existing campaign entry with only name/path/type/last_access. That silently drops org/tags/status during routine re-register/sync flows, effectively undoing the organization layer.

**What's Done Well:** The read-side normalization and denormalized persistence strategy are coherent, and the focused tests cover the new command behavior, JSON shapes, filtering, grouping, and default-key omission well.

**Staff Standard:** I would not merge until the shared registry update path preserves the new fields and has a regression test. After that, the PR is close.

Verification: `go test ./internal/config ./cmd/camp/festivals ./cmd/camp/org ./cmd/camp/tag ./cmd/camp/lifecycle`

obey-agent (commented):
**Verdict:** Comment

**Overview:** Re-reviewed current head `9dcc0a8904ddf25864729da8d5f6e48dcea7b9b2`. The prior blocking issues I had flagged are addressed in this revision: `camp tag rm` now validates tag names before writing, `camp festivals` skips `fest` lookup when no selected campaign has a `festivals/` workspace, and `Registry.Register` now preserves existing org/tags/status metadata on re-register with a regression test.

**Key Findings:**
- No new blocking findings on this head.
- This is not a final approval because the branch is still actively changing; please re-run review and the full gate set after the final push.

**What's Done Well:** The registry normalization/persistence model is now better protected by tests, and the command-level tests cover the important no-write and no-op paths for org/tag/lifecycle. The `camp festivals` aggregator also has focused coverage for process composition, missing workspaces, and JSON shape.

**Staff Standard:** For the current head only, I would be comfortable advancing the quality gate if the full gates stay green and no additional commits land. If more commits are pushed, this review should be treated as a checkpoint, not final merge approval.

Verification: `go test ./internal/config ./cmd/camp/festivals ./cmd/camp/org ./cmd/camp/tag ./cmd/camp/lifecycle ./cmd/camp`

cmd/camp/list.go
```diff
@@ -132,21 +146,34 @@ func runList(cmd *cobra.Command, args []string) error {
 	}
 
 	sortBy, _ := cmd.Flags().GetString("sort")
+	campaigns := filterEntries(sortCampaigns(reg.Campaigns, sortBy), filter)
```
obey-agent: This filtered slice needs to be computed before the `--count` branch above and used there. As written, `camp list --count` still uses `reg.Len()`, so it ignores `--org`, repeated `--tag`, `--status`, `--all`, and the new active-only default. For example, if the registry has 10 campaigns but only 6 active, `camp list --count` reports `10 campaigns` while `camp list` shows 6; `camp list --org obey --count --json` also reports the global registry size instead of the filtered org size.

Please move the sort/filter step before the count branch and emit `len(campaigns)` for both text and JSON count output, with a regression test covering the active-only default and at least one explicit filter.

obey-agent (commented):
**Verdict:** Request Changes

**Overview:** Re-reviewed current head `23201631817ce10ab779e98070dc1d283f8ffb52`. The prior blocking `Registry.Register` metadata-loss issue is fixed and covered by `TestRegister_PreservesOrgTagsStatus`; the earlier tag/festivals concerns also remain addressed. The new problem is in the integration between this PR's list filters and the merged `--count` behavior.

**Key Findings:**
- Inline on `cmd/camp/list.go`: `camp list --count` still reports `reg.Len()` before applying PR 336's filters, so it ignores the active-only default plus `--org`, `--tag`, `--status`, and `--all`. The count output should be based on `len(filterEntries(sortCampaigns(...), filter))` for both text and JSON modes, with regression coverage.

**Additional Note:** The generated CLI docs are still missing the `--count` flags, which is already tracked separately as `camp#338` from the post-merge review of PR 337.

**What's Done Well:** The registry normalization and metadata preservation paths are in much better shape now. The org/tag/lifecycle commands and `camp festivals` aggregator remain well scoped, and the focused tests cover most of the new command behavior and JSON shapes.

**Staff Standard:** I would not merge this head until the filter-aware count semantics are fixed. After that, this looks close, assuming the full local gates stay green.

Verification:
- `go test ./internal/config ./cmd/camp/festivals ./cmd/camp/org ./cmd/camp/tag ./cmd/camp/lifecycle ./cmd/camp ./cmd/camp/project ./internal/project ./internal/ui`
- `just lint`

lancekrogers:
Addressed in de2e440.

`camp list --count` now computes the filtered slice (`filterEntries(sortCampaigns(...), filter)`) before the count branch and emits `len(campaigns)` for both text and JSON output, so it honors the active-only default plus `--org`, `--tag`, `--status`, and `--all`. The empty-registry help text still keys off the raw `reg.Len() == 0`, so `--count` on an empty registry prints `0 campaigns` rather than the onboarding hint.

Regression coverage in `cmd/camp/list_count_test.go` drives `runList` end-to-end against a 4-entry registry (2 active, 1 inactive, 1 reference, across `default`/`obey`):
- `TestListCount_TextRespectsFilters`: default = `2 campaigns` (not 4), `--all` = 4, `--org obey` = 1, `--status reference` = 1.
- `TestListCount_JSONRespectsFilters`: default `{"count":2}`, `--org obey` `{"count":1}`, `--all` `{"count":4}`.

Gates at this head: the verification test set is green, `just lint` reports 0 issues (both ratchets clean), and `git diff --check origin/main...HEAD` is clean. The generated-docs `--count` gap remains tracked separately as camp#338.

obey-agent (commented):
**Verdict:** Approve

**Overview:** Re-reviewed current head `de2e44014428775e369a70e1d50a25b1efe37d89`. The previous blocking issue is fixed: `camp list --count` now counts the filtered campaign slice, so it respects the active-only default plus `--org`, `--status`, and `--all` in both text and JSON modes.

**Key Findings:**
- No new blocking findings on this head.
- The generated CLI docs are still missing the already-merged `--count` flags, but that drift exists on `main` from PR 337 and is already tracked as `camp#338`; I am not treating it as a PR 336 blocker.

**What's Done Well:** The fix is placed at the right integration point, before the count branch, and the new regression tests cover the exact semantics that were previously wrong. The earlier registry metadata preservation, tag validation, and `camp festivals` no-workspace behavior remain addressed.

**Staff Standard:** Yes, I would be comfortable merging this head, assuming the project owner is satisfied with the broader gate coverage and the separate docs issue remains tracked.

Verification:
- `go test ./internal/config ./cmd/camp/festivals ./cmd/camp/org ./cmd/camp/tag ./cmd/camp/lifecycle ./cmd/camp ./cmd/camp/project ./internal/project ./internal/ui`
- `just lint`

PR #339, opened 2026-06-19
feat(workitem): camp workitem promote + shared internal/promote seam + shelve fold-in

lancekrogers:
Implements Issue 1 of the unified promotion flow (design: workflow/design/camp-workitem-promote-2026-06-16).

## Scope
- [x] Extract the festival-create seam into `internal/promote` (FindAndCreateFestival, CopyIntoFestivalIngest, CopyTree, promoted_to/promoted_at writer over a Promotable interface), lifted verbatim with no `internal/intent` dependency
- [x] Repoint `internal/intent/promote` at `internal/promote` with no behavior change; intent promote tests green
- [x] `camp workitem promote [id]` with id/cwd/current resolution and flags `--target/--dest/--goal/--keep/--force/--dry-run/--no-commit/--json`
- [x] Dungeon targets completed/archived/someday via `intdungeon.Service.MoveToDungeonStatus`; `active` rejected
- [x] festival target: create festival, copy whole workitem dir into `001_INGEST/input_specs/<slug>/`, record promoted_to, shelve source unless `--keep`, graceful fest-missing fallback (no partial corruption)
- [x] doc target: copy whole workitem dir to `docs/<dest>` (default slug), refuse non-empty dest without `--force`, record promoted_to, shelve source unless `--keep`
- [x] promoted_to/promoted_at in `.workitem` (minimal-diff atomic write), workitem audit promote event, single-object `--json` contract, auto-commit unless `--no-commit`
- [x] `camp shelve` folded into a Hidden deprecated alias delegating to the shared move; `--json`/`--no-commit` preserved byte-for-byte; one-line stderr deprecation note
- [x] Tests: resolution, all five targets, every flag, fest-missing fallback, double-promote error; shelve alias hidden-assertion + existing integration tests

## Notes
- The cwd workitem detector was relocated from `cmd/camp/promote` to a shared `internal/workitem/locate` package (proper internal->internal layering; its `fmt.Errorf` calls converted to `camperrors`).
- Hidden commands are now excluded from the agent `__manifest` (mirrors their absence from `--help`).
- No behavior change in the `internal/promote` extraction; shelve output is byte-for-byte unchanged for existing scripts.
- Issue 2 (camp promote router) depends on this branch.

### DIFF
--- a/internal/commands/workitem/promote.go
+++ b/internal/commands/workitem/promote.go
@@ -1,0 +1,476 @@
+package workitem
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	dungeoncmd "github.com/Obedience-Corp/camp/cmd/camp/dungeon"
+	"github.com/Obedience-Corp/camp/internal/config"
+	intdungeon "github.com/Obedience-Corp/camp/internal/dungeon"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/intent"
+	navindex "github.com/Obedience-Corp/camp/internal/nav/index"
+	promotepkg "github.com/Obedience-Corp/camp/internal/promote"
+	"github.com/Obedience-Corp/camp/internal/ui"
+	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
+	wkaudit "github.com/Obedience-Corp/camp/internal/workitem/audit"
+	"github.com/Obedience-Corp/camp/internal/workitem/links"
+	"github.com/Obedience-Corp/camp/internal/workitem/locate"
+)
+
+type runWorkitemPromoteOptions struct {
+	ID       string
+	Target   string
+	Dest     string
+	Goal     string
+	Keep     bool
+	Force    bool
+	DryRun   bool
+	NoCommit bool
+	JSON     bool
+}
+
+type workitemPromoteResult struct {
+	ID            string   `json:"id"`
+	Type          string   `json:"type"`
+	Target        string   `json:"target"`
+	From          string   `json:"from"`
+	To            string   `json:"to"`
+	PromotedTo    string   `json:"promoted_to"`
+	SourceShelved string   `json:"source_shelved,omitempty"`
+	Committed     bool     `json:"committed"`
+	CommitMessage string   `json:"commit_message,omitempty"`
+	Warnings      []string `json:"warnings,omitempty"`
+}
+
+type commitInputs struct {
+	description string
+	sourcePaths []string
+	destPaths   []string
+	rewritten   []string
+}
+
+func newPromoteCommand() *cobra.Command {
+	var (
+		target   string
+		dest     string
+		goal     string
+		keep     bool
+		force    bool
+		dryRun   bool
+		noCommit bool
+		jsonOut  bool
+	)
+
+	cmd := &cobra.Command{
+		Use:   "promote [id] --target <target>",
+		Short: "Promote a workitem to a festival, doc, or dungeon status",
+		Long: `Promote the workitem identified by [id], by cwd, or by the current pointer.
+
+TARGETS:
+  festival    Create a festival from the workitem and shelve the source
+  doc         Copy the workitem doc into docs/ and shelve the source
+  completed   Move the workitem to its local dungeon/completed
+  archived    Move the workitem to its local dungeon/archived
+  someday     Move the workitem to its local dungeon/someday`,
+		Args: cobra.RangeArgs(0, 1),
+		Annotations: map[string]string{
+			"agent_allowed": "true",
+			"agent_reason":  "Fully specified by flags; only the bare camp promote selector is interactive",
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			id := ""
+			if len(args) == 1 {
+				id = args[0]
+			}
+			return runWorkitemPromote(cmd, runWorkitemPromoteOptions{
+				ID: id, Target: target, Dest: dest, Goal: goal,
+				Keep: keep, Force: force, DryRun: dryRun, NoCommit: noCommit, JSON: jsonOut,
+			})
+		},
+	}
+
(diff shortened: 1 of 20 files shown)


internal/commands/workitem/promote.go
```diff
@@ -0,0 +1,481 @@
...
+	default:
+		return camperrors.New("invalid target: " + opts.Target + " (use festival, doc, completed, archived, someday)")
+	}
+
+	cfg, root, err := config.LoadCampaignConfigFromCwd(ctx)
+	if err != nil {
+		return camperrors.Wrap(err, "not in a campaign directory")
+	}
+
+	loc, err := resolveWorkitem(ctx, root, opts.ID)
```
obey-agent, line 130: `camp workitem promote` does not actually work from cwd in a real initialized campaign on macOS-style temp paths. `camp workitem resolve --json` correctly resolves `workflow/design/feat`, but `camp workitem promote --target completed --no-commit` from that same directory returns `no workitem in context`; promoting by explicit id succeeds. The issue is this path uses the raw `campaignRoot` from `LoadCampaignConfigFromCwd`, while the main workitem command normalizes it with `pathutil.ResolveRoot` before resolving items. Please normalize `root` here before calling `resolveWorkitem`, or reuse the existing workitem resolver path so cwd/current/id resolution has one implementation.

internal/commands/workitem/promote.go
```diff
@@ -0,0 +1,481 @@
...
+	}
+	if !opts.Force && strings.TrimSpace(docContent) == "" {
+		return nil, camperrors.New("workitem doc is empty; use --force to promote anyway")
+	}
+
+	relDest := opts.Dest
+	if relDest == "" {
+		relDest = loc.Slug
+	}
+	destDir := filepath.Join(campaignRoot, "docs", relDest)
```
obey-agent: `--dest` needs the same boundary validation as the existing docs move path. As written, `camp workitem promote <id> --target doc --dest ../escaped --no-commit --json` succeeds, copies the workitem to `<campaign>/escaped`, records `promoted_to: escaped`, and shelves the source. That violates the documented `docs/<dest>` target and bypasses the existing `internal/dungeon.ResolveDocsDestination` protections for absolute/traversal destinations. Please reject escaping paths before copying and before recording/shelving.

internal/commands/workitem/promote.go
```diff
@@ -0,0 +1,481 @@
...
+			return runWorkitemPromote(cmd, runWorkitemPromoteOptions{
+				ID: id, Target: target, Dest: dest, Goal: goal,
+				Keep: keep, Force: force, DryRun: dryRun, NoCommit: noCommit, JSON: jsonOut,
+			})
+		},
+	}
+
+	f := cmd.Flags()
+	f.StringVar(&target, "target", "", "Promotion target: festival, doc, completed, archived, someday")
+	f.StringVar(&dest, "dest", "", "Destination path override for doc/festival targets")
```
obey-agent: This flag is advertised as a destination override for `doc/festival`, but the festival path never reads `opts.Dest`; `doFestivalPromote` always creates a standard festival through `FindAndCreateFestival` and uses `fr.Dest/fr.Dir`. Either wire the flag into festival promotion semantics or narrow the help text/validation so users do not think festival destination is supported.

obey-agent (commented):
**Verdict: Request Changes**

This PR is moving in the right direction: the shared promotion seam is a good extraction, the workitem audit/metadata path is coherent, and the test suite is broad. I found three issues that should be fixed before merge:

- `camp workitem promote` does not resolve the current workitem from cwd in a real initialized campaign, even though `camp workitem resolve --json` does; explicit id promotion works. This breaks one of the advertised resolution modes.
- `--target doc --dest ../escaped` can write outside `docs/`, record the escaped path in metadata, and shelve the source. The new path should reuse or mirror the existing docs destination boundary validation.
- `--dest` is advertised for doc/festival targets, but the festival path ignores it. Either implement festival destination semantics or narrow the flag/help/validation.

What is done well: the feature uses the existing dungeon move plumbing for status promotions, preserves the shelve alias path, adds a single-object JSON contract, and has useful unit coverage for the main happy paths.

Staff standard: I would not merge this as-is because the default cwd workflow and docs path safety need to be production-correct before this command becomes agent-allowed.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** I rechecked PR #339 and the head is unchanged (`e77b0f4`). My prior review still applies: the shared promotion seam and coverage are promising, but the command has unresolved correctness and path-safety issues.

**Key Findings:**
- `camp workitem promote` still needs to resolve the current workitem from cwd in a real initialized campaign the same way `camp workitem resolve --json` does.
- `--target doc --dest ../escaped` still needs docs-root boundary validation so promotion cannot write outside `docs/`.
- `--dest` is still advertised for festival targets but ignored.

**What's Done Well:**
- The shared promotion extraction is useful.
- Dungeon moves reuse existing plumbing.
- The JSON result shape and audit path are directionally good.

**Validation:**
- Earlier review at this same head passed `just test unit` and `just build-camp` locally.
- Earlier manual probes reproduced the cwd-resolution failure and docs path escape.

**Staff Standard:** I would not merge this head as-is.

obey-agent (commented):
**Verdict: Approve**

**Overview:** I re-reviewed the current #339 head after the follow-up fix commit. The prior cwd-resolution, docs boundary, and festival `--dest` concerns are addressed, and the shared promotion seam plus workitem promote command now line up with the intended workflow.

**Key Findings:**
- No remaining blocking findings in this pass.
- The prior review issues appear fixed: cwd resolution handles symlinked campaign roots, doc promotion validates the docs boundary, and `--dest` is rejected for festival targets.

**What's Done Well:**
- Reuses the existing dungeon move plumbing instead of creating a parallel mover.
- Adds focused coverage for resolution, targets, dry-run, JSON output, fest-missing fallback, and path safety.
- Keeps the deprecated shelve alias hidden while preserving its scripting surface.

**Validation:**
- `just test unit` passed: 91 packages, 2789/2789 tests.
- `just lint` passed.

**Staff Standard:** Yes. I would be comfortable submitting this version; the remaining risk is mostly the normal blast radius of a large CLI surface addition, not an identified correctness blocker.

PR #341, opened 2026-06-19
feat(promote): camp promote universal router (CW0004 Issue 2)

lancekrogers:
## Summary

Adds the top-level `camp promote` router: the single discoverable front door over the three promotion pipelines. It resolves an item from `[id]` or cwd, detects its kind (intent / workitem / festival), and dispatches to the existing `camp intent promote`, `camp workitem promote`, or `fest promote`. With no context in an interactive terminal it opens a selector over promotable intents (inbox/ready), active workitems, and festivals (planning/ready/active), then routes the choice through a per-kind target picker. Adds shell completion over the promotable pool and a non-interactive guard requiring an explicit id and `--target`.

## Design

Implements the unified promotion router from the `camp-workitem-promote-2026-06-16` design (campaign workitem `WI-c62894`), specifically the `camp promote` router/selector in `03-command-spec.md`. Issue 2 of the three-issue breakdown in festival `camp-workitem-promote-CW0004`.

## Stacking base (D003)

**This PR is stacked on Issue 1 (#339, `workitem-promote`), which is still open.** It is based on `workitem-promote` so the diff shows only the router delta. **Retarget the base to `main` after #339 merges** (GitHub usually auto-retargets on merge).

## Implementation notes

- Workflow workitems carry no `SourceID`, so they dispatch via `cmd.Dir` cwd-resolution (reliable) rather than an id; intents dispatch by `SourceID`; festivals run `fest promote` in the festival dir (`--dungeon <status>` for dungeon targets, bare `promote` for the next stage).
- Dispatch goes through a stubbable runner seam.

## Tests

- Dispatch argv tests (stubbed runner) for all three kinds.
- Pool-correctness table (`isPromotable`): inbox/ready intents, active workitems incl. custom types, planning/ready/active festivals; excludes terminal/none-staged.
- Non-interactive guard test (no id, or id without target under `--json`).
- Shell-completion test. `just build` / `just test` / `just lint` green.

### DIFF
--- a/cmd/camp/promote/dispatch.go
+++ b/cmd/camp/promote/dispatch.go
@@ -1,0 +1,110 @@
+package promote
+
+import (
+	"context"
+	"os"
+	"os/exec"
+
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/fest"
+	"github.com/Obedience-Corp/camp/internal/workitem"
+)
+
+type commandRunner interface {
+	run(ctx context.Context, dir, bin string, args []string) error
+}
+
+type execRunner struct{}
+
+func (execRunner) run(ctx context.Context, dir, bin string, args []string) error {
+	cmd := exec.CommandContext(ctx, bin, args...)
+	cmd.Dir = dir
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return camperrors.Wrapf(err, "dispatching %s %v", bin, args)
+	}
+	return nil
+}
+
+var runner commandRunner = execRunner{}
+
+func campBinary() (string, error) {
+	exe, err := os.Executable()
+	if err != nil {
+		return "", camperrors.Wrap(err, "locating camp binary")
+	}
+	return exe, nil
+}
+
+func festBinary() (string, error) {
+	p, err := fest.FindFestCLI()
+	if err != nil {
+		return "", camperrors.Wrap(err, "locating fest binary")
+	}
+	return p, nil
+}
+
+type promoteKind int
+
+const (
+	kindIntent promoteKind = iota
+	kindWorkitem
+	kindFestival
+)
+
+func kindForType(wt workitem.WorkflowType) promoteKind {
+	switch wt {
+	case workitem.WorkflowTypeIntent:
+		return kindIntent
+	case workitem.WorkflowTypeFestival:
+		return kindFestival
+	default:
+		return kindWorkitem
+	}
+}
+
+func dispatchIntent(ctx context.Context, id, target string, pass []string) error {
+	bin, err := campBinary()
+	if err != nil {
+		return err
+	}
+	args := []string{"intent", "promote", id, "--target", target}
+	args = append(args, pass...)
+	return runner.run(ctx, "", bin, args)
+}
+
+func dispatchWorkitem(ctx context.Context, dir, target string, pass []string) error {
+	bin, err := campBinary()
+	if err != nil {
+		return err
+	}
+	args := []string{"workitem", "promote", "--target", target}
+	args = append(args, pass...)
+	return runner.run(ctx, dir, bin, args)
+}
+
+func dispatchFestival(ctx context.Context, dir string, festArgs []string) error {
+	bin, err := festBinary()
+	if err != nil {
+		return err
+	}
+	args := []string{"promote"}
+	args = append(args, festArgs...)
+	return runner.run(ctx, dir, bin, args)
+}
+
+func festPassthrough(target string, pass []string) []string {
+	var out []string
+	if target != "" {
+		out = append(out, "--dungeon", target)
+	}
+	for _, f := range pass {
+		switch f {
+		case "--force", "--json", "--no-commit":
+			out = append(out, f)
+		}
+	}
+	return out
+}
--- a/cmd/camp/promote/dispatch.go
+++ b/cmd/camp/promote/dispatch.go
@@ -2,6 +2,7 @@ package promote
 
 import (
 	"context"
+	"io"
 	"os"
 	"os/exec"
 
@@ -11,16 +12,16 @@ import (
 )
 
 type commandRunner interface {
-	run(ctx context.Context, dir, bin string, args []string) error
+	run(ctx context.Context, dir, bin string, args []string, stdout io.Writer) error
 }
 
 type execRunner struct{}
 
-func (execRunner) run(ctx context.Context, dir, bin string, args []string) error {
+func (execRunner) run(ctx context.Context, dir, bin string, args []string, stdout io.Writer) error {
 	cmd := exec.CommandContext(ctx, bin, args...)
 	cmd.Dir = dir
 	cmd.Stdin = os.Stdin
-	cmd.Stdout = os.Stdout
+	cmd.Stdout = stdout
 	cmd.Stderr = os.Stderr
 	if err := cmd.Run(); err != nil {
 		return camperrors.Wrapf(err, "dispatching %s %v", bin, args)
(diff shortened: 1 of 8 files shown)


cmd/camp/promote/dispatch.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+func dispatchIntent(ctx context.Context, id, target string, pass []string) error {
+	bin, err := campBinary()
+	if err != nil {
+		return err
+	}
+	args := []string{"intent", "promote", id, "--target", target}
+	args = append(args, pass...)
```
obey-agent, line 77: `camp promote --json` is advertised as the non-interactive/agent-safe path, but for intents this forwards `--json` to `camp intent promote`, which does not define that flag. I reproduced this with a temporary campaign: `camp promote <intent-id> --target ready --json` exits through the child command with `unknown flag: --json`. Please either implement a router-owned JSON envelope for intent promotions or filter/validate passthrough flags per target kind before dispatching; as written the manifest marks this route agent-allowed but the intent case fails.

cmd/camp/promote/dispatch.go
```diff
@@ -0,0 +1,112 @@
...
+	args = append(args, festArgs...)
+	return runner.run(ctx, dir, bin, args)
+}
+
+func festPassthrough(target string, pass []string) []string {
+	var out []string
+	if target != "" {
+		out = append(out, "--dungeon", target)
+	}
+	for _, f := range pass {
```
obey-agent, line 117: This silently drops `--dry-run` for festival promotions even though the top-level router flag says "Preview without making changes". That is a production-safety issue: `camp promote <festival> --target completed --dry-run` will dispatch `fest promote --dungeon completed` without any dry-run protection. Please fail closed for unsupported passthrough flags or forward a real dry-run flag if `fest promote` supports one.

cmd/camp/promote/router.go
```diff
@@ -0,0 +1,165 @@
...
+	cwd, err := os.Getwd()
+	if err != nil {
+		return workitem.WorkItem{}, false, camperrors.Wrap(err, "getting current directory")
+	}
+	if loc, derr := locate.DetectFromCwd(campaignRoot, cwd); derr == nil {
+		for _, it := range pool {
+			if it.WorkflowType == workitem.WorkflowTypeIntent || it.WorkflowType == workitem.WorkflowTypeFestival {
+				continue
+			}
+			if filepath.Base(it.RelativePath) == loc.Slug {
```
obey-agent: Cwd resolution should match the exact located workitem path/type, not just the directory slug. Two workflow collections can legitimately contain the same basename, e.g. `workflow/design/auth` and `workflow/explore/auth`; from inside the explore item this loop can pick the design item depending on discovery/sort order and dispatch promotion in the wrong directory. Compare against `loc.SourcePath`/campaign-relative path (or reuse the existing selector/resolver) so cwd dispatch is unambiguous.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The universal `camp promote` front door is the right product direction, and the router/selector/runner split is understandable. I found three correctness issues in the router delta that should be fixed before this becomes an agent-allowed command.

**Key Findings:**
- `camp promote <intent-id> --target ready --json` forwards `--json` to `camp intent promote`, which does not support that flag; I reproduced the command failing in a temporary campaign. Inline: `cmd/camp/promote/dispatch.go`.
- Festival promotions silently drop `--dry-run`, so a flag documented as "Preview without making changes" can still dispatch a mutating `fest promote`. Inline: `cmd/camp/promote/dispatch.go`.
- Cwd workitem resolution matches only the basename, so duplicate slugs across workflow types can promote the wrong item. Inline: `cmd/camp/promote/router.go`.

**What's Done Well:**
- The command runner seam makes dispatch tests straightforward.
- The promotable pool rules are explicit and covered.
- The stacked PR keeps the router delta narrow, which made review tractable.

**Validation:**
- `go test ./cmd/camp/promote ./cmd/camp/intent ./internal/commands/workitem`
- `just test unit` - 91 packages, 2790/2790 tests passed
- `just build-camp` - build/vet passed across 107 packages
- Manual probe: built the PR binary, initialized a temp campaign, created an intent, and confirmed `camp promote <intent-id> --target ready --json` fails via child `unknown flag: --json`.

**Staff Standard:** Not yet. The command is positioned as the safe universal front door, so passthrough validation and cwd resolution need to be exact before merge.

cmd/camp/promote/router.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+		if dryRun {
+			return printDryRun(cmd, item, target)
+		}
+		return dispatchWorkitem(ctx, item.AbsPath(campaignRoot), target, pass)
+	}
+	return camperrors.New("unknown promote kind")
+}
+
+func printDryRun(cmd *cobra.Command, item workitem.WorkItem, target string) error {
```
obey-agent: `--json` is accepted by the router as “machine-readable output”, but this path still emits plain text. The same issue exists for intent dispatch: `dispatchIntent` drops the router passthrough flags because `camp intent promote` has no `--json`, so `camp promote <intent> --target festival --json` will succeed with human output instead of JSON. Please either have the router own JSON responses for dry-run/intent dispatch, or reject `--json` for kinds that cannot satisfy it; otherwise the new universal front door has an unreliable scripting contract.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The router is a good direction: it resolves the promotable pool, dispatches through a clean runner seam, and the stacked fixes handle dry-run safety and cwd workitem disambiguation. I found one scripting-contract issue that should be fixed before this becomes the universal front door.

**Key Findings:**
- Inline comment on `cmd/camp/promote/router.go`: `camp promote --json` can still emit human-readable text. Router-level dry-run always prints text, and intent dispatch drops `--json` because `camp intent promote` does not support it, so callers cannot trust the advertised machine-readable contract.

**What's Done Well:**
- Dispatch argv is narrowly tested across intent, workitem, and festival cases.
- The promotable-pool rules are explicit and covered.
- The runner seam keeps the command routing testable without spawning real child commands.

**Validation:**
- `just test unit` passed: 91 packages, 2795/2795 tests.
- `just lint` passed.

**Staff Standard:** Not yet. The implementation is close, but I would not merge a new scriptable front door while one of its global output flags is unreliable across supported item kinds.

cmd/camp/promote/dispatch.go
```diff
@@ -0,0 +1,112 @@
...
+		return kindWorkitem
+	}
+}
+
+func dispatchIntent(ctx context.Context, id, target string, stdout io.Writer) error {
+	bin, err := campBinary()
+	if err != nil {
+		return err
+	}
+	args := []string{"intent", "promote", id, "--target", target}
```
obey-agent, line 76: `camp promote` exposes `--force` and `--no-commit`, and `camp intent promote` supports both, but this intent dispatch path drops them. That means `camp promote <intent> --target design --force` still fails readiness checks, and worse, `camp promote <intent> --target ready --no-commit` can create the child auto-commit even though the caller explicitly disabled commits. Please thread the safe intent passthrough flags (`--force`, `--no-commit`) into `dispatchIntent` while continuing to suppress unsupported `--json`, and add tests for both flags.

cmd/camp/promote/router.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+		}
+		return workitem.WorkItem{}, false, camperrors.Wrapf(camperrors.New("not a promotable item"), "id %q", args[0])
+	}
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return workitem.WorkItem{}, false, camperrors.Wrap(err, "getting current directory")
+	}
+	if loc, derr := locate.DetectFromCwd(campaignRoot, cwd); derr == nil {
```
obey-agent, line 104: The non-interactive guard only fires when nothing resolves, but `resolveItem` still accepts cwd context before dispatch. In a workflow directory, `camp promote --json --target completed` can mutate without an explicit id even though the command annotation and PR contract say non-interactive use requires an explicit id and `--target`. Please enforce `len(args) == 1` whenever `!interactive` before cwd resolution, or narrow cwd resolution to interactive mode, and add a regression test from inside a workitem directory.

obey-agent (commented):
**Verdict:** Request Changes

**Overview:** The current head is much improved from the earlier review: the router has a clear dispatch seam, dry-run is intercepted safely, and the JSON success path is now more deliberate. I still found two production-affecting contract gaps around mutation safety and intent flag forwarding.

**Key Findings:**
- Inline: intent dispatch drops `--force` and `--no-commit` even though the top-level router exposes those flags and `camp intent promote` supports them. The `--no-commit` case is especially risky because callers can explicitly request no commit and still get a child auto-commit.
- Inline: non-interactive mode can still resolve from cwd and mutate without an explicit id, contradicting the command annotation/PR contract that agents/scripts must pass an explicit id and `--target`.

**What's Done Well:** The implementation keeps the universal command as a router over existing intent/workitem/festival promote implementations instead of reimplementing the promotion logic. The stubbable runner seam and focused tests make the dispatch behavior easy to reason about, and the recent JSON/dry-run fixes address the previous review direction.

**Staff Standard:** Not yet. I would want the intent passthrough and non-interactive explicit-id guard fixed before I would be comfortable merging this.

obey-agent (commented):
**Verdict: Approve**

**Overview:** This PR adds a focused top-level `camp promote` router over the existing intent, workitem, and festival promotion paths. The implementation keeps mutation in the child commands, constrains non-interactive execution to explicit selection, and has useful coverage for dispatch argv, pool filtering, JSON handling, completion, and cwd-resolution safety.

**Key Findings:** No blocking findings. I did not find an issue worth an inline comment.

**What's Done Well:** The router is deliberately thin: it resolves the item, filters/normalizes per-kind flags, and delegates to the existing promote implementations instead of duplicating promotion mechanics. The regression tests cover the risky parts of this command surface, especially non-interactive cwd behavior and intent flag filtering.

**Validation:**
- `go test ./cmd/camp/promote ./internal/commands/workitem`
- `just build default-build`
- `just test unit`
- `just lint`
- `just gate-fast`

**Staff Standard:** Yes. I would be comfortable submitting this as my own work.

PR #343, opened 2026-06-23
Superseded quest metadata update command

lancekrogers:
## Summary

- Add `camp quest update <quest>` for non-interactive purpose/description edits
- Preserve omitted fields and allow explicit empty values to clear metadata
- Keep default quest metadata protected and auto-commit updates by default

## Validation

- `BUILD_TAGS=dev just test unit`

### DIFF
--- a/cmd/camp/quest/update.go
+++ b/cmd/camp/quest/update.go
@@ -1,0 +1,77 @@
+//go:build dev
+
+package quest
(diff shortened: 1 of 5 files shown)


lancekrogers:
Replaced by #344 to use neutral branch and PR naming.

PR #344, opened 2026-06-23
Add quest metadata update command

lancekrogers:
## Summary

- Add `camp quest update <quest>` for non-interactive purpose/description edits
- Preserve omitted fields and allow explicit empty values to clear metadata
- Keep default quest metadata protected and auto-commit updates by default

## Validation

- `BUILD_TAGS=dev just test unit`

### DIFF
--- a/cmd/camp/quest/update.go
+++ b/cmd/camp/quest/update.go
@@ -1,0 +1,77 @@
+//go:build dev
+
+package quest
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/git/commit"
+	questsvc "github.com/Obedience-Corp/camp/internal/quest"
+)
+
+var questUpdateCmd = &cobra.Command{
+	Use:   "update <quest>",
+	Short: "Update quest metadata",
+	Long: `Update quest metadata without opening an editor.
+
+Omitted fields are left unchanged. Provide an empty value to clear a field.
+
+Examples:
+  camp quest update platform-launch --purpose "harden launch path"
+  camp quest update platform-launch --description ""
+  camp quest update platform-launch --purpose "..." --description "..."`,
+	Args: cobra.ExactArgs(1),
+	Annotations: map[string]string{
+		"agent_allowed": "true",
+		"agent_reason":  "Non-interactive quest metadata update",
+	},
+	RunE: runQuestUpdate,
+}
+
+func init() {
+	Cmd.AddCommand(questUpdateCmd)
+	flags := questUpdateCmd.Flags()
+	flags.String("purpose", "", "Short purpose statement")
+	flags.String("description", "", "Full description")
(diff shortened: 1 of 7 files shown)


obey-agent (commented):
## Staff Review: camp #344 (quest metadata update command)

**Verdict: Comment (clean, no blockers). Leaving the formal Approve for Lance since this is still a draft.**

### Overview
Adds a non-interactive `camp quest update <quest> [--purpose] [--description]` as the scriptable counterpart to the interactive `quest edit`. Small, focused, well-tested, and consistent with the existing quest command/service patterns. Build (`-tags dev`) and `internal/quest` + `cmd/camp/quest` tests pass locally.

### What's done well
- Correct pointer-options pattern (`*string`) on `MetadataUpdateOptions` so "field omitted" is distinct from "field cleared to empty". Verified the clear-to-empty path end to end (trims server-side, default-quest protected).
- `UpdateMetadata` mirrors `Rename`: same `ensureInitialized` -> `Resolve` -> `IsDefault` guard -> `Save` shape, returns only the quest file in `MutationResult.Files` so the auto-commit stays selective.
- `commit.QuestUpdate` is fully wired (the action string flows straight into `doCommit`, no missing switch arm).
- Error style matches the file: `errors.New` for static validation, `camperrors.Wrap` for context wrapping. No stray code comments.
- Tests cover trim, preserve-untouched-field, clear, required-field, and default-quest-read-only.

### Minor (non-blocking)
- Success output prints `result.Quest.Name`, whereas sibling `quest edit` prints `quest.RelativePath(...)`. Harmless (the app calls `quest show` to read state, it does not parse this line), but matching `edit` would be more consistent.

### Staff standard
Yes, I would be comfortable submitting this. It extends the existing system rather than duplicating it, and the tests assert behavior rather than implementation.

obey-agent (commented):
## Staff Re-Review: camp #344 (quest metadata update command)

**Verdict: Approve. No blockers.** (Out of draft, fully self-contained.)

### What changed since the last review
One new commit: "refactor: modularize quest metadata updates". This is a pure, behavior-preserving move:
- `MetadataUpdateOptions` + `UpdateMetadata` relocated from `service.go` into a new `internal/quest/service_metadata.go` (logic byte-identical).
- The tests moved into `internal/quest/service_metadata_test.go`, and the default-quest protection is now its own dedicated test (`TestServiceUpdateMetadataRejectsDefaultQuest`) instead of being appended to the lifecycle test.

### Verified this round (at head 081414f1)
- `go build -tags dev ./...`: clean
- `go test -tags dev ./internal/quest/... ./cmd/camp/quest/...`: pass
- `go vet -tags dev ./internal/quest/`: clean

### Assessment
Good modularization that keeps `service.go` smaller without touching behavior. The pointer-options pattern, selective `MutationResult.Files`, default-quest guard, and error style all carry over unchanged from the version I already validated.

The earlier cosmetic note (the success line prints `result.Quest.Name` where `quest edit` prints the relative path) is unchanged. It is not gating, fine to leave as is.

### Staff standard
Yes. I would submit this as my own work.

PR #345, opened 2026-06-23
feat(org,list): interactive org browser, org create, and list --sort org

lancekrogers:
## Summary

Three UX follow-ons to the shipped org/tags/status feature (`camp org`, `camp tag`, `camp lifecycle`). Designed in `workflow/design/camp-org-tui-and-list-sort/`. Each thread routes through the existing registry write path; no new persistence, no schema change, no database.

### 1. Interactive org browser (TUI)
A two-pane bubbletea browser of orgs and their member campaigns:
- **Entry point:** bare `camp org` opens the browser **when run in a terminal**. Piped or with `--json` it prints the current campaign's org (script-safe). `camp org which` (alias `current`) prints the org unconditionally; `-i` opens the browser explicitly (and prints the org list when stdout is not a terminal, since bubbletea needs a TTY).
- Navigate orgs (left) with a live member preview (right); `enter`/`l` drills into members, `h`/`esc` backs out.
- `m` moves the selected campaign to another org, **`c` creates a new org** for it, `d` returns it to the fallback org, `r` renames the focused org.
- Every mutation goes through `config.UpdateRegistry` (the same atomic, file-locked path the flat commands use), then reloads from disk. Errors (rename collision, same-name) surface in a styled footer with no state change.
- Styling: top bar with org/campaign totals, colored campaign/active counts, semantic status badges (active=green, inactive=amber, reference=blue), accent-focused rounded pane borders, current-campaign marker, responsive single-pane collapse below a width floor, and explicit loading/empty/success/error states. Adaptive `theme.TUI()` palette, consistent with the intent/promote TUIs. Degrades to `camp org list` when stdout is not a TTY.

### 2. `camp org create <org>` — friendly create verb
- With no campaign args inside a campaign, it joins the **current** campaign to `<org>` (fills the gap that `camp org add` requires `MinimumNArgs(2)`).
- With campaign args, it behaves like `add`.
- Orgs stay **derived**: `create` only assigns membership, never persists an empty org. Idempotent (no "already exists" error); errors clearly outside a registered campaign. Reuses `reassignOrg`.

### 3. `camp list --sort org` + org-aware output
- New `--sort org`: org-major ordering (fallback org first, then alphabetical, then by name), matching the grouped-view ordering.
- The flat (ungrouped) table gains an accent-colored `ORG` column so `camp list --no-group` and single-org output are org-aware. Grouped headers are now styled and show a per-org count (e.g. `obey (3 campaigns)`). JSON shape is unchanged.
- `camp org list` / `camp org show` tables are also styled (accent org names, green active counts, status badges).

## Tests
- `camp list`: `--sort org` ordering (incl. custom fallback), invalid-sort fallthrough, updated flat-table and grouped goldens.
- `camp org create`: current-campaign join, explicit args, idempotent join, no-op when already a member, invalid name (no write), not-in-campaign and unregistered-current errors.
- Entry point: `shouldOpenOrgTUI` decision matrix (TTY / `-i` / `--json`), `camp org which`, bare-org print forced non-TTY.
- TUI: model-level tests driving `Update` with synthetic keys — open/navigate/enter, move (incl. typed new org), **create**, return-to-default, rename, error surfacing, quit, and a View render smoke test.
- Full suite green; `just lint` clean (0 issues), gofmt clean, no `fmt.Errorf` or host-fs-test regressions.

## Notes for reviewers
- The TUI lives in `package org` (`cmd/camp/org/tui*.go`) rather than a separate `internal/org/tui` package, so it can reuse the unexported `computeOrgCounts`/`buildOrgShow`/`renameOrgInRegistry`/`validateOrgName` directly — matching how `promote` keeps its selector inline.
- `camp org` help no longer states "There is no org create" (that contradicted the new verb); the derived-org model is preserved and documented in the `create` help instead.
- The TTY decision is behind a `stdoutIsTTY` package var + a pure `shouldOpenOrgTUI` helper so the dispatch is unit-tested without launching a real program.
- Remaining manual acceptance item from the design's experience bar: a human eyeball/screenshot of the TUI under each terminal theme at a real terminal (model + View tests cover structure; rendered frames look consistent with the existing camp TUIs).

### DIFF
--- a/cmd/camp/org/tui_update.go
+++ b/cmd/camp/org/tui_update.go
@@ -61,6 +61,11 @@ func (m orgTUIModel) updateBrowse(key tea.KeyMsg) (tea.Model, tea.Cmd) {
 			m.openOverlay(overlayMove, "target org")
 		}
 		return m, nil
+	case "c":
+		if m.pane == paneMembers && len(m.members) > 0 {
+			m.openOverlay(overlayCreate, "new org name")
+		}
+		return m, nil
 	case "d":
 		if m.pane == paneMembers && len(m.members) > 0 {
 			member := m.members[m.memCursor]
@@ -150,6 +155,16 @@ func (m orgTUIModel) commitOverlay() (tea.Model, tea.Cmd) {
 			m.setError(err)
 		} else {
 			m.setStatus(fmt.Sprintf("moved %q to org %q", member.Name, value))
+		}
+	case overlayCreate:
+		member := m.members[m.memCursor]
+		existed := m.orgExists(value)
+		if err := m.assignOrg(member.ID, value); err != nil {
+			m.setError(err)
+		} else if existed {
+			m.setStatus(fmt.Sprintf("added %q to existing org %q", member.Name, value))
+		} else {
+			m.setStatus(fmt.Sprintf("created org %q with %q", value, member.Name))
 		}
 	}
 	m.overlay = overlayNone
--- a/cmd/camp/org/tui_update.go
+++ b/cmd/camp/org/tui_update.go
@@ -69,9 +69,6 @@ func (m orgTUIModel) updateBrowse(key tea.KeyMsg) (tea.Model, tea.Cmd) {
 	case "d":
 		if m.pane == paneMembers && len(m.members) > 0 {
 			member := m.members[m.memCursor]
-			if member.Status == "" || m.fallback == "" {
-				return m, nil
-			}
 			if err := m.assignOrg(member.ID, m.fallback); err != nil {
 				m.setError(err)
 			} else {
--- a/cmd/camp/org/tui_update.go
+++ b/cmd/camp/org/tui_update.go
@@ -1,0 +1,158 @@
+package org
+
+import (
+	"fmt"
+	"strings"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+func (m orgTUIModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	switch msg := msg.(type) {
+	case tea.WindowSizeMsg:
+		m.width, m.height = msg.Width, msg.Height
+		return m, nil
+	case tea.KeyMsg:
+		if m.overlay != overlayNone {
+			return m.updateOverlay(msg)
+		}
+		return m.updateBrowse(msg)
+	}
+	return m, nil
+}
+
+func (m orgTUIModel) updateBrowse(key tea.KeyMsg) (tea.Model, tea.Cmd) {
+	m.status = ""
+	switch key.String() {
+	case "ctrl+c", "q":
+		m.quitting = true
+		return m, tea.Quit
+	case "esc":
+		if m.pane == paneMembers {
+			m.pane = paneOrgs
+			return m, nil
+		}
+		m.quitting = true
+		return m, tea.Quit
+	case "down", "j":
+		m.moveDown()
+		return m, nil
+	case "up", "k":
+		m.moveUp()
+		return m, nil
+	case "right", "l", "enter":
+		if m.pane == paneOrgs && len(m.members) > 0 {
+			m.pane = paneMembers
+			m.memCursor = 0
+		}
+		return m, nil
+	case "left", "h", "backspace":
+		if m.pane == paneMembers {
+			m.pane = paneOrgs
+		}
+		return m, nil
+	case "r":
+		if m.pane == paneOrgs && len(m.orgs) > 0 {
+			m.openOverlay(overlayRename, "new org name")
+		}
+		return m, nil
+	case "m":
+		if m.pane == paneMembers && len(m.members) > 0 {
+			m.openOverlay(overlayMove, "target org")
+		}
+		return m, nil
+	case "d":
+		if m.pane == paneMembers && len(m.members) > 0 {
+			member := m.members[m.memCursor]
+			if member.Status == "" || m.fallback == "" {
+				return m, nil
+			}
+			if err := m.assignOrg(member.ID, m.fallback); err != nil {
+				m.setError(err)
+			} else {
+				m.setStatus(fmt.Sprintf("returned %q to org %q", member.Name, m.fallback))
+			}
+		}
+		return m, nil
+	}
+	return m, nil
+}
+
+func (m *orgTUIModel) moveDown() {
+	if m.pane == paneOrgs {
+		if len(m.orgs) == 0 {
+			return
+		}
+		m.orgCursor = (m.orgCursor + 1) % len(m.orgs)
+		m.memCursor = 0
+		m.syncFocusedOrg()
+		return
+	}
+	if len(m.members) == 0 {
+		return
+	}
+	m.memCursor = (m.memCursor + 1) % len(m.members)
+}
+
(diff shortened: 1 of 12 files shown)


cmd/camp/org/tui_update.go
```diff
@@ -0,0 +1,158 @@
...
+		return m, nil
+	case "m":
+		if m.pane == paneMembers && len(m.members) > 0 {
+			m.openOverlay(overlayMove, "target org")
+		}
+		return m, nil
+	case "d":
+		if m.pane == paneMembers && len(m.members) > 0 {
+			member := m.members[m.memCursor]
+			if member.Status == "" || m.fallback == "" {
```
obey-agent: Optional, non-blocking clarity note: this guard is dead code. `LoadRegistry` normalizes every entry's status to `active` (internal/config/registry.go:52 -> types.go:377-378), so `member.Status` is never `""` here, and `FallbackOrg()` always returns at least `"default"` (types.go:336-340), so `m.fallback` is never `""`. Neither branch can fire.

It also reads as an asymmetry: the `m` (move) handler has no equivalent precondition, which will make a future reader wonder what status-less member `d` is defending against. The `len(m.members) > 0` check above already covers the real precondition, so consider dropping this line. If it is intentional future-proofing for a possible empty-status tier, a one-line note would keep the `d`/`m` divergence from looking accidental. Does not block.

obey-agent (commented):
**Verdict: Comment (merge-ready; leaving the Approve to you)**

**Overview:** Three clean UX follow-ons to the org/tags/status feature (`org -i` browser, `org create`, `list --sort org`). The work integrates correctly: every mutation routes through the existing `config.UpdateRegistry` atomic, file-locked path and reloads from disk; the TUI reuses the unexported org helpers in-package (matching the promote selector pattern) instead of reimplementing them; the project error package is used throughout. No new persistence, no schema change, no database. I traced the registry/normalize internals and ran the gate at head.

**Verified at head (45650d3, which merges current main; the only delta since the org/list work is the unrelated quest-metadata command from #344, no org/list files touched):**
- `just lint` clean: 0 issues, no new host-fs-test violators, no fmt.Errorf ratchet regressions.
- `go test ./cmd/camp/org/... ./cmd/camp/` green.
- Backward compatibility holds: bare `camp org` unchanged; `--json` and `--format simple` shapes unchanged; grouped list output unchanged (the ORG column lands only on the flat table, golden updated). Scripting paths are preserved.
- Reuse claims check out: `reassignOrg`, `validateOrgName`, `renameOrgInRegistry`, `computeOrgCounts`, `buildOrgShow`, `currentCampaignID` are reused, not duplicated. Orgs stay derived (`create` only assigns membership, never persists an empty org).
- Adaptive theme is safe: `theme.TUI()` returns `AdaptiveColor`s that resolve at render, and the package-level-var pattern matches the existing intent/quest/status TUIs.

**Tests:** Genuinely behavioral. The TUI tests drive `Update` with synthetic keys and assert on the real persisted registry (navigate/enter, move, typed-new-org, return-to-default, rename, and error-surfaced-without-mutation, plus a narrow-width View smoke). `create` covers the no-arg current-campaign join plus not-in-campaign, unregistered-current, and invalid-name-no-write paths with byte-level no-write checks.

**One optional, non-blocking note (inline on tui_update.go):** the `d` return-to-default guard `member.Status == "" || m.fallback == ""` is dead code (both values are always populated after `LoadRegistry` normalization) and reads as an asymmetry vs the `m` move path. Harmless; drop or annotate at your discretion. Does not gate.

**Staff standard:** Would I be proud to submit this? Yes. It is correct, well-scoped, contract-stable, and tested to a staff standard. Merge-ready; I am leaving the actual Approve to you since you asked for a review rather than an approval.

cmd/camp/org/views.go
```diff
@@ -68,14 +76,26 @@ var orgShowCmd = &cobra.Command{
...
 	Cmd.AddCommand(orgRenameCmd)
 	Cmd.AddCommand(orgListCmd)
 	Cmd.AddCommand(orgShowCmd)
+	Cmd.AddCommand(orgWhichCmd)
 	orgRenameCmd.Flags().Bool("json", false, "Output as JSON")
 	orgListCmd.Flags().Bool("json", false, "Output as JSON")
 	orgShowCmd.Flags().Bool("json", false, "Output as JSON")
+	orgWhichCmd.Flags().Bool("json", false, "Output as JSON")
 	Cmd.Flags().Bool("json", false, "Output as JSON")
+	Cmd.Flags().BoolP("interactive", "i", false, "Force the interactive org browser even when output is piped")
```
obey-agent: Minor, non-blocking: this help says the browser opens "even when output is piped", but `runOrgTUI` returns `runOrgList(cmd, nil)` when stdout is not a terminal (bubbletea needs a TTY), so `-i` with a piped stdout prints the org list rather than forcing the browser. `-i` still does real work when TTY auto-detection is wrong on an actual terminal, so the flag is worth keeping; consider rewording to match the degrade-to-list behavior (e.g. "Force the interactive browser when a terminal is available").

obey-agent (commented):
**Re-review. Verdict: Comment (merge-ready; leaving the Approve to you)**

**What landed since the first review** (head is now 0a0a7b9):
- `fa8d640` — `camp org` (no args) now opens the interactive browser in a terminal, prints the current org when piped or with `--json`, and adds `camp org which` (alias `current`) to print the org unconditionally. Also: in-TUI create (`c`), and richer (TTY-gated) styling for the TUI, `org list`, `org show`, and grouped `list` output.
- `0a0a7b9` — drops the dead return-to-default guard I flagged last round. Resolved: the `d` handler is now symmetric with `m`/`c` (guard on `len(members) > 0`, then `assignOrg`, which still validates), with no behavior change.

**Backward compatibility on the default-command change (the thing I scrutinized hardest) holds, and is tested:**
- `camp org` piped, in `$(...)`, or redirected (non-TTY) still prints the current org; `camp org --json` still prints `{campaign, org}` and never opens the TUI. `shouldOpenOrgTUI` is covered for all five json/interactive/tty combinations (`TestShouldOpenOrgTUI`), plus `TestOrgWhich` and a `forceNoTTY`-guarded `TestOrgBare_PrintsCurrentOrg`. The only behavior change is interactive: a human typing `camp org` at a terminal now gets the browser instead of a printed line, which is the point of the feature.
- The new styling routes through the existing TTY-gated `ui.*` helpers (same mechanism `camp list` already used); piped output stays plain. Grouped `list` headers gain a `(N campaigns)` suffix — additive, plain when piped, golden updated.

**Verified at head (0a0a7b9):** `just lint` clean (0 issues, no host-fs-test or fmt.Errorf ratchet regressions); `go test ./cmd/camp/org/... ./cmd/camp/` green. New behavior is genuinely tested (in-TUI create drives the registry write; dispatch table covers the contract paths).

**One minor, non-blocking note (inline):** the `-i` flag help says "even when output is piped," but a piped stdout degrades to `org list` (bubbletea needs a TTY), so `-i` can't force the browser when piped. Reword at your discretion; does not gate.

**Staff standard:** Correct, well-scoped, contract-stable (scripting paths preserved and tested), and the one prior finding is cleanly resolved. Merge-ready; leaving the Approve to you since you asked for a review.

PR #346, opened 2026-06-24
feat(intent): inherit ambient workitem context on intent and note commits

lancekrogers:
## Problem

Intent and note auto-commits go through `commit.Intent` (internal/git/commit/intent.go), which only populates `CampaignRoot` / `CampaignID` / `Files`. It never sets `QuestID`, `FestivalRef`, or `WorkitemRef`, so every intent and note commit emits only the bare `[OBEY-CAMPAIGN-<id>]` tag.

By contrast `camp commit` resolves ambient context (cmd/camp/commit_workitem.go) and `camp workitem commit` sets the fields directly (internal/git/commit/workitem.go). Those carry the FE-/WI-/qst_ segments; the intent path did not. A note captured with `cnote` while working inside an active festival/quest/workitem recorded none of that context.

This also surfaced that notes are not workitems and therefore have no `WI-` ref of their own. This PR makes intent/note commits inherit the *ambient* context. Giving notes their own `NT-` segment, and deriving the `OBEY-CAMPAIGN` prefix from config, are tracked as separate intents.

## Change

- Add `ResolveCommitContext` in `internal/commands/workitem` (commit_context.go): a staging-free resolver returning the ambient `QuestID` / `FestivalRef` / `WorkitemRef`, mirroring how `camp workitem commit` derives those segments. Resolution is best-effort: any failure yields zero-valued fields so callers fall back to the bare campaign tag.
- Wire it into the note capture path (cmd/camp/intent/note.go) and the intent-add path (cmd/camp/intent/add.go).

## Verification

Manual smoke test on a throwaway campaign:

```
# note captured inside a workitem dir
[OBEY-CAMPAIGN-fe270033-WI-WI-2f47a0] Create: check the daemon socket path
# note captured at campaign root (no context)
[OBEY-CAMPAIGN-fe270033] Create: loose thought
```

- `just lint` clean (golangci-lint, host-fs ratchet, fmt-errorf ratchet, `go vet -tags=integration`).
- Unit tests: `ResolveCommitContext` ancestor-ref, no-context, and cancelled-context cases.
- Integration tests: a note inside a workitem carries `WI-<ref>`; a root note stays bare. These compile under `-tags=integration` but were not executed locally (Docker unavailable in this environment; CI Actions are paused).

## Scope notes

Only the note + intent-add capture paths are wired here. The remaining `commit.Intent` call sites (edit/move/archive/promote/gather/rename/convert/crawl) are trivial one-line follow-ups using the same helper.

### DIFF
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/Obedience-Corp/camp/cmd/camp/cmdutil"
+	wkcmd "github.com/Obedience-Corp/camp/internal/commands/workitem"
 	"github.com/Obedience-Corp/camp/internal/concept"
 	"github.com/Obedience-Corp/camp/internal/config"
 	"github.com/Obedience-Corp/camp/internal/editor"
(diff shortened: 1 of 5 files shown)


obey-agent (commented):
**Verdict:** Approve

**Overview:** This cleanly threads ambient workitem context into the intent/note auto-commit paths without adding staging behavior or changing the commit schema. The resolver helper mirrors the existing workitem commit context enough for these capture paths, and its best-effort fallback is documented and tested.

**Key Findings:**
- No blocking findings.

**What's Done Well:** The new helper is small, cancellation is preserved through the resolver call, errors stay non-fatal by design for capture flow, and the added comments explain contracts that are not obvious from the call sites. The tests cover ancestor context, no-context fallback, cancellation, and note commit integration shape.

**Staff Standard:** Yes. I would be comfortable submitting this as my own work. I ran focused unit tests and `just lint`; I did not run the Docker-backed integration suite locally.

PR #347, opened 2026-06-24
feat(list): interactive campaign browser, lifecycle toggle, and tilde paths

lancekrogers:
## Summary

Three improvements to `camp list`, all surfaced while using it. Designed in `workflow/design/camp-list-tui-and-tilde-paths/`. Every mutation routes through the existing `config.UpdateRegistry` atomic write path; no new persistence, no schema change, no database.

### 1. Interactive campaign browser
Bare `camp list` in a terminal opens an interactive browser (grouped by org), mirroring the `camp org` browser from #345.
- **Dispatch:** opens only for a plain, interactive invocation. `--json`, `--count`, a non-table `--format`, **or any filter/sort flag** (`--org/--tag/--status/--all/--sort/--group/--no-group`) prints the table unchanged. `-i` forces, degrading to the table when stdout is not a terminal. Behind a `listTUIRequested` helper + `stdoutIsTTY` var, fully unit-tested.
- Shows **all statuses by default** (so a just-deactivated campaign doesn't vanish), with color-coded badges (active=green, inactive=amber, reference=blue) and windowed scrolling that keeps the cursor in view.

### 2. Lifecycle toggle (the discoverable deactivation UX)
`camp lifecycle set <campaign> inactive` already existed but was buried and couldn't be done while browsing. In the browser, **`s`** cycles the selected campaign's status active → inactive → reference → active. One keypress to deactivate. `f` toggles an active-only filter.

Also in the browser: **`m`** reassigns the campaign's org (implicit-create on a new name, like `camp org add`); **`y`** copies the path to the clipboard.

### 3. Tilde paths
Home paths display as `~/Dev/...` instead of `/Users/<name>/Dev/...` — shorter and doesn't leak the username, and `~` re-expands. Via a new tested `pathutil.AbbreviateHome`. **Display-only**: the registry and `--json` keep absolute paths. Applied to the static `camp list` table (flat + grouped) and the browser; paths outside `$HOME` (e.g. `/srv/...`) stay absolute.

## Tests
- `pathutil.AbbreviateHome`: under-home, exact-home, not-under-home, sibling-prefix (no false match), empty-HOME.
- Dispatch matrix (`listTUIRequested`): TTY/piped, `--json`, `--count`, non-table format, `-i` force, shaping-flag-prints-table.
- Browser model (synthetic keys): opens org-major all-statuses, cycle-status deactivation persists and stays visible, active-only filter, move-org (valid + invalid-name-no-mutation), copy-uses-tilde (clipboard injected), quit, and a constrained-size View smoke.
- Full suite green; `just lint` clean (0 issues); gofmt clean; no `fmt.Errorf` or host-fs-test regressions.

## Notes for reviewers
- The browser lives in `package main` (`cmd/camp/list_tui*.go`), reusing `campaignEntry`/`sortCampaigns`/`filterEntries` and writing status/org via `config.UpdateRegistry` + `config.ValidateStatus`/`ValidateName` directly (no cross-package access to the unexported `lifecycle`/`org` handlers needed). Matches how `cmd/camp/status_all.go` already hosts a TUI.
- `writeClipboard` is a package var so tests don't touch the real clipboard.
- Per the design, this surfaces deactivation **only** in the TUI (no `camp deactivate` alias) and the browser does lifecycle/org/copy (no open/`cd` picker) — both deferred by choice.
- Remaining manual acceptance item: a human eyeball of the browser under each terminal theme (model + View tests cover structure; rendered frames look consistent with the org browser).

### DIFF
--- a/cmd/camp/list_tui.go
+++ b/cmd/camp/list_tui.go
@@ -18,8 +18,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// stdoutIsTTY reports whether stdout is an interactive terminal. A package var so
-// the dispatch is deterministic under test.
+// Overridden in tests so dispatch does not depend on the runner's terminal.
 var stdoutIsTTY = func() bool { return term.IsTerminal(int(os.Stdout.Fd())) }
 
 type listOverlay int
@@ -33,8 +32,8 @@ type listTUIModel struct {
 	ctx context.Context
 
 	fallback   string
-	all        []campaignEntry // every campaign, org-major sorted
-	visible    []campaignEntry // after the activeOnly filter
+	all        []campaignEntry
+	visible    []campaignEntry
 	cursor     int
 	activeOnly bool
 
@@ -70,8 +69,6 @@ func listTUIRequested(cmd *cobra.Command, isTTY bool) bool {
 	return isTTY
 }
 
-// runListTUI launches the browser, degrading to the table when stdout is not a
-// terminal (bubbletea needs a TTY).
 func runListTUI(cmd *cobra.Command) error {
 	ctx := cmd.Context()
 	if !term.IsTerminal(int(os.Stdout.Fd())) {
@@ -144,8 +141,6 @@ func (m *listTUIModel) reload() error {
 
 func (m listTUIModel) Init() tea.Cmd { return textinput.Blink }
 
-// cycleStatus advances the selected campaign active -> inactive -> reference ->
-// active through the shared atomic registry write path.
 func (m *listTUIModel) cycleStatus() error {
 	e := m.visible[m.cursor]
 	next := nextLifecycleStatus(e.Status)
@@ -186,8 +181,7 @@ func (m *listTUIModel) mutate(id string, apply func(*config.RegisteredCampaign))
 	return m.reload()
 }
 
-// writeClipboard copies s to the system clipboard. A package var so tests can
-// intercept it instead of touching the real clipboard.
+// Package var so tests do not touch the real clipboard.
 var writeClipboard = func(s string) error {
 	var c *exec.Cmd
 	switch runtime.GOOS {
--- a/cmd/camp/list_tui.go
+++ b/cmd/camp/list_tui.go
@@ -77,11 +77,15 @@ func runListTUI(cmd *cobra.Command) error {
 	if !term.IsTerminal(int(os.Stdout.Fd())) {
 		return renderListTable(cmd)
 	}
-	reg, err := config.LoadRegistry(ctx)
+	reg, report, err := loadVerifiedListRegistry(ctx)
 	if err != nil {
 		return camperrors.Wrap(err, "failed to load registry")
 	}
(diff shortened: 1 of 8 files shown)


cmd/camp/list_tui.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// runListTUI launches the browser, degrading to the table when stdout is not a
+// terminal (bubbletea needs a TTY).
+func runListTUI(cmd *cobra.Command) error {
+	ctx := cmd.Context()
+	if !term.IsTerminal(int(os.Stdout.Fd())) {
+		return renderListTable(cmd)
+	}
+	reg, err := config.LoadRegistry(ctx)
```
obey-agent: This changes the default `camp list` path in terminals from `renderListTable` to the TUI, but the TUI loads the registry directly and skips the existing `VerifyAndRepair` pass in `renderListTable`. That is a behavioral regression for plain `camp list`: stale/missing campaign paths, ID mismatches, and name/type drift no longer get repaired before display, and a later TUI mutation can persist the stale entry back through `UpdateRegistry`. Please factor the load+verify+save step into a shared helper and use it before constructing the TUI model so the new default path keeps the old self-healing contract.

obey-agent (commented):
**Verdict:** Request Changes

**Overview:** This is a useful TUI addition and the implementation largely reuses the existing registry mutation path, sort/filter helpers, and local TUI patterns. The main issue is that the new default terminal path bypasses the existing `camp list` registry verification/repair contract.

**Key Findings:**
- `cmd/camp/list_tui.go:80`: `runListTUI` loads the registry directly, so plain interactive `camp list` no longer removes stale entries or repairs ID/name/type drift before display. Please share the existing load/verify/save path with the TUI before constructing the model.

**What's Done Well:** The dispatch matrix is explicit and covered, mutations go through `config.UpdateRegistry`, and the comments are short contract notes rather than long narration. The `pathutil.AbbreviateHome` helper is scoped and tested.

**Staff Standard:** I would be comfortable with the feature after the self-healing regression is fixed; as written I would not merge it because it changes the default `camp list` behavior.

obey-agent (commented):
**Verdict:** Approve

**Overview:** The previous self-healing regression is fixed: both the table path and the default TUI path now use the shared `loadVerifiedListRegistry` flow before rendering. The TUI still preserves the new interactive behavior, and registry cleanup is surfaced in the model status line instead of silently diverging from the old `camp list` contract.

**Key Findings:**
- No blocking findings on re-review.
- Prior finding on `cmd/camp/list_tui.go:80` is addressed by loading through `loadVerifiedListRegistry` before constructing the TUI model.

**What's Done Well:** The fix keeps verification/repair centralized, adds a focused regression test for stale ID and missing-path repair, and trims redundant comments while keeping the useful contract/test-seam comments. Mutation behavior still routes through `config.UpdateRegistry`.

**Staff Standard:** Yes. I would be comfortable submitting this as my own work after the latest changes.

**Verification:** `go test ./cmd/camp ./internal/pathutil`; `just lint`; `git diff --check main...HEAD`.

PR #349, opened 2026-06-24
feat(list): g/enter in the browser to switch to the selected campaign

lancekrogers:
## Summary

In the `camp list` browser, **`g`** (go) or **enter** navigates the shell to the selected campaign — the same outcome as `camp switch`, but from the browser. Find a campaign, hit `g`, land in it. Designed in `workflow/design/camp-list-tui-go-to-campaign/`.

## How it works (and why)

A TUI is a child process and can't `cd` the parent shell. camp already solves this for the `workitem` browser with a hidden `--path-output <file>` flag: the TUI writes the chosen path to a temp file, and the generated `camp()` shell function reads it and `cd`s after the TUI exits. This reuses that exact contract (stdout isn't usable — the browser owns it via the alt-screen):

- **Hidden `--path-output` flag** on `camp list`. On `g`/enter the browser records the selected campaign's **absolute** path and quits; `runListTUI` writes it to the file (`writeGotoSelection`).
- **New `list`/`ls` branch** in the `camp()` shell function (`bash.sh.tmpl`, `zsh.sh.tmpl`, `fish.sh.tmpl`), mirroring the `workitem` branch but `cd "$dest"` directly — campaign registry paths are absolute, so no `camp root` prefix (unlike workitem's root-relative paths).
- **Graceful fallback:** without shell integration (no `--path-output`), `g` shows a one-line hint to run `camp shell-init` rather than silently closing.

`g` and enter both go: `g` is the mnemonic (consistent with `cgo`/`camp go`), enter is the intuitive select. Footer now shows `g: go`. The move overlay still uses enter to confirm a move.

## Tests
- Browser model: `g`/enter set `gotoPath` + quit when enabled; `g` shows the hint when disabled; enter inside the move overlay still moves (not go).
- `writeGotoSelection` seam: writes the absolute path when chosen; no file when the user just quits; no-op without `--path-output`.
- Shell render: bash/zsh/fish init contain the `list` branch with `--path-output` and the `camp-list` temp file; the existing `bash -n`/`zsh -n`/`fish -n` syntax checks validate the new branches.
- Full suite green; `just lint` clean (0 issues); gofmt clean; no new dependency.

## Notes for reviewers
- The browser owns stdout (alt-screen), so this uses the temp-file `--path-output` contract exactly like `workitem`, not `switch`'s `--print`/stdout.
- **Shell-integration change:** existing users must re-run `eval "$(camp shell-init zsh|bash|fish)"` (or restart their shell) to get the new `list` navigation. Until then bare `camp list` still opens the browser and `g` shows the hint.
- Deferred (design): `@tab` sub-location navigation (`campaign@p` -> projects/) like `camp switch`, and updating last-access time on go. v1 goes to the campaign root.
- Remaining manual acceptance item: end-to-end interactive `cd` at a real terminal (`eval "$(camp shell-init zsh)"; camp list`, select, `g`). The shell branch is syntax-checked and asserted; the actual interactive cd needs a human TTY.


obey-agent (commented):
**Verdict: Approve**

**Overview:** This PR adds `g`/enter navigation from the `camp list` browser by reusing the existing shell temp-file handoff pattern instead of inventing a second mechanism. The implementation keeps machine-readable list modes intact, keeps the TUI path explicit through the hidden `--path-output` flag, and preserves the move overlay's enter behavior.

**Key Findings:**
- No blocking findings. I did not find correctness, integration, or maintainability issues worth an inline comment.

**What's Done Well:** The shell integration is mirrored consistently across bash, zsh, and fish; the browser model tests cover enabled, disabled, enter, and overlay cases; the write seam is small and easy to reason about. The hidden flag stays out of normal help while giving the shell wrapper a reliable contract.

**Staff Standard:** Yes. I would be comfortable submitting this after the documented manual TTY acceptance check is completed by a human shell session.

PR #353, opened 2026-06-26
fix(quest): make default quest editable and ensure it always exists

lancekrogers:
## Problem

The `default` quest (`qst_default`) exists for one reason: to guarantee a campaign always has a quest to work in. But it was incorrectly locked down as **read-only** — renaming it, editing its purpose/description, or changing its lifecycle all failed with `ErrDefaultQuestReadOnly`. That surfaced in festival-app as "can't close quest" and "can't update the default quest details including name."

Separately, **freshly-opened campaigns got no quest at all**: `camp quest list` ran with `ensureScaffold=false`, so a campaign without a `.campaign/quests/` directory failed with `ErrNotInitialized` instead of returning a quest.

## Changes

1. **Default quest is now a normal quest.** Removed all 8 `if q.IsDefault() { return ErrDefaultQuestReadOnly }` guards (Edit, Rename, Restore, Link, Unlink, `updateInPlace` → pause/resume, `moveToStatus` → complete/archive, and `UpdateMetadata`), plus the now-dead `ErrDefaultQuestReadOnly` definition. Renaming keeps the immutable slug/ID, so the fallback-attribution contract is preserved.

2. **`camp quest list` ensures the scaffold** (`loadQuestCommandContext(ctx, true)`). A campaign with no quests directory now gets the default quest created on list instead of erroring — this is what guarantees "you always have a quest." Combined with #1, completing the last quest lets the next list recreate a fresh default.

3. **Tests** inverted from "default is protected" to "default is mutable" (`TestServiceDefaultQuestIsMutable`, `TestServiceUpdateMetadataAllowsDefaultQuest`).

## Verification

- `go build -tags dev ./...` — clean
- `go test -tags dev ./internal/quest/ ./cmd/camp/quest/` — pass
- `go vet -tags dev` — clean; `gofmt` clean
- Manual CLI smoke (fresh campaign): `quest rename default`, `quest update default`, `quest complete default` all succeed; deleting `.campaign/quests/` then `quest list` recreates `qst_default`.

## Note

This is the `camp` half. The festival-app UI also gates the default quest (`slug === "default"` disables edit/close) — a companion festival-app PR removes those gates. The bundled `camp` sidecar must be rebuilt from this branch for the end-to-end fix.

### DIFF
--- a/cmd/camp/quest/list.go
+++ b/cmd/camp/quest/list.go
@@ -38,7 +38,10 @@ func init() {
 
 func runQuestList(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
-	qctx, err := loadQuestCommandContext(ctx, false)
+	// Ensure the quest scaffold (and the default quest) exists before listing so
+	// a freshly-opened campaign always has at least one quest instead of failing
+	// with ErrNotInitialized. This is what guarantees "you always have a quest".
+	qctx, err := loadQuestCommandContext(ctx, true)
 	if err != nil {
 		return err
 	}
--- a/internal/quest/quest.go
+++ b/internal/quest/quest.go
@@ -16,17 +16,16 @@ import (
 )
 
 var (
-	ErrNotInitialized       = camperrors.Wrap(camperrors.ErrNotFound, "quest system not initialized")
-	ErrQuestNotFound        = camperrors.Wrap(camperrors.ErrNotFound, "quest not found")
-	ErrQuestAmbiguous       = camperrors.Wrap(camperrors.ErrInvalidInput, "quest identifier is ambiguous")
-	ErrDefaultQuestReadOnly = camperrors.Wrap(camperrors.ErrInvalidInput, "default quest cannot change lifecycle state")
-	ErrInvalidTransition    = camperrors.Wrap(camperrors.ErrInvalidInput, "invalid quest status transition")
-	ErrInvalidQuest         = camperrors.Wrap(camperrors.ErrInvalidInput, "invalid quest")
-	ErrMissingID            = camperrors.Wrap(camperrors.ErrInvalidInput, "quest id is required")
-	ErrMissingName          = camperrors.Wrap(camperrors.ErrInvalidInput, "quest name is required")
-	ErrInvalidStatus        = camperrors.Wrap(camperrors.ErrInvalidInput, "quest status is invalid")
-	ErrMissingCreatedAt     = camperrors.Wrap(camperrors.ErrInvalidInput, "quest created_at is required")
-	ErrMissingUpdatedAt     = camperrors.Wrap(camperrors.ErrInvalidInput, "quest updated_at is required")
+	ErrNotInitialized    = camperrors.Wrap(camperrors.ErrNotFound, "quest system not initialized")
+	ErrQuestNotFound     = camperrors.Wrap(camperrors.ErrNotFound, "quest not found")
+	ErrQuestAmbiguous    = camperrors.Wrap(camperrors.ErrInvalidInput, "quest identifier is ambiguous")
+	ErrInvalidTransition = camperrors.Wrap(camperrors.ErrInvalidInput, "invalid quest status transition")
+	ErrInvalidQuest      = camperrors.Wrap(camperrors.ErrInvalidInput, "invalid quest")
+	ErrMissingID         = camperrors.Wrap(camperrors.ErrInvalidInput, "quest id is required")
+	ErrMissingName       = camperrors.Wrap(camperrors.ErrInvalidInput, "quest name is required")
+	ErrInvalidStatus     = camperrors.Wrap(camperrors.ErrInvalidInput, "quest status is invalid")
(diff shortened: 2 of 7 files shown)


cmd/camp/quest/list.go
```diff
@@ -38,7 +38,10 @@ func init() {
 
 func runQuestList(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
-	qctx, err := loadQuestCommandContext(ctx, false)
+	// Ensure the quest scaffold (and the default quest) exists before listing so
+	// a freshly-opened campaign always has at least one quest instead of failing
+	// with ErrNotInitialized. This is what guarantees "you always have a quest".
+	qctx, err := loadQuestCommandContext(ctx, true)
```
obey-agent, line 44: `quest list` now guarantees a root `default/quest.yaml`, but the default quest is also allowed to move into the dungeon. That means completing `qst_default` and then listing quests creates a second quest with the same fixed ID while the completed one still exists under `dungeon/completed/default`. `Resolve` searches by ID/slug across root + dungeon and returns the first match, so the historical default becomes ambiguous/unaddressable by `qst_default` or `default` (for example, restore will hit the newly-created open default and fail the status transition). Preserve the “always have a quest” behavior without duplicating the fixed default identity: either do not recreate while a dungeon default exists, give recreated defaults unique IDs, or make resolution/status operations handle multiple default generations explicitly.

obey-agent (commented):
**Verdict: Request Changes**

The intent is right: the default quest should be mutable, and `quest list` should be a safe first operation for a fresh campaign. The current implementation creates a duplicate fixed default identity after `qst_default` is completed and `quest list` scaffolds a new root default.

**Key Findings**
- Inline comment on `cmd/camp/quest/list.go`: completing the default quest and listing again can leave two `qst_default` quests, making the historical default ambiguous/unaddressable by ID or slug.

**What’s Done Well**
- The read-only guard removal is consistently applied across the service methods.
- The targeted tests correctly invert the old default-readonly expectations.

**Staff Standard**
I would not merge this as-is because the new lifecycle behavior breaks quest identity invariants.

lancekrogers:
Addressed the duplicate-default-identity finding (root-cause backend bug for festival-app#111).

## Root cause
Making the default quest mutable lets it be completed/archived, which moves its directory into the quest dungeon while keeping the fixed `qst_default` identity. `EnsureScaffold` -> `ensureDefaultQuest` only checked for the root `default/quest.yaml`, so the next `quest list`/`quest create` recreated a second open `qst_default` at root. `Resolve` returns root matches before dungeon matches, so the historical default became ambiguous and unaddressable, and `restore qst_default` resolved to the new open default and failed the status transition.

## Fix (option a: never create a second default)
- `ensureDefaultQuest` now skips recreation when a default quest already exists anywhere, including the dungeon (`defaultQuestExists` scans root + dungeon by the fixed `qst_default` ID). This keeps `qst_default`/`default` globally unique, so it always resolves to the historical default and restore works.
- `ensureInitialized` now treats the quest system as initialized when the quests directory exists, rather than requiring the root default file. That preserves the "fresh campaign always has a quest / `quest list` is a safe first op" guarantee (a fresh campaign has no default anywhere, so one is created), while a campaign whose default sits in the dungeon no longer errors and no longer mints a duplicate.
- Removed the now-misleading `IsInitialized` predicate (its "root default present" meaning is no longer the correct initialization signal, and it had no other callers).

Behavior after the fix: completing the default leaves zero active quests (recoverable via `restore`), which correctly reflects the user's action instead of silently resurrecting a duplicate.

## Regression test
`TestEnsureScaffoldDoesNotDuplicateCompletedDefault` completes the default, re-runs `EnsureScaffold`, and asserts exactly one (completed) default remains, that it resolves by `qst_default`, and that it restores to open. Verified it fails on the prior code (found 2 defaults) and passes now.

## Gate
`just gate` green (head `3374c76`): stable+dev builds, `go vet` stable/dev/integration, lint both profiles (0 issues), no-host-fs-tests and no-fmt-errorf clean, unit tests 2889/2889 (dev) and 2863/2863 (stable).

obey-agent (commented):
**Verdict: Approve**

This re-review is against head `3374c76`. The new `defaultQuestExists` check addresses the earlier correctness risk: completing or archiving the default quest no longer causes `EnsureScaffold` to mint a second `qst_default` at the root.

**Key Findings**
- No blocking findings on this head.

**What's Done Well**
- The fix preserves identity uniqueness by resolving across root and dungeon state before recreating the fallback default.
- The regression test exercises the important lifecycle: complete default, re-run scaffold, resolve the completed default, and restore it.
- Removing the read-only guards is now coherent with the global identity invariant.

**Staff Standard**
I would be comfortable submitting this as my own work.

Validation note: I reviewed the changed quest implementation and tests in full. I did not run the touched Go tests directly because they mutate filesystem state through host-side `t.TempDir`; this repo's guidance requires filesystem-effect tests to run through a containerized harness, and I did not find a targeted containerized unit entrypoint for these quest tests.

PR #356, opened 2026-06-29
fix(quest): real default-quest timestamps + skip stray quest dirs (QD0001)

lancekrogers:
## Summary

Festival **QD0001** (quest-default-recovery-and-date-fix), camp portion. Closes two of the three quest defects on the camp side:

- **Bug 2 — default quest fake timestamp.** A freshly initialized campaign stamped the default quest with `2000-01-01T00:00:00Z`, and existing campaigns kept that sentinel forever.
- **Bug 3 — stray quest directories.** `camp quest list` emitted `warning: skipping unreadable quest ...` for any directory under `.campaign/quests/` that lacked a `quest.yaml` (for example a festival-app per-quest `ui-state.json` directory).

## Changes

**Bug 2 (timestamps)**
- `internal/scaffold/campaign/templates/.campaign/quests/default/quest.yaml.tmpl`: render `created_at`/`updated_at` from a real timestamp instead of the hardcoded sentinel.
- `internal/scaffold/init.go` and `repair.go`: inject a UTC `now` into the scaffold `Vars` so the template renders a real date through the engine.
- `camp init --repair` now backfills an existing default quest still carrying the `2000-01-01` sentinel, using the quest file's mtime (fallback: now). Implemented as plan/apply: `ComputeRepairPlan` detects and stages a `RepairChange` without writing; `Init` performs the write only on apply, so declining the repair prompt never mutates the file. Idempotent.

**Bug 3 (stray dirs)**
- `internal/quest/quest.go`: `List` now skips a directory with no `quest.yaml` silently, and reserves `warnUnreadableQuest` for the malformed case (a `quest.yaml` that is present but unparseable). Applied to both the root and dungeon scans. The contract is documented on `List` and in `docs/tribal-knowledge/quest-directory-contract.md`.

## Also fixed: README template variable nesting

A fresh `camp init` wrote `# <no value>` into `README.md`: `README.md.tmpl` referenced top-level `{{ .campaign_name }}` etc., but the guild-scaffold engine nests vars under `.vars` (which is why `AGENTS.md.tmpl`/`SKILL.md.tmpl` already use `{{ .vars.X }}`). Fixed the template and corrected `readme_test.go`, which had been rendering with a flat map and so never caught it.

## Testing

- Filesystem-altering behavior runs in the containerized integration harness (`tests/integration/quest_default_test.go`): fresh-init real timestamp, repair backfill + auto-commit + idempotency, stray-dir silent skip vs malformed warn, and README var rendering.
- `internal/scaffold` unit tests are pure-logic only (context-cancellation, nil handling), no host FS writes; `cmd/camp/init` unit tests cover the repair-commit file/message construction.
- `just lint` clean (0 issues; `lint-no-host-fs-tests` and `lint-no-fmt-errorf` both clean); unit tests pass under dev and stable; integration tests pass.

## Notes

- Quests are dev-profile-scoped (festival decision D001); no stable-profile behavior changes beyond the quiet-listing fix.
- Festival-app sequences (CLI currency bump + silent-failure recovery) are a separate PR that depends on this one merging to `main` first.


internal/scaffold/repair.go
```diff
@@ -72,6 +76,16 @@ type RepairPlan struct {
 	Migrations []MigrationAction
 	// IntentMigrations lists legacy intent-root moves handled internally by Init.
 	IntentMigrations []MigrationAction
+	// QuestDateBackfill, when set, rewrites a default quest's sentinel timestamps
+	// to a real date during apply.
+	QuestDateBackfill *QuestDateBackfill
```
obey-agent, line 81: This new repair action needs to be included in the repair auto-commit scope. `camp init --repair` uses `commitRepairChanges`, which builds a selective file list from `InitResult` files/dirs, migrations, intent migrations, and skill projections; it does not currently add `plan.QuestDateBackfill.Path`. The apply path will rewrite `.campaign/quests/default/quest.yaml`, but the auto-commit will stage only the existing config/jumps paths and omit the actual repaired quest file, leaving the user with an uncommitted backfill despite a successful repair run. Please thread this path into `buildRepairCommitFiles` (and ideally the repair commit message/test) so repair remains operationally complete.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** This is a focused fix for the default quest timestamp sentinel and noisy stray quest directories. The core quest-listing behavior is clean, and the plan/apply split for the date backfill is the right shape, but the repair path is not fully integrated with camp's selective repair auto-commit.

**Key Findings:**
- Inline: `QuestDateBackfill.Path` is not included in `buildRepairCommitFiles`, so `camp init --repair` can rewrite `.campaign/quests/default/quest.yaml` but omit that modified file from the repair auto-commit. That leaves the main repair artifact dirty/uncommitted even though the repair completed.

**What's Done Well:** The PR preserves the preview-before-apply repair model, keeps stray directories silent while malformed quest files still warn, and adds targeted tests for the new quest listing and date backfill helpers. Context cancellation and idempotency are covered in the helper-level tests.

**Staff Standard:** I would be comfortable with the direction, but I would not merge it as-is because the repair feature needs to carry its modified file through the existing commit integration before it is production-complete.

lancekrogers:
Addressed the review, plus two related fixes. New head `1808e80`.

## Review finding: repair backfill not in the auto-commit scope (fixed)

`camp init --repair` could rewrite `.campaign/quests/default/quest.yaml` but leave it uncommitted. Fixed in `cmd/camp/init/repair.go`:

- `buildRepairCommitFiles` now includes `plan.QuestDateBackfill.Path`.
- `commitRepairChanges`'s `hasChanges` guard now counts the backfill, so a campaign whose **only** repair is the sentinel backfill still auto-commits (previously it would have skipped the commit entirely).
- `buildRepairCommitMessage` lists the backfill.

Verified end-to-end in a containerized integration test: inject the sentinel, run `--repair`, assert the file is backfilled **and** `git status --porcelain` for `quest.yaml` is clean (committed), and that a second `--repair` is a no-op.

## Bonus fix: README template variable nesting

While verifying scaffolding I found a pre-existing bug: a fresh `camp init` wrote `# <no value>` into `README.md`. `README.md.tmpl` referenced top-level `{{ .campaign_name }}` etc., but the guild-scaffold engine nests vars under `.vars` (which is why `AGENTS.md.tmpl` and `SKILL.md.tmpl` already use `{{ .vars.X }}`). Fixed the README template and updated `readme_test.go` to render with the real nested shape (the old test passed a flat map, which is why this slipped through).

## Test placement: FS-altering tests moved to the container harness

Per team guidance that filesystem-mutating tests must run in the containerized integration harness (not host `t.TempDir`, which can strand temp files on a developer's machine), I moved the quest timestamp / repair-backfill / stray-dir / README behaviors into `tests/integration/quest_default_test.go` (CLI black-box, FS mutations isolated in the container). The remaining `internal/scaffold` unit tests are pure-logic only (context-cancellation, nil handling), with no host FS writes.

`just lint` clean (0 issues; `lint-no-host-fs-tests` and `lint-no-fmt-errorf` clean); unit tests pass under dev and stable; the four new integration tests pass.

obey-agent (commented):
**Verdict: Approve**

Re-reviewed current head `1808e80efaa9d1e63e8532836a44ab108896de70`.

**Key Findings**
- No blocking findings.
- The prior repair auto-commit issue is fixed: `QuestDateBackfill.Path` is now included in `buildRepairCommitFiles`, a backfill-only repair counts as a change, and the repair commit message lists the backfilled default quest path.

**What's Done Well**
- The repair path keeps the right preview/apply boundary: the sentinel backfill is detected during `ComputeRepairPlan`, but the quest file is only rewritten after confirmation/`--yes`.
- `quest.List` now treats directories without `quest.yaml` as expected stray state while still warning for malformed quest metadata, and the contract is documented.
- The README template fix uses the real scaffold variable shape, and the test was corrected so it would catch the previous `<no value>` regression.
- Filesystem-mutating coverage was placed in the containerized integration harness instead of adding new host-FS tests.

**Validation**
- `just lint` passed.
- `BUILD_TAGS=dev just lint` passed.
- `go test ./cmd/camp/init ./internal/scaffold ./internal/quest` passed.
- `go test -tags dev ./cmd/camp/init ./internal/scaffold ./internal/quest` passed.
- `go test -v -tags=integration ./tests/integration/... -run 'TestQuestDefault|TestQuestList_SkipsStrayDirWithoutWarning|TestReadme_RendersCampaignVars' -timeout 5m` passed.

**Staff Standard**
I would be comfortable merging this head.

PR #358, opened 2026-06-30
feat(git): use campaign name in commit tags instead of hardcoded OBEY-CAMPAIGN

lancekrogers:
## What

The campaign commit tag's leading token was the hardcoded literal `OBEY-CAMPAIGN`, so every campaign produced `[OBEY-CAMPAIGN-<id>]` regardless of its actual name. This makes the tag use the **campaign name** instead.

- **Before:** `[OBEY-CAMPAIGN-8deed8b4] Update navigation`
- **After:** `[obey-campaign:8deed8b4] Update navigation` (and e.g. `[shrapnel:a1b2c3d4]`, `[job-search:...]`)

## Tag format

```
[<campaign-name-slug>:<id8>[-qst_<...>][-FE-<ref>][-WI-WI-<ref>]]
```

- The leading token is the slugified campaign name (via the shared `internal/slug` generator), followed by `:` and the short campaign id.
- The **colon** separates the name from the structured remainder. Campaign names legitimately contain hyphens (`job-search`, `brainshare-planning`), so a `-` boundary would be ambiguous. An explicit delimiter is the right call per the repo's "explicit formats over heuristics" rule.
- The **short id is kept** after the colon. It is the stable machine identity (names can change, the registry keys on id) and is what every existing commit in history carries.

## Backward compatibility

- The parser recognizes **both** the new `[<name>:<id>...]` form and the legacy `[OBEY-CAMPAIGN-<id>...]` form, so the entire pre-existing commit history still resolves correctly (verified by the `camp workitem commits` query path).
- When the campaign name is unavailable or slugifies to nothing, the formatter **falls back to the legacy form**.
- The new-style parse gates the id on `[0-9a-f]{1,8}` (real ids are UUID-derived), so ordinary bracket prefixes like `[wip]` or `[scope: note]` are not misread as campaign tags.

## Implementation

- `internal/git/campaign_tag.go`: `FormatContextTagsFull` now takes `campaignName`; parser handles both forms; `TagComponents` gains an additive `CampaignName` field (`--json` gets a new `campaign_name`, omitempty).
- `internal/git/commit/commit.go`: `Options.CampaignName` (new) is resolved from the campaign config at `CampaignRoot` when callers don't supply it, so the ~18 `doCommit` call sites pick up the name with no per-site edits.
- The direct `commitkit` callers (`camp commit`, `camp p commit`, worktree commit, workitem staging) and the submodule-ref / refs-sync paths thread `cfg.Name` explicitly.
- `pkg/commitkit` public **id-only** functions (`FormatCampaignTag` / `PrependCampaignTag`, consumed by `fest`) keep their signatures and emit the legacy form. **Follow-up:** thread the name through `fest` so its commits also use the name-style tag.

## Tests / gates

- `go build ./...`, `go test ./...`, dev build, and integration-test compilation all green.
- `just lint`: 0 issues.
- Tag contract tests extended with name-style + negative (non-tag bracket) + legacy-fallback cases; round-trip property test now exercises the name.
- Integration assertions updated for the new prefix (`[commit-tags:...]`, `[proj-link-commit:...]`).

### DIFF
--- a/internal/git/campaign_tag.go
+++ b/internal/git/campaign_tag.go
@@ -28,8 +28,11 @@ func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemR
 		shortID = shortID[:campaignTagMaxIDLen]
 	}
 
+	// Only emit the name-style head when shortID has the hex shape the parser
+	// requires (isNameStyleHead); otherwise fall back to the legacy form so the
+	// emit and parse sides cannot diverge.
 	head := legacyTagMarker + "-" + shortID
-	if nameSlug := slug.Generate(campaignName); nameSlug != "" {
+	if nameSlug := slug.Generate(campaignName); nameSlug != "" && tagNameStyleIDRe.MatchString(shortID) {
 		head = nameSlug + ":" + shortID
 	}
 
--- a/internal/git/campaign_tag.go
+++ b/internal/git/campaign_tag.go
@@ -3,18 +3,31 @@ package git
 import (
 	"regexp"
 	"strings"
+
+	"github.com/Obedience-Corp/camp/internal/slug"
 )
 
 const campaignTagMaxIDLen = 8
 
-// FormatContextTagsFull composes the consolidated campaign tag from any
-// subset of the four components. Component order inside the bracket is
-// fixed: campaign id, then quest id (qst_<...>), then festival ref
-// (FE-<ref>), then workitem ref (WI-<ref>). Absent components are
-// omitted entirely; their separators do not appear in the output.
+// legacyTagMarker is the fixed leading token campaign tags carried before they
+// embedded the campaign name. Tags formatted without a resolvable campaign name
+// fall back to "[<legacyTagMarker>-<id>...]", and ParseTag still recognizes this
+// form so the entire pre-existing commit history resolves correctly.
+const legacyTagMarker = "OBEY-CAMPAIGN"
+
+// FormatContextTagsFull composes the consolidated campaign tag from any subset
+// of (campaign name, campaign id, quest id, festival ref, workitem ref).
+//
+// The leading token is the slugified campaign name followed by ":" and the
+// short campaign id, e.g. "[obey-campaign:8deed8b4]". When campaignName has no
+// usable slug (empty or unslugifiable), it falls back to the legacy
+// "[OBEY-CAMPAIGN-<id>]" form. After the leading token the component order is
+// fixed: quest id (qst_<...>), then festival ref (FE-<ref>), then workitem ref
+// (WI-<ref>). Absent components are omitted entirely; their separators do not
+// appear in the output.
 //
 // Returns "" when campaignID is empty (no tag without a campaign).
-func FormatContextTagsFull(campaignID, questID, festRef, workitemRef string) string {
+func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemRef string) string {
 	if campaignID == "" {
 		return ""
 	}
@@ -22,7 +35,13 @@ func FormatContextTagsFull(campaignID, questID, festRef, workitemRef string) str
 	if len(shortID) > campaignTagMaxIDLen {
 		shortID = shortID[:campaignTagMaxIDLen]
 	}
-	parts := []string{"OBEY-CAMPAIGN", shortID}
+
+	head := legacyTagMarker + "-" + shortID
+	if nameSlug := slug.Generate(campaignName); nameSlug != "" {
+		head = nameSlug + ":" + shortID
+	}
+
+	parts := []string{head}
 	if questID != "" {
 		parts = append(parts, questID)
 	}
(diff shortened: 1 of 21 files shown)


pkg/commitkit/commitkit.go
```diff
@@ -49,17 +50,20 @@ func PrependCampaignTag(campaignID, message string) string {
...
-// subset of (campaign id, quest id, festival ref, workitem ref). Component
-// order is fixed: campaign → quest → festival → workitem.
-func FormatContextTagsFull(campaignID, questID, festRef, workitemRef string) string {
-	return git.FormatContextTagsFull(campaignID, questID, festRef, workitemRef)
+// subset of (campaign name, campaign id, quest id, festival ref, workitem
+// ref). The leading token is the slugified campaign name plus the short id
+// ("[{name}:{id}]"), falling back to the legacy "[OBEY-CAMPAIGN-{id}]" form
+// when campaignName is empty. Remaining order is fixed: quest → festival →
+// workitem.
+func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemRef string) string {
```
obey-agent: `pkg/commitkit` is documented as the stable public API for external tools, but this changes `FormatContextTagsFull` from four args to five. Any downstream tool that updates camp without a matching source change now fails to compile, which is avoidable here. Please keep the old id-only signature as a compatibility wrapper and add a new named API for the campaign-name form, or move this to an options struct/versioned helper before updating internal callers.

pkg/commitkit/commitkit.go
```diff
@@ -31,15 +31,16 @@ type CommitOptions struct {
...
+// string. Truncates campaignID to 8 characters. Returns empty string if
+// campaignID is empty. Callers that have the campaign name should use
+// FormatContextTagsFull to emit a name-style "[{name}:{id}]" tag instead.
 func FormatCampaignTag(campaignID string) string {
 	return git.FormatCampaignTag(campaignID)
 }
 
-// PrependCampaignTag prepends the campaign tag to a commit message.
-// If campaignID is empty, returns the message unchanged.
+// PrependCampaignTag prepends the legacy id-only campaign tag to a commit
```
obey-agent: Keeping the id-only helper for legacy/public callers is reasonable, but one in-repo command still accidentally depends on it: `cmd/camp/refs/commands.go:99` uses `git.PrependCampaignTag(cfg.ID, msg)`, so `camp refs-sync` will continue producing `[OBEY-CAMPAIGN-...]` commits after this PR. That undercuts the stated goal for campaign-created commits. Please thread `cfg.Name` through that path with `PrependContextTagsFull` and add an integration assertion on the refs-sync commit subject so this caller does not drift again.

internal/git/campaign_tag.go
```diff
@@ -120,36 +156,50 @@ type TagParseWarning struct {
...
 func ParseTag(subject string) TagComponents {
 	tc, _ := ParseTagDetailed(subject)
 	return tc
 }
 
-// ParseTagDetailed is the warnings-aware peer of ParseTag. It continues
-// scanning past unknown segments, applies shape checks to extracted
-// component values, zeros out failures, and records each fixup in the
-// returned warnings slice.
+// ParseTagDetailed is the warnings-aware peer of ParseTag. It recognizes both
```
obey-agent: This doc block has grown into a walkthrough of the implementation. The parser below also has long comments for the regex, the state walk, name-style detection, and the branch for non-tags, so the file now repeats the same contract in several places. Please trim these comments to the durable contract and the non-obvious invariants only; the current prose density makes future parser changes harder because maintainers have to audit narrative text as well as tests.

obey-agent (commented):
**Verdict:** Request Changes

**Overview:** This PR moves campaign commit tags in the right direction and covers the core formatter/parser cases well, including legacy parsing and false-positive bracket prefixes. I am requesting changes because it currently breaks the exported `pkg/commitkit` helper signatures and misses at least one in-repo campaign commit path.

**Key Findings:**
- `pkg/commitkit.FormatContextTagsFull` / `PrependContextTagsFull` changed exported stable API signatures; keep compatibility wrappers or add new named APIs before updating internal callers.
- `camp refs-sync` still uses the legacy id-only tag path at `cmd/camp/refs/commands.go:99`, so it will continue emitting `[OBEY-CAMPAIGN-...]` commits. Please thread `cfg.Name` there and assert the subject in integration coverage.
- The parser comments are unnecessarily long and repeat the same contract in multiple places. Trim them to durable invariants so the implementation and tests carry the detail.

**What's Done Well:** The parser keeps legacy tags readable, adds explicit gating for name-style tags, and has useful adversarial coverage for non-tag bracket prefixes and embedded/revert subjects. The main `camp commit`, `camp p commit`, worktree commit, and workitem commit paths are wired to the campaign name.

**Staff Standard:** Not yet. The direction is sound, but I would not ship this as my own until the public API compatibility and missed `refs-sync` path are fixed.

**Verification:** Ran `go test ./internal/git ./pkg/commitkit ./cmd/camp/refs` and `go test ./internal/git/commit ./internal/commands/workitem`; both passed.

lancekrogers:
Addressed the requested changes in 4682f35:

1. **Public API compat** — reverted `commitkit.FormatContextTagsFull` / `PrependContextTagsFull` to their original id-only signatures (they now emit the legacy form), and added `FormatContextTagsFullNamed` / `PrependContextTagsFullNamed` for the campaign-name form. Internal callers use the `*Named` variants, so no downstream tool breaks on recompile.
2. **refs-sync** — `cmd/camp/refs/commands.go` now threads `cfg.Name` via `PrependContextTagsFull`, and `TestIntegration_RefsSyncAtomic` asserts the commit subject starts with `[refs-sync:<id>]`.
3. **Comments** — trimmed the doc/inline prose in `campaign_tag.go` and `commit.go` down to durable invariants.

`go test ./...` and `just lint` (0 issues) are green.

internal/git/campaign_tag.go
```diff
@@ -3,26 +3,37 @@ package git
...
 		return ""
 	}
 	shortID := campaignID
 	if len(shortID) > campaignTagMaxIDLen {
 		shortID = shortID[:campaignTagMaxIDLen]
 	}
-	parts := []string{"OBEY-CAMPAIGN", shortID}
+
+	head := legacyTagMarker + "-" + shortID
+	if nameSlug := slug.Generate(campaignName); nameSlug != "" {
```
obey-agent: The formatter switches to the name-style head (`<slug>:<shortID>`) whenever `slug.Generate(campaignName)` is non-empty, but it never validates that `shortID` is hex. The parser side gates name-style detection on `tagNameStyleIDRe = ^[0-9a-f]{1,8}$` (see `isNameStyleHead`). Today this round-trips only because campaign IDs are always `uuid.New()` hex (`internal/config/campaign.go`), so the formatter and parser agree by coincidence rather than by contract.

If a non-hex id ever reached this path, the formatter would emit a name-style tag that the parser silently drops, losing campaign attribution on those commits. Consider gating this branch on the same hex shape the parser requires (reuse `tagNameStyleIDRe` or assert it), so the emit and parse sides can't diverge if the id format ever changes.

Non-blocking, since it can't manifest with the current uuid ids, but it tightens a latent coupling in a function whose whole job is to round-trip.

obey-agent (commented):
## Staff review (reviewed at head `4682f352`)

**Verdict: Comment, merge-ready** (one optional hardening inline; leaving the Approve click for Lance)

### Overview
Replaces the hardcoded `OBEY-CAMPAIGN` tag token with the campaign name in `<slug>:<shortID>` form, while keeping the short id (the stable machine identity history keys on). The backward-compat story is genuinely well done: dual-form parser, body-scan regex, a round-trip property test, and negative cases. It extends the existing tag functions and threads the name through `commit.Options`, so ~18 call sites needed no edits rather than a parallel implementation.

### Resolved since opening
An earlier-head concern was that the `camp refs-sync` path still emitted the legacy `OBEY-CAMPAIGN-<id>` tag (`cmd/camp/refs/commands.go`), which contradicted the PR description's claim that "the submodule-ref / refs-sync paths thread `cfg.Name` explicitly." This is **already fixed** in the latest commit ("fix(commitkit): preserve public API, fix refs-sync tag, trim comments"): line 99 now calls `git.PrependContextTagsFull(cfg.Name, cfg.ID, "", "", "", msg)`. Good catch by whoever pushed that.

### Remaining finding (optional hardening, inline on `internal/git/campaign_tag.go:32`)
- The formatter emits name-style for any non-empty slug without validating `shortID` is hex, while the parser gates on `^[0-9a-f]{1,8}$`. They agree today only because ids are always uuid hex. Worth gating the emit side on the same shape so the two can't silently diverge. Non-blocking.

### Verified clean
- `camp workitem commits` still resolves both tag forms (greps the unchanged `-WI-<ref>` segment).
- `campaign_name,omitempty` is additive, so no `--json` contract break.
- Legacy / revert / `[wip]` / `[scope:msg]` cases handled; no `fmt.Errorf` introduced; ctx propagated; file under 500 lines.

### Staff standard
Yes, I would submit this. The dual-form compatibility work and the decision to thread the name through `commit.Options` instead of forking the tag logic are exactly the right calls. Merge-ready.

lancekrogers:
Addressed the remaining hardening finding in fd69f97: the formatter now gates the name-style head on the same `^[0-9a-f]{1,8}$` shape the parser requires, so a non-hex id falls back to the legacy form instead of producing a tag the parser would drop. Added an `AllCombinations` case covering it. The stacked follow-up #359 has been rebased onto this so it carries the fix too.

PR #359, opened 2026-06-30
refactor(git): collapse workitem tag prefix from WI-WI- to WI-

lancekrogers:
Follow-up to #358. **Targets the #358 branch** (`tag-use-campaign-name`); GitHub will retarget this to `main` once #358 merges.

## What

The workitem segment of the commit tag was doubled: `WI-WI-abcdef`. The workitem ref is already self-identifying (it starts with `WI-`, per `internal/workitem/ref.go::Derive`), so the extra `WI-` segment marker was redundant. This emits the ref verbatim as a single `WI-` segment, the same way quest ids lead with `qst_`.

- **Before:** `[obey-campaign:8deed8b4-WI-WI-abcdef]`
- **After:** `[obey-campaign:8deed8b4-WI-abcdef]`

## Backward compatibility

The parser accepts **both** the new single-prefix `WI-abcdef` and the historical doubled `WI-WI-abcdef`, so every commit written before this change still resolves. `TagComponents.WorkitemRef` is unchanged (`WI-abcdef`) for both forms.

## Tests / gates

- `go build ./...`, `go test ./...`, dev build, integration compile, and `just lint` (0 issues) all green.
- Tag tests keep the doubled-form parse cases (back-compat) and add single-form format + parse + round-trip coverage.
- Integration assertions and plan fixtures updated to the single-prefix form.

### DIFF
--- a/internal/git/campaign_tag.go
+++ b/internal/git/campaign_tag.go
@@ -47,9 +47,9 @@ func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemR
 		if !strings.HasPrefix(workitemRef, "WI-") {
 			workitemRef = "WI-" + workitemRef
 		}
-		// The ref already starts with WI-, so the WI- segment marker produces
-		// the intentional WI-WI- double prefix.
-		parts = append(parts, "WI-"+workitemRef)
+		// The ref already carries WI-, so it is self-identifying and embedded
+		// verbatim (no extra marker).
+		parts = append(parts, workitemRef)
 	}
 	return "[" + strings.Join(parts, "-") + "]"
 }
@@ -199,20 +199,24 @@ func ParseTagDetailed(subject string) (TagComponents, []TagParseWarning) {
 			}
 			rest = after
 		case strings.HasPrefix(rest, "WI-"):
-			// WI- is always the last segment, so the remainder is the ref.
-			payload := rest[len("WI-"):]
-			if !tagWorkitemRefRe.MatchString(payload) {
+			// WI- is the last segment. Accept the single-prefix ref (WI-abcdef)
+			// and the historical doubled form (WI-WI-abcdef).
+			ref := rest
+			if strings.HasPrefix(ref, "WI-WI-") {
+				ref = ref[len("WI-"):]
+			}
+			if !tagWorkitemRefRe.MatchString(ref) {
 				warnings = append(warnings, TagParseWarning{
-					Field: "workitem_ref", Value: payload,
+					Field: "workitem_ref", Value: ref,
 					Reason: "shape check failed (want WI-<6 hex>)",
 				})
 			} else if out.WorkitemRef != "" {
 				warnings = append(warnings, TagParseWarning{
-					Field: "workitem_ref", Value: payload,
+					Field: "workitem_ref", Value: ref,
 					Reason: "duplicate workitem_ref segment",
 				})
 			} else {
-				out.WorkitemRef = payload
+				out.WorkitemRef = ref
 			}
 			rest = ""
 		default:
--- a/docs/workitem-commit-reference.md
+++ b/docs/workitem-commit-reference.md
@@ -118,7 +118,7 @@ skipped:
   <path> (out of scope)
   <path> (submodule pointer; use --include-submodule-pointer)
   <path> (--exclude)
-tag:    [<campaign-name>:<8hex>[-qst_<...>][-FE-<festival-ref>][-WI-WI-<6hex>]]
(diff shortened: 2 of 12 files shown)


internal/git/campaign_tag.go
```diff
@@ -47,9 +47,9 @@ func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemR
 		if !strings.HasPrefix(workitemRef, "WI-") {
 			workitemRef = "WI-" + workitemRef
 		}
-		// The ref already starts with WI-, so the WI- segment marker produces
-		// the intentional WI-WI- double prefix.
-		parts = append(parts, "WI-"+workitemRef)
+		// The ref already carries WI-, so it is self-identifying and embedded
+		// verbatim (no extra marker).
+		parts = append(parts, workitemRef)
```
obey-agent, line 52: This changes newly written commit subjects to the single-prefix form, but `camp workitem commits` still prefilters history with `grep := "-WI-" + ref` in `internal/commands/workitem/commits.go`. With `ref == "WI-abc123"`, that searches for `-WI-WI-abc123`, so new commits like `[...-WI-abc123]` are never returned and the parser compatibility path never gets a chance to run.

Please update the query path to find the new format while preserving historical doubled tags. I confirmed the current PR head fails:

`just test integration-run 'TestIntegration_WorkitemCommits_(CrossRepo|IncludesCampaignAndLinkedProject)'`

Both tests return empty commit result sets after creating new single-prefix workitem commits.

obey-agent (commented):
**Verdict: Request Changes**

This PR correctly updates the formatter and parser unit coverage for the new single-prefix `WI-<hex>` tag shape, and it keeps parser compatibility for historical `WI-WI-<hex>` subjects. The remaining issue is in the cross-repo history lookup path: newly written single-prefix commits are no longer discoverable because the `camp workitem commits` grep prefilter still searches for the old doubled shape before parsing candidates.

**Key Findings**
- `camp workitem commits` still builds its prefilter as `"-WI-" + ref`, which becomes `-WI-WI-abc123` for a canonical `WI-abc123` ref. New subjects emitted by this PR contain `-WI-abc123`, so they are filtered out before `commitkit.ParseTag` can validate them. I left an inline comment on the formatter change that introduces the new emitted subject shape.

**Validation**
- Failed on current head `8d041af`: `just test integration-run 'TestIntegration_WorkitemCommits_(CrossRepo|IncludesCampaignAndLinkedProject)'`
- Failures: `TestIntegration_WorkitemCommits_IncludesCampaignAndLinkedProject` returned no root/project repos, and `TestIntegration_WorkitemCommits_CrossRepo` returned an empty commits array.

**What's Done Well**
- The parser-side backward compatibility is straightforward and covered for both legacy doubled tags and new single-prefix tags.
- The docs and plan fixtures are aligned with the new emitted format.

**Staff Standard**
Not ready to merge yet. I would want the query path fixed and the targeted integration test passing before submitting this as my own work.

lancekrogers:
Good catch, fixed in 948ab70. The `camp workitem commits` prefilter built `git log --grep="-WI-"+ref` → `-WI-WI-<ref>`, which no longer matched the new single-prefix `-WI-<ref>` subjects. Changed it to anchor on `"-"+ref`, which is a substring of both the new `-WI-abc123` and the legacy `-WI-WI-abc123` forms (the post-grep `parts.WorkitemRef != ref` exact check already handled both).

Verified against real `git log --grep`: old prefilter matched only the legacy commit; the fix matches both forms and still skips unrelated commits. Docker isn't available in my env to run `integration-run` directly, but this is the exact filter those two tests fail on.

obey-agent (commented):
**Verdict: Approve**

I re-reviewed the current head `948ab70` after the query-path fix. The previous blocker is resolved: `camp workitem commits` now prefilters with `"-" + ref`, which finds newly emitted single-prefix subjects like `-WI-abc123` while still matching historical doubled subjects like `-WI-WI-abc123` before `commitkit.ParseTag` normalizes them.

**Key Findings**
- No merge-blocking findings in this pass.

**Validation**
- Passed: `go test ./internal/git ./pkg/commitkit`
- Passed: `just test integration-run 'TestIntegration_WorkitemCommits_(CrossRepo|IncludesCampaignAndLinkedProject)'`

**What's Done Well**
- The formatter/parser change is small and targeted.
- Backward compatibility for historical doubled tags is preserved in parsing, and the query prefilter now reaches both old and new forms.
- The previously failing cross-repo query path is covered by the targeted integration tests.

**Staff Standard**
Yes, I would be comfortable submitting this version as my own work.

PR #360, opened 2026-06-30
feat(commitkit): expose campaign name accessors

lancekrogers:
## Why

After #358, `commitkit` can emit name-style tags via `FormatContextTagsFullNamed` / `PrependContextTagsFullNamed`, but it only exposes the campaign **id** (`DetectCampaign`, `LoadCampaignID`). External callers — specifically `fest commit` — have no public way to get the campaign **name**, so they can't build name-style tags.

This adds the missing accessors so `fest` can stop emitting the hardcoded `[OBEY-CAMPAIGN-...]` form.

## What

- `commitkit.DetectCampaignName(ctx)` — walks up from cwd, returns the campaign name (mirrors `DetectCampaign`).
- `commitkit.LoadCampaignName(ctx, campaignRoot)` — reads the name from `campaign.yaml` (mirrors `LoadCampaignID`).

Both are thin, additive wrappers over `config.LoadCampaignConfig`. No existing signatures change.

## Follow-up

This unblocks the `fest` PR: bump the camp dep to a commit containing this, then replace the four hardcoded tag sites (`internal/commands/commit/commit.go:471` + `:476`, `status/handlers_festival.go:507`, `ritual/run.go:200`) with name-aware calls.

## Tests / gates

`go test ./pkg/commitkit/...`, `go build ./...`, and `just lint` (0 issues) green. Added `TestLoadCampaignName` / `TestDetectCampaignName` mirroring the existing id-accessor tests.

### DIFF
--- a/pkg/commitkit/commitkit.go
+++ b/pkg/commitkit/commitkit.go
@@ -112,6 +112,33 @@ func LoadCampaignID(ctx context.Context, campaignRoot string) (string, error) {
 	return cfg.ID, nil
 }
 
+// DetectCampaignName is DetectCampaign for the campaign name, for callers that
+// build name-style tags (FormatContextTagsFullNamed).
+func DetectCampaignName(ctx context.Context) (string, error) {
+	root, err := campaign.DetectFromCwd(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	cfg, err := config.LoadCampaignConfig(ctx, root)
+	if err != nil {
+		return "", camperrors.Wrapf(err, "commitkit: load campaign config at %s", root)
+	}
+
+	return cfg.Name, nil
+}
+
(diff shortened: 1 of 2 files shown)


obey-agent (commented):
**Verdict: Approve**

This is a small, additive commitkit API expansion that mirrors the existing campaign ID accessors and keeps errors wrapped through the existing internal error package. I did not find blocking issues.

Key findings:
- None.

What is done well:
- The new functions reuse `campaign.DetectFromCwd` and `config.LoadCampaignConfig` instead of adding a parallel config path.
- Tests mirror the existing ID-accessor behavior for valid and missing campaign config cases.

Verification: `go test ./pkg/commitkit/...` passed against PR head `df9bc95298438c3e4406c8e9a4529e3b66bb224f`.

Staff standard: Yes, I would be comfortable submitting this as my own work.

PR #361, opened 2026-06-30
fix(worktrees): keep git worktrees out of campaign status and commits

lancekrogers:
## Problem

In a campaign, `camp status` and `camp commit` at the campaign root surfaced and staged git worktrees living under `projects/worktrees/`.

Root cause: nothing ignored the worktrees directory at the campaign root.

- `camp status` just runs `git status` on the root (`cmd/camp/status.go`). Worktrees are ordinary untracked directories in the superproject working tree, so git reports them.
- `camp commit` at the root stages with `git add -- . :(exclude,literal)<submodule>` (`internal/git/commit.go` `StageAllExcluding`). The exclude list is only registered submodules, so worktrees got swept in and staged as **embedded git repositories**.
- The scaffold never writes a root `.gitignore`, so a fresh `camp init` campaign has no worktrees ignore rule at all. Campaigns that don't show this (e.g. the obey campaign) only avoid it via a hand-added `worktrees/` line.

## Fix

Make the fix git-native (one rule fixes status, commit, add, and raw git) plus a defense-in-depth guard on commit:

1. **`camp init` ensures the campaign root `.gitignore` ignores the worktrees directory** with an anchored rule (`/projects/worktrees/`, derived from the configured worktrees path). It creates the file when absent and honors any existing equivalent rule (anchored, unanchored, or a bare `worktrees/`) so it never duplicates.
2. **`camp init --repair` backfills the rule** for existing campaigns (added to the repair plan + applied on init).
3. **`camp commit` excludes the worktrees directory from root staging** as defense-in-depth for campaigns with a customized worktrees path or a hand-edited ignore file.

`camp status` is fixed transitively by the gitignore rule; its git-flag passthrough is left untouched.

## Tests

- `internal/scaffold/worktree_gitignore_test.go`: rule formatting, equivalence detection across rule forms, and create/append/no-op/idempotency for the ensure helper.
- `cmd/camp/commit_test.go`: `worktreesExcludeSpec` default fallback.
- Updated `TestComputeMiscFileChanges_FilesExist` fixture to include the root `.gitignore`.

`just gate` green (gate-fast: stable+dev build, vet, fmt, lint; plus stable unit tests). All 2902 tests pass.

## Notes

Existing campaigns surfacing worktrees today are fixed by running `camp init --repair`, which writes the missing rule.

### DIFF
--- a/internal/scaffold/init.go
+++ b/internal/scaffold/init.go
@@ -363,6 +363,20 @@ workitems/current.yaml
 		}
 	}
 
+	if !opts.DryRun {
+		worktreesPath := jumps.Paths.Worktrees
+		if worktreesPath == "" {
+			worktreesPath = config.DefaultCampaignPaths().Worktrees
+		}
+		created, _, giErr := ensureRootGitignoreWorktrees(absDir, worktreesPath)
+		if giErr != nil {
+			return nil, giErr
+		}
+		if created {
+			result.FilesCreated = append(result.FilesCreated, filepath.Join(absDir, ".gitignore"))
+		}
+	}
+
 	// Create CLAUDE.md symlink to AGENTS.md (AGENTS.md is the source of truth)
 	claudePath := filepath.Join(absDir, "CLAUDE.md")
 
--- a/internal/scaffold/init.go
+++ b/internal/scaffold/init.go
@@ -67,6 +67,8 @@ type InitResult struct {
 	DirsCreated []string
 	// FilesCreated lists files that were created.
 	FilesCreated []string
+	// FilesModified lists existing files that were modified in place.
+	FilesModified []string
 	// Skipped lists items that were skipped (already exist).
 	Skipped []string
 	// GitInitialized indicates if a git repository was initialized.
@@ -368,12 +370,15 @@ workitems/current.yaml
 		if worktreesPath == "" {
 			worktreesPath = config.DefaultCampaignPaths().Worktrees
 		}
-		created, _, giErr := ensureRootGitignoreWorktrees(absDir, worktreesPath)
+		created, modified, giErr := ensureRootGitignoreWorktrees(absDir, worktreesPath)
 		if giErr != nil {
 			return nil, giErr
 		}
+		gitignoreAbs := filepath.Join(absDir, ".gitignore")
 		if created {
-			result.FilesCreated = append(result.FilesCreated, filepath.Join(absDir, ".gitignore"))
+			result.FilesCreated = append(result.FilesCreated, gitignoreAbs)
+		} else if modified {
+			result.FilesModified = append(result.FilesModified, gitignoreAbs)
 		}
 	}
 
--- a/internal/scaffold/repair.go
+++ b/internal/scaffold/repair.go
@@ -634,6 +634,9 @@ func computeMiscFileChanges(absDir string, plan *RepairPlan) {
 
 	rootGitignorePath := filepath.Join(absDir, ".gitignore")
 	worktreesPath := config.DefaultCampaignPaths().Worktrees
+	if plan.MergedJumps != nil && plan.MergedJumps.Paths.Worktrees != "" {
+		worktreesPath = plan.MergedJumps.Paths.Worktrees
+	}
 	if raw, err := os.ReadFile(rootGitignorePath); os.IsNotExist(err) {
 		plan.Changes = append(plan.Changes, RepairChange{
 			Type:        RepairAdd,
--- a/internal/scaffold/repair.go
+++ b/internal/scaffold/repair.go
@@ -632,6 +632,24 @@ func computeMiscFileChanges(absDir string, plan *RepairPlan) {
 		}
 	}
 
+	rootGitignorePath := filepath.Join(absDir, ".gitignore")
+	worktreesPath := config.DefaultCampaignPaths().Worktrees
+	if raw, err := os.ReadFile(rootGitignorePath); os.IsNotExist(err) {
+		plan.Changes = append(plan.Changes, RepairChange{
+			Type:        RepairAdd,
+			Category:    "file",
+			Key:         ".gitignore",
+			Description: "missing worktrees ignore rule",
+		})
+	} else if err == nil && !gitignoreIgnoresWorktrees(string(raw), worktreesPath) {
+		plan.Changes = append(plan.Changes, RepairChange{
+			Type:        RepairModify,
+			Category:    "file",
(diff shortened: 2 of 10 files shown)


internal/scaffold/repair.go
```diff
@@ -632,6 +632,24 @@ func computeMiscFileChanges(absDir string, plan *RepairPlan) {
 		}
 	}
 
+	rootGitignorePath := filepath.Join(absDir, ".gitignore")
+	worktreesPath := config.DefaultCampaignPaths().Worktrees
```
obey-agent, line 636: Requesting changes: repair planning checks the default worktrees path, but repair application writes the configured path. If a campaign has `paths.worktrees: custom/wt/` and an existing root `.gitignore` with only `/projects/worktrees/`, `camp init --repair` reports up to date and never adds `/custom/wt/`, so `camp status` still surfaces the custom worktrees directory. Please derive this check from the loaded jumps/campaign config and add a customized-path regression test.

obey-agent (commented):
**Verdict: Request Changes**

The worktrees fix is directionally right: using a root `.gitignore` plus a staging exclude covers both `camp status` and `camp commit`, and the focused tests pass. I found one correctness gap in repair planning for customized worktrees paths; see the inline comment on `internal/scaffold/repair.go`.

Key finding:
- Repair planning checks `config.DefaultCampaignPaths().Worktrees` while repair application uses the configured `jumps.Paths.Worktrees`, so `camp init --repair` can incorrectly report “up to date” and skip the `.gitignore` fix for custom worktrees layouts.

What is done well:
- The fix uses git-native ignore behavior instead of only masking `camp status` output.
- The commit path has a defense-in-depth exclude.

Staff standard: I would merge this after the customized-path repair case is fixed and covered by a regression test.

internal/scaffold/repair.go
lancekrogers: Fixed in 6731567. Confirmed this was a real functional skip, not just a wrong preview: with the plan reporting no `.gitignore` change, `camp init --repair` hits the `!plan.HasChanges()` early return in `cmd/camp/init/command.go` and never calls `Init`, so the apply never runs and the custom rule is never written.

`computeMiscFileChanges` now derives the worktrees path from `plan.MergedJumps.Paths.Worktrees` (populated from the existing config via `computeJumpsChanges`), which is the exact same source the apply reads (`jumps.Paths.Worktrees` in `Init`), so plan and apply agree for custom layouts.

Regression test added: `TestComputeMiscFileChanges_CustomWorktreesPath` sets `paths.worktrees: custom/wt/` with a root `.gitignore` containing only the default `/projects/worktrees/` rule and asserts the plan emits a `.gitignore` RepairModify. `just gate` green.

internal/scaffold/init.go
```diff
@@ -363,6 +363,20 @@ workitems/current.yaml
 		}
 	}
 
+	if !opts.DryRun {
+		worktreesPath := jumps.Paths.Worktrees
+		if worktreesPath == "" {
+			worktreesPath = config.DefaultCampaignPaths().Worktrees
+		}
+		created, _, giErr := ensureRootGitignoreWorktrees(absDir, worktreesPath)
```
obey-agent: Requesting changes: this handles the on-disk modification, but repair auto-commit loses the common existing-file case. `ensureRootGitignoreWorktrees` returns `modified=true` when it appends to an existing root `.gitignore`, but this call discards that value and only records the path when the file was newly created. `camp init --repair` commits a selective file list built from `InitResult.FilesCreated` / `DirsCreated`, so an existing `.gitignore` that gets the worktrees rule appended is left modified but uncommitted. Please thread the modified path into the repair commit file list and add a regression test for repairing an existing root `.gitignore`.

obey-agent (commented):
**Verdict: Request Changes**

The original customized-worktrees repair planning issue is fixed: the plan now derives the check from `MergedJumps.Paths.Worktrees`, and the added regression covers the previous false “up to date” case. I found one remaining repair-flow issue in the existing-root `.gitignore` path; see the inline comment on `internal/scaffold/init.go`.

Key finding:
- Existing root `.gitignore` updates are applied on disk but not included in the selective repair auto-commit because `modified` is discarded and only newly created `.gitignore` files are recorded.

What is done well:
- The git-native ignore approach is still the right fix for both `camp status` and raw git behavior.
- The customized-path repair planning regression is targeted and addresses the prior review finding.

Verification: `go test ./cmd/camp ./internal/scaffold ./internal/git` passed on PR head `673156771036c369159fa53e702b7b95daf2d2e2`.

Staff standard: I would merge after the existing-file repair commit path is fixed and covered.

internal/scaffold/init.go
lancekrogers: Fixed in 7c557a1. You're right: the existing-file append was written to disk but dropped from the selective repair commit.

- `InitResult` now has a `FilesModified` field; `Init` records the root `.gitignore` there when `ensureRootGitignoreWorktrees` returns `modified=true` (the `created` case still goes to `FilesCreated`).
- `buildRepairCommitFiles` now appends `FilesModified` to the staged list, `buildRepairCommitMessage` emits a `Files updated:` section, and the `commitRepairChanges` `hasChanges` guard counts it. Repair output prints a matching `Files updated:` block.

Regression coverage:
- `TestInit_RepairRecordsModifiedRootGitignore` (end-to-end): repairs an existing root `.gitignore` that lacks the rule, asserts the path lands in `FilesModified`, the worktrees rule is appended, and prior content is preserved.
- `TestBuildRepairCommitFiles_IncludesModifiedGitignore` and `TestBuildRepairCommitMessage_IncludesFilesModified` cover the commit-list and message plumbing.

`just gate` green (2880 tests).

obey-agent (commented):
**Verdict: Approve**

The two prior repair issues are now addressed. The repair plan uses the configured worktrees path, and in-place root `.gitignore` edits are recorded through `FilesModified`, included in the selective repair commit file list, shown in repair output, and covered by regression tests.

Key findings:
- None blocking.

What is done well:
- The fix stays git-native: the root `.gitignore` rule resolves `camp status`, `camp commit`, and raw git behavior consistently.
- The repair path now covers both newly created and existing `.gitignore` files.
- The tests cover helper behavior, customized worktrees paths, and the repair commit file-list plumbing.

Verification: `go test ./cmd/camp ./cmd/camp/init ./internal/scaffold ./internal/git` passed on PR head `7c557a1d95db53823bef6953685bd6544cf8b0a9`.

Staff standard: Yes, I would be comfortable merging this.

PR #363, opened 2026-07-01
feat(workitem-tui): bind custom workflow types to filter keys 5-9

lancekrogers:
## Problem

The `camp workitem` TUI can only filter by the four builtin workflow types (intent, design, explore, festival) via hardcoded keys 1-4. Campaigns tracking custom workitem types (`camp workitem create --type bug`) show those items in the all-items view but offer no way to filter down to them. The CLI `--type` flag already accepts custom slugs; the TUI key layer was the only closed surface.

Intent: `camp-workitems-filtering-only-filters-20260701-145447` (obey-campaign)
Design: `workflow/design/workitem-tui-custom-type-filters` (obey-campaign, WI-26efc9)

## Approach

- Keys `1-4` stay pinned to the builtins in canonical order, preserving muscle memory and existing docs.
- Custom types discovered in the loaded item set bind to keys `5-9` in deterministic alphabetical order, recomputed whenever a refresh replaces the item set.
- Footer hint becomes dynamic (`1-6 filter` with two custom types) and the help screen renders one row per bound key from the same binding source the key handler uses, so help cannot drift from behavior.
- Unbound digits are no-ops. A refresh that removes the actively filtered type keeps the filter (empty list + visible pill) rather than silently reshuffling under the cursor.

## Tradeoffs

- Digits cap custom bindings at five. Campaigns with more custom types get the first five alphabetically; the rest stay reachable via `/` search and CLI `--type`. A picker overlay can be layered on later if real campaigns outgrow this.
- Bindings for custom types can shift when the custom type set itself changes between refreshes; the help screen always shows current truth.

## Testing

- New model tests: custom key filtering, binding determinism/dedup/cap (table-driven), unbound-digit no-op, refresh rebinding, help and footer rendering.
- `just gate-push` green (build, vet, lint ratchets, full test suite).

### DIFF
--- a/internal/workitem/tui/model.go
+++ b/internal/workitem/tui/model.go
@@ -3,6 +3,8 @@ package tui
 
 import (
 	"context"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


obey-agent (commented):
**Verdict: Approve**

This is a focused TUI improvement that closes the custom workflow type filtering gap without disturbing the established builtin key bindings. The implementation keeps one source of truth for key bindings, uses deterministic discovery for custom types, and updates refresh/help/footer behavior coherently.

**Key Findings**
- No blocking findings.

**What's Done Well**
- Builtin filters stay pinned to `1-4`, preserving existing muscle memory.
- Custom type binding is deterministic, deduped, and capped to the available `5-9` key space.
- Help rendering and key handling share the same binding source, which prevents drift.
- Tests cover binding order, cap behavior, no-op unbound digits, refresh rebinding, help text, and footer text.

**Staff Standard**

Checks run:
- `go test ./internal/workitem/tui`
- `go test ./internal/workitem ./internal/workitem/tui`
- `just gate-push`

obey-agent (approved):
Approved. No blocking findings; gate-push and focused workitem TUI tests passed.

PR #364, opened 2026-07-01
fix(commit): unstage worktrees after staging instead of exclude pathspec

obey-agent:
## Problem

#362 was incomplete. It skipped the worktrees exclude pathspec only when `git check-ignore` reported the worktrees **directory itself** as ignored. That predicate breaks in campaigns where the ignored worktrees dir contains **tracked content** (a `.gitkeep`, committed worktree gitlinks): tracked content keeps the directory itself 'not ignored' per check-ignore, so the exclude pathspec stayed - and `git add` still refuses it because ignored entries (e.g. `.DS_Store`, pruned worktree dirs matched by a dir-only `projects/worktrees/` rule) exist beneath it. Root `camp commit` kept failing with:

```
Error: git add failed: The following paths are ignored by one of your .gitignore files:
projects/worktrees
```

Reproduced against a real user campaign in exactly this shape. No exclude pathspec variant dodges the advice check - `:(exclude,literal)path`, trailing-slash, and `:(exclude,glob)path/**` all fail identically on git 2.54.

## Fix

Stop naming the worktrees path in the add pathspec at all:

1. `git add -- . :(exclude,literal)<submodule>...` - submodule excludes only. Submodule paths are tracked gitlinks and can never be gitignored, so they cannot trip the advice check.
2. `git reset -q -- <worktrees>` afterward (new `git.UnstagePath`, lock-retried like the other index operations; verified safe on unborn branches and a no-op when nothing under the path is staged).

`camp stage` gets the same flow. `git.PathIgnored` and the #362 predicate are removed.

Behavior notes:
- A dirty worktree gitlink is staged then unstaged within the command; if the reset fails the command errors before committing.
- Anything a user manually pre-staged under the worktrees path is also unstaged - consistent with #361's intent that worktrees content never lands in campaign root commits.

## Testing

- Containerized integration tests (no host filesystem): new `TestIntegration_CommitSucceedsWithTrackedContentUnderIgnoredWorktrees` models the failing user campaign (init-written `/projects/worktrees/` rule + tracked .gitkeep + dirty committed gitlink + ignored child dir) and asserts commit succeeds, worktrees stay out of the commit, and the dirty gitlink remains unstaged. The three #362 tests still pass unchanged.
- Removed the host-`t.TempDir` unit test for the path helper per the no-host-FS-tests rule; the container suite covers it.
- `just gate`: builds/vet/lint green; `internal/git` and `internal/doctor/checks` hit the dashboard's 120s per-package cap under machine load but pass standalone in 21.8s/9.0s with zero test failures - pre-existing harness margin issue, tracked separately.

Follow-up after merge: verify with the installed binary against the real affected campaign, then retry the v0.2.16 release.


obey-agent (commented):
**Verdict: Approve**

I reviewed this as a correctness-sensitive git/index change rather than a simple pathspec tweak. The new approach preserves the root-commit safety contract by no longer naming the worktrees path in `git add`, then removing any staged worktrees entries with the same lock-retry pattern used by other index mutations.

**Key Findings**

- No blocking issues found.
- I specifically checked the two-step index behavior for ignored worktrees, unignored worktrees, tracked content under an ignored worktrees directory, unborn branch reset behavior, and the default root `commit` / `stage` paths with refs excluded.

**What's Done Well**

- The fix removes the fragile `git check-ignore` predicate instead of layering another heuristic onto it.
- `git.UnstagePath` is small, context-aware through `exec.CommandContext`, uses the existing retry/lock classification pattern, and keeps the working tree untouched.
- The new integration test models the real regression shape: tracked content under the ignored worktrees directory, a dirty gitlink, and ignored child content beneath the same directory.
- Removing the host-TempDir unit test is consistent with the repository's filesystem-test policy.

**Verification**

- `go test -tags integration ./tests/integration -run 'TestIntegration_(Commit|Stage).*Worktrees' -count=1` passed in 73.401s.
- `go test ./internal/git -run 'TestStageAllExcluding|TestHasChanges|TestHasStagedChanges|TestExecutor_HasChanges' -count=1` passed.
- `go test ./cmd/camp -run 'TestEffectiveCommitAll|TestCommitHint|Test.*Stage|Test.*Commit' -count=1` passed.
- `just gate-push` passed. golangci emitted warnings about stale sibling worktree paths, but reported `0 issues`.

**Staff Standard**

PR #365, opened 2026-07-01
feat(workitem-tui): replace 5-9 custom type bindings with f filter mode

obey-agent:
## Problem

#363 bound custom workflow types to digit keys 5-9. Review feedback rejected that design on two grounds:

1. Campaigns commonly track more than five custom workflow types, and the digit slots cap at five.
2. A number-to-type mapping that reshuffles alphabetically whenever a workflow type is added or removed can never become muscle memory. Digits only work for the builtins because that set is fixed.

Design revision: `workflow/design/workitem-tui-custom-type-filters` 02-design.md rev 2 (obey-campaign, WI-26efc9).

## Approach

`f` enters a filter mode following the footer-mode pattern the dashboard already uses for priority (`P`) and stage (`S`):

- The footer becomes a chip row listing "all" plus every workflow type present in the loaded items, each with its item count, active chip highlighted: `[all 37]  intent 12  design 9  bug 2`.
- `j/k`, `h/l`, and arrow keys step through the chips, refiltering the list live on every step, so the user watches real results instead of reading a menu.
- `enter` (or `f`) commits; `esc` restores the filter active at mode entry, mirroring search mode's snapshot/restore semantics; `0` jumps to the all chip.
- Chip options rebuild from the item set on every mode entry (and on refresh mid-mode), so adding/removing workflow types self-heals. No cap: the chip row windows around the active chip when it overflows the terminal width.
- The 5-9 custom bindings are removed. `0` and `1-4` stay as stable builtin accelerators.

`internal/intent/tui/filterchip` was considered for reuse but its contract is dropdown-select (focus/open/highlight/commit), not live-refilter stepping, so filter mode is implemented natively alongside the existing modes.

## Testing

- New model tests: chip option derivation (presence-based, canonical builtin order, customs sorted, uncapped), live stepping with clamping, enter commit, esc restore, 0 jump, index positioning on the active filter at entry, refresh-mid-mode rebuild, 5-9 regression (digits now unbound), footer chip rendering with counts, chip window math (table-driven).
- `just gate-push` green (build, vet, lint ratchets, full test suite).


internal/workitem/tui/model.go
```diff
@@ -126,55 +134,80 @@ func New(ctx context.Context, items []workitem.WorkItem, campaignRoot string, re
...
+
+func (m Model) isFilterMode() bool {
+	return m.filterMode
+}
+
+// rebuildFilterOptions refreshes filter-mode chips after the item set
+// changed, re-locating the active chip by value.
+func (m *Model) rebuildFilterOptions() {
+	m.filterOptions = filterOptions(m.allItems)
+	m.filterIndex = 0
```
obey-agent, line 178: This rebuild can leave the model in an internally inconsistent filter state. If the active type is no longer present after refresh, `filterIndex` falls back to 0 but `typeFilter` remains on the removed value; the footer then highlights `[all ...]` while `refilter()` still applies the stale type and shows an empty list. The new test at `TestModel_FilterModeRefreshRebuildsOptions` actually encodes that stale `bug` filter by expecting zero items after `bug` disappears.

Please make the chip selection and applied filter a single source of truth after rebuilding: either keep the stale active type as an explicit zero-count chip, or normalize `typeFilter` to the selected rebuilt option and refilter. This also needs a regression assertion that the highlighted chip, `typeFilter`, and `filteredItems` agree after refresh and after entering filter mode from an absent builtin accelerator.

internal/workitem/tui/view.go
```diff
@@ -98,15 +98,81 @@ func (m Model) renderFooter() string {
...
+	for i, opt := range m.filterOptions {
+		if opt == "" {
+			labels[i] = fmt.Sprintf("all %d", total)
+		} else {
+			labels[i] = fmt.Sprintf("%s %d", opt, counts[opt])
+		}
+	}
+
+	const prefix = "filter: "
+	avail := m.width - len(prefix) - 4
```
obey-agent: This width budget is too small to guarantee the footer stays within the terminal width. `chipWindow` budgets only label lengths plus a fixed separator, but the rendered row also adds active-chip brackets, optional left/right ellipses, and the join separators around those ellipses. For example, at width 20 with an active middle chip like `bug 10`, `chipWindow` can return only the active chip, then `renderFilterChips` still renders `filter: …  [bug 10]  …`, which is wider than the terminal and will wrap the footer.

Because the PR explicitly promises that overflowing chip rows window around the active chip, please either reserve space for ellipses/separators/brackets before choosing the window or measure the final rendered width with `lipgloss.Width` and shrink/truncate until it fits. Add a narrow-width regression test where the active chip is in the middle and both ellipses would render.

obey-agent (commented):
**Verdict: Request Changes**

This PR is directionally aligned with the design revision: replacing unstable custom digit bindings with an explicit `f` filter mode is the right interaction model, and the tests cover the basic stepping/commit/cancel behavior. However, the implementation currently lets the visible chip selection diverge from the applied `typeFilter`, and the overflow math does not actually guarantee the chip footer fits in narrow terminals.

**Key Findings**

- `rebuildFilterOptions` can leave a stale `typeFilter` active while highlighting the `all` chip after refresh removes the active type. The same mismatch is reachable by entering filter mode after a builtin accelerator selects a type not present in the options. The model needs one source of truth between `filterIndex`, `filterOptions`, and `typeFilter`, plus regression assertions for refresh and absent-active-filter entry.
- `renderFilterChips` / `chipWindow` under-budget the rendered footer width because the calculation omits active brackets, ellipses, and separators around ellipses. The promised overflow behavior can still wrap the footer in a narrow terminal when the active chip is in the middle.

**What's Done Well**

- Removing the 5-9 custom bindings avoids the unstable muscle-memory problem from the previous design.
- The mode ordering and live refiltering are simple and localized, and the existing search snapshot/cancel semantics were mirrored cleanly.
- The test suite added useful coverage for normal stepping, commit, escape restore, zero jump, removed custom digit bindings, and basic chip rendering.

**Verification**

- `go test ./internal/workitem/...` passed.
- `just gate-push` passed. golangci emitted warnings about stale sibling worktree paths, but reported `0 issues`.

**Staff Standard**

Not yet. I would be comfortable with this once the filter-state mismatch is fixed and the footer width guarantee is tested against the actual rendered output.

lancekrogers:
Both findings addressed in 4b949f3.

**Chip/filter divergence**: took the zero-count chip option. `filterOptions` now always includes the active filter (as an explicit chip with count 0 when no items carry that type), so `filterOptions[filterIndex] == typeFilter` holds by construction after every rebuild. The kept filter renders as `[bug 0]` over the empty list instead of silently highlighting `all`, and the chip drops out on the next mode entry once the filter moves off it. Regression tests: `TestModel_FilterModeRefreshRebuildsOptions` (refresh removes the active type) and `TestModel_FilterModeEntryWithAbsentBuiltin` (accelerator selects an empty builtin, then `f`), both asserting the highlighted chip, applied filter, and filtered list agree via a shared `assertFilterStateConsistent` helper.

**Footer width**: `chipWindow` now validates every candidate expansion against `chipRowWidth`, which counts the exact rendered row: labels, active brackets, ellipsis markers, and all join separators. Labels truncate when the terminal is narrower than a single bracketed chip flanked by two ellipses, and below that the row falls back to plain text truncated before styling, so no width can wrap the footer (styles are applied after truncation, so no ANSI splitting). `TestModel_FilterChipsFitNarrowWidth` asserts `lipgloss.Width(footer) <= width` for the middle-active double-ellipsis case at width 30 and across widths 12 through 60.

`just gate-push` green at 4b949f3.

internal/workitem/tui/model.go
```diff
@@ -126,55 +129,74 @@ func New(ctx context.Context, items []workitem.WorkItem, campaignRoot string, re
...
+
+func (m Model) isFilterMode() bool {
+	return m.filterMode
+}
+
+// rebuildFilterOptions derives chips from the current item set. The active
+// filter is always among the options (filterOptions ensures it), so the
+// highlighted chip and the applied filter cannot diverge.
+func (m *Model) rebuildFilterOptions() {
+	m.filterOptions = filterOptions(m.allItems, m.typeFilter)
```
obey-agent: This still derives the available chips from `m.allItems`, but after a refresh `m.allItems` is the raw discovered set while the actual list and displayed counts apply `ShowParked` in `refilter()` / `visibleTypeCounts()`. The initial TUI load gets pre-filtered by `runTUI`, but refresh does not; so with the default `showParked=false`, a workflow type that only exists on parked items will appear as a non-active `type 0` chip. Stepping onto it live-refilters to an empty list, even though the chip was presented as an available option.

That breaks the filter-mode contract that chips represent real visible result sets, and it is the same class of state drift we fixed for stale active filters. Please build the base chip list from the same visible item set used by `refilter`/`visibleTypeCounts` (or otherwise exclude parked-only types), then add `m.typeFilter` as the explicit zero-count exception only when needed to keep the current active filter representable. A regression should cover refresh with `showParked=false` and a parked-only custom type: it should not appear as a selectable zero-count chip unless it is the active filter.

obey-agent (commented):
Request changes on current head `386e3d95abc2a7c92038dc8f4fc37be6b0fafc71`.

The two issues from my previous review look materially fixed: the active type filter is now preserved as a zero-count chip when it disappears from the refreshed data, and the chip/footer width budgeting now accounts for brackets, separators, and ellipses with narrow-width coverage.

I found one remaining correctness issue and left it inline:

- `internal/workitem/tui/model.go:177` builds filter options from raw `m.allItems`, but refresh stores the unfiltered discovered item set while visible rows/counts still apply `ShowParked` via `refilter()` and `visibleTypeCounts()`. With the default `showParked=false`, a custom type that exists only on parked work items can appear as a selectable zero-count chip after refresh even though it is not the active preserved filter. Selecting it yields an empty list. The option source should be aligned with the same visible item set used by counts/refiltering, with the active type retained as the explicit zero-count exception.

Verification performed:

- `go test ./internal/workitem/...` passed.
- Targeted TUI regression tests passed: `go test ./internal/workitem/tui -run 'TestModel_FilterModeRefreshRebuildsOptions|TestModel_FilterModeEntryWithAbsentBuiltin|TestModel_FilterChipsFitNarrowWidth|TestChipWindow' -count=1 -v`.
- `just gate-push` passed.

Once the parked-hidden type leakage is fixed and covered with a refresh regression, this should be in good shape.

lancekrogers:
Parked-type leakage fixed in aa1811e (rebased onto the main merge at 386e3d9).

Verified the finding first: the command layer pre-filters the initial item set through `FilterAdvanced` (internal/commands/workitem/workitem.go:98), but the refresh path stores raw `Discover` output, so parked-only types could indeed leak into the chip row.

Fix: `visibleBaseItems` now centralizes the parked visibility rule, and both `rebuildFilterOptions` and `visibleTypeCounts` derive from it, so chips and counts always describe result sets the list can actually show. The active filter stays the explicit zero-count exception via `filterOptions`' ensure argument.

Regression tests: `TestModel_FilterModeRefreshHidesParkedOnlyTypes` (refresh introduces a parked-only custom type with a real store-driven parked stage, chip must not appear, counts must exclude it) and `TestModel_FilterModeShowParkedOffersParkedTypes` (with --show-parked the type is offered and stepping onto it shows the item).

`just gate-push` green at aa1811e.

lancekrogers:
Orchestrated review round one complete at 2323fda. Two independent Opus reviewers (adversarial correctness + quality/UX/tests) reviewed the diff; the correctness pass traced all three mode invariants end to end with no findings, and the quality pass surfaced two real issues, both fixed:

- **Chip counts ignored an active committed search.** Options and counts now derive from `FilterAdvanced` with the search query and parked rule, exactly the list minus the type filter, so a chip's count always equals the rows selecting it shows and query-dead types are not offered.
- **Filter mode was the only mode whose footer taught none of its keys.** It now appends `j/k step  Enter apply  Esc cancel` like the priority/stage footers, dropped first under width pressure.

Also addressed from the reviews: digits 1-4 inside the mode jump to their builtin chip when offered (matching the existing 0 jump), `filterOptions` the function renamed `deriveFilterOptions` to stop shadowing the model field, and new tests for the search interaction, digit jumps, h/l aliases, and f-as-commit.

`just gate-push` green. Round two with fresh reviewers is running against this head.

lancekrogers:
Review round two (fresh reviewer, no prior context) at 2323fda: clean pass, no blocking findings. All three invariants re-verified independently, including a brute force of the footer width guarantee across widths 1-220 with every active-chip position (zero overflows), and an audit of the new tests confirming they fail if any fix regresses. Two non-blocking notes for the record: the pre-existing byte-based `truncate` helper can cosmetically split a multibyte rune in exotic custom type names at narrow widths (width safety unaffected), and the help screen's four builtin rows are literals that duplicate `builtinFilterTypes`. Neither warrants blocking.

From my side this is merge-ready: three review-loop findings and two orchestrated-review findings fixed with regression tests, gates green at head.

obey-agent (commented):
**Verdict: Approve**

Re-reviewed current head `2323fdab245d9dd642a1649ecec882e05269a498`. The prior blockers are addressed: the active filter is preserved as an explicit zero-count chip when the underlying type disappears, the footer width math now budgets brackets/separators/ellipses against rendered width, and the parked-only type leakage after refresh is fixed by deriving chips and counts from the same `FilterAdvanced` base as the visible list.

**Key Findings**

- No new blocking findings on this head.
- I specifically re-traced refresh, active-filter preservation, parked visibility, committed-search chip derivation, digit jumps inside filter mode, and narrow footer rendering.

**What's Done Well**

- The latest `visibleBaseItems`/`deriveFilterOptions` split gives the filter mode a coherent source of truth: options and counts are computed from the list-before-type-filter, while the active type remains the only intentional zero-count exception.
- The tests now cover the failure modes this PR previously missed: active type removed on refresh, absent builtin filters, parked-only custom types, `--show-parked`, committed search, builtin digit jumps, h/l aliases, f-as-commit, and narrow footer overflow.
- The implementation stays localized to the TUI model/update/view surface and avoids reintroducing unstable custom digit bindings.

**Verification**

- `go test ./internal/workitem/tui -run 'TestModel_FilterMode|TestFilterOptions|TestChipWindow|TestModel_FilterChipsFitNarrowWidth' -count=1 -v` passed.
- `go test ./internal/workitem/tui -count=1` passed.
- `go test ./internal/workitem/...` passed.
- `just gate-push` passed. Lint emitted stale sibling-worktree warnings, but reported `0 issues`.

**Staff Standard**

Yes. I would be comfortable submitting this version as my own work.

obey-agent (commented):
**Verdict: Approve / no new findings**

Re-reviewed current head `2323fdab245d9dd642a1649ecec882e05269a498`. The head is unchanged from the prior approved review, and this fresh pass over the changed TUI files did not find a new blocker.

**Key Findings**

- No new findings.
- Re-checked the previously risky paths: filter chip derivation from committed search + parked visibility, active zero-count filter preservation after refresh, builtin digit jumps inside filter mode, Esc restore semantics, and narrow footer width accounting.

**What's Done Well**

- The model keeps `filterOptions[filterIndex]` aligned with `typeFilter`, including absent active filters.
- Chip options/counts now derive from the same pre-type-filter visible item set used by the list, so parked and search behavior stay consistent.
- Regression coverage is broad for the bugs found in earlier review rounds.

**Verification**

- `go test ./internal/workitem/tui -run 'TestModel_FilterMode|TestFilterOptions|TestChipWindow|TestModel_FilterChipsFitNarrowWidth' -count=1 -v` passed.
- `go test ./internal/workitem/...` passed.
- Full `just gate-push` had already passed on this same head in the prior review pass.

**Staff Standard**

Yes. I would be comfortable submitting this version as my own work.

PR #367, opened 2026-07-02
fix(list-tui): make camp list TUI responsive at small terminal sizes

lancekrogers:
## Problem

The `camp list` TUI rendered a fixed-width layout regardless of terminal size: 22-char name columns, untruncated paths, a fixed footer, and a bordered frame were emitted even when the window was narrower or shorter than the content. In a small window or terminal split the picker wrapped, overflowed, and became unusable.

## Fix

All in `cmd/camp/list_tui_*.go` — the list TUI is self-contained (it does not use `internal/listview`, which serves the workitem list output paths), so the fix is scoped here.

- A per-render `layout()` derives content width, list row budget, and whether chrome fits: the rounded border drops below 30x8, the footer below 6 rows.
- Row width splits into name/status/path columns that clamp and truncate (rune-safe ellipsis), dropping rightmost columns first as space runs out.
- Org-grouped rendering degrades to a flat, cursor-centered scroll window with a `[x-y of N]` position indicator when the list exceeds the height budget - the selection always stays on screen.
- Footer help shortens progressively when wider than the content.
- A final `frame()` guard hard-caps rendered line count and ANSI-safe-clamps line widths (`lipgloss MaxWidth`), so no size can overflow the window.
- The move-org overlay is bounded by the same frame rules and no longer indexes `m.visible[m.cursor]` unguarded (latent panic with an empty filtered list).
- The filter input resizes with the window.

Behavior at full size is unchanged (unknown size = unbounded layout, preserving the first frame before WindowSizeMsg and existing tests).

## Testing

Pure in-memory model tests only (no host filesystem, no subprocesses): bounds asserted at every size from 120x40 down to 1x1 (line count <= height, `lipgloss.Width` <= width per line), cursor visibility when scrolled, empty-list rendering, overlay bounding, column budget invariants, exhaustive `windowRange` property sweep, rune-safe truncation incl. multibyte.

`just gate` green: stable+dev builds, vet, lints, 2921 dev + 2895 stable unit tests.

## Note

Started by a background agent that hit its session limit mid-task; reviewed, completed (rune-safe truncation + multibyte tests), and verified by the main session.

### DIFF
--- a/cmd/camp/list_tui_responsive_test.go
+++ b/cmd/camp/list_tui_responsive_test.go
@@ -1,0 +1,191 @@
+package main
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/charmbracelet/bubbles/textinput"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+)
+
+func responsiveModel() listTUIModel {
+	ti := textinput.New()
+	ti.Prompt = "> "
+	entries := []campaignEntry{
+		{ID: "A", Name: "alpha", Org: "default", Status: config.StatusActive, Path: "/tmp/a"},
+		{ID: "B", Name: "this-is-an-extremely-long-campaign-name-that-should-truncate", Org: "obey", Status: config.StatusInactive, Path: "/Users/someone/very/deep/nested/directory/structure/inside/project-repo"},
+		{ID: "C", Name: "gamma", Org: "obey", Status: config.StatusReference, Path: "/Users/someone/work/gamma"},
+		{ID: "D", Name: "delta", Org: "platform", Status: config.StatusActive, Path: "/srv/delta"},
+		{ID: "E", Name: "epsilon", Org: "platform", Status: config.StatusActive, Path: "/srv/epsilon/really/long/path/segment/that/keeps/going/on"},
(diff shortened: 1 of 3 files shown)


obey-agent (commented):
**Verdict: Approve**

Reviewed current head `7463346553ea7f8c009c073722f07809f89cf75b`. This is a focused fix to make the `camp list` TUI bounded and usable at small terminal sizes, and the implementation is appropriately scoped to the self-contained list TUI files.

**Key Findings**

- No blocking findings.
- I specifically checked the frame/layout budget, scroll-window cursor visibility, footer collapse behavior, empty-list and overlay rendering, and the move-org overlay guard against empty visible lists.

**What's Done Well**

- The size-dependent `layout()` and final `frame()` clamp provide a clear last line of defense against both width and height overflow.
- The row column budget drops rightmost content first and keeps the selected row visible in normal scrollable small-window cases.
- The tests cover bounds across small sizes, empty views, overlays, footer collapse, column budget invariants, window range properties, and multibyte-safe truncation.

**Verification**

- `go test ./cmd/camp -run 'TestListView|TestListColumns|TestWindowRange|TestTruncate' -count=1 -v` passed.
- `go test ./cmd/camp -count=1` passed.
- `just gate-push` passed. Lint emitted stale sibling-worktree warnings but reported `0 issues`.

**Staff Standard**

Yes. I would be comfortable submitting this version as my own work.

PR #368, opened 2026-07-02
feat(dungeon): add --dry-run and batch support to camp dungeon move

lancekrogers:
## What

Add `--dry-run` and multi-item batch support to `camp dungeon move`, built on an explicit plan/apply refactor of the move path.

## Motivation

Scripted and agent-driven dungeon crawls need a preflight: a way to see exactly what a move would do before it touches the filesystem or writes a commit. Separately, a campaign-hygiene sweep of ~53 stale items surfaced the lack of any batch form — moving items one at a time is slow and gives no all-or-nothing guarantee. This PR addresses both, and does the refactor that makes them clean rather than bolted on.

## Design: plan/apply seam

The mutation used to be entangled with destination computation inside `statusmove.Move`. That is now split:

- **`statusmove`**: `PlanMove(ctx, src, dst, opts)` validates the source and computes the final destination (including the dated-bucket path and the existing-item check) without mutating anything; `Plan.Apply(ctx)` performs the on-disk move. `Move` is now `PlanMove` then `Apply`, so all nine existing callers are unaffected.
- **`internal/dungeon`**: a `MovePlan` value (kind, item, source, destination, status) plus `PlanMoveTo{Status,Dungeon,DungeonStatus,Docs}` and `ApplyMove`. The four existing `MoveTo*` service methods are now thin `plan + apply` wrappers, so `crawl`, `triage`, `promote`, and the TUI keep identical behavior (verified: sentinel errors preserved, all existing dungeon integration tests pass).
- **command**: `--dry-run` renders plans and returns without applying; batch builds N plans, validates them all, then applies. Single-item output and exit codes are byte-for-byte unchanged.

Dry-run is a faithful preflight: it runs the same resolution and validation as a real move, so an invalid item (missing source, bad status, existing destination) fails with the real error. For the non-dated modes the plan now also does an explicit destination-existence check, so `already exists` is caught at plan time rather than only at mutation time (same `ErrAlreadyExists` sentinel, so downstream behavior is unchanged).

While refactoring `statusmove`, two pre-existing tautological `err != nil` conditions in `existingDatedItemPath` (inside `else` branches of `err == nil`, so always true — flagged by gopls/nilness) were simplified. Behavior-preserving.

## Batch shape decision: variadic positional items + one trailing status

Chosen over a `--from-file` manifest because it is the camp-consistent option:

- camp already uses variadic positional args in several commands (`sync`, `run`, `org *`); there is **no** existing `--from-file`/manifest convention to match.
- It generalizes the existing `move <item> [status]` grammar directly: with two or more positionals the last is the shared status and the rest are items, so `move a b c archived` moves three items to `archived`. One positional stays a lone item, exactly as today.
- In `--to-docs` mode there is never a status token, so every positional is an item.

Tradeoff: because the trailing positional is the shared status, batch-moving several items into the *dungeon root* with no status is not expressible in one call (that stays one-at-a-time). This mirrors the existing single-item grammar and is surfaced by `--dry-run`. `--workitem` resolves N independent dungeons, so it stays single-item per invocation (rejected with a clear error for multiple items); workitem batch is a possible follow-up.

Batch is all-or-nothing at the planning stage: every item is validated (and intra-batch destination collisions are detected) before any move is applied, and **all** invalid items are reported, not just the first. A successful batch produces one aggregated `Dungeon sweep: N items -> status` commit.

## `--json`

`--dry-run --json` emits the plan as a small JSON object (`schema_version`, `dry_run`, `would_commit`, `count`, `moves[]`), matching the dungeon package's existing plain-JSON convention (`dungeon list --json`). It is scoped to dry-run in this PR so the applied-move output path stays byte-for-byte unchanged; a machine-readable envelope for applied results is a possible follow-up.

## Test coverage

- **Host unit** (`cmd/camp/dungeon/move_batch_test.go`): table-driven tests for item/status arg splitting, mode validation, and dry-run text/JSON rendering. Pure logic, no filesystem.
- **Integration** (`tests/integration/dungeon_test.go`, containerized harness): dry-run mutates nothing (tree unchanged + no new commit), dry-run `--json` shape, batch all-or-nothing on a mixed valid/invalid set (nothing moves, no commit), batch round-trip (three items to `archived`, single sweep commit), and workitem dry-run (no move, no dungeon init, no commit).

## Gates (run locally at branch head, rebased on fresh origin/main)

`just gate-fast` — **PASSED** (exit 0):
- stable build: OK
- dev build: OK
- vet stable / dev / integration: OK
- lint stable + dev: `0 issues`; `lint-no-host-fs-tests: clean`; `lint-no-fmt-errorf: clean` (both ratchets)
- dev unit tests: **ALL 2925 TESTS PASSED**

`gopls check` on all changed files: clean (no nilness/tautological-condition diagnostics).

Integration tests (Docker available locally, containerized harness): the new `TestDungeonMove_DryRun*`, `TestDungeonMove_Batch*`, and `TestDungeonMove_WorkitemDryRun` pass, and the full existing `TestDungeonMove_*` suite (25 tests) passes with the service refactor.

Note: GitHub Actions is intentionally paused for this repo; all enforcement above was run locally.


tests/integration/dungeon_test.go
```diff
@@ -916,3 +916,149 @@ func TestManifest_DungeonCommandsAgentAllowed(t *testing.T) {
 	// dungeon crawl should NOT be agent_allowed
 	assert.False(t, cmdMap["dungeon crawl"], "dungeon crawl should be agent_allowed=false")
 }
+
+// --- dungeon move --dry-run / batch ---
+
+func TestDungeonMove_DryRunMutatesNothing(t *testing.T) {
```
obey-agent: This file is now over 1,000 lines and this PR appends another full dungeon-move suite to it. That crosses the maintainability bar for the integration tests: unrelated dungeon list/move/docs/workitem/intent/manifest coverage is all in one file, so future reviewers have to scan too much state to understand a focused behavior change. Please split this before merge, at minimum by moving the new dry-run/batch coverage into a focused file such as `tests/integration/dungeon_move_batch_test.go` (and preferably peeling the existing move-specific helpers/tests into `dungeon_move_test.go` / shared helper files). The tests themselves are valuable; the issue is that adding more cases to this monolith makes the suite harder to navigate and maintain.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The implementation is much stronger than I expected for this size of change: the plan/apply split is coherent, existing `Move` callers stay behind the old API, dry-run uses the same validation path as apply, and the batch path correctly validates every item before mutation. I did not find a behavioral regression in the move, docs routing, statusmove, or link-rewrite paths.

**Key Findings:**
- Inline on `tests/integration/dungeon_test.go:922`: this file is now over 1,000 lines and the PR appends another full dungeon-move suite to it. Please split the new dry-run/batch tests into a focused integration test file before merge, and ideally peel existing move helpers/tests into smaller dungeon-specific files as well.

**What's Done Well:** The refactor keeps compatibility by making `statusmove.Move` call `PlanMove` + `Apply`, preserves sentinel-style error translation in `internal/dungeon`, batches external markdown link rewrites through the existing service API, and adds meaningful host/unit plus container integration coverage for dry-run, JSON output, batch validation, batch success, and workitem dry-run.

**Verification Run:**
- `go test ./cmd/camp/dungeon ./internal/dungeon ./internal/statusmove -count=1` passed.
- `go test ./cmd/camp/dungeon -run 'TestResolveItemsAndStatus|TestValidateMoveModes|TestRenderDryRun' -count=1 -v` passed.
- `go test -tags=integration ./tests/integration -run 'TestDungeonMove_(DryRun|Batch|WorkitemDryRun)' -count=1 -v` passed.
- `just gate-fast` passed: stable/dev builds, vet stable/dev/integration, lint stable/dev with `0 issues`, and 2,935/2,935 dev tests passed. Lint emitted only stale sibling-worktree warnings.

**Staff Standard:** I would be comfortable with the production code after this review, but I would not submit the test change in its current monolithic file layout. Please make the integration tests modular, then this is approvable from my side.

obey-agent (changes requested):
Requesting changes for the integration test organization issue called out inline: `tests/integration/dungeon_test.go` is now over 1,000 lines, and the new dry-run/batch coverage should be split into focused dungeon move test files before merge. Behavior and gates otherwise looked solid in this review.

tests/integration/dungeon_test.go
lancekrogers: Done in 18a5831. Split into three files: all dungeon-move tests (inner/triage/docs/workitem) now live in `dungeon_move_test.go` with the move-only helpers, the new dry-run/batch tests in `dungeon_move_batch_test.go`, and `dungeon_test.go` keeps list/commands/intent/manifest plus the shared setup helper (down from ~1,060 to 290 lines). Pure test relocation, no behavior change; the full dungeon integration suite passes.

tests/integration/dungeon_move_test.go
```diff
@@ -0,0 +1,591 @@
...
+		assert.Contains(t, body, wantBody, "crawl commit body should describe the move")
+	}
+
+	diff := tc.GitOutput(t, repoPath, "diff-tree", "--no-commit-id", "--name-status", "--no-renames", "-r", "HEAD")
+	for _, want := range wantNameStatus {
+		assert.Contains(t, diff, want, "crawl commit should include expected name-status entry")
+	}
+	return diff
+}
+func TestDungeonMove_ToStatus(t *testing.T) {
```
obey-agent, line 46: The split is a real improvement, but this extracted file is still 591 lines and now mixes several separable behaviors: core status moves, triage moves, workitem moves, docs routing, and nested-context cases. Our staff-review standard for Go files is under 500 lines, and this is exactly the kind of test file that keeps growing. Please split this one more step before merge, for example into `dungeon_move_test.go` for core/status+triage, `dungeon_move_workitem_test.go`, and `dungeon_move_docs_test.go` (keeping the shared helpers in one small helper file if needed). That preserves the modularity fix from this update instead of moving the monolith into a new filename.

obey-agent (commented):
**Verdict: Request Changes**

**Overview:** The update fixed the original `dungeon_test.go` monolith: that file is now 290 lines, the new dry-run/batch tests live in `dungeon_move_batch_test.go`, and move-specific coverage was pulled into `dungeon_move_test.go`. The production implementation still looks sound from this pass; I did not find a behavioral regression in the plan/apply refactor, docs routing, statusmove, or batch link-rewrite path.

**Key Findings:**
- Inline on `tests/integration/dungeon_move_test.go:46`: the split is not quite finished. `dungeon_move_test.go` is still 591 lines and mixes core status/triage moves, workitem moves, docs routing, and nested-context cases. Please split it one more step so the modularity fix does not just move the large-file problem to a new filename.

**What's Done Well:** The requested `dungeon_test.go` issue was addressed materially, `dungeon_move_batch_test.go` is a good focused file, and the code path continues to use the same planning validation for dry-run and apply. The tests added for batch validation, batch success, JSON dry-run, and workitem dry-run remain valuable and pass in the container harness.

**Verification Run:**
- `gofmt -l` on all changed Go files returned clean.
- `go test ./cmd/camp/dungeon ./internal/dungeon ./internal/statusmove -count=1` passed.
- `go test ./cmd/camp/dungeon -run 'TestResolveItemsAndStatus|TestValidateMoveModes|TestRenderDryRun' -count=1 -v` passed.
- `go test -tags=integration ./tests/integration -run 'TestDungeonMove_(DryRun|Batch|WorkitemDryRun)' -count=1 -v` passed.
- `just gate-fast` passed: stable/dev builds, vet stable/dev/integration, lint stable/dev with `0 issues`, and 2,943/2,943 dev tests passed. Lint emitted only stale sibling-worktree warnings.

**Staff Standard:** Not yet. The production code is approvable, but the integration test organization still misses the file-size/modularity standard. Once `dungeon_move_test.go` is split into smaller focused files, I expect this to be approvable.

obey-agent (changes requested):
Requesting changes for the remaining integration test organization issue called out inline: `tests/integration/dungeon_move_test.go` is still 591 lines and should be split into smaller focused move/workitem/docs test files before merge. Behavior and gates otherwise looked clean.

obey-agent:
Pushed ba949ce to finish the requested integration-test split. The remaining move coverage is now divided across focused files: dungeon_move_test.go (283 lines), dungeon_move_workitem_test.go (111), dungeon_move_docs_test.go (231), dungeon_move_batch_test.go (152), and dungeon_test.go (290). Verification passed: gofmt/git diff checks, go test -tags=integration ./tests/integration -run 'TestDungeonMove_' -count=1 -v, and just gate-fast (2943/2943 tests).

PR #369, opened 2026-07-02
feat(switch): add org-aware campaign selection

lancekrogers:
## Summary

Implements org-aware `camp switch` selection for `WI-cdfacb`.

- Adds scoped selector parsing for `org/campaign` and `org/campaign@tab`.
- Adds `--org`, `--status`, `--all`, and `--json` to `camp switch`.
- Filters direct switch, picker candidates, and completion candidates by org/status before applying the existing fuzzy matching behavior.
- Keeps inactive/reference campaigns out of default switch resolution; they are reachable via `--all` or explicit `--status`.
- Regenerates CLI reference docs.

## User impact

Users with many registered campaigns can narrow switch resolution and completion by organization without losing the fuzzy matching behavior they already use. Agents can use `camp switch --org <org> <campaign> --json` for a structured selected-campaign payload.

## Validation

- `go test ./cmd/camp ./cmd/camp/cmdutil`
- `just docs`
- `just gate-push`

### DIFF
--- a/cmd/camp/cmdutil/campaign_scope.go
+++ b/cmd/camp/cmdutil/campaign_scope.go
@@ -1,0 +1,187 @@
+package cmdutil
+
+import (
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/nav/fuzzy"
+)
+
+// CampaignScope describes the candidate set for switch resolution.
+type CampaignScope struct {
+	Org    string
+	Status string
+	All    bool
+}
+
+// ParsedSwitchSelector is the parsed campaign[@tab] or org/campaign[@tab]
+// selector accepted by camp switch.
+type ParsedSwitchSelector struct {
+	Org      string
+	Campaign string
+	Tab      string
+	HasTab   bool
+}
+
+// ParseSwitchSelector splits the switch selector into org, campaign, and tab
+// parts. Validation that depends on completion/direct mode stays with callers.
+func ParseSwitchSelector(raw string) ParsedSwitchSelector {
+	left := raw
+	parsed := ParsedSwitchSelector{}
+	if at := strings.IndexByte(raw, '@'); at >= 0 {
+		left = raw[:at]
+		parsed.Tab = raw[at+1:]
+		parsed.HasTab = true
+	}
+	if slash := strings.IndexByte(left, '/'); slash >= 0 {
+		parsed.Org = left[:slash]
+		parsed.Campaign = left[slash+1:]
+		return parsed
+	}
+	parsed.Campaign = left
+	return parsed
+}
+
+// FilterCampaigns returns the registry entries visible under scope. Direct
+// switch, picker, and completion all default to active campaigns unless --all or
+// --status changes the lifecycle policy.
+func FilterCampaigns(reg *config.Registry, scope CampaignScope) []config.RegisteredCampaign {
+	all := reg.ListAll()
+	out := make([]config.RegisteredCampaign, 0, len(all))
+	fallbackOrg := reg.FallbackOrg()
+	for _, c := range all {
+		if c.Org == "" {
+			c.Org = fallbackOrg
+		}
+		if c.Status == "" {
+			c.Status = config.StatusActive
+		}
+		if scope.Org != "" && c.Org != scope.Org {
+			continue
+		}
+		if scope.Status != "" {
+			if c.Status != scope.Status {
+				continue
+			}
+		} else if !scope.All && c.Status != config.StatusActive {
+			continue
+		}
+		out = append(out, c)
+	}
+	return out
+}
+
+// ResolveCampaignSelectionScoped applies the existing switch lookup order
+// inside the filtered candidate set: exact ID, unique ID prefix, exact name, and
+// fuzzy name.
+func ResolveCampaignSelectionScoped(query string, reg *config.Registry, scope CampaignScope, matchWriter io.Writer) (config.RegisteredCampaign, error) {
+	candidates := FilterCampaigns(reg, scope)
+	return resolveCampaignFromCandidates(query, candidates, scope, matchWriter)
+}
+
+func resolveCampaignFromCandidates(query string, candidates []config.RegisteredCampaign, scope CampaignScope, matchWriter io.Writer) (config.RegisteredCampaign, error) {
+	if query == "" {
+		return config.RegisteredCampaign{}, camperrors.New("campaign name required" + scopeDescription(scope))
+	}
+
+	for _, c := range candidates {
+		if c.ID == query {
+			return c, nil
+		}
+	}
+
+	var idPrefixMatches []config.RegisteredCampaign
+	for _, c := range candidates {
+		if strings.HasPrefix(c.ID, query) {
+			idPrefixMatches = append(idPrefixMatches, c)
+		}
+	}
+	switch len(idPrefixMatches) {
+	case 1:
+		return idPrefixMatches[0], nil
+	case 0:
+	default:
+		return config.RegisteredCampaign{}, camperrors.New(fmt.Sprintf("campaign ID prefix %q is ambiguous%s: %s",
+			query, scopeDescription(scope), campaignIDs(idPrefixMatches)))
+	}
+
+	var exactNameMatches []config.RegisteredCampaign
+	for _, c := range candidates {
+		if c.Name == query {
+			exactNameMatches = append(exactNameMatches, c)
+		}
+	}
+	switch len(exactNameMatches) {
+	case 1:
+		return exactNameMatches[0], nil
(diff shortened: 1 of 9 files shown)


cmd/camp/cmdutil/campaign_selection.go
```diff
@@ -13,39 +13,33 @@ import (
...
-	all := reg.ListAll()
+	return PickCampaignWithOptions(ctx, reg, PickCampaignOptions{Scope: CampaignScope{All: true}})
+}
+
+// PickCampaignWithOptions opens the shared campaign picker UI used by
+// `camp switch` with scoped candidate filtering.
+func PickCampaignWithOptions(ctx context.Context, reg *config.Registry, opts PickCampaignOptions) (config.RegisteredCampaign, error) {
+	all := FilterCampaigns(reg, opts.Scope)
+	if len(all) == 0 {
+		return config.RegisteredCampaign{}, fmt.Errorf("no campaigns found%s", scopeDescription(opts.Scope))
```
obey-agent: Use the project error package here (`camperrors.New(fmt.Sprintf("no campaigns found%s", scopeDescription(opts.Scope)))`) rather than `fmt.Errorf`. `camperrors` is already imported and every other error you added in this PR (campaign_scope.go, switch.go) uses it — this is the lone `fmt.Errorf` newly introduced by the change and it breaks the repo's "internal error package, not fmt.Errorf in production code" rule.

cmd/camp/cmdutil/campaign_scope.go
```diff
@@ -0,0 +1,187 @@
...
+			c.Status = config.StatusActive
+		}
+		if scope.Org != "" && c.Org != scope.Org {
+			continue
+		}
+		if scope.Status != "" {
+			if c.Status != scope.Status {
+				continue
+			}
+		} else if !scope.All && c.Status != config.StatusActive {
```
obey-agent, line 70: This default makes `camp switch <name>` resolve only `active` campaigns unless `--all`/`--status` is given, so switching to an `inactive`/`reference` campaign by name now fails where it previously succeeded (contract change for real users). It correctly mirrors the existing `camp list` default, so the behavior is consistent and intentional — but please confirm it's intended and add a release/changelog note, since there's no CHANGELOG in the repo to surface the change otherwise.

cmd/camp/cmdutil/campaign_selection.go
```diff
@@ -13,39 +13,33 @@ import (
...
-	bestName := matches[0].Target
-	c, ok = reg.GetByName(bestName)
-	if !ok {
-		return config.RegisteredCampaign{}, fmt.Errorf("campaign %q not found in registry", query)
-	}
-
-	if matchWriter != nil {
-		_, _ = fmt.Fprintf(matchWriter, "Matched: %s -> %s\n", query, c.Name)
-	}
+	return resolveCampaignFromCandidates(query, reg.ListAll(), CampaignScope{All: true}, matchWriter)
```
obey-agent, line 21: Routing `ResolveCampaignSelection` through `resolveCampaignFromCandidates` changes resolution semantics for the three non-switch callers (project/add.go, attach/resolver.go, intent/add.go): an ambiguous ID-prefix or duplicate exact name now returns a hard "ambiguous" error instead of the old fall-through / arbitrary first-match. This is more correct and `scope.All=true` preserves all-status visibility, but the change is untested for those three commands — consider a regression test asserting the new ambiguity error so a future refactor can't silently revert it.

obey-agent (commented):
## Staff Review — Verdict: COMMENT (merge-ready)

**Overview**
Adds org/status-aware selection to `camp switch`: a `ParseSwitchSelector` for `org/campaign[@tab]`, a `FilterCampaigns` + `ResolveCampaignSelectionScoped` candidate pipeline reused across direct switch, picker, and completion, plus `--org/--status/--all/--json` flags. Cleanly factored into a new `cmdutil/campaign_scope.go`; the new resolution/completion/JSON paths are well covered by table tests. I traced the registry load path (`LoadRegistry` sets `c.ID` and normalizes `Org`/`Status`) and found no correctness or security defects.

**Key Findings** (all minor — see inline)
- `campaign_selection.go:41` introduces a lone `fmt.Errorf` in production while `camperrors` is imported and used everywhere else in the PR — breaks the project error-package rule.
- Default switch resolution now excludes `inactive`/`reference` campaigns (`campaign_scope.go:70`). Intentional and consistent with `camp list`, but a user-visible default change (camp has real users) worth a changelog/release note.
- `ResolveCampaignSelection` now routes three other callers (`project/add`, `attach/resolver`, `intent/add`) through `resolveCampaignFromCandidates`, changing ambiguous-prefix/duplicate-name behavior to a hard error — more correct, but untested for those callers.
- `runSwitch` is ~67 lines (>50 guideline); cohesive, acceptable, noting for awareness.

**What's Done Well**
Single shared candidate pipeline (no parallel resolver), flag axes consistent with `camp list`, sensible mutual-exclusion guards, versioned `--json` schema (`camp-switch/v1`) with no nil-slice fields, output routed through `cmd.OutOrStdout()`, and thorough table tests.

**Staff Standard**
Yes — I'd be comfortable submitting this. It extends existing systems rather than duplicating them and is well-tested on the changed paths. The one behavior change is intentional and consistent. Cleanups (the `fmt.Errorf` slip and a changelog note) don't block merge. Leaving the Approve for Lance.

obey-agent (commented):
## Staff Review — Verdict: COMMENT (merge-ready)

**Overview**
The follow-up commit addresses the prior review notes. The user-visible `camp switch` default change is now documented in a migration note and README text, and the shared resolver behavior has regression coverage for ambiguous ID prefixes, duplicate exact names, and legacy all-status visibility.

**Key Findings**
- No remaining blocking findings. I did not find a new correctness, contract, or maintainability issue on the current head.

**What's Done Well**
- The scoped resolver remains a single shared pipeline across direct switch, picker, and completion paths.
- The added cmdutil tests lock down the non-switch caller semantics that changed when `ResolveCampaignSelection` was routed through the shared resolver.
- `camp switch --json` is versioned and writes through Cobra's output stream, which keeps it testable and script-friendly.

**Verification**
- `go test ./cmd/camp ./cmd/camp/cmdutil -run 'TestSwitch|TestParseSwitch|TestResolveCampaignSelection'`
- `go test ./cmd/camp ./cmd/camp/cmdutil`

**Staff Standard**
Yes. I would be comfortable submitting this as my own work. The behavior change is now called out, the resolver edge cases are tested, and the implementation extends the existing registry/navigation model without parallel lookup logic.

cmd/camp/switch.go
```diff
@@ -193,10 +302,91 @@ func runSwitch(cmd *cobra.Command, args []string) error {
...
 	if printOnly {
-		fmt.Println(selected.Path)
-	} else {
-		fmt.Printf("cd %s\n", selected.Path)
+		_, err := fmt.Fprintln(cmd.OutOrStdout(), targetPath)
+		return err
+	}
+	if jsonOut {
+		out := switchOutput{
+			SchemaVersion: "camp-switch/v1",
```
obey-agent: `SchemaVersion: "camp-switch/v1"` is a magic string duplicated between this emit site and `TestSwitchJSONOutput`. Optional: promote it to a package const (e.g. `const switchSchemaVersion = "camp-switch/v1"`) so the JSON contract version and its test stay in sync as the schema evolves. Non-blocking.

obey-agent (commented):
## Staff Re-Review (head 6005b8b) — Verdict: COMMENT (merge-ready)

**Overview**
Follow-up since my prior review (26a6e0c): one production line changed (the `no campaigns found` error now uses `camperrors.New`), a new `campaign_scope_test.go`, a README clarification, and a migration guide. No new production logic beyond the error-package fix.

**Prior findings — all resolved**
1. FIXED — the `fmt.Errorf` slip is now `camperrors.New(fmt.Sprintf("no campaigns found%s", ...))`; grep confirms zero new `fmt.Errorf` in the added prod lines.
2. FIXED — the active-only default is now documented in `docs/migrations/switch-org-aware-navigation-2026-07.md` and the README axes section; legacy callers keep all-status access via `CampaignScope{All:true}`.
3. FIXED — `campaign_scope_test.go` adds `TestResolveCampaignSelection_AmbiguousIDPrefixErrors`, `_DuplicateExactNamesError`, and `_LegacyCallersSeeAllStatuses`, covering the shared resolver all three legacy callers route through.

**New findings**
- (minor, inline) `SchemaVersion: "camp-switch/v1"` is a magic string duplicated between the emit site and the test — promote to a package const.
- Out of scope (unchanged legacy line): `PickCampaignWithOptions` still returns `fmt.Errorf("cancelled")` on `fuzzyfinder.ErrAbort` — worth a future cleanup for full error-package compliance, not introduced here.

**What's Done Well**
Legacy callers preserved at `All:true` with a test locking it in; scalar-only `--json` (no nil-slice contract risk); output routed through `cmd.OutOrStdout()`; fuzzy matching reused, not reimplemented; the contract change documented in two places.

**Staff Standard**
All three prior minors cleared and the broader surface is now genuinely well-tested. Merge-ready. Leaving the Approve for Lance.

PR #372, opened 2026-07-02
fix(scaffold): campaign-commit skill taught raw git commit at campaign root

lancekrogers:
## Problem

The scaffolded `campaign-commit` skill (`.campaign/skills/campaign-commit/SKILL.md`) has told agents to use **raw `git commit`** for campaign root files since the template was first authored:

```
- Campaign root files: `git commit -m "msg"`
```

This directly contradicts campaign commit discipline (root commits go through `camp commit`), and it survived `camp init --repair` regeneration because the template source itself carries the error. An agent that follows the skill literally loses the `[<CAMPAIGN>-<id>]` context tag prefix, the accidental-submodule-ref staging guard, and stale-lock handling.

The skill was also silent on the scoped-commit case (one artifact at root amid unrelated dirty files), which is what pushed agents toward raw git in practice — even though `camp commit --all=false` already commits staged-only and has existed since early in the command's history.

## Change

Rewrite the decision table in the skill template:

- Campaign root files: `camp commit -m "msg"` (stages all root-level changes; submodule refs excluded by default)
- Campaign root, scoped: `git add <paths>` then `camp commit --all=false -m "msg"` (commits only what is staged)
- New rule: never run raw `git commit` anywhere in a campaign workspace; staging with `git add` is fine.

Frontmatter description updated to stop advertising `root git commit` as an option.

## Verification

- `go build ./...` clean, `go test ./internal/scaffold/...` passes (tests assert skill presence, not content).
- `camp commit --all=false` staged-only behavior verified empirically in an isolated scratch repo: with one path staged and another dirty, the commit contained only the staged path and still received the context tag prefix (`effectiveCommitAll` gating in `cmd/camp/commit.go`).

Content-only template change; no code paths touched.

### DIFF
--- a/internal/scaffold/campaign/templates/.campaign/skills/campaign-commit/SKILL.md.tmpl
+++ b/internal/scaffold/campaign/templates/.campaign/skills/campaign-commit/SKILL.md.tmpl
@@ -1,6 +1,6 @@
 ---
 name: campaign-commit
-description: Choose the correct commit command in a campaign workspace. Use when you are about to commit and need to select `camp p commit`, `fest commit`, root `git commit`, or intentional root pointer sync via `camp refs-sync`.
+description: Choose the correct commit command in a campaign workspace. Use when you are about to commit and need to select `camp commit`, `camp p commit`, `fest commit`, or intentional root pointer sync via `camp refs-sync`.
 ---
 
 # Campaign Commit Decision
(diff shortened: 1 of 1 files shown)


obey-agent (commented):
## Staff Review: camp #372

**Verdict: Comment (merge-ready).** Not approving in this review pass; leaving the formal approve for the maintainer.

### Overview
This correctly fixes a real, harmful bug. The scaffolded `campaign-commit` skill told agents to run raw `git commit` at the campaign root, bypassing camp's context-tag prefix, submodule-ref staging guard, and lock handling. The rewritten decision table and the `--all=false` scoped-commit guidance are accurate, verified against `cmd/camp/commit.go` (root staging excludes `projects/*` refs by default around commit.go:143-167; `--all=false` commits only pre-staged paths via `effectiveCommitAll` around commit.go:256-261 while still applying the tag prefix and the pre-staged-ref guard), not just read from the diff.

### Second commit (worktrees)
The follow-up commit broadens `camp p commit` to cover "submodule, linked project, or worktree." This is correct and matches the intended direction that `camp p commit` is the single project-commit entry point for every project type. From inside a submodule-worktree, `camp p commit` commits to the correct repo.

One separate, non-blocking note (a pre-existing bug outside this PR's diff): `internal/git/submodule.go` `isWorktreeGitDir` (submodule.go:40-44) only matches the `/.git/worktrees/` segment. A worktree of a submodule project has gitdir `.git/modules/projects/<name>/worktrees/<wt>`, which does not match, so `IsSubmodule` returns true and the printed project label resolves to "worktrees". The commit still lands in the right repo, so it is not destructive; only the displayed label is wrong. This should be tracked as its own camp workitem rather than fixed in this docs change.

### What's done well
- Removes the harmful raw-git guidance and replaces it with a single strong "never run raw `git commit` anywhere" rule.
- Honest, accurate PR verification section: build, tests, and the `--all=false` staged-only behavior all reproduce.
- Minimal, well-scoped, content-only diff.

### Staff standard
Yes, I would be comfortable submitting this docs fix as my own work. The core change is correct and well-verified. The only caveat is a separate pre-existing code bug that belongs in its own workitem.

PR #374, opened 2026-07-06
feat: campaign workflow categories

lancekrogers:
## Campaign workflow categories

Adds a campaign-level `workflow_category` classification layered over workitem
types, so `camp workitem` (and the TUI / festival app) can filter and group work
by *kind of work* at a glance without changing the existing `workflow_type` /
`--type` contract. Implements the `campaign-workflow-categories` design
(`design-campaign-workflow-categories-2026-07-01`, WI-4eca07).

### Model
- Categories are **config-derived per type**, never stored per-item, and never
  mutate the loaded config.
- Shipped vocabulary: **`plan` / `research` / `pipeline` / `review`** (`plan` is
  the catch-all; `code_reviews` maps to `review`). Unmapped types resolve to
  `uncategorized` and still appear in listings.
- **No forced setup ceremony:** custom tracked workflows are just
  `workflow/<type>/<slug>/` mkdir + `.workitem`; they're mapped with a one-line
  `category_by_type` edit or `camp workflow create --category`.
- Goal: `camp workitem` as a campaign dashboard, with category as the filter axis.

### What's included (whole feature)
- **Config** (`internal/config/workflows.go`): `WorkflowsConfig` on
  `CampaignConfig`, defaults, `WorkflowCategoryForType`, ordered keys, slug-safe
  validation (reuses `pathutil.ValidateSegment`), `DefaultWorkflowsConfig`.
- **Workitem** : `WorkItem.WorkflowCategory` (omitempty), `ApplyWorkflowCategories`
  enrichment, `FilterOptions.Categories` + query matching.
- **CLI**: `camp workitem --category` (repeatable) and `--group-by category`;
  enrichment wired into the command.
- **JSON `workitems/v1alpha8`**: `workflow_category` per item,
  `category_vocabulary`, `category_counts`, `category` in `available_group_by`
  and section rows (v1alpha7 output byte-stable until enrichment runs).
- **Workflow commands**: `camp workflow create --category` (default `plan`,
  validated, writes the mapping); `list` CATEGORY column; `show` category.
- **Init/repair**: `camp init` writes the default block; `camp init --repair`
  backfills missing defaults while preserving user overrides.
- **TUI**: `c` cycles the category filter (Esc clears; header shows
  `category:<x>`), composing with type filter + live search; type keys unchanged.
- **Docs**: json-contracts, campaign-settings-files, workflow-reference, and
  regenerated CLI reference.

### Verification
Driven end-to-end against real campaigns: item enrichment, `--category`
filtering, `--group-by category` sections, empty-but-valid category, invalid-input
errors, `workflow create --category` (+ unknown-category rejection),
`list`/`show` category, fresh-init block, and `init --repair` backfill with a
preserved `design: research` override.

`just gate-fast` green: stable + dev build, `go vet` (stable/dev/integration),
`just lint` (stable/dev), full unit suite (**2975/2975**).

### Notes
- The built-in `code_reviews` collection isn't discovered as workitems today, so
  `review` is forward-compatible but only populates once review work is tracked
  with `.workitem` markers — a separate follow-up, not a blocker.

### DIFF
--- a/internal/workitem/tui/model.go
+++ b/internal/workitem/tui/model.go
@@ -10,6 +10,7 @@ import (
 	"github.com/charmbracelet/bubbles/textinput"
 	tea "github.com/charmbracelet/bubbletea"
 
+	"github.com/Obedience-Corp/camp/internal/config"
 	"github.com/Obedience-Corp/camp/internal/paths"
 	"github.com/Obedience-Corp/camp/internal/workitem"
 	"github.com/Obedience-Corp/camp/internal/workitem/priority"
@@ -25,6 +26,48 @@ var builtinFilterTypes = []string{
 	string(workitem.WorkflowTypeDesign),
 	string(workitem.WorkflowTypeExplore),
 	string(workitem.WorkflowTypeFestival),
+}
+
+// builtinFilterCategories pins the canonical category cycle order.
+var builtinFilterCategories = []string{
+	config.WorkflowCategoryPlan,
+	config.WorkflowCategoryResearch,
+	config.WorkflowCategoryPipeline,
+	config.WorkflowCategoryReview,
+}
+
+// deriveCategoryOptions returns the category cycle values: "" (all) first, then
+// builtin categories present in items in canonical order, then custom categories
+// (including uncategorized) alphabetically. ensure is always included so the
+// active filter is always a cycle stop.
+func deriveCategoryOptions(items []workitem.WorkItem, ensure string) []string {
+	builtin := make(map[string]bool, len(builtinFilterCategories))
+	for _, c := range builtinFilterCategories {
+		builtin[c] = true
+	}
+	present := make(map[string]bool, len(items))
+	for _, item := range items {
+		if c := item.WorkflowCategory; c != "" {
+			present[c] = true
+		}
+	}
+	if ensure != "" {
+		present[ensure] = true
+	}
+	opts := []string{""}
+	for _, c := range builtinFilterCategories {
+		if present[c] {
+			opts = append(opts, c)
+		}
+	}
+	var customs []string
+	for c := range present {
+		if !builtin[c] {
+			customs = append(customs, c)
+		}
+	}
+	sort.Strings(customs)
+	return append(opts, customs...)
 }
 
 // deriveFilterOptions returns the filter-mode chip values: "" (all) first,
@@ -83,6 +126,7 @@ type Model struct {
 
 	// Filters
 	typeFilter      string // empty = all, or any workflow type
+	categoryFilter  string // empty = all, or any workflow category
 	showParked      bool
 	filterMode      bool
 	filterOptions   []string // chip values while filter mode is active; "" = all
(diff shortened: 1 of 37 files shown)


obey-agent (commented):
**Verdict: Approve**

**Overview:** This is a clean first slice for workflow categories. The config model keeps the category taxonomy derived from effective campaign config, merges defaults without mutating loaded state, validates category/type keys through the existing path-segment contract, and keeps production workitem JSON unchanged until a later slice wires enrichment into the read path.

**Key Findings:** None blocking. I did not leave inline comments because I did not find a correctness, compatibility, or maintainability issue worth calling out on the diff.

**What's Done Well:** The tests cover default vocabulary, user overrides, deterministic ordering, validation failures, nil/empty config behavior, and in-place workitem enrichment. The `workflows` block omission test is especially useful for keeping existing campaign saves conservative.

**Staff Standard:** Yes. I would be comfortable merging this slice as my own work.

**Verification:** `go test ./internal/config ./internal/workitem`, `go test ./...`, `go vet ./...`, `go build ./...`, and `just gate-push` all pass in a detached PR worktree. `just gate-push` emitted stale-worktree golangci-lint warnings while reporting `0 issues`; the gate completed with `PASSED`.

obey-agent (approved):
Approved. I did not find blocking issues in this config/enrichment slice, and the local gates pass: focused config/workitem tests, full go test, go vet, go build, and just gate-push.

obey-agent (commented):
A few product/API clarity notes before this lands:

1. The `DefaultWorkflowCategoryByType` entries for `feature`, `bug`, and `chore` are surprising in this campaign/workitem model. The current built-in workflow types appear to be `intent`, `design`, `explore`, and `festival`, with repo-local custom workflow types discovered under `workflow/<type>/`. Unless `feature`/`bug`/`chore` are intentionally supported workflow types for `camp workitem create --type`, I would avoid baking them into the default category mapping. Otherwise the defaults imply a workflow vocabulary that the campaign methodology has not actually established.

2. The customization story should be explicit in the PR description/docs: users add category vocabulary under `.campaign/campaign.yaml` at `workflows.categories`, then map workflow/workitem types under `workflows.category_by_type`. For example, a user-created `workflow/interviews/` collection would be categorized by adding `category_by_type.interviews: research` or by defining a new category and mapping to it.

3. The current PR body says user-added workflows are uncategorized until mapped with a one-line `category_by_type` edit. That answers the data-model side, but the PR should make clear whether this slice includes any CLI path for that edit. If the answer is “not yet,” say that directly so users know the file to edit is `.campaign/campaign.yaml`.

4. There are several explanatory comments in the new code that restate the obvious shape of the types/functions. I would trim those and keep only comments that document real contracts: config-derived categories, user mappings overriding defaults, unknown type keys being allowed for pre-mapping, and `uncategorized` fallback behavior.

internal/workitem/tui/model.go
```diff
@@ -83,6 +126,7 @@ type Model struct {
 
 	// Filters
 	typeFilter      string // empty = all, or any workflow type
+	categoryFilter  string // empty = all, or any workflow category
```
obey-agent, line 125: This model now owns category filter state, but the refresh path does not preserve the category enrichment. `refreshCmd` calls `workitem.Discover` and returns those raw items without the `ApplyWorkflowCategories(cfg.WorkflowCategoryForType)` step that `camp workitem` runs before constructing the model. After pressing `r`, `m.allItems` has empty `WorkflowCategory` values, so the `c` cycle loses the configured categories and any active category filter can collapse the list to zero items. Please carry the category resolver into the model or otherwise enrich refreshed items before returning `refreshMsg`, and add a refresh-path test.

obey-agent (commented):
**Verdict: Request Changes**

This is a solid, broad implementation of config-derived workflow categories: the model, JSON surface, workflow create/list/show, init/repair backfill, docs, and most tests line up cleanly. The remaining blocker is the TUI refresh path: refreshed items lose their `workflow_category` enrichment, so the new category filter can break after pressing `r`.

**Key Findings**
- Inline: `internal/workitem/tui/model.go` needs to preserve/reapply workflow category enrichment on refresh. The initial command path enriches before `wktui.New`, but `refreshCmd` replaces `m.allItems` with raw `Discover` results.

**What's Done Well**
- Category derivation is config-based and avoids storing per-item category state.
- `init --repair` preserves user mappings while backfilling defaults.
- JSON v1alpha8 and CLI docs are updated coherently.
- Validation and focused tests cover the main config/filter/create paths.

**Staff Standard**
Not quite ready as-is because the refresh regression affects the interactive workflow this PR extends. I would be comfortable with this after the TUI refresh enrichment path is fixed and covered.

Validation run: `go test ./...` passed in the detached PR worktree.

internal/workitem/tui/model.go
obey-agent: Fixed in 74b0f07.

The model now carries a category resolver (`SetCategoryResolver`), which `runTUI` sets from `cfg.WorkflowCategoryForType`. The `refreshMsg` handler re-applies `ApplyWorkflowCategories` to the re-discovered raw items before they replace `allItems`, so `workflow_category` (and the category filter) survive a refresh (`r`).

Covered by `TestRefreshReappliesCategoryEnrichment`, and `just gate-fast` is green (2976/2976). Resolving.

obey-agent (commented):
**Verdict: Approve**

I re-reviewed the current PR head (`3a42f60`) including the follow-up fix for the earlier TUI refresh finding. No blocking issues found in this pass.

The previous blocker is addressed: refreshed TUI items are now re-enriched through the configured category resolver before priority overlay/filtering, and `TestRefreshReappliesCategoryEnrichment` covers the regression path. The rest of the feature hangs together cleanly across config defaults/validation, workflow create/list/show, init/repair backfill, workitem filtering/grouping, TUI category cycling, JSON contract versioning, and docs.

**User Experience**

For users, this turns `camp workitem` into a kind-of-work dashboard without changing how workitems are created or stored. Existing workitems get their category at read time from `.campaign/campaign.yaml`, so changing `workflows.category_by_type` immediately reclassifies all items of that type; no `.workitem` files need migration. Out of the box, shipped camp types resolve into `plan`, `research`, `pipeline`, and `review`, while custom workflow types remain visible as `uncategorized` until the user maps them.

In day-to-day CLI use, users can filter with `camp workitem --category research`, repeat `--category` for multiple buckets, or group structured output with `camp workitem --json --group-by category`. JSON consumers get the new `workitems/v1alpha8` shape with `workflow_category` on items, `category_counts`, and ordered `category_vocabulary` metadata.

In the TUI, category filtering behaves like a first-class dashboard control: pressing `c` cycles through categories present in the current view, composes with type/search/stage filters, and now survives refresh correctly. New workflow collections can opt into the right bucket immediately with `camp workflow create <type> --category <category>`, and `camp init --repair` backfills the default category vocabulary/mappings for existing campaigns without clobbering user-owned custom categories.

**Verification**

- `go test ./internal/config ./internal/workitem ./internal/workitem/tui ./internal/commands/workitem ./internal/commands/workflow ./internal/scaffold`
- `go test ./...`

PR #375, opened 2026-07-06
Fix worktrees list to show registered worktrees

lancekrogers:
## Summary

`camp worktrees list` was scanning every directory under `projects/worktrees/<project>` and treating immediate child directories as worktrees. That made copied checkout leftovers such as `bin`, `src`, `scripts`, and `node_modules` appear as stale worktrees.

This change scopes the global list command to known campaign projects, asks Git for each project)'s registered worktrees, and only renders filesystem candidates that match those registered worktree paths.

## Validation

- `go test ./cmd/camp/worktrees`
- `go test ./internal/worktree`
- `go test ./...`
- `go run ./cmd/camp worktrees list`
- `go run ./cmd/camp worktrees list --project camp`

### DIFF
--- a/cmd/camp/worktrees/list.go
+++ b/cmd/camp/worktrees/list.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/charmbracelet/lipgloss"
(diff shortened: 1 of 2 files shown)


obey-agent (commented):
**Verdict: Approve**

This is a focused fix for `camp worktrees list`: it stops treating arbitrary copied checkout children under `projects/worktrees/<project>/` as worktrees and filters filesystem candidates against Git's registered worktree list. The implementation is small, follows existing project/worktree helpers, and has a targeted regression test for the copied-checkout case.

**Key Findings**
- No blocking issues found.

**What's Done Well**
- Uses Git's registered worktree state instead of directory-name heuristics.
- Keeps `--project` errors explicit while allowing the global list to skip projects that cannot report worktrees.
- Adds a focused unit test for the bad listing behavior described in the PR.

**Staff Standard**
Yes. I would be comfortable submitting this change.

Validation run: `go test ./cmd/camp/worktrees ./internal/worktree` passed in the detached PR worktree.

PR #381, opened 2026-07-08
fix: improve intent action responsiveness

obey-agent:
## Summary
- Caches intent explorer searchable text so live search avoids rebuilding content on each keypress.
- Moves best-effort TUI auto-commits into a background goroutine after successful intent/audit updates.

## Validation
- go test ./internal/intent/tui/explorer
- just gate-fast
- just test unit
- git diff --check

## Notes
- Non-TUI CLI intent commands still commit synchronously; a durable background worker would be a separate design.

### DIFF
--- a/cmd/camp/intent/explore.go
+++ b/cmd/camp/intent/explore.go
@@ -1,10 +1,12 @@
 package intent
 
 import (
+	"fmt"
 	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
+	"time"
 
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 
@@ -134,8 +136,31 @@ func runIntentExplore(cmd *cobra.Command, args []string) error {
 		return camperrors.Wrap(err, "running explorer")
 	}
 
+	// The explorer fires intent auto-commits asynchronously to keep the UI
+	// responsive, so drain any still in flight before returning; otherwise a
+	// change already written to disk could exit without ever being committed.
+	// A fast drain stays silent. If commits are still running (e.g. contending
+	// for the git lock) tell the user why the terminal is pausing, then wait a
+	// bounded amount so a wedged lock cannot hang the exit.
+	if !explorer.WaitForAutoCommits(quickDrainTimeout) {
+		_, _ = fmt.Fprintln(cmd.ErrOrStderr(), "Finalizing intent commits...")
+		if !explorer.WaitForAutoCommits(autoCommitDrainTimeout) {
+			_, _ = fmt.Fprintln(cmd.ErrOrStderr(), "warning: some intent auto-commits did not finish; run 'camp status' to check for uncommitted intent changes")
+		}
+	}
+
 	return nil
 }
+
+// quickDrainTimeout is the silent grace period for in-flight auto-commits to
+// finish on exit before the user is told the shutdown is waiting on them.
+const quickDrainTimeout = 300 * time.Millisecond
+
+// autoCommitDrainTimeout bounds how long the explorer waits for background
+// auto-commits to finish on exit. The commit path retries on git lock
+// contention for several seconds (internal/git/retry.go), so this caps the wait
+// generously while still ensuring a wedged lock cannot hang process shutdown.
+const autoCommitDrainTimeout = 15 * time.Second
 
 // quietSlogDuringTUI swaps slog.Default with a handler that writes to a log
 // file under <campaignRoot>/.campaign/logs/ instead of stderr. It returns a
--- a/cmd/camp/intent/explore.go
+++ b/cmd/camp/intent/explore.go
@@ -1,12 +1,10 @@
 package intent
 
 import (
-	"fmt"
 	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
-	"time"
 
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 
@@ -127,50 +125,19 @@ func runIntentExplore(cmd *cobra.Command, args []string) error {
 	}
 	defer restoreLogger()
 
-	// The explorer fires intent auto-commits asynchronously to keep the UI
-	// responsive, so any still in flight must be drained before this function
-	// returns; otherwise a change already written to disk could exit without
-	// ever being committed. Deferred (not inline after p.Run) so the drain also
-	// covers error-exit paths, and registered after restoreLogger so LIFO runs
-	// it first, while slog is still routed to the TUI log file.
-	defer drainExplorerAutoCommits(cmd.ErrOrStderr())
-
-	// Create and run the TUI
 	model := explorer.NewModel(ctx, svc, conceptSvc, intentsDir, campaignRoot, cfg.ID, author, shortcuts).
 		WithAvailableTags(cfg.IntentTags())
-	p := tea.NewProgram(model, tea.WithAltScreen())
 
-	if _, err := p.Run(); err != nil {
+	// Deferred after restoreLogger so LIFO drains in-flight auto-commits (on both
+	// success and error exits) while slog is still routed to the TUI log file.
+	defer model.DrainAutoCommits(cmd.ErrOrStderr())
+
+	if _, err := tea.NewProgram(model, tea.WithAltScreen()).Run(); err != nil {
 		return camperrors.Wrap(err, "running explorer")
 	}
 
 	return nil
 }
-
-// drainExplorerAutoCommits waits for the explorer's background intent
-// auto-commits to finish on exit. A fast drain stays silent; if commits are
-// still running (e.g. contending for the git lock) it tells the user why the
-// terminal is pausing, then waits a bounded amount so a wedged lock cannot hang
-// the exit.
-func drainExplorerAutoCommits(w io.Writer) {
-	if explorer.WaitForAutoCommits(quickDrainTimeout) {
-		return
-	}
-	_, _ = fmt.Fprintln(w, "Finalizing intent commits...")
-	if !explorer.WaitForAutoCommits(autoCommitDrainTimeout) {
-		_, _ = fmt.Fprintln(w, "warning: some intent auto-commits did not finish; run 'camp status' to check for uncommitted intent changes")
-	}
-}
-
-// quickDrainTimeout is the silent grace period for in-flight auto-commits to
-// finish on exit before the user is told the shutdown is waiting on them.
-const quickDrainTimeout = 300 * time.Millisecond
-
-// autoCommitDrainTimeout bounds how long the explorer waits for background
-// auto-commits to finish on exit. The commit path retries on git lock
-// contention for several seconds (internal/git/retry.go), so this caps the wait
-// generously while still ensuring a wedged lock cannot hang process shutdown.
-const autoCommitDrainTimeout = 15 * time.Second
 
 // quietSlogDuringTUI swaps slog.Default with a handler that writes to a log
 // file under <campaignRoot>/.campaign/logs/ instead of stderr. It returns a
--- a/cmd/camp/intent/explore.go
+++ b/cmd/camp/intent/explore.go
@@ -127,6 +127,14 @@ func runIntentExplore(cmd *cobra.Command, args []string) error {
 	}
 	defer restoreLogger()
 
+	// The explorer fires intent auto-commits asynchronously to keep the UI
+	// responsive, so any still in flight must be drained before this function
+	// returns; otherwise a change already written to disk could exit without
+	// ever being committed. Deferred (not inline after p.Run) so the drain also
+	// covers error-exit paths, and registered after restoreLogger so LIFO runs
+	// it first, while slog is still routed to the TUI log file.
+	defer drainExplorerAutoCommits(cmd.ErrOrStderr())
+
 	// Create and run the TUI
 	model := explorer.NewModel(ctx, svc, conceptSvc, intentsDir, campaignRoot, cfg.ID, author, shortcuts).
 		WithAvailableTags(cfg.IntentTags())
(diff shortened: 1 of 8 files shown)


internal/intent/tui/explorer/audit.go
```diff
@@ -44,5 +55,6 @@ func (m *Model) autoCommitIntent(action commit.IntentAction, title, description
 		Action:      action,
 		IntentTitle: title,
 		Description: description,
-	})
+	}
```
obey-agent: **Blocker — concurrent git commits:** `autoCommitIntent` now fires unbounded background goroutines with no serialization. Rapid intent actions (move/archive/promote) can run overlapping `commit.Intent` calls against the same campaign repo; failures are swallowed (`_ = commit.Intent`). Before this change commits were synchronous and naturally serialized. Please queue or single-flight background commits (channel/worker, or sync.Mutex around intent commits) so audit trails cannot be dropped on index.lock races.

obey-agent (commented):
## Staff Review — Request Changes (comment; obey-agent cannot self-request-changes)

**Verdict:** Request Changes

**Overview:** Search corpus caching and async auto-commits are sensible responsiveness wins; explorer tests pass.

**Key Findings:**
1. **Concurrent background commits** (inline on `audit.go:58`) — parallel goroutines can race on git index lock; errors are discarded. Queue or single-flight intent commits.

**Staff Standard:** Not yet — fix commit serialization before merge.

internal/intent/tui/explorer/audit.go
```diff
@@ -44,5 +58,10 @@ func (m *Model) autoCommitIntent(action commit.IntentAction, title, description
...
+	go func() {
```
obey-agent: Prior blocker resolved: `autoCommitIntentMu` serializes background commits and `TestAutoCommitIntentSerializesBackgroundCommits` proves the second goroutine waits for the first. Non-blocking UI path preserved.

obey-agent (commented):
## Staff Review (re-check) — Approve (comment; obey-agent cannot self-approve)

**Verdict:** Approve

**Overview:** Re-reviewed at `3146b5c`. Follow-up adds `autoCommitIntentMu` around the background commit path, addressing the prior concurrent git race. Explorer tests pass (`go test ./internal/intent/tui/explorer/...`), including new async + serialization coverage.

**Key Findings:** None blocking.

**What's Done Well:** Mutex held only around `commit.Intent` in the worker goroutine; UI still returns immediately; search corpus cache rebuilds on intent reload; PR notes clarify non-TUI paths remain synchronous.

**Staff Standard:** Yes — responsiveness win without sacrificing commit serialization.

internal/intent/tui/explorer/audit.go
obey-agent: **Blocking — auto-commits are lost on quit.** These background commits are raw `go func(){}()` goroutines, invisible to bubbletea (they are not `tea.Cmd`s), and nothing drains them before the process exits. On quit (`update_keys.go:55` → `tea.Quit` → `p.Run()` returns in `cmd/camp/intent/explore.go:133` → `runIntentExplore` returns → process exits), an in-flight commit goroutine is abandoned: the intent change is already written to disk synchronously, but the `git commit` never happens.

This is a **regression** from the previous synchronous commit — bubbletea processes one message at a time, so the old code guaranteed every action was committed before the quit message could be handled. The loss window is not sub-millisecond either: the lock-aware retry path (`internal/git/retry.go`: 2 cycles + up to 2s backoff + 5s `ActiveLockWait`) means a contended commit can run for several seconds, and because the mutex serializes, a slow first commit plus a queued second commit can both be lost.

Auto-commit exists precisely to record intent transitions in git history — silently dropping it defeats the feature. **Fix:** add a package-level `sync.WaitGroup` alongside `autoCommitIntentMu` (`Add(1)` before the `go`, `defer Done()` inside), and drain it (bounded, so a stuck lock cannot hang exit) after `p.Run()` returns in `explore.go`.

internal/intent/tui/explorer/audit.go
```diff
@@ -1,12 +1,20 @@
...
+	"context"
 	"strings"
+	"sync"
 
 	"github.com/Obedience-Corp/camp/internal/git/commit"
 	"github.com/Obedience-Corp/camp/internal/intent/audit"
 )
 
+var runAutoCommitIntent = func(ctx context.Context, opts commit.IntentOptions) {
+	_ = commit.Intent(ctx, opts)
```
obey-agent: **Discarded error (prior finding only half-addressed).** The round-1 review flagged two things: the git-index race *and* discarded errors. The mutex fixed the race, but the error here is still fully dropped (`_ = commit.Intent(...)`), so a failed auto-commit is now completely invisible — not even in the new `intent-explore.log`.

Since slog is already routed to that logfile for the duration of the TUI (`explore.go:122`), surface the failure at zero UI cost:

```go
if res := commit.Intent(ctx, opts); res.Err != nil {
    slog.WarnContext(ctx, "intent auto-commit failed",
        "action", opts.Action, "intent", opts.IntentTitle, "error", res.Err)
}
```

`commit.Result.Err` is nil on a no-op (no changes), so this will not log spurious warnings. Best-effort is fine; silent-and-unrecoverable is not.

obey-agent (commented):
## Staff Review (re-check #2) — Request Changes (comment; obey-agent cannot self-request-changes)

**Verdict:** Request Changes — the prior Approve was premature.

**Overview:** The `autoCommitIntentMu` mutex genuinely fixes the round-1 git-index race (verified: `go test -race ./internal/intent/tui/explorer/...` is clean). But converting the auto-commit to fire-and-forget goroutines introduced a **new durability regression that neither review round evaluated**, and the added tests only exercise the mutex — not the failure mode — so the green suite reinforced false confidence.

**Key Findings:**
1. **Auto-commits are silently lost on quit** (inline on `audit.go:62`). Background commits are raw goroutines invisible to bubbletea; nothing drains them before `p.Run()` returns and the process exits. Action-then-quit — normal TUI usage — leaves the intent change on disk but never committed. Regression from the previous synchronous path. Window widened to seconds by the lock-aware retry subsystem. Fix: `sync.WaitGroup` alongside the mutex, drained (bounded) after `p.Run()`.
2. **Discarded commit error** (inline on `audit.go:13`). Round 1 flagged the race *and* dropped errors; only the race was fixed. Log the failure via slog (already routed to `intent-explore.log` during the TUI).

**What's Done Well:** `context.WithoutCancel` is the right call to avoid aborting an in-flight commit on ctx cancellation. Mutex serialization is correct and race-clean. Search-corpus caching is correct — every mutation path reloads via `Update` → `rebuildSearchCorpus()`, with a cheap length-guard fallback. Tests are real channel choreography, not stubs.

**Staff Standard:** Not yet — a feature whose whole purpose is recording intent transitions in git must not silently drop those commits when the user quits. Add the drain-on-exit and error logging, and this is a clean merge.

obey-agent:
**Fix pushed in `75f9606`** — both review findings addressed on this branch:

1. **Drain-on-exit (blocking):** background auto-commits are now tracked with a package-level `sync.WaitGroup` (via `WaitGroup.Go`, alongside the existing serialization mutex) and drained by a new `explorer.WaitForAutoCommits(timeout)`, called in `cmd/camp/intent/explore.go` after `p.Run()` returns. Fast drains stay silent; if commits are still contending for the git lock the user sees `Finalizing intent commits...` and the wait is bounded (15s) so a wedged lock cannot hang the exit. An intent change written to disk can no longer exit without being committed.
2. **Discarded error:** `runAutoCommitIntent` now logs failures via `slog.WarnContext` (routed to `intent-explore.log` during the TUI) instead of dropping them.

**Validation:**
- New `TestWaitForAutoCommitsDrainsInFlightCommit` asserts the drain blocks while a commit is in flight and returns once it completes.
- `go test -race -count=5 ./internal/intent/tui/explorer/...` clean.
- `just gate-fast` green (2982/2982 tests, lint clean).

The prior async-serialization tests and search-corpus caching are unchanged.

internal/intent/tui/explorer/audit.go
```diff
@@ -44,5 +72,31 @@ func (m *Model) autoCommitIntent(action commit.IntentAction, title, description
...
+		autoCommitIntentMu.Lock()
+		defer autoCommitIntentMu.Unlock()
+		runAutoCommitIntent(ctx, opts)
 	})
 }
+
+// WaitForAutoCommits blocks until every background intent auto-commit started by
+// autoCommitIntent has finished, or until timeout elapses. The explorer fires
+// commits asynchronously to keep the UI responsive, so the process must drain
+// them on exit; otherwise an intent change already written to disk could be lost
```
obey-agent: Round-2: both prior blockers are resolved here.

- `autoCommitIntentWg` + `WaitForAutoCommits` with the two-stage drain in `explore.go` closes the quit-before-commit gap.
- `runAutoCommitIntent` now logs failures via `slog.WarnContext` (routed to `intent-explore.log` during the TUI) instead of discarding them.
- Serialization via `autoCommitIntentMu` + `TestAutoCommitIntentSerializesBackgroundCommits` / `TestWaitForAutoCommitsDrainsInFlightCommit` give good coverage.

Verified locally: `go test -race -count=3 ./internal/intent/tui/explorer/...` and `just gate-fast` (2982/2982).

cmd/camp/intent/explore.go
```diff
@@ -134,9 +136,32 @@ func runIntentExplore(cmd *cobra.Command, args []string) error {
 		return camperrors.Wrap(err, "running explorer")
 	}
 
+	// The explorer fires intent auto-commits asynchronously to keep the UI
+	// responsive, so drain any still in flight before returning; otherwise a
+	// change already written to disk could exit without ever being committed.
+	// A fast drain stays silent. If commits are still running (e.g. contending
+	// for the git lock) tell the user why the terminal is pausing, then wait a
+	// bounded amount so a wedged lock cannot hang the exit.
+	if !explorer.WaitForAutoCommits(quickDrainTimeout) {
```
obey-agent: Non-blocking follow-up: the drain currently runs only after a successful `p.Run()`. If the program returns early on `running explorer` error, any in-flight auto-commit goroutine can still be cut off at process exit.

Consider a `defer` drain (same two-stage messaging) immediately after `tea.NewProgram` so commits drain on both success and error paths. Low probability in practice, but cheap insurance.

obey-agent (commented):
## Staff Re-Review: camp #381 — intent action responsiveness

**Verdict:** Approve

**Overview:** Commit `75f9606` fully addresses the round-2 blockers from the async auto-commit work. Background intent commits are now serialized, logged on failure, tracked with a package-level WaitGroup, and drained on explorer exit with a sensible two-stage timeout + user messaging. The search-corpus cache remains a clean win for live fuzzy filtering.

**Resolved since round 2:**
- **Quit-before-commit gap:** `WaitForAutoCommits` + drain in `runIntentExplore` (300ms silent grace, then 15s bounded wait with stderr guidance)
- **Discarded errors:** `runAutoCommitIntent` logs via `slog.WarnContext` → `intent-explore.log` during TUI
- **Concurrent git index races:** `autoCommitIntentMu` serialization preserved from round 1

**Verified locally:**
- `go test -race -count=3 ./internal/intent/tui/explorer/...`
- `just gate-fast` — 2982/2982 tests, lint clean

**Non-blocking follow-up (inline on `explore.go`):** drain only runs after successful `p.Run()`; a `defer` drain would also cover error exits. Cheap insurance, not a merge blocker.

**What's Done Well:**
- `context.WithoutCancel` keeps in-flight commits alive after TUI ctx cancellation
- `sync.WaitGroup.Go` (Go 1.25) is the right primitive here
- Search corpus rebuilt on `intentsLoadedMsg`; length-mismatch guard is sufficient because mutations reload intents
- Test stubs for async/serialize/drain behavior are thorough

**Staff Standard:** Yes — merge-ready.

obey-agent:
**Follow-up addressed in `eb1a695`.**

The non-blocking follow-up from the re-review (drain only ran after a *successful* `p.Run()`, so an error exit could still abandon in-flight commits) is fixed: the drain is now a `defer drainExplorerAutoCommits(cmd.ErrOrStderr())` registered right after `defer restoreLogger()`, so it runs on every return path (success **and** error) and, by LIFO ordering, drains while slog is still routed to `intent-explore.log`. Extracted the two-stage drain into `drainExplorerAutoCommits`, with `TestDrainExplorerAutoCommits_SilentWhenNothingPending` guarding against spurious "Finalizing..." output on a normal quit.

`just gate-fast` green (2983/2983 tests, lint clean); explorer + intent packages race-clean.

**Deliberately not folded into this PR:** the PR-notes item — non-TUI `camp intent` CLI commands still commit synchronously and the in-process mutex does not serialize across processes (only the git retry/lock subsystem mitigates that today). That is a real cross-process design effort, so it is tracked as its own intent (`durable-cross-process-auto-commit-...`) rather than bloating this responsiveness PR.

cmd/camp/intent/explore.go
```diff
@@ -125,6 +127,14 @@ func runIntentExplore(cmd *cobra.Command, args []string) error {
 	}
 	defer restoreLogger()
 
+	// The explorer fires intent auto-commits asynchronously to keep the UI
+	// responsive, so any still in flight must be drained before this function
+	// returns; otherwise a change already written to disk could exit without
+	// ever being committed. Deferred (not inline after p.Run) so the drain also
+	// covers error-exit paths, and registered after restoreLogger so LIFO runs
+	// it first, while slog is still routed to the TUI log file.
+	defer drainExplorerAutoCommits(cmd.ErrOrStderr())
```
obey-agent: Round-4 follow-up resolved. Moving the drain into a `defer` (registered after `restoreLogger` so LIFO runs drain first while slog still routes to `intent-explore.log`) closes the error-exit gap flagged in round 3. Success and `running explorer` error paths now both wait for in-flight commits with the same two-stage timeout + user messaging.

obey-agent (commented):
## Staff Re-Review (round 4) — camp #381

**Verdict:** Approve (comment; obey-agent cannot self-approve)

**Overview:** Re-reviewed at `eb1a695`. This commit closes the last open item from round 3 by deferring the auto-commit drain so it runs on both success and `running explorer` error exits, with correct LIFO ordering against `restoreLogger`. Combined with the earlier mutex serialization, WaitGroup tracking, failure logging, and search-corpus cache, the PR is a solid responsiveness improvement without sacrificing commit durability.

**Resolved since round 3:**
- **Error-exit drain gap:** `defer drainExplorerAutoCommits(...)` after `restoreLogger` — drain runs first (slog still on `intent-explore.log`), covers all exit paths
- All round-1/2 blockers remain fixed: serialized commits, quit-before-commit drain, `slog.WarnContext` on failure

**Verified locally:**
- `go test -race -count=2 ./internal/intent/tui/explorer/... ./cmd/camp/intent/...`
- `just gate-fast` — 2983/2983 tests, lint clean

**Non-blocking notes:**
- Package-level `autoCommitIntentWg` is fine today (no `t.Parallel()` in explorer tests); reset/isolation would matter only if parallel explorer tests are added later
- `autoCommitDrainTimeout` (15s) is a reasonable cap vs. `internal/git/retry.go`; the stderr warning on timeout is the right escape hatch if two serialized commits contend heavily

**What's Done Well:**
- `context.WithoutCancel` keeps in-flight commits alive after TUI ctx cancellation
- Two-stage drain UX (300ms silent, then bounded wait + guidance) is operator-friendly
- Search corpus rebuilt on every `intentsLoadedMsg`; length guard is sufficient because mutations reload intents

**Staff Standard:** Yes — merge-ready.

obey-agent:
**Non-blocking notes addressed in `fca0784`.**

- **Package-global auto-commit state (note 1):** the serialization mutex, WaitGroup, and run-hook now live in an `autoCommitter` that the `Model` owns by pointer, drained via `Model.DrainAutoCommits`. This removes the global mutable state and isolates the tests — each `Model` gets its own committer, so the cross-test "absorb stragglers" workaround is gone and parallel explorer tests would be safe. (The pointer field means bubbletea's value copies still share one committer, so the deferred drain in `runIntentExplore` works unchanged.)
- **15s drain timeout (note 2):** left as-is — that note endorsed the cap + stderr warning as the right escape hatch, so no change.

Also trimmed the oversized comment blocks the earlier commits added, per feedback. Net −25 lines. `just gate-fast` green (2983/2983, lint clean); explorer + intent packages race-clean at `-count=3`.

PR #383, opened 2026-07-09
feat(fresh): add --list to cycle a set of projects in one run

lancekrogers:
## Summary

`camp fresh` previously offered only two granularities: one project at a time (positional arg, `--project`, or cwd auto-detect) or **every** submodule via `camp fresh all`. This PR adds the middle ground via an explicit `--list` flag:

```bash
camp fresh --list camp,fest,festival    # cycle just these three
```

An explicit flag (rather than variadic positional args) keeps the single-project positional form unambiguous — a stray shell token or typo can't silently become a project target.

## Behavior

- **`--list a,b,c`** runs the full fresh cycle (checkout default, pull, prune, optional branch/push) against each named project, honoring all existing flags (`--branch`, `--no-prune`, `--no-push`, `--dry-run`) and per-project `fresh.yaml` overrides.
- **Mutually exclusive**: `--list` cannot be combined with a positional project name or with `--project` (both rejected with a clear error rather than silently ignoring one).
- **Fail-fast validation**: every name is resolved up front, so an unknown project name aborts the batch *before* any project is mutated. Per-project *execution* errors (e.g. uncommitted changes in one repo) are collected and reported in an aggregate summary, matching `camp fresh all`.
- **Robust parsing**: entries are trimmed and empties dropped (`--list=camp, fest,` → `camp`, `fest`); duplicates collapse to a single target, preserving first-seen order.
- Single-project and `camp fresh all` behavior is unchanged. The batch runner is now shared: `all` delegates to it, so its output is identical.

## Implementation

- New `internal/commands/fresh/batch.go` holds the shared `runFreshBatch` runner, `freshTarget`/`freshFlagSet`, target resolution, and list sanitization.
- `fresh.go` adds the `--list` `StringSlice` flag (mutually exclusive with `--project`), routes a non-empty list through the batch path, and keeps `Args: MaximumNArgs(1)` so a stray extra positional still errors.
- `all.go` is reduced to building the target list and delegating to `runFreshBatch` (removes duplicated loop/summary code).
- Dropped a pre-existing unused `freshStepRed` style variable while touching the file.

## Testing

Three container-backed integration tests in `tests/integration/fresh_test.go`:

- `TestFresh_ListCyclesEachProject` — `camp fresh --list test-a,test-b --branch develop` cycles both onto the new branch and announces the project count.
- `TestFresh_ListFailFastOnUnknownName` — an unknown name in the list aborts before mutating any project (asserts `test-a` stays on `main`, no `develop` branch created).
- `TestFresh_ListRejectsPositionalArg` — combining a positional name with `--list` is rejected and mutates nothing.

All 11 `TestFresh_*` integration tests pass (including the pre-existing single-project and worktree suites). `just lint` reports `0 issues`; `just vet` and both lint ratchets (`no-fmt-errorf`, `no-host-fs-tests`) are clean.

### Real CLI output (dry-run)

```
$ camp fresh --list camp,fest --dry-run
Running fresh across 2 project(s)...

  camp (dry-run)
  ...
  fest (dry-run)
  ...

──────────────────────────────────────────────────
All done! Fresh completed for 2 project(s)
```

### DIFF
--- a/internal/commands/fresh/batch.go
+++ b/internal/commands/fresh/batch.go
@@ -1,0 +1,152 @@
+package fresh
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/charmbracelet/lipgloss"
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/project"
+	"github.com/Obedience-Corp/camp/internal/ui"
+)
+
+// freshFlagSet captures the resolved persistent flags shared by every fresh
+// invocation. It is passed to runFreshBatch so per-project settings can be
+// resolved against the fresh config.
+type freshFlagSet struct {
+	branch   string
+	noBranch bool
+	noPush   bool
+	noPrune  bool
+	dryRun   bool
+}
+
+// freshTarget is a resolved project to run the fresh cycle against.
+type freshTarget struct {
+	name string
+	path string
+}
+
+// getFreshFlagSet extracts the fresh persistent flags from the parent command.
+// These are stored on the parent (freshCmd) and inherited by subcommands.
+func getFreshFlagSet(freshCmd *cobra.Command) freshFlagSet {
+	branch, _ := freshCmd.PersistentFlags().GetString("branch")
+	noBranch, _ := freshCmd.PersistentFlags().GetBool("no-branch")
+	noPush, _ := freshCmd.PersistentFlags().GetBool("no-push")
+	noPrune, _ := freshCmd.PersistentFlags().GetBool("no-prune")
+	dryRun, _ := freshCmd.PersistentFlags().GetBool("dry-run")
+	return freshFlagSet{
+		branch:   branch,
+		noBranch: noBranch,
+		noPush:   noPush,
+		noPrune:  noPrune,
+		dryRun:   dryRun,
+	}
+}
+
+// cleanProjectList trims surrounding whitespace from each entry and drops
+// empties, so '--list=camp, fest,' resolves to ["camp", "fest"] rather than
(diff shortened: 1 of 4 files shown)


internal/commands/fresh/batch.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		seen[result.Name] = struct{}{}
+		targets = append(targets, freshTarget{name: result.Name, path: result.Path})
+	}
+	return targets, nil
+}
+
+// runFreshBatch runs the fresh cycle across multiple targets. It continues on
+// per-project execution errors and reports an aggregate summary, returning an
+// error if any target failed. This backs both `camp fresh <a> <b> ...` and
```
obey-agent, line 99: **Maintainability:** This comment says the batch runner backs `camp fresh <a> <b> ...`, but the implementation only exposes `--list` (variadic positionals are explicitly rejected). Update the doc comment to `camp fresh --list` and `camp fresh all` so future readers don't hunt for a code path that doesn't exist.

obey-agent (commented):
## Verdict: Approve (comment)

## Overview
Clean middle-ground for `camp fresh`: `--list` cycles an explicit project set without the ambiguity of variadic positionals. The shared `runFreshBatch` refactor deduplicates `fresh all` and list paths, fail-fast target resolution is correct, and container integration tests cover the happy path, unknown-name abort, and positional conflict.

## Key Findings
- Stale doc comment on `runFreshBatch` (inline) — references variadic `camp fresh <a> <b>` that does not exist; only `--list` and `all`.

## What's Done Well
- `MarkFlagsMutuallyExclusive(project, list)` plus explicit positional+list rejection.
- Up-front `resolveFreshTargets` with duplicate collapse and project-list hint on not-found.
- `fresh all` delegates to the same batch runner — output parity preserved.
- Integration tests use container isolation per repo standards.

## Staff Standard
Yes — small, focused CLI improvement with the right safety properties. Fix the doc comment and merge.

PR #384, opened 2026-07-09
Multi-machine campaign navigation (MM0001): camp list --remote + camp switch machine:campaign

lancekrogers:
## Multi-machine campaign navigation (MM0001)

Brings the festival app's multi-machine feature to the terminal: `camp list --remote`
sees campaigns on every ssh/Tailscale-reachable machine, and `camp switch
machine:campaign` (via `csw`) lands a login shell in that campaign **on that
machine**, in one command. camp stays daemon-free; local commands stay
byte-identical and no slower.

Ships slices 0–3 of the spec at `workflow/design/multi-machine-campaign-navigation/`.

### What's in it

**Slice 0 — `~/.obey/machines.yaml` + typed library** (`internal/machines`)
- Schema mirrors the festival app's `RemoteMachine` field-for-field
  (`id`/`label`/`host`/`auth_method`/`ssh_user`/`identity_file`) so the app can
  adopt the same file with a serde rename, not a redesign.
- `Load` (absent file ⇒ zero machines, nil error — the degradation guarantee),
  atomic `Save` (0600, never persists the reserved `local`), `Lookup`, `Upsert`.
- `MachinesPath()` mirrors `internal/config/registryfile.Path`'s resolution
  (env override → `XDG_CONFIG_HOME` → `~/.obey`).

**Slice 1 — `camp switch machine:campaign`** (`cmd/camp/switch.go`, `internal/remote`)
- `cmdutil.ParseMachineSelector` splits an optional `machine:` prefix before the
  existing `org/campaign[@tab]` grammar. Colon is collision-free (the shipped
  parser never inspects `:`), and a selector with no colon is byte-identical.
- `internal/remote` ssh primitives mirror the app's construction
  (`Target`/`Opts`/`ShellQuote`/`Run`/`ResolveRoot`, `StrictHostKeyChecking=accept-new`,
  `BatchMode=yes`, `IdentitiesOnly`, ControlMaster).
- Remote root resolved by the remote's OWN `camp switch --print` over ssh (its
  registry decides the path). Hidden `--shell-connect` emits one eval-able line:
  `cd -- <path>` locally, `exec ssh -t … "cd <root> && exec $SHELL -l"` remotely
  (injection-safe quoting). zsh/bash/fish `switch|sw` branches now eval it.
  ControlMaster means resolve + hop share one connection.
- v1 is agent/key auth only; a `ssh-password` machine errors with a clear
  "configure key auth" message instead of hopping.

**Slice 2 — `camp list --remote`** (`cmd/camp/list_remote.go`)
- Concurrent, per-machine-bounded fan-out of ssh-exec `camp list --json`;
  re-tags rows with the machine id; `ConnectTimeout=8` so a dead host fails fast.
- `campaignEntry.Machine` is `omitempty` ⇒ **default `camp list --json` is
  byte-for-byte unchanged**; local rows tag `local` only under `--remote`.
- Human table gains a leading `MACHINE` column only when a remote machine is
  present (single-machine output unchanged); unreachable machines render as one
  muted labeled row and never fail the command (exit 0). `--json` keeps stdout
  clean (unreachable → stderr). `tags` `[]`-not-`null` guard preserved.

**Slice 3 — completion + `camp machine` CLI**
- Machine-dimension tab completion: `<machine>:` candidates always; per-machine
  campaign completion from a short-TTL cache warmed by `camp list --remote`. The
  keystroke path **never** does a live ssh (structural; miss ⇒ machine id only).
- `camp machine list` (+`--json`, synthetic `local` row) / `add` (idempotent via
  `Upsert`, `local` reserved, `--auth` validated) / `remove`, and `add --discover`
  which parses `tailscale status --json` (a faithful mirror of the app's
  `interpret_tailscale_status` / `into_discovered_device`) and picks a device.

### The hard constraint (verified, not just intended)

Multi-machine is pure enrichment. The local path gains no latency and no network
call — enforced structurally (the `machine:` split short-circuits without a colon,
`--remote` defaults off, the `machine` json field is `omitempty`), not by tuning.
`TestPlainListMakesNoSshCall` proves that with a machines file **present** but no
`--remote`, `camp list` makes zero ssh calls (ssh shim records nothing) and
`camp list --json` is byte-identical.

### Testing

- Host unit tests across all packages (selector collision guard, ssh helpers,
  fan-out re-tag/isolation, completion cache logic, tailscale parse vs a committed
  fixture, machine CLI validation/round-trip).
- **12 container integration tests** (`tests/integration/`, `//go:build integration`)
  over real loopback ssh, all passing together: the `self:` switch hop lands in the
  right dir with a single shared ControlMaster socket; `list --remote` degrades on
  an unreachable host; completion reads the ssh-warmed cache; `camp machine`
  round-trips all six fields; `--discover` via a stubbed tailscale; `--no-remote`
  makes zero ssh calls.
- `just build` / `just lint` (0 issues) / `lint-no-fmt-errorf` / `lint-no-host-fs-tests`
  green. No `fmt.Errorf` in production; file-mutating tests are container-only.

### Tradeoffs / known limitations

- **Owning repo = camp** (`internal/machines`), not obey-shared. The file + schema
  is the cross-tool contract regardless of package home (the app reads the YAML via
  serde, never the Go package), and `MachinesPath` must share camp's own
  `registryfile` resolution, which obey-shared can't reference without importing
  camp or duplicating it. Promotion to obey-shared stays a cheap future move.
- **Remote `camp` must be on the non-interactive ssh PATH.** The remote *resolve*
  runs a non-login shell, so a `~/go/bin` / `/opt/homebrew/bin` / npm-global install
  needs a symlink onto `/usr/bin`/`/usr/local/bin`. The resolve error surfaces the
  remote `camp: not found` stderr. (A login-shell wrapper is a portability minefield
  — busybox `ash`/`dash` reject `-l` — so it's deliberately not done in v1.)
- **Shell integration changed** — after merge, re-run `eval "$(camp shell-init zsh|bash|fish)"`.
- **Version-skew write:** a v1 writer re-stamps `version: 1` and drops unknown
  fields; fine while camp is the sole writer, to be co-designed when the app adopts
  the file.

Out of scope (deferred by design): daemon-forward session-count enrichment (slice 4);
password-auth remote switch.


cmd/camp/list_remote.go
```diff
@@ -0,0 +1,167 @@
...
+
+// enumerateRemote runs the remote machine's OWN `camp list --json` over ssh, so
+// its registry, org config, and absolute paths are authoritative, then re-tags
+// each row with the machine id. Remote paths are left verbatim (meaningful only on
+// the far machine).
+func enumerateRemote(ctx context.Context, m *machines.Machine) ([]campaignEntry, error) {
+	if err := remote.EnsureKeyAuth(m); err != nil {
+		return nil, err
+	}
+	out, err := remote.Run(ctx, remote.Target(m), remote.Opts(m), "camp list --json")
```
obey-agent: **Bug: local list filters are not applied to remote rows**

`enumerateRemote` always runs bare `camp list --json` (remote’s own default active set). In `renderListTable`, local rows are `filterEntries(...)`’d first, then remote rows are appended **unfiltered**.

Consequences:

1. `camp list --remote --org obey` shows only local obey campaigns but **all** remote orgs.
2. `camp list --remote --tag x` / `--status` same leak.
3. `camp list --remote --all` still only gets whatever the remote default emits (usually active-only), because `--all` is never passed over ssh.

**Fix direction:**

- Pass equivalent filter flags into the remote command (`--org`, `--tag`, `--status`, `--all` as set), **and/or**
- Re-run `filterEntries` on the combined slice after fan-out (idempotent for already-filtered local rows; required for org/tag at minimum).

Prefer both: remote-side filter reduces payload; local re-filter is the correctness backstop. Add a unit test with a fake enumerate returning mixed orgs under `--org`.

cmd/camp/switch.go
```diff
@@ -304,9 +336,29 @@ func runSwitch(cmd *cobra.Command, args []string) error {
...
+// forces a PTY, and $SHELL expands on the REMOTE side because the whole remote
+// command is single-quoted here so the local shell never touches the '$'.
+func emitShellConnect(w io.Writer, isRemote bool, path string, m *machines.Machine) error {
+	if !isRemote {
+		_, err := fmt.Fprintf(w, "cd -- %s\n", remote.ShellQuote(path))
+		return err
+	}
+	inner := "cd " + remote.ShellQuote(path) + " && exec $SHELL -l"
+	_, err := fmt.Fprintf(w, "exec ssh -t %s %s %s\n",
+		strings.Join(remote.Opts(m), " "), remote.ShellQuote(remote.Target(m)), remote.ShellQuote(inner))
```
obey-agent: **Robustness: ssh opts are joined unquoted for shell eval**

Path and target are correctly `ShellQuote`’d, but `strings.Join(remote.Opts(m), " ")` leaves `ControlPath=...` and `-i <identity>` as bare tokens. A home directory or identity file path containing spaces (or other shell metacharacters) breaks the eval’d line before ssh runs.

`remote.Run` is fine (exec args, no shell). Only the `--shell-connect` emission path is affected.

**Fix:** quote every opt token the same way as the target:

```go
opts := remote.Opts(m)
quoted := make([]string, len(opts))
for i, o := range opts {
  quoted[i] = remote.ShellQuote(o)
}
fmt.Fprintf(w, "exec ssh -t %s %s %s\n",
  strings.Join(quoted, " "), remote.ShellQuote(remote.Target(m)), remote.ShellQuote(inner))
```

Add a test with `IdentityFile: "/tmp/my key"` or a ControlPath-bearing home with a space.

cmd/camp/list.go
```diff
@@ -125,6 +132,25 @@ func renderListTable(cmd *cobra.Command) error {
 	sortBy, _ := cmd.Flags().GetString("sort")
 	campaigns := filterEntries(sortCampaigns(reg.Campaigns, sortBy, reg.FallbackOrg()), filter)
 
+	var remoteResults []remoteResult
+	if listRemote {
```
obey-agent, line 136: **Bug: `--remote` is swallowed by the interactive list TUI**

`runList` checks `listTUIRequested` *before* this block. On a TTY, `camp list --remote` has no shaping flag in the current `listTUIRequested` set (`sort`, `org`, `tag`, `status`, `all`, `group`, `no-group` — **not** `remote`), so it opens the local campaign browser and never reaches the fan-out below.

Operators typing `camp list --remote` get the local TUI with no fleet listing and no indication multi-machine ran.

**Fix (minimal):** treat `--remote` as a shaping flag in `listTUIRequested` (add `"remote"` to the `Changed` loop), or short-circuit before the TUI gate:

```go
if listRemote {
  return renderListTable(cmd)
}
```

Add a unit test: `--remote` + isTTY ⇒ TUI not requested.

obey-agent (commented):

## Overview
Ambitious, well-structured MM0001 delivery: `~/.obey/machines.yaml`, `camp switch machine:campaign` hop via `--shell-connect` + ControlMaster, concurrent `camp list --remote`, completion cache, and `camp machine` CLI with Tailscale discover. The hard constraint (local path stays inert without `--remote` / no colon) is designed and tested structurally (`TestPlainListMakesNoSshCall`, omitempty machine field). Container coverage over real loopback ssh is a major confidence booster.

## Key Findings (must fix)
1. **`--remote` swallowed by list TUI** (`list.go` / `list_tui.go`) — on a TTY, `camp list --remote` opens the local interactive browser because `listTUIRequested` does not treat `remote` as a shaping flag. Fan-out never runs. Force the text path when `--remote` is set.
2. **List filters not applied to remote rows** (`list_remote.go` + append site) — local rows are `filterEntries`’d; remote rows from bare `camp list --json` are appended unfiltered. `--org`/`--tag`/`--status` leak remote campaigns; `--all` is never passed over ssh. Pass filter flags remotely and/or re-filter the combined slice; add a fake-enumerate unit test.

## Secondary (should fix)
3. **`emitShellConnect` joins ssh opts unquoted** (`switch.go`) — path/target are `ShellQuote`’d, but `ControlPath` / `-i` identity paths with spaces break the eval’d line. Quote each opt token; `remote.Run` (exec) is already safe.

## What's Done Well
- Schema parity with festival-app `RemoteMachine`; absent machines file ⇒ zero machines, nil error.
- Selector is collision-free and no-colon is byte-identical; `--print` refuses remote so `cd "$(camp switch --print)"` cannot mis-cd.
- Fan-out isolation (fixed-index results, ConnectTimeout, unreachable as labeled rows / stderr for JSON).
- Completion keystroke path never sshs (cache warmed by `list --remote` only).
- Password auth rejected before hop; remote remainder single-quoted for injection safety (tested).
- Shell templates eval `--shell-connect` with local/remote uniformity; re-init called out in PR body.
- Integration suite (12 tests) including ControlMaster reuse and `--no-remote` inert proof.

## Staff Standard
Not yet — architecture and testing bar are staff-level, but the two list bugs mean the flagship `camp list --remote` entry point is wrong on a normal TTY and filter semantics are inconsistent across machines. Fix those (and preferably shell-connect opt quoting) and this is mergeable.

Reviewed against: https://github.com/Obedience-Corp/camp/pull/384 @ `e9f9ac4`

obey-agent:
Addressed all three findings in `62b1a57` (pushed to `mm0001-multi-machine`).

**1. `--remote` swallowed by the list TUI** — `listTUIRequested` now treats `remote` as a shaping flag, so on a TTY `camp list --remote` forces the text fan-out path instead of opening the local campaign browser. Unit test: `--remote` + TTY ⇒ TUI not requested.

**2. List filters not applied to remote rows** — `enumerateRemoteFor(filter)` re-expresses `--org` / `--tag` / `--status` / `--all` as flags on the remote `camp list --json`. Passing `--all` / `--status` is a fetch-completeness requirement, not just payload reduction: the remote default emits active-only, and a local re-filter cannot recover rows the remote never sent. `renderListTable` then re-filters the combined slice as the authoritative backstop against a version-skewed or looser remote. Unit tests: a `remoteListArgs` table and a fake-enumerate returning mixed orgs under `--org`, narrowed to one row by the backstop.

**3. `emitShellConnect` joined ssh opts unquoted** — every opt token is now `ShellQuote`'d the same way as target/inner, so a `ControlPath` or `-i` identity path containing spaces no longer splits the eval-ed line before ssh runs. Unit test: `IdentityFile: "/tmp/my key"` emits a single quoted token.

**Bonus (caught by the full suite):** running the whole integration suite, not just the 12 multi-machine tests, surfaced a pre-existing regression from the switch-to-`--shell-connect` rewire: the `csw` shell-init tests still assumed the old `--print` wrapper. The camp stub now models `--shell-connect`'s local `cd -- '<dir>'` output that the rewired wrapper evals, so `TestShellInit_{Zsh,Bash}AliasWrappers/csw_delegates_to_camp_switch` pass again (fish stub updated to match).

Gates at `62b1a57`: `just build`, `just lint` (0 issues), both ratchets (`lint-no-fmt-errorf`, `lint-no-host-fs-tests`), the full unit suite, and the full container integration suite all green.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Re-reviewed `62b1a57` against the prior Request Changes. All three must-fix findings are addressed with matching unit tests, and the shell-init stub fix is a real integration-suite regression catch rather than drive-by noise.

## Findings closed
1. **`--remote` + TTY** — `"remote"` is a shaping flag in `listTUIRequested`; new test forces the fan-out path.
2. **Filter semantics** — `remoteListArgs` re-expresses `--org/--tag/--status/--all` with `ShellQuote` (correct: `remote.Run` passes one remote shell string to ssh). Local `filterEntries` re-applied as backstop; tests cover args table + org leak.
3. **`emitShellConnect` opt quoting** — every opt token is `ShellQuote`’d; identity path with spaces covered by unit test.

## Residual notes (non-blocking)
- `ShellQuote` on filter values is the right model for OpenSSH’s single remote-command arg; no argv-vs-shell confusion.
- Integration suite green per author comment; no reason to hold merge on residual nits.

## Staff Standard
Yes — flagship multi-machine entry points now match the hard constraint and the documented filter contract. Merge when ready.

Reviewed against: https://github.com/Obedience-Corp/camp/pull/384 @ `62b1a57`

internal/remote/ssh.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// authArgs mirrors the app's ssh_auth_args (connection.rs:217-239) for the
+// agent/key case: BatchMode=yes (never prompt), plus IdentitiesOnly and -i when
+// an identity file is configured. Password auth is rejected upstream by
+// EnsureKeyAuth, so its interactive-prompt args are never emitted.
+func authArgs(m *machines.Machine) []string {
+	args := []string{"-o", "BatchMode=yes"}
+	if m.IdentityFile != "" {
+		args = append(args, "-o", "IdentitiesOnly=yes", "-i", m.IdentityFile)
```
obey-agent, line 41: If identity_file is stored as ~/path/to/key (common when users type tilde in config), ssh -i will not expand ~. Consider expanding with os.UserHomeDir / filepath.Expand-style helper at authArgs time so camp machine add --identity "~/..." works the same as users expect from OpenSSH config.

internal/remote/ssh.go
```diff
@@ -0,0 +1,129 @@
...
+		"-o", "ControlPersist=30s",
+	}
+	return append(opts, authArgs(m)...)
+}
+
+// controlPath returns the per-machine ssh ControlMaster socket path under
+// ~/.obey/ssh-ctl and best-effort-creates the directory so ControlMaster=auto can
+// bind the socket. A short, per-id name keeps the path under the OS socket-length
+// limit for typical home directories.
+func controlPath(m *machines.Machine) string {
```
obey-agent, line 67: ControlMaster sockets live under ~/.obey/ssh-ctl/<id>.sock with 0700 dir — good. Worth documenting that a stuck ControlMaster (e.g. after sleep/network flap) can leave later hops hanging until ControlPersist expires or the socket is removed; a doctor/machine diagnose tip would help operators.

obey-agent (commented):
## Verdict: **Approve** (with minor follow-ups)

## Overview
Excellent multi-machine navigation delivery. The hard constraint—local path stays byte-identical and never does SSH without an explicit machine dimension—is enforced structurally and covered by unit + container integration tests. Selector grammar, remote list fan-out, ControlMaster hop, and completion cache design are coherent and production-minded.

## Key Findings
- **`identity_file` tilde expansion** (inline on `internal/remote/ssh.go`): OpenSSH config expands `~`; `-i` does not. Worth expanding before use.
- **ControlMaster operational note** (inline): stuck multiplex sockets after network flaps can confuse operators; a short doctor/docs tip would help.
- Neither blocks merge; remote `camp` PATH requirement is already well documented in the PR body.

## What's Done Well
- Collision-free `machine:` prefix before existing switch grammar; no-colon path byte-identical.
- `ShellQuote` + `BatchMode=yes` + password auth rejected at the gate.
- `--remote` `omitempty` machine field preserves local `--json` stability; unreachable hosts degrade without failing the command.
- Completion never SSHs on the keystroke path (cache warm from `list --remote` only).
- Broad host unit tests and 12 container integration tests over real loopback SSH.

## Staff Standard
Yes — I would be proud to ship this as my own. It is large, but the seams are clean and the non-regression story for local camp is exemplary.

PR #386, opened 2026-07-09
fix(intent): let notes be archived/restored in the explorer

lancekrogers:
## Problem

In the intent explorer's **notes view**, a note could not be dungeoned. Both the action menu and the `a` key stubbed note archiving with the message _"Note archiving is not available in the explorer yet"_, so the only lifecycle action available on a note was converting/promoting it into an intent. There was no way to shelve a note you were done with.

## Change

Wire note archiving through the existing `IntentService.ArchiveNote` path (`notes/` → `notes/archived/`) and add a symmetric `RestoreNote` so an archived note is never a dead end.

- **Action menu** (`Enter` → Actions) on a note now offers:
  - **Archive** — enabled for active notes
  - **Restore** — enabled for archived notes
- **`a` key** archives the selected active note directly.
- Archiving a note is **reason-free**, matching the lightweight note model (notes carry no lifecycle metadata or decision records), unlike dungeoning a lifecycle intent which still requires a reason.

Archive emits an `EventArchive` audit event and an `IntentArchive` auto-commit; restore emits an `EventMove` audit event and an `IntentMove` auto-commit — consistent with the existing intent flows.

## Tests

- `RestoreNote` service tests: archived→active restore, and active-note no-op.
- TUI dispatch tests exercising the real `handleActionMenuSelection` path for both `archive` and `restore`.
- Action-menu enablement test: active notes offer Archive (not Restore); archived notes offer Restore (not Archive).

Full local gate passes (`just gate-fast`): build, vet, lint (no new host-fs or `fmt.Errorf` violators), and all 2983 unit tests.

## Notes / tradeoffs

- Note archive is intentionally reason-free to stay consistent with the note model. If reasons are wanted for note dungeoning later, that would be a follow-up.
- Restore is exposed via the action menu only (no dedicated key), since it is a rarer operation than archive.

### DIFF
--- a/internal/intent/notes.go
+++ b/internal/intent/notes.go
@@ -194,6 +194,49 @@ func (s *IntentService) ArchiveNote(ctx context.Context, id string) (*Intent, er
 	return note, nil
 }
 
+// RestoreNote moves a note back from notes/archived/ into the active notes/
+// store, reversing ArchiveNote so an archived note is never a dead end.
+func (s *IntentService) RestoreNote(ctx context.Context, id string) (*Intent, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, camperrors.Wrap(err, "context cancelled")
+	}
+
+	note, err := s.GetNote(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	if note.Status == StatusNote {
+		return note, nil
+	}
+
+	oldPath := note.Path
+	note.Status = StatusNote
+	note.UpdatedAt = time.Now()
+
+	newPath := s.getIntentPath(StatusNote, note.ID)
+	if err := os.MkdirAll(filepath.Dir(newPath), 0755); err != nil {
+		return nil, camperrors.Wrap(err, "creating directory")
+	}
+
+	data, err := SerializeIntent(note)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "serializing note")
+	}
+	if err := fsutil.WriteFileAtomically(newPath, data, 0644); err != nil {
+		return nil, camperrors.Wrap(err, "writing note file")
+	}
+
+	if err := os.Remove(oldPath); err != nil {
+		_ = os.Remove(newPath)
+		return nil, camperrors.Wrap(err, "removing old note file")
+	}
+
+	note.Path = newPath
+	s.invalidateIDIndex()
+	return note, nil
+}
+
 // MoveIntentToNote moves a lifecycle intent into the active notes/ store. The
 // file keeps its title, body, author, tags, and timestamps, but drops lifecycle
 // metadata that notes do not carry.
--- a/internal/intent/notes.go
+++ b/internal/intent/notes.go
@@ -180,6 +180,11 @@ func (s *IntentService) ArchiveNote(ctx context.Context, id string) (*Intent, er
 	if err != nil {
 		return nil, camperrors.Wrap(err, "serializing note")
 	}
+	if _, statErr := os.Stat(newPath); statErr == nil {
+		return nil, camperrors.Wrap(ErrFileExists, newPath)
+	} else if !os.IsNotExist(statErr) {
+		return nil, camperrors.Wrap(statErr, "checking destination note file")
+	}
 	if err := fsutil.WriteFileAtomically(newPath, data, 0644); err != nil {
 		return nil, camperrors.Wrap(err, "writing note file")
 	}
@@ -222,6 +227,11 @@ func (s *IntentService) RestoreNote(ctx context.Context, id string) (*Intent, er
 	data, err := SerializeIntent(note)
 	if err != nil {
 		return nil, camperrors.Wrap(err, "serializing note")
+	}
+	if _, statErr := os.Stat(newPath); statErr == nil {
+		return nil, camperrors.Wrap(ErrFileExists, newPath)
+	} else if !os.IsNotExist(statErr) {
+		return nil, camperrors.Wrap(statErr, "checking destination note file")
 	}
 	if err := fsutil.WriteFileAtomically(newPath, data, 0644); err != nil {
 		return nil, camperrors.Wrap(err, "writing note file")
--- a/internal/intent/tui/explorer/update.go
+++ b/internal/intent/tui/explorer/update.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/Obedience-Corp/camp/internal/intent"
 	"github.com/Obedience-Corp/camp/internal/intent/tui"
 	"github.com/Obedience-Corp/camp/internal/intent/tui/filterchip"
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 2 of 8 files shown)


internal/intent/tui/explorer/update.go
```diff
@@ -336,10 +337,19 @@ func (m Model) handleActionMenuSelection(msg tui.ActionMenuSelectedMsg) (tea.Mod
...
+			}
+			m.statusMessage = "Archiving note..."
+			return m, m.archiveNote(selected)
 		}
 		return m.handleArchiveAction()
+	case "restore":
+		if selected.Status == intent.StatusNoteArchived {
+			m.statusMessage = "Restoring note..."
+			return m, m.restoreNote(selected)
+		}
```
obey-agent, line 352: The \`restore\` case is missing an explicit return when the selected note is not archived. Execution falls through into the \`delete\` branch — for lifecycle intents that could surface the delete confirmation dialog on an unrelated action. Mirror the \`archive\` branch: return early with a status message (e.g. \"Note is not archived\") when \`selected.Status != intent.StatusNoteArchived\`, even though the menu disables this path today.

obey-agent (commented):
## Verdict: Approve

## Overview
Focused, well-scoped fix that wires note archive/restore through the existing `IntentService` path and mirrors the established `ArchiveNote` pattern with a symmetric `RestoreNote`. TUI integration is consistent with other note actions (audit events, auto-commits, action-menu enablement), and test coverage spans service, menu, and explorer dispatch paths.

## Key Findings
- **Switch fall-through in `restore` handler** (`internal/intent/tui/explorer/update.go`): missing early return when restore is selected on a non-archived item. Low practical risk today because the menu disables the path, but the fall-through into `delete` is a latent correctness bug. One-line fix before merge would be ideal.

## What's Done Well
- `RestoreNote` is a clean mirror of `ArchiveNote` — same error wrapping, atomic writes, rollback on remove failure, index invalidation.
- Reason-free note archiving is the right model choice and is documented in code.
- Tests exercise real `handleActionMenuSelection` dispatch, not just helpers.
- Reuses `archiveFinishedMsg` / `moveFinishedMsg` patterns already understood by the explorer reload logic.

## Staff Standard
Yes — this is production-ready with one small defensive-coding fix. I would be comfortable shipping this after adding the explicit return in the restore case.

internal/intent/tui/explorer/update.go
```diff
@@ -336,10 +337,19 @@ func (m Model) handleActionMenuSelection(msg tui.ActionMenuSelectedMsg) (tea.Mod
...
-			return m, nil
+			if selected.Status != intent.StatusNote {
+				m.statusMessage = "Note is already archived"
+				return m, nil
+			}
+			m.statusMessage = "Archiving note..."
+			return m, m.archiveNote(selected)
 		}
 		return m.handleArchiveAction()
+	case "restore":
```
obey-agent, line 348: **Fall-through into delete (correctness).** When `restore` is selected but the note is not `StatusNoteArchived`, this case has no `return` / no status message and control falls through into `case "delete"`. Today the action menu usually disables restore for non-archived notes, so the path is latent — but Go switch fall-through is a real footgun: any menu enablement bug or future keybinding that fires `restore` on a live note will open the Delete Intent confirmation for that note.

Add an explicit else branch (status message + `return m, nil`) before `case "delete"`, matching the archive case's defensive style. Cover with a unit test that dispatches restore on a non-archived note and asserts delete is not entered.

obey-agent (commented):

## Overview
Re-review under the stricter gate (no Approve with open correctness issues). The feature is still well-scoped — `RestoreNote` mirrors `ArchiveNote`, audit/auto-commit wiring is consistent — but the switch fall-through in the explorer action handler is a latent destructive path and must not ship.

## Key Findings
1. **`restore` falls through into `delete`** (`update.go`) when the selection is not `StatusNoteArchived` — missing early return. Menu disablement is not a sufficient guard for a delete confirmation path. Fix + unit test required.

## What's Done Well
- Service-layer `RestoreNote` is a clean mirror of `ArchiveNote` (atomic write, rollback, index invalidation).
- Reason-free note archiving is the right model and is documented.
- Tests exercise real `handleActionMenuSelection` dispatch for the happy path.

## Staff Standard
Not until the fall-through is closed. This is a one-line defensive fix plus a regression test.

Reviewed against: https://github.com/Obedience-Corp/camp/pull/386 @ `d4c395cd0dafd26cb1f4ccf7f63780e6f8577792`

obey-agent (commented):
## Verdict: Merge-ready (supersedes the prior Request Changes)

## What changed
Re-review at `4d243cd`. The prior Request Changes (against `d4c395c`) flagged the `restore` case as "falling through into `delete`" for non-archived notes. That finding does not hold: Go `switch` statements break implicitly and only continue with an explicit `fallthrough` keyword, which this handler never had. On a non-archived note the `restore` case was already a no-op and never reached the delete confirmation.

## State now
- The author pinned the no-op with `TestRestore_TUIFlow_NonArchivedNoOp`, asserting `cmd == nil`, `focus != focusConfirm`, and `pendingAction != "delete"`.
- `4d243cd` makes the no-op explicit: an early `return m, nil` with a `"Note is not archived"` status message, mirroring the `archive` branch. This drops any dependence on menu enablement for safety and self-documents the intent. The test now also asserts the surfaced status message.

## Assessment
The single blocking finding rested on incorrect Go switch semantics and no longer stands; the residual silent-no-op UX nit is closed. `RestoreNote` remains a clean mirror of `ArchiveNote` (atomic write, rollback on remove failure, index invalidation). No open correctness issues.

Leaving the merge to Lance.

Reviewed against: https://github.com/Obedience-Corp/camp/pull/386 @ `4d243cd`

internal/intent/notes.go
```diff
@@ -194,6 +194,49 @@ func (s *IntentService) ArchiveNote(ctx context.Context, id string) (*Intent, er
...
+	newPath := s.getIntentPath(StatusNote, note.ID)
+	if err := os.MkdirAll(filepath.Dir(newPath), 0755); err != nil {
+		return nil, camperrors.Wrap(err, "creating directory")
+	}
+
+	data, err := SerializeIntent(note)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "serializing note")
+	}
+	if err := fsutil.WriteFileAtomically(newPath, data, 0644); err != nil {
```
obey-agent, line 236: Mirror ArchiveNote: if Serialize+Write succeeds but Remove(oldPath) fails, you correctly delete the new file. One small gap: if an active note file already exists at newPath (corrupt double-presence), WriteFileAtomically may overwrite without checking. Unlikely under normal IDs, but a Stat(newPath) guard with a clear error would make RestoreNote safer to run twice under races.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Small, focused fix that closes a real UX hole: notes in the intent explorer can finally be archived and restored. Wires through existing `ArchiveNote`, adds symmetric `RestoreNote`, and correctly updates action menu + `a` key without inventing a parallel lifecycle model.

## Key Findings
- **Optional guard on restore target path** (inline on `RestoreNote`): extremely unlikely ID collision, but a pre-write existence check would make double-presence races clearer.
- Reason-free note archive is a deliberate product choice and is documented; consistent with lightweight notes.
- Dispatch tests cover archive, restore, and non-archived restore no-op (including no fallthrough into delete)—excellent.

## What's Done Well
- Symmetric service API with context cancellation and atomic write + rollback on remove failure (mirrors ArchiveNote).
- Action menu enablement: active → Archive; archived → Restore; convert disabled when archived.
- Audit + auto-commit hooks aligned with existing intent patterns.
- Status messaging instead of silent no-ops.

## Staff Standard
Yes. Clean, well-tested, and appropriately small. Ship it.

PR #387, opened 2026-07-10
feat: first-class orgs (registry, CLI, create/init --org, TUI)

lancekrogers:
## Summary

Make orgs a first-class, machine-wide concept in camp (festival CF0002).

- **Registry v3:** persist `orgs` on both `config.Registry` and `registryfile.File`, with load-time reconcile/v2 backfill and fallback synthesis (not written to disk).
- **CLI:** `camp org create --empty`, `camp org delete` (empty-only / `--force` reassign), empty-aware `list` / `show` / `rename`.
- **Create/init:** `--org` on `camp create` and `camp init` via atomic `RegisterWithOrg` + `EnsureOrg`.
- **TUI:** `n` create empty org, `x`/`D` confirm-delete empty org, `N` new campaign into focused org (async create seam).

## Test plan

- [x] `just build-camp` green
- [x] Unit tests for registry round-trip, v2 backfill, create/delete, list/show/rename, create `--org` validation
- [x] Org TUI model tests for empty create, delete guards/confirm, new-campaign seam
- [ ] CI green on this PR
- [ ] Manual smoke: `camp org create foo --empty` → `list`/`show`/`rename`/`delete`
- [ ] Manual smoke: `camp create demo --org obey` (or dry-run) assigns org
- [ ] Optional: TUI theme screenshots for experience bar (PR attach)

Festival: `camp-first-class-orgs-CF0002` (`projects/worktrees/camp/first-class-orgs`)

### DIFF
--- a/cmd/camp/create.go
+++ b/cmd/camp/create.go
@@ -88,10 +88,6 @@ func runCreate(cmd *cobra.Command, args []string) error {
 		displayName = name
 	}
 
-	if dryRun && org != "" {
-		_, _ = fmt.Fprintf(w.HumanOut, "would assign org: %s\n", org)
-	}
-
 	p := initcmd.Params{
 		Dir:         target,
 		Name:        displayName,
--- a/cmd/camp/create.go
+++ b/cmd/camp/create.go
@@ -44,12 +44,19 @@ func init() {
 	createCmd.Flags().Bool("no-skills", false, "Skip linking campaign skills into .claude/skills and .agents/skills")
 	createCmd.Flags().Bool("dry-run", false, "Show what would be done without creating anything")
 	createCmd.Flags().String("path", "", "Override the base campaigns directory (campaign created at <path>/<name>/)")
+	createCmd.Flags().String("org", "", "Assign the new campaign to this org (created if new; defaults to the fallback org)")
 }
 
 func runCreate(cmd *cobra.Command, args []string) error {
 	name := args[0]
 	if err := validateCampaignName(name); err != nil {
 		return err
+	}
+	org := cmdutil.GetFlagString(cmd, "org")
+	if org != "" {
+		if err := config.ValidateName("org", org); err != nil {
+			return err
+		}
 	}
 	ctx := cmd.Context()
 	dryRun := cmdutil.GetFlagBool(cmd, "dry-run")
@@ -81,6 +88,10 @@ func runCreate(cmd *cobra.Command, args []string) error {
 		displayName = name
 	}
 
+	if dryRun && org != "" {
+		_, _ = fmt.Fprintf(w.HumanOut, "would assign org: %s\n", org)
+	}
+
 	p := initcmd.Params{
 		Dir:         target,
 		Name:        displayName,
@@ -90,6 +101,7 @@ func runCreate(cmd *cobra.Command, args []string) error {
 		NoGit:       cmdutil.GetFlagBool(cmd, "no-git"),
 		NoSkills:    cmdutil.GetFlagBool(cmd, "no-skills"),
 		DryRun:      dryRun,
+		Org:         org,
 		// force, noRegister, repair, yes stay zero — create deliberately does not support them.
 	}
 	return initcmd.RunFlow(ctx, p, w, tui.IsTerminal())
--- a/cmd/camp/org/create.go
+++ b/cmd/camp/org/create.go
@@ -74,27 +74,29 @@ type orgCreateEmptyResult struct {
 
 func createEmptyOrg(cmd *cobra.Command, org string, asJSON bool) error {
 	created := false
+	members := 0
 	err := config.UpdateRegistry(cmd.Context(), func(reg *config.Registry) error {
 		if !orgExists(reg, org) {
 			created = true
 		}
 		ensureOrg(reg, org)
+		members = len(membersOf(reg, org))
 		return nil
 	})
 	if err != nil {
 		return err
 	}
-	result := orgCreateEmptyResult{Org: org, Created: created, Members: 0}
+	result := orgCreateEmptyResult{Org: org, Created: created, Members: members}
 	if asJSON {
 		enc := json.NewEncoder(cmd.OutOrStdout())
 		enc.SetIndent("", "  ")
 		return enc.Encode(result)
 	}
 	if created {
-		_, err := fmt.Fprintf(cmd.OutOrStdout(), "created org %q (0 members)\n", org)
+		_, err := fmt.Fprintf(cmd.OutOrStdout(), "created org %q (%d members)\n", org, members)
 		return err
 	}
-	_, err = fmt.Fprintf(cmd.OutOrStdout(), "org %q already exists (0 members added)\n", org)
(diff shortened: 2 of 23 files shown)


cmd/camp/org/tui_create.go
```diff
@@ -0,0 +1,60 @@
...
+		return camperrors.Wrap(err, "loading global config")
+	}
+	base, err := cfg.ResolvedCampaignsDir(ctx)
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(base, 0o755); err != nil {
+		return camperrors.Wrapf(err, "ensure campaigns dir %s", base)
+	}
+	target := filepath.Join(base, name)
```
obey-agent, line 35: This path is built directly from the TUI text input, but this code path bypasses the `camp create` `validateCampaignName` guard. A value like `../outside` or `nested/name` will escape or nest under the configured campaigns directory before `RunFlow` initializes it. Please apply the same single-segment campaign-name validation used by `runCreate` before joining with `base`, and add a test that exercises the real `defaultCreateCampaignInOrg` path rather than only the stubbed seam.

obey-agent (changes requested):
**Verdict: Request Changes**

This is a strong first-class org implementation overall: the registry v3 migration path is explicit, empty-org CLI behavior is well covered, and create/init registration uses a single atomic registry update. I found one path-validation gap in the new TUI campaign creation action that should be fixed before merge.

**Key Findings**
- `cmd/camp/org/tui_create.go`: `defaultCreateCampaignInOrg` builds the target path from raw TUI input without the `camp create` single-segment name validation, so `../outside` or `nested/name` can write outside or below the configured campaigns directory. See the inline comment.

**What’s Done Well**
- Registry reconciliation/backfill keeps v2 files compatible while preserving empty orgs.
- The CLI create/delete/list/show/rename behavior has good focused unit coverage, including JSON shapes and empty-org semantics.
- The create/init `--org` path uses the shared `RegisterWithOrg` mutation instead of a separate write.

**Staff Standard**
I would be comfortable with the registry and CLI pieces, but I would not submit the TUI `N` action without reusing the same campaign-name validation and adding a real-path test for it.

cmd/camp/create.go
```diff
@@ -81,6 +88,10 @@ func runCreate(cmd *cobra.Command, args []string) error {
 		displayName = name
 	}
 
+	if dryRun && org != "" {
+		_, _ = fmt.Fprintf(w.HumanOut, "would assign org: %s\n", org)
```
obey-agent: **[suggestion]** Dry-run prints org assignment twice for `camp create --org X --dry-run`.

`runCreate` prints `would assign org:` here, then `initcmd.RunFlow` prints the same line again when `p.DryRun && p.Org != ""` (init/command.go). Drop this block and let `RunFlow` own the dry-run messaging so CLI output stays single-source.

cmd/camp/org/tui_create.go
```diff
@@ -0,0 +1,81 @@
...
+			return camperrors.New(fmt.Sprintf("target %s exists and is not empty", target))
+		}
+	}
+	p := initcmd.Params{
+		Dir:         target,
+		Name:        name,
+		TypeStr:     string(config.CampaignTypeProduct),
+		Description: "Created from camp org browser",
+		Mission:     "Created from camp org browser",
+		NoGit:       true,
```
obey-agent: **[suggestion]** TUI "N new campaign" hardcodes `NoGit: true`, `NoSkills: true`, product type, and placeholder Description/Mission.

That diverges from `camp create` defaults (git + skills) and will surprise operators who create a campaign from the org browser and find no repo / no skills projection. Prefer matching `camp create` defaults, or surface these choices in the overlay (even as fixed defaults documented in the footer help). At minimum, call out the intentional divergence in the help text / status line so the TUI path is not a silent subset of create.

cmd/camp/org/create.go
```diff
@@ -52,6 +66,38 @@ func runOrgCreate(cmd *cobra.Command, args []string) error {
...
+		if !orgExists(reg, org) {
+			created = true
+		}
+		ensureOrg(reg, org)
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	result := orgCreateEmptyResult{Org: org, Created: created, Members: 0}
```
obey-agent: **[suggestion]** `Members` is always forced to `0` here, even when the org already exists and has campaigns.

For machine consumers, `members: 0` reads as "current membership count", not "members added by this call". On the already-exists path, either report `len(membersOf(reg, org))` or rename the field (e.g. `members_added`) so the JSON contract is not misleading.

cmd/camp/org/membership.go
```diff
@@ -105,6 +106,52 @@ func reassignOrg(cmd *cobra.Command, target func(*config.Registry) string, campa
...
+			reg.Orgs[i].Name = newName
+			return
+		}
+	}
+}
+
+// membersOf returns campaigns whose Org equals name.
+func membersOf(reg *config.Registry, name string) []config.RegisteredCampaign {
+	var members []config.RegisteredCampaign
+	for _, c := range reg.Campaigns {
```
obey-agent, line 149: **[nit]** `membersOf` ranges `reg.Campaigns` (a map), so `--force` delete `reassigned` order is non-deterministic.

Not a correctness bug, but JSON/`--json` snapshots and operator logs will shuffle across runs. Sort by campaign name or ID before returning (or when building `result.Reassigned`) for stable CLI output.

obey-agent (commented):
## Verdict: Comment (approve with small cleanups)

## Overview
Solid first-class org work. Registry v3 (`Orgs` + load-time reconcile, fallback synthesis not persisted), CLI (`create --empty`, `delete`/`--force`), `--org` on create/init via `RegisterWithOrg`, and TUI overlays for empty create / confirm-delete / new campaign form a coherent vertical slice. Tests cover round-trip, backfill, delete guards, list semantics, and TUI seams well.

## Key Findings
- **Dry-run double-print**: `camp create --org X --dry-run` prints "would assign org" in both `create.go` and `init.RunFlow` — drop the create-side print.
- **TUI create is a silent subset of `camp create`**: hardcodes `NoGit`/`NoSkills`/placeholder mission. Either match create defaults or document the divergence in the TUI help/status.
- **JSON `members: 0` always**: `createEmptyOrg` reports membership zero even for already-populated orgs — rename or report actual count.
- **Nit**: `membersOf` map iteration makes `--force` `reassigned` order unstable.

None of these block the architecture. CI shows no checks on the branch — confirm gates green before merge.

## What's Done Well
- Clear migration story (v2 → v3 via reconcile; no heavy migration).
- Fallback org cannot be deleted; empty delete vs `--force` reassign is the right safety model.
- Rename now keys off first-class org existence (empty orgs rename correctly; no implicit merge).
- TUI async create seam is testable without real filesystem mutation.
- Errors use `camperrors`; mutations go through `UpdateRegistry` (locked load-mutate-save).

## Staff Standard
Yes — I would be proud to submit this as my own after addressing the dry-run duplication and deciding intentionally on TUI create defaults. The core model is production-grade.

PR #388, opened 2026-07-10
MM0001 follow-ups: identity ~ expansion + camp machine diagnose (stale ControlMaster)

lancekrogers:
## Summary

Follow-ups from the MM0001 review on #384 (multi-machine campaign navigation), which merged with two non-blocking inline notes on `internal/remote/ssh.go`. Both are addressed here.

### 1. Tilde expansion for `-i` identity files
`authArgs` now resolves a leading `~`/`~/` in a machine's `identity_file` to the current user's home before handing it to `ssh -i`.

OpenSSH's `IdentityFile` config directive expands `~`, but whether `ssh -i` expands it is client- and platform-dependent (OpenSSH 10.2 here does; not guaranteed elsewhere). Expanding in camp makes `camp machine add --identity '~/.ssh/id'` behave the way users expect from `ssh_config` regardless of the ssh build, and keeps the path camp hands off unambiguous. `~otheruser/` and mid-path `~` are left untouched for ssh to resolve. This is hardening, not a fix for a guaranteed break.

### 2. `camp machine diagnose` for stale ControlMaster sockets
The review flagged that a stuck ControlMaster socket (left behind by a sleep or network flap) can hang the next `camp switch machine:…` / `camp list --remote` hop until `ControlPersist` expires. New subcommand:

```
camp machine diagnose [id]          # report socket state: none | live | stale
camp machine diagnose [id] --reset  # tear down STALE sockets (live/absent left alone)
camp machine diagnose --json        # scriptable output
```

`live`/`stale` are distinguished with `ssh -O check`; `--reset` runs `ssh -O exit` then removes the socket file so a truly stuck socket cannot survive. Read path is agent-allowed; `--reset` is not (mirrors `camp doctor --fix`).

`camp doctor` was intentionally not used: it is campaign/submodule-scoped and requires a campaign root, whereas machines are a global `~/.obey/machines.yaml` concern — `camp machine` is the right home.

## Tests
- **Unit** (`internal/remote/ssh_test.go`): tilde expansion table (incl. `~user`/mid-path untouched), `authArgs` end-to-end expansion, side-effect-free `ControlSocketPath`, and the no-socket `none`/no-op reset paths.
- **Container integration** (`tests/integration/machine_diagnose_test.go`): real loopback ssh lifecycle — warm→`live`, `--reset` preserves a live master, SIGKILL the master→`stale`, `--reset` clears→`none` and removes the socket file.
- `just gate-fast` green (3064/3064 dev tests, lint clean); full multi-machine container integration suite passes.

## Docs
CLI reference regenerated for the `camp machine` family (add/list/remove/diagnose) and `list --remote` — surface that #384 introduced but did not regenerate. Kept scoped; the combined `camp-reference.md` full sweep (which also picks up unrelated `fresh`/`festivals` doc debt on main) is left for a dedicated docs pass.

### DIFF
--- a/cmd/camp/machine.go
+++ b/cmd/camp/machine.go
@@ -11,6 +11,7 @@ import (
 	"github.com/Obedience-Corp/camp/internal/config"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 	"github.com/Obedience-Corp/camp/internal/machines"
+	"github.com/Obedience-Corp/camp/internal/remote"
 	"github.com/Obedience-Corp/camp/internal/ui"
 )
 
@@ -58,11 +59,17 @@ var machineCmd = &cobra.Command{
 and 'camp list --remote'.
 
 Machines are stored in ~/.obey/machines.yaml. The current machine is always
-implicitly available as "local" and is never written to that file.`,
+implicitly available as "local" and is never written to that file.
+
+'camp machine diagnose' inspects the per-machine ssh ControlMaster sockets and
+can clear a stale one (the state a sleep or network flap can leave behind, which
+would otherwise hang the next hop until ControlPersist expires).`,
 	Example: `  camp machine list
   camp machine add devbox --host devbox.tailnet.ts.net --auth tailscale-ssh
   camp machine add --discover
-  camp machine remove devbox`,
+  camp machine remove devbox
+  camp machine diagnose
+  camp machine diagnose devbox --reset`,
 }
 
 var machineListCmd = &cobra.Command{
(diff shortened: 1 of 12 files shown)


obey-agent (approved):
**Verdict: Approve**

This PR cleanly addresses the MM0001 follow-ups: identity paths are normalized before reaching `ssh -i`, and the new `camp machine diagnose` surface gives operators a bounded way to identify and clear stale ControlMaster sockets. The implementation is narrow, uses the existing machine and remote packages, and the CLI docs were regenerated for the new surface.

**Key Findings**
- No blocking findings.

**What’s Done Well**
- The ControlMaster probes are context-bounded and reuse the existing SSH option construction.
- `--reset` only acts after a stale diagnosis, leaving live masters alone.
- Tests cover tilde expansion, side-effect-free socket path inspection, no-socket behavior, and the real loopback stale-socket lifecycle.

**Staff Standard**
Yes. I would be comfortable submitting this change as my own work.

**Verification**
- `go test ./cmd/camp ./internal/remote` passed.
- `go test -tags integration ./tests/integration -run TestMachineDiagnoseControlMasterLifecycle -count=1` passed.

PR #389, opened 2026-07-10
feat(quest): workitem-aware checklist for quests

lancekrogers:
## Quest checklist — workitem-aware todos on a quest

Adds a thin, quest-owned checklist: the layer between a quest ("what am I working on") and its workitems. Each checklist item is an ordered unit of work that can optionally link a workitem, so one design/explore packet can back several slices and an initiative can be divided across agents/sessions **without** promoting everything into a festival.

This is the first of two quest PRs. Companion (separate): terminal-local quest activation (`camp quest use`/`clear`).

### Design source
Ported from the My_Tools feedback docs (grok-festival-feedback-2026-07-09):
- `feature-quest-activity-list.md`
- `quest-checklist-data-model.md`

Naming follows the doc's guidance — **checklist**, not `activity` (avoids collision with `camp activity`).

### What's included

**Storage — `internal/quest` (untagged logic layer)**
- `checklist.yaml` sibling file per quest (`schema_version: quest-checklist/v1alpha1`), kept separate from `quest.yaml` so frequent item churn never rewrites quest metadata.
- `qci_YYYYMMDD_<hex>` item ids generated with the collision-scan pattern from `docs/id-schemes.md`.
- Load/save (atomic), rank-based ordering with tie-break on created/id, selector resolution by exact id / unique suffix / substring.

**CLI — `cmd/camp/quest` (dev-tagged, matching the existing quest command tree)**
- `camp quest checklist <quest> [--open] [--json]`
- `camp quest item add|done|reopen|edit|rank|link-workitem|unlink-workitem` — all non-interactive with `--json`.
- Workitem links store the **stable id** (+ derived `WI-` ref) and resolve the path/stage at **read time**, so moving a workitem — including to the dungeon — re-resolves the row instead of breaking it; a truly deleted workitem surfaces as `missing: true`.

### Dev-gating
`camp quest` is `//go:build dev` today, so these commands are dev-tagged too and do not appear in the stable binary or the generated `docs/cli-reference` until quest graduates. No stable-surface or doc changes.

### Testing
- **Unit** (`internal/quest/checklist_test.go`, host-safe, no FS): status/parse, id format + collision, resolve/rank/sort, YAML round-trip.
- **Integration** (`tests/integration/quest_checklist_test.go`, `//go:build integration`): full CLI round-trip in the container harness — add/list/complete/reopen, `--open` filter, and workitem link with read-time resolution.
- `just gate-fast` green: stable+dev builds, vet (stable/dev/integration), lint (stable+dev, incl. no-fmt-errorf and no-host-fs-test ratchets), 3088/3088 unit tests.
- Manually driven end-to-end against a throwaway campaign (create → add → list table/JSON → done/reopen/rank/edit → link/unlink → missing-after-delete).

### Notes for review
- New agent-allowed commands are registered in `manifestAgentAllowedReasons`.
- Checklist mutations stage only `checklist.yaml` for auto-commit; `quest.yaml` is untouched. `--json` mode silences the auto-commit line to keep stdout clean.


cmd/camp/quest/checklist_render.go
```diff
@@ -0,0 +1,217 @@
...
+			workitemCell(ctx, root, item.Workitem),
+		})
+	}
+
+	t := table.New().
+		Border(lipgloss.HiddenBorder()).
+		Headers(headers...).
+		Rows(rows...).
+		StyleFunc(func(row, _ int) lipgloss.Style {
+			if row == table.HeaderRow {
```
obey-agent, line 240: **[suggestion]** Checklist JSON/table output hardcodes `os.Stdout` / `fmt.Printf` instead of `cmd.OutOrStdout()`.

Elsewhere quest commands (e.g. activation status) use cobra writers so tests and callers can capture output. Using `os.Stdout` here:
1. Bypasses cobra's output redirection
2. Makes unit-level CLI tests harder
3. Is inconsistent with `jsoncontract` + activation patterns in the same package

Prefer `cmd.OutOrStdout()` (thread the writer through the render helpers).

internal/quest/checklist_service.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	cl, err := LoadChecklist(ctx, ChecklistPathForQuest(q), q.ID)
+	if err != nil {
+		return nil, nil, err
+	}
+	return q, cl, nil
+}
+
+// loadForMutation resolves the quest and loads its checklist for a write.
+func (s *Service) loadForMutation(ctx context.Context, identifier string) (*Quest, *Checklist, string, error) {
```
obey-agent: **[suggestion]** Load-mutate-save for checklist has no exclusive lock.

`loadForMutation` → mutate in memory → `SaveChecklist` (atomic write only). Two concurrent `camp quest item …` processes (multi-agent / parallel sessions on the same quest) can both load the same YAML and last-writer-wins, dropping the other agent's item.

Registry mutations already use `fsutil.AcquireFileLock` around the full cycle. Consider the same pattern on `checklist.yaml` (or a sibling `.lock`) for the mutation path — especially since checklist is meant for dividing work across agents/sessions.

cmd/camp/quest/checklist_render.go
```diff
@@ -0,0 +1,217 @@
...
+		CompletedAt: item.CompletedAt,
+	}
+}
+
+func outputChecklistJSON(ctx context.Context, root string, q *quest.Quest, items []quest.ChecklistItem) error {
+	jsonItems := make([]checklistItemJSON, 0, len(items))
+	for _, item := range items {
+		jsonItems = append(jsonItems, checklistItemToJSON(ctx, root, item))
+	}
+	enc := json.NewEncoder(os.Stdout)
```
obey-agent: **[suggestion]** Checklist JSON output uses `os.Stdout` (and the table path uses `fmt.Printf`/`fmt.Println`) instead of `cmd.OutOrStdout()`.

Elsewhere quest commands (e.g. activation status) use cobra writers so tests and callers can capture output. Hardcoding `os.Stdout` bypasses cobra redirection and is inconsistent with `jsoncontract` + activation patterns in the same package. Prefer threading `cmd.OutOrStdout()` through the render helpers.

(Note: an earlier comment landed on an unrelated table StyleFunc line — this is the actual site.)

internal/quest/checklist_service.go
```diff
@@ -0,0 +1,257 @@
...
+		} else {
+			item.CompletedAt = nil
+		}
+	}
+	item.UpdatedAt = nowUTC()
+	return s.saveResult(ctx, q, cl, path, item)
+}
+
+// RankChecklistItem sets an item's explicit sort rank.
+func (s *Service) RankChecklistItem(ctx context.Context, identifier, selector string, rank int) (*ChecklistMutationResult, error) {
```
obey-agent, line 222: **[nit]** `RankChecklistItem` accepts any int, including negative ranks.

Not a correctness bug for sorting, but negative ranks sort before the auto-assigned `10, 20, …` sequence and can confuse operators. Either reject `rank < 0` with `ErrInvalidInput`, or document that negative ranks are intentional "pin to top" behavior.

internal/quest/checklist.go
```diff
@@ -0,0 +1,270 @@
...
+		return nil, ErrChecklistItemNotFound
+	}
+	for i := range cl.Items {
+		if cl.Items[i].ID == selector {
+			return &cl.Items[i], nil
+		}
+	}
+	var matches []*ChecklistItem
+	for i := range cl.Items {
+		if strings.HasSuffix(cl.Items[i].ID, selector) || strings.Contains(cl.Items[i].ID, selector) {
```
obey-agent: **[nit]** Substring resolution (`strings.Contains` on the full id) can match the date segment (`20260710`), which is rarely what a human intends and often ambiguous across items created the same day.

Exact + unique suffix already cover the common UX. Consider restricting fuzzy match to the hex suffix only (after the last `_`), or requiring a minimum selector length.

obey-agent (commented):
## Verdict: Comment (lean approve — address lock + output writers before multi-agent use)

## Overview
Well-scoped first quest-checklist slice. Sibling `checklist.yaml` (schema `quest-checklist/v1alpha1`), stable `qci_` ids, rank-ordered items, workitem links by stable id with read-time resolution, and a complete non-interactive CLI surface under the existing dev-tagged `camp quest` tree. Separation of storage (internal/quest) from CLI (cmd/camp/quest) and workitem coupling only at the CLI boundary is the right shape.

## Key Findings
- **Lost-update race on mutate**: load → mutate → atomic write has no file lock. Concurrent agents editing the same quest checklist can clobber each other. Mirror registry’s `AcquireFileLock` around the mutation cycle before this is multi-agent production path.
- **Output writers**: JSON/table use `os.Stdout` / `fmt.Printf` instead of `cmd.OutOrStdout()` — inconsistent with other quest commands and harder to test.
- **Nits**: unbounded/negative rank; substring resolve can match date segments of the id (prefer suffix-only fuzzy match).
- Manifest agent-allowed entries and selective auto-commit of only `checklist.yaml` (quiet under `--json`) are correct.

## What's Done Well
- Workitem link stores id/ref and resolves path/stage at read time — dungeon moves surface as `missing: true` instead of stale paths.
- ID generation follows the collision-scan pattern from id-schemes docs.
- Unit tests cover status/id/resolve/rank/YAML; integration covers add/list/done/reopen/open-filter and link/unlink.
- Dev-gated so stable binary and cli-reference stay untouched.
- `autoCommitQuestQuiet` keeps JSON stdout clean.

## Staff Standard
Yes, with the concurrency lock treated as the main follow-up before leaning on multi-agent checklist edits. The model and CLI design are clean and production-minded for a v1alpha1 surface.

obey-agent:
Addressed the review findings in `2b698d3`.

**Lost-update race (the multi-agent blocker)** — the `load → mutate → atomic write` cycle now runs under an exclusive file lock. Introduced `Service.mutateChecklist`, which acquires `checklist.yaml.lock` via `fsutil.AcquireFileLock` around the full transaction (load, mutate, save) and every mutating method (`add`/`status`/`edit`/`rank`/`link`/`unlink`) now routes through it. This mirrors the workitem-links `WithLock` pattern, so concurrent `camp quest item …` processes on the same quest serialize instead of clobbering each other. Read-only `Checklist` stays lock-free (atomic writes give it a consistent snapshot).

**Output writers** — `outputChecklistJSON`, `outputChecklistItemResultJSON`, and `outputChecklistTable` now take an `io.Writer` and the callers thread `cmd.OutOrStdout()`; the human line in `emitChecklistMutation` uses `fmt.Fprintln(cmd.OutOrStdout(), …)`. No more hardcoded `os.Stdout` / `fmt.Printf`, so output honors cobra redirection and is capturable in tests.

**Negative rank (nit)** — `RankChecklistItem` (and `item add --rank`) reject `rank < 0` with a new `ErrNegativeChecklistRank` (wraps `ErrInvalidInput`).

**Substring resolution (nit)** — fuzzy matching is now scoped to the hex suffix after the last `_`. Suffix matches still run against the full id; substring matches no longer reach the shared `YYYYMMDD` date segment, so a selector like `2026` resolves to nothing instead of ambiguously matching every same-day item.

**Tests added**
- `TestConcurrentChecklistItemAdd_NoLostUpdate` — 8 parallel adds on one quest must all persist (fails without the lock).
- `TestRankChecklistItem_RejectsNegative` — negative rank rejected, non-negative still succeeds.
- `TestChecklist_Resolve` — new case asserting the date segment does not fuzzy-match.

`just gate-fast` green (stable+dev builds, vet stable/dev/integration, lint both profiles incl. no-host-fs-test and no-fmt-errorf ratchets, 3090/3090 unit tests). Quest checklist integration suite (`TestQuestChecklist_*`) passing against the container harness.

cmd/camp/quest/checklist_render.go
```diff
@@ -0,0 +1,217 @@
...
+// source of truth, and the path/stage are looked up now so a workitem moving to
+// the dungeon surfaces as missing rather than a stale path.
+func resolveChecklistWorkitem(ctx context.Context, root string, wi *quest.ChecklistWorkitem) *checklistWorkitemJSON {
+	if wi == nil {
+		return nil
+	}
+	out := &checklistWorkitemJSON{ID: wi.ID, Ref: wi.Ref}
+	w, err := selector.Resolve(ctx, root, wi.ID, selector.ResolveOptions{})
+	if err != nil {
+		out.Missing = true
```
obey-agent, line 107: **Bug: any Resolve error becomes \`missing: true\`.**

\`selector.Resolve\` returns not only \`ErrSelectorNotFound\` but also campaign-config load failures, discovery FS errors, context cancellation, and ambiguous matches. Folding every error into \`Missing\` means a transient I/O or config problem will look identical to a deleted workitem in both the table (\`… (missing)\`) and the JSON contract agents will key off.

Please distinguish:
- \`errors.Is(err, selector.ErrSelectorNotFound)\` → \`missing: true\`
- other errors → return/surface the error (or at least a separate field like \`resolve_error\`)

Otherwise agents will treat "campaign config unreadable" as "workitem gone" and may recreate or re-link incorrectly.

cmd/camp/quest/checklist_render.go
```diff
@@ -0,0 +1,217 @@
...
+	Item          checklistItemJSON  `json:"item"`
+}
+
+func questJSON(q *quest.Quest) checklistQuestJSON {
+	return checklistQuestJSON{ID: q.ID, Name: q.Name, Status: string(q.Status)}
+}
+
+// resolveChecklistWorkitem performs the read-time join: the stored id is the
+// source of truth, and the path/stage are looked up now so a workitem moving to
+// the dungeon surfaces as missing rather than a stale path.
```
obey-agent: **Product/contract gap: dungeon and deleted are indistinguishable.**

This comment is accurate today — \`workitem.Discover\` skips dungeon directories, so a dungeoned link fails resolve and becomes \`missing\`. That conflicts with the PR body claim that dungeon moves "re-resolve the row" while only true deletes surface as missing.

If the intended UX is "dungeon still resolves with a dungeon path/stage", resolution needs an option that includes dungeon inventory (or a locate-by-id path). If the intended UX is "dungeon == missing", update the PR description and consider a richer signal (\`missing\` vs \`dungeoned\` / lifecycle) so operators can tell "archived packet" from "deleted packet". As written, agents cannot.

cmd/camp/quest/checklist_render.go
```diff
@@ -0,0 +1,217 @@
...
+		CreatedAt:   item.CreatedAt,
+		UpdatedAt:   item.UpdatedAt,
+		CompletedAt: item.CompletedAt,
+	}
+}
+
+func outputChecklistJSON(ctx context.Context, w io.Writer, root string, q *quest.Quest, items []quest.ChecklistItem) error {
+	jsonItems := make([]checklistItemJSON, 0, len(items))
+	for _, item := range items {
+		jsonItems = append(jsonItems, checklistItemToJSON(ctx, root, item))
```
obey-agent: **Perf: N full discoveries per checklist list.**

Each linked item calls \`resolveChecklistWorkitem\` → \`selector.Resolve\` → \`discover(ctx, root)\`, which reloads campaign config and walks the full workitem tree. A 20-item checklist with 15 links does ~15 full discoveries.

Cache one discovery (or a \`map[id]*WorkItem\`) per checklist render and resolve against that. Same correctness, much cheaper, and avoids amplifying the error-as-missing issue under partial FS failures mid-loop.

tests/integration/quest_checklist_test.go
```diff
@@ -0,0 +1,114 @@
...
+	linkOut, err := tc.RunCampInDir(path, "quest", "item", "link-workitem", "billing",
+		item.Item.ID, "billing-dunning", "--no-commit")
+	require.NoError(t, err, linkOut)
+
+	// Read model joins the workitem back in by id.
+	listOut, err := tc.RunCampInDir(path, "quest", "checklist", "billing", "--json")
+	require.NoError(t, err, listOut)
+	assert.Contains(t, listOut, "feature-billing-dunning")
+	assert.Contains(t, listOut, `"resolved_path"`)
+	assert.Contains(t, listOut, "workflow/feature/billing-dunning")
```
obey-agent, line 106: **Test gap: the PR claims missing-after-delete but does not lock it.**

This test covers link → resolve path → unlink. It never asserts \`missing: true\` after the linked workitem is deleted (or dungeoned). That is the load-bearing read-time-join property called out in the design.

Please add an integration case: create + link → delete/rm workitem dir (or promote to dungeon) → \`quest checklist --json\` shows the stored id/ref with \`"missing": true\` and no stale \`resolved_path\`. Without that, a future change that stops re-resolving will still pass CI.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Solid first cut of a quest-owned checklist: separate `checklist.yaml`, lock-guarded mutations, stable `qci_` ids, rank ordering, agent-friendly JSON, and a concurrent lost-update test. The architecture (service layer free of workitem coupling; resolve at read time) is the right shape. The review blockers are in the read-time join contract and its tests, not the storage model.

## Key Findings
- **Any `selector.Resolve` error is reported as `missing: true`** (inline on `checklist_render.go`). Config/FS/cancel failures will look like deleted workitems to agents. Classify `ErrSelectorNotFound` vs other errors.
- **Dungeon vs deleted conflated; PR body disagrees with code** (inline). Discover skips dungeon dirs, so dungeon links are missing — document or fix product behavior, and consider a richer signal than a boolean.
- **N full workitem discoveries per checklist list** (inline). Cache discovery per render.
- **Missing integration coverage for missing-after-delete** (inline), despite the PR claiming that path. Link/unlink alone does not lock the join contract.

## What's Done Well
- File lock around load→mutate→save with a real concurrent add test.
- Atomic writes, collision-scan id generation matching existing schemes, selector resolution that avoids date-segment ambiguity.
- Auto-commit only stages `checklist.yaml`; `--json` quiets commit noise.
- Clean separation: CLI resolves workitems, service stores only id/ref.
- Unit + container integration coverage for the happy path CRUD flow.

## Staff Standard
Would be proud of the storage/mutation design after the join contract is tightened. Not ready to ship the agent-facing `missing` signal as-is — that field will be trusted by automation and currently lies under non-not-found errors.

internal/quest/checklist_service.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if err := SaveChecklist(ctx, path, cl); err != nil {
+		return nil, err
+	}
+	return &ChecklistMutationResult{
+		Quest:     q,
+		Checklist: cl,
+		Item:      item,
```
obey-agent, line 104: **Bug (new): mutation result Item pointer is invalidated by SaveChecklist Sort.**

fn returns a pointer into cl.Items. SaveChecklist then calls cl.Sort(), which reorders the slice in place. After that, the returned Item pointer still addresses the same index slot which may now hold a different item.

Concrete failure mode: items=[A rank=10, B rank=20]; camp quest item rank Q A 25.
1. resolve A -> *Items[0]
2. set A.Rank=25
3. Sort -> [B,A] so Items[0] is now B
4. JSON/human emit uses res.Item -> reports B as ranked

Affects any mutation that changes rank order (explicit --rank / rank, and equal-rank reorders). Copy the item by value before save (or re-resolve by id after sort) and return that stable snapshot.

cmd/camp/quest/checklist_render.go
obey-agent, line 107: **Still open: any selector.Resolve error becomes missing: true.**

Resolve returns not only not-found but also config load failures, discovery FS errors, context cancellation, and ambiguous matches. Folding every error into Missing means a transient I/O/config problem is reported as "linked workitem gone," and agents will act on a false signal.

Classify at least:
- errors.Is(err, selector.ErrSelectorNotFound) -> missing: true
- other errors -> return/propagate (or a distinct resolve_error field) so the list path fails loudly instead of lying.

cmd/camp/quest/checklist_render.go
obey-agent: **Still open: dungeon and deleted remain indistinguishable — and this comment documents the wrong contract.**

workitem.Discover skips dungeon directories, so a dungeoned link fails resolve and is labeled missing. That conflicts with treating the stored id as a durable join key that should re-resolve after a dungeon move. Operators/agents cannot tell "archived to dungeon" from "hard-deleted / never existed."

Either include dungeon in the checklist join (and surface lifecycle/attention as dungeon/archived) or rename the field/docs to something like unresolved and stop claiming dungeon re-resolves.

cmd/camp/quest/checklist_render.go
```diff
@@ -0,0 +1,217 @@
...
+		Notes:       item.Notes,
+		CreatedAt:   item.CreatedAt,
+		UpdatedAt:   item.UpdatedAt,
+		CompletedAt: item.CompletedAt,
+	}
+}
+
+func outputChecklistJSON(ctx context.Context, w io.Writer, root string, q *quest.Quest, items []quest.ChecklistItem) error {
+	jsonItems := make([]checklistItemJSON, 0, len(items))
+	for _, item := range items {
```
obey-agent, line 139: **Still open: N full discoveries per checklist list.**

Each linked item calls resolveChecklistWorkitem -> selector.Resolve -> full config load + tree walk. A 20-item checklist with 15 links does ~15 full discoveries (and the table path can double that via workitemCell).

Cache one discovery (or a map[id]*WorkItem built once per render) and resolve against it. This is not optional for multi-agent checklist usage on real campaigns.

tests/integration/quest_checklist_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// Unlink drops the reference.
+	_, err = tc.RunCampInDir(path, "quest", "item", "unlink-workitem", "billing", item.Item.ID, "--no-commit")
+	require.NoError(t, err)
+	afterOut, err := tc.RunCampInDir(path, "quest", "checklist", "billing", "--json")
+	require.NoError(t, err, afterOut)
+	assert.NotContains(t, afterOut, "resolved_path", "unlinked item must have no workitem block")
```
obey-agent, line 124: **Still open: no missing-after-delete (or dungeon) integration coverage.**

This test covers link -> resolve path -> unlink only. It never asserts missing: true after the linked workitem is deleted or dungeoned — the load-bearing read-time-join property. Please add an integration case that:
1. links a real workitem
2. deletes (or dungeons) it
3. lists checklist --json and asserts the link is present with missing: true (and, if dungeon is distinguished, the correct stage)

obey-agent (changes requested):
## Verdict: Request Changes

## Overview
The follow-up commit (`2b698d3`) correctly lands the multi-agent mutation lock, OutOrStdout plumbing, negative-rank rejection, and hex-suffix selector scoping. Those earlier findings are fixed and tested. However the **read-time join contract** (the load-bearing product surface of this PR) is still wrong in four places, and a new correctness bug was introduced/exposed by the lock refactor: mutation result `Item` pointers are invalidated by `SaveChecklist` → `Sort`.

## Fixed since first pass
- Exclusive `checklist.yaml.lock` around load→mutate→save + concurrent lost-update test
- `cmd.OutOrStdout()` / `io.Writer` for checklist render
- Negative rank rejection (`ErrNegativeChecklistRank`)
- Fuzzy id match restricted to hex suffix (date segment no longer matches)

## Remaining blockers
1. **Item pointer after Sort (NEW)** — `mutateChecklist` returns `*ChecklistItem` into `cl.Items`; `SaveChecklist` sorts in place and can make that pointer alias a different item. Ranking/JSON emit can report the wrong item.
2. **Any Resolve error → `missing: true`** — config/FS/ctx/ambiguous errors are silently reclassified as missing.
3. **Dungeon vs deleted conflated** — Discover skips dungeon; comment still claims dungeon "surfaces as missing."
4. **N full discoveries per list** — one `selector.Resolve` (full walk) per linked item.
5. **Missing-after-delete integration test still absent** — only link→resolve→unlink is covered.

## Tests run (worktree)
```
go test -tags dev ./internal/quest/... ./cmd/camp/quest/... -count=1
```
PASS (`internal/quest`, `internal/quest/tui`, `cmd/camp/quest`).

## Required before merge
- Return a by-value item snapshot (or re-resolve by id after sort) from `mutateChecklist`
- Classify Resolve errors properly; do not map non-not-found errors to `missing`
- Decide and implement dungeon join semantics (or stop claiming them)
- Cache discovery once per checklist render
- Add integration coverage for missing-after-delete (and dungeon if distinguished)

internal/quest/checklist_service.go
obey-agent, line 104: **Still open:** `Item` is a pointer into `cl.Items`, then `SaveChecklist` calls `cl.Sort()` which reorders the slice by value. After sort, `Item` can point at a different checklist row than the one just mutated (especially when `--rank` places a new item mid-board, or status/rank updates reorder).

Re-resolve by id after save, or return a value copy:

```go
id := item.ID
if err := SaveChecklist(...); err != nil { ... }
resolved, _ := cl.Resolve(id) // or copy *item before Sort
return &ChecklistMutationResult{..., Item: resolved}, nil
```

This was the remaining deep-review blocker; the workitem-resolution follow-up looks solid.

obey-agent (changes requested):
## Re-review verdict: Request Changes

**HEAD:** `112fd595e193` · commits since last review: harden workitem resolution + worktree ignore · `go test -tags dev ./internal/quest/...` **PASS**

### Fixed since last CHANGES_REQUESTED
1. **N discoveries per row** → single `loadChecklistWorkitemIndex` per render
2. **Resolve error → missing** → discovery/config failures now return errors; missing means not in inventory
3. **missing-after-delete integration test** → added to `quest_checklist_test.go`
4. Dungeon/deleted as missing is now an explicit product contract (documented in comments)

### Still open (blocker)
1. **`mutateChecklist` returns `*ChecklistItem` into `cl.Items` before `SaveChecklist`→`Sort()`** — pointer can observe the wrong item after reorder (inline on `checklist_service.go`).

### Staff standard
Almost there. Fix the post-sort result pointer (value copy or re-resolve by id) and this is approve-ready.

obey-agent (approved):
## Re-review verdict: Approve

**HEAD:** `02f831f9490d` · fix(quest): preserve checklist mutation result identity · `go test -tags dev ./internal/quest/...` **PASS**

### Prior blocker — verified fixed
**`mutateChecklist` Item pointer after `SaveChecklist`→`Sort()`**  
- Captures `itemID` before save  
- Re-`Resolve(itemID)` after sort so the result never retains a stale slice slot  
- Regression test `TestRankChecklistItem_ReturnsAffectedItemAfterSort` forces a mid-board reorder (rank 30) and asserts result ID/rank match the affected item while checklist order is `[second, first]`

### Earlier findings (already fixed on prior heads)
- Mutation lock, OutOrStdout, negative ranks, hex-only fuzzy id  
- Single workitem discovery index per render; missing vs load errors  
- missing-after-delete integration coverage  

### Staff standard
The last correctness hole is closed with a targeted test. Approve.

obey-agent (approved):
## Verdict: Approve

## Overview
Re-review of head `6671b49`. Follow-up replaces `fmt.Errorf` with `camperrors.Wrap` in `loadChecklistWorkitemIndex` (campaign config load + workitem discover). Matches camp's error-wrapping convention on the checklist render path; no behavioral change.

## Key Findings
- Scope is two wrap calls in `checklist_render.go` only.
- Prior checklist correctness (Sort pointer preservation, locking, workitem resolution hardening) remains intact.

## What's Done Well
Small, targeted hygiene fix rather than a drive-by refactor.

## Staff Standard
Yes — still ship.

PR #390, opened 2026-07-10
feat(quest): terminal-local quest activation (use/clear/status)

lancekrogers:
## Terminal-local quest activation — `camp quest use` / `clear` / `status`

Makes a quest the working context for **one terminal** by setting `CAMP_QUEST`, so child processes (including camp itself — e.g. `camp workitem create` stamping the active quest) inherit it. It's terminal-local by design: a campaign-global "active quest" file would make every terminal in the same workspace overwrite each other.

Companion to #389 (quest checklist). Independent branch; the only overlap is one line in `zz_manifest_annotations.go` (trivial merge).

### Design source
`workflow/explore/camp-terminal-local-quest-context` (README / UX_WORKFLOW / RELIABILITY). This is the explore's **constrained first version**: explicit activation only — no prompt-hook autoloading, no PATH manipulation yet.

### What's included

**`camp quest use <quest> [--shell posix|fish] [--allow-dungeon]`**
Resolves the quest, then emits shell code on **stdout** — `export CAMP_QUEST='<id>'` (posix) / `set -gx CAMP_QUEST '<id>'` (fish) — with the confirmation on **stderr**. Validation happens before any output, so stdout stays empty on failure and the wrapper never evals a partial result. Dungeoned quests are rejected unless `--allow-dungeon`.

**`camp quest clear [--shell ...]`** — emits `unset` / `set -e` for `CAMP_QUEST`.

**`camp quest status [--json]`** — reports the three states from the design: active+valid (name/id/source/root), none, and invalid (`CAMP_QUEST` set but unresolvable). Read-only, agent-allowed.

**Shell wrapper** — `camp shell-init` (bash/zsh/fish) now intercepts `quest use`/`quest clear` and evals the emitted line; every other quest subcommand passes through unchanged. `--help` passes through too.

**`internal/quest`** — `QuestEnvVar` const (now also used by the existing `ResolveContext`) + an injection-safe shell-quoting renderer.

### Dev-gating
`camp quest` is `//go:build dev`; the new commands are dev-tagged too, and the shell-template additions are inert in the stable binary (the stable `camp` has no `quest` subcommand to intercept). No stable-surface or `docs/cli-reference` changes.

### Testing
- **Unit** (`internal/quest/shellenv_test.go`, pure): dialect parsing, render output for posix/fish, and an **injection-safety** case (a hostile value stays a single quoted literal).
- **Integration** (`tests/integration/quest_activate_test.go`, `//go:build integration`): `use` emits the right dialect line; `status` reflects env (none / valid `--json` / invalid).
- **Live wrapper check**: sourced `camp shell-init` in real bash and zsh — `camp quest use billing` sets `CAMP_QUEST` in the shell, `clear` unsets it, `quest list` still passes through.
- `just gate-fast` green: stable+dev builds, vet (stable/dev/integration), lint (stable+dev incl. ratchets), 3084/3084 unit tests. Existing wrapper-transparency tests unaffected (the `quest)` case sits outside the go/switch section they inspect).

### Deliberately out of scope (v1)
Prompt-hook autoloading and PATH manipulation — the explore flags both as later work once explicit activation proves useful.


cmd/camp/quest/activate.go
```diff
@@ -0,0 +1,190 @@
...
+	return nil
+}
+
+func printQuestStatusHuman(w io.Writer, qctx *questCommandContext, p questStatusPayload) {
+	switch {
+	case !p.Active:
+		fmt.Fprintln(w, "No terminal quest active.")
+		fmt.Fprintln(w, "\nUse:")
+		fmt.Fprintln(w, "  camp quest use <quest>")
+	case !p.Valid:
```
obey-agent, line 175: **[nit]** `qctx` is unused in `printQuestStatusHuman` (payload already carries everything).

Drop the parameter or use it (e.g. relative path helpers) to avoid dead surface and keep the signature honest.

internal/shell/templates/bash.sh.tmpl
```diff
@@ -191,6 +191,30 @@ camp() {
         rm -f "$tmp"
       fi
       ;;
+    quest)
```
obey-agent, line 194: **[suggestion]** Unit + integration cover emit/status, but not the shell-init intercept path (`case quest` / `use|clear` → eval).

The dangerous surface is exactly that wrapper: wrong dialect flag, eval of multi-line output, or swallowing exit codes. You already have `shell_init_test.go` harnesses for wrapper transparency. A small integration case that **sources** `camp shell-init` and runs `camp quest use` / `clear` in real bash (and ideally fish) would lock the v1 contract the PR claims was manually verified.

internal/shell/templates/bash.sh.tmpl
```diff
@@ -191,6 +191,30 @@ camp() {
...
+            case "$arg" in
+              --help|-h)
+                command camp "$@"
+                return
+                ;;
+            esac
+          done
+          # camp emits shell code on stdout only after validation succeeds and
+          # diagnostics on stderr, so eval only a successful non-empty result.
+          line=$(command camp "$@" --shell bash) || return "$?"
```
obey-agent, line 210: **[nit]** Wrapper always appends `--shell bash` after `"$@"`.

If a user (or script) already passes `--shell fish` while somehow going through the bash wrapper, cobra last-flag-wins makes bash win — correct for *this* shell, but surprising when debugging dialect flags. Low risk. Optional: document that the wrapper always forces the host shell dialect.

obey-agent (commented):
## Verdict: Comment (lean approve — add wrapper integration test if easy)

## Overview
Clean constrained v1 of terminal-local quest activation: `CAMP_QUEST` env (not a campaign-global active file), stdout shell code after validation, stderr human confirmation, dungeon guard with `--allow-dungeon`, and shell-init intercept for bash/zsh/fish. Correctly independent of #389 aside from a one-line manifest addition.

## Key Findings
- **Missing automated wrapper test**: emit/status are covered; the eval intercept in `shell-init` templates is the highest-risk path and only manually verified. A sourced-wrapper integration case in the existing shell_init harness would lock the contract.
- **Nit**: wrapper always appends `--shell <host>` after `"$@"` (last-flag-wins) — correct for the host shell, worth a one-line doc note.
- **Nit**: unused `qctx` param on `printQuestStatusHuman`.
- Injection-safe quoting (posix + fish) with an explicit hostile-value unit test — good.
- Validation-before-emit so failed `use` leaves stdout empty for the wrapper — correct safety model.
- `quest status` three-state model (none / valid / invalid) with agent-allowed JSON is the right read surface; use/clear stay non-agent (env mutation) — correct.

## What's Done Well
- Terminal-local design matches the explore constraint and avoids multi-terminal thrash.
- `QuestEnvVar` const shared with `ResolveContext` (no magic string drift).
- Fish/bash/zsh templates stay consistent and pass `--help` through.
- Integration covers dialect emit + status env reflection.

## Staff Standard
Yes. Small, intentional, and safe for the stated v1 (no prompt hooks / PATH). Wrapper integration test would bring this from “manually green” to “CI-locked.”

tests/integration/quest_activate_test.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func TestQuestUse_EmitsShellCodeForDialect(t *testing.T) {
+	path := "/campaigns/quest-activate-use"
+	tc, id := setupActivationQuest(t, path)
+
+	posix, err := tc.RunCampInDir(path, "quest", "use", "billing")
+	require.NoError(t, err, posix)
+	assert.Contains(t, posix, "export CAMP_QUEST='"+id+"'", "posix export line")
+	assert.Contains(t, posix, "Activated quest for this terminal", "human confirmation")
```
obey-agent, line 36: **Test does not lock the stdout/stderr contract this feature depends on.**

Production correctness of `eval "$(camp quest use …)"` and the shell wrapper requires shell code on **stdout only** and the confirmation on **stderr**. `RunCampInDir` merges streams (`2>&1`), so asserting both the export line and `"Activated quest…"` in the same string would still pass if confirmation were accidentally moved to stdout — which would break `eval`.

Prefer `ExecCommand` (or a helper that captures streams separately) and assert:
- stdout is exactly one shell line (export/set)
- stderr contains the confirmation
- failure paths leave stdout empty

That locks the eval-safety property the design document is built on.

cmd/camp/quest/activate.go
```diff
@@ -0,0 +1,190 @@
...
+	case raw == "":
+		payload.Valid = true
+	default:
+		payload.Active = true
+		payload.Source = quest.QuestEnvVar
+		payload.RawEnv = raw
+		q, findErr := qctx.service.Find(ctx, raw)
+		if findErr != nil {
+			payload.Valid = false
+			payload.Reason = "quest not found"
```
obey-agent, line 151: **Any Find failure becomes "quest not found".**

Same pattern as the checklist missing-join: `Find` can fail for reasons other than absence (corrupt quest.yaml, FS errors). Surfacing those as `valid: false` / `reason: "quest not found"` misleads both humans and agents consuming `--json`.

Prefer:
- not-found → current invalid payload
- other errors → return the error from `runQuestStatus` (fail the command)

Status is agent-allowed; lying about why resolution failed will produce bad recovery behavior (`quest clear` when the real problem is a corrupt on-disk quest).

obey-agent (commented):
## Verdict: Comment

## Overview
Clean constrained v1 of terminal-local quest activation: emit shell code on stdout after validation, confirm on stderr, wrap `use`/`clear` in shell-init, and expose a three-state `status`. Design matches the explore docs and correctly rejects campaign-global active-quest files.

## Key Findings
- **Integration tests merge stdout/stderr**, so they do not lock the eval-safety contract (inline on `quest_activate_test.go`). Separate streams and assert empty stdout on failure.
- **`status` maps every Find error to "quest not found"** (inline on `activate.go`). Propagate non-not-found errors; status is agent-allowed.
- Minor gaps: no automated coverage for `--allow-dungeon` rejection/allow, and no wrapper-level integration test (manual bash/zsh check noted in the PR is good but not CI-guarded).

## What's Done Well
- Validation-before-emit so failed `use` never produces partial shell for `eval`.
- Injection-safe quoting tests for posix and fish.
- `QuestEnvVar` centralized and reused from ResolveContext.
- Shell wrappers intercept only `use`/`clear`, pass through `--help` and other quest subcommands.
- Dev-gated with the rest of `camp quest`; stable surface untouched.

## Staff Standard
Yes — I would be comfortable shipping this after the status error classification and a stream-separated integration assertion. No architectural redesign needed; the shape is right.

obey-agent (approved):
## Verdict: Approve

## Overview
Re-review of head `dbf38fd`. Terminal-local quest activation via `CAMP_QUEST` remains the right design (no campaign-global active-quest file). The follow-up hardens `quest status` so non-not-found Find errors propagate, and adds an integration test that locks stdout=shell / stderr=human / empty-stdout-on-failure — the exact eval contract the shell wrapper depends on.

## Key Findings
- Quoting helpers stay injection-safe (unit-tested).
- Dungeon guard + `--allow-dungeon` still correct.
- Status three-state (inactive / invalid / active) is clear for humans and JSON (`quest-status/v1alpha1`).
- `//go:build dev` on the command package is an intentional alpha gate; fine if that matches the rest of the quest surface rollout plan.

## What's Done Well
Stdout-only-after-validation is the right shell-eval contract. Follow-up tests address the previous review gap on stream separation.

## Staff Standard
Yes — ready to ship for the terminal-local activation path.

obey-agent (approved):
## Verdict: Approve

## Overview
Re-review of head `f14e3aa` (merge main into quest terminal activation). No feature regressions on the terminal-local quest activation surface: `activate.go` / `shellenv` / integration tests for use/clear/status are unchanged relative to previously approved `dbf38fd`. Delta is main integration (checklist from #389, org create/TUI, etc.).

## Key Findings
- Stream separation, dungeon guard, injection-safe quoting, and `//go:build dev` gate remain as previously approved.
- Merge does not rewrite the activation eval contract (stdout shell / stderr human / empty stdout on failure).

## What's Done Well
Clean restack onto current main before merge.

## Staff Standard
Yes — still ship.

PR #391, opened 2026-07-10
fix: eliminate terminal background query stall at startup

lancekrogers:
## Summary
- seed Lipgloss background detection from `COLORFGBG` before Bubble Tea initializes
- make explicit light/dark/high-contrast themes refine the cached background without terminal queries
- route adaptive Glamour selection through the seeded Lipgloss cache instead of calling termenv directly

This ports the startup portion of fest #254 to camp and closes the fresh campaign intent for mute PTYs used by recorders, CI, and agent sessions.

## Reproduction and verification
Controlled PTY that advertises `TERM=xterm-256color` but never answers OSC 11/DSR:

```text
Before: OSC 11 + DSR emitted; camp version output after 5.03s
After:  no terminal query bytes; output after 0.40s
```

## Gates
- focused `go test` for bginit, theme, intent TUI, and cmd/camp
- `just build dev`
- `just test unit`
- `just lint` (0 issues; no new host-filesystem or fmt.Errorf ratchet violations)
- `just gate`

### DIFF
--- a/cmd/camp/main.go
+++ b/cmd/camp/main.go
@@ -8,6 +8,9 @@ import (
 	"os/signal"
 	"syscall"
 
+	// bginit must initialize before the bubbletea subtree under the command
+	// packages; its path keeps it first under gofmt.
+	_ "github.com/Obedience-Corp/camp/internal/bginit"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 )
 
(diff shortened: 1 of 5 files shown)


obey-agent (approved):
Verdict: Approve

This is a faithful, camp-aware port of the proven fest startup fix. The seed package has a deliberately narrow dependency boundary, explicit themes refine the cached background, and the intent explorer no longer bypasses the cache with a direct termenv query.

Verification covered the exact failure mode and the broader repository contract:
- mute PTY with TERM=xterm-256color: 5.03s plus OSC/DSR before, 0.40s with no query bytes after
- focused packages passed
- full unit suite and dev build passed
- just lint: 0 issues and both ratchets clean
- just gate passed

No blocking findings. I would be comfortable submitting and operating this change.

PR #393, opened 2026-07-11
feat(workitem): record create/adopt events in workitem ledger

lancekrogers:
## Summary

camp-graph needs a reliable origin event for directory-backed workitems before later promote/gather movement events. The campaign-wide ledger `.campaign/workitems/.workitems.jsonl` previously recorded only `promote` and `gather`, leaving workitem creation and adoption invisible.

This PR adds two lifecycle-origin event types across every directory-backed workflow type (design, explore, bug, audit, and custom):

- **`create`** (`camp workitem create`): `id`, `ref`, `type`, `title`, and `to` (new path)
- **`adopt`** (`camp workitem adopt` / `camp workitem init`): the same shape, with a distinct event type because it brings an existing directory under management

The `Event` struct gains optional `ref` and `title` fields. Both use `omitempty`, so existing promote/gather lines remain backward compatible; a mixed old/new decode test locks that behavior in.

## Design notes

- Event identity uses the newly written `.workitem` metadata (`id` and `ref`).
- Ledger append failures are best-effort and warn on stderr without failing the completed create/adopt mutation, matching existing post-write bookkeeping behavior.
- Create/adopt deliberately do not stage or auto-commit the ledger. The next explicit campaign commit picks up the tracked ledger alongside the workitem mutation.
- Attention `stage` and `group` changes are intentionally excluded. They are not path movement, and their persistence/sharing policy should be resolved separately before they enter the shared ledger.
- No actor field is added because the existing workitem ledger schema has no actor-detection plumbing.

## Scope

Promote and gather already emit movement events and remain unchanged. Priority, stage, group, link, and unlink audit policy is separate follow-up scope.

## Verification

- [x] Targeted containerized tests: `go test ./internal/commands/workitem ./internal/workitem/audit`
- [x] Full compile: `go build ./...`
- [x] Vet: `go vet ./...`
- [x] `just lint` (0 issues; filesystem-test and `fmt.Errorf` ratchets clean)
- [x] Backward-compatible mixed old/new ledger decode test


internal/commands/workitem/stage.go
```diff
@@ -95,6 +98,25 @@ func runStage(ctx context.Context, cmd *cobra.Command, selectorArg, stageArg str
...
+	auditID := wi.StableID
+	if auditID == "" {
+		auditID = wi.Key
+	}
+	newStage := stage
+	if clear {
+		newStage = priority.AttentionNone
+	}
+	auditRef, _ := wi.SourceMetadata["ref"].(string)
+	appendWorkitemAuditEvent(ctx, cmd, root, wkaudit.Event{
```
obey-agent: **Design risk: machine-local attention is written into a git-synced ledger.**

Attention stage lives in the priority store (\`current.yaml\` / attention map), which is intentionally gitignored as per-machine state. Appending \`stage\` (and \`group\`) transitions to \`.campaign/workitems/.workitems.jsonl\` — a file that *is* tracked and already committed in real campaigns — means:

1. Personal "parked/next" thrash becomes shared history and multi-machine merge noise on an append-only JSONL.
2. There is no \`actor\` / machine / host field (acknowledged in the PR), so camp-graph cannot tell whose attention change it was.
3. Unlike \`promote\`/\`gather\`, this path never stages the ledger for commit, so consumers that only read committed history see a different subset than on-disk readers.

If camp-graph needs stage/group for *local* projection, consider a separate non-git ledger (or mark these events \`local: true\` and keep them out of the shared file). If they truly belong in the shared ledger, add actor/source metadata and a commit/staging story so promote/create/stage are consistent. As written, create/adopt (directory lifecycle) fit the shared ledger; stage/group are a different consistency domain bolted onto the same file.

internal/commands/workitem/create.go
```diff
@@ -122,9 +123,17 @@ func runCreate(ctx context.Context, cmd *cobra.Command, slug, typeFlag, title, i
 		return err
 	}
 	markerWritten = true
+	rel := filepath.Join(parent, slug)
 	invalidateNavigationCache(cmd, campaignRoot)
+	appendWorkitemAuditEvent(ctx, cmd, campaignRoot, wkaudit.Event{
```
obey-agent, line 128: **create/adopt append the ledger but never stage it (unlike promote/gather).**

Promote appends *and* adds \`.workitems.jsonl\` to the commit destPaths. Create/adopt only best-effort-append. That is defensible under the audit-trail explore ("ledger on disk is source of truth; git sync is eventual"), but it means:

- A user who creates workitems and never promotes can leave a dirty tracked file indefinitely.
- camp-graph (or any consumer reading committed history) will not see create events until some later commit happens to include the ledger.

Not a blocker if intentional — please state the commit policy in the PR or a one-line comment next to the append: e.g. "ledger dirty until next explicit commit / promote; no auto-stage by design." Without that, the asymmetry looks like a miss relative to promote.go.

obey-agent (commented):
## Verdict: Comment

## Overview
Straightforward, well-tested extension of the workitem audit ledger with `create` / `adopt` / `stage` / `group` events and optional `ref`/`title` fields. Backward-compat decode of old promote/gather lines is locked in. Implementation reuses existing helpers and follows create/adopt's best-effort bookkeeping pattern deliberately.

## Key Findings
- **stage/group write per-machine attention into a git-synced ledger** without actor/machine metadata (inline on `stage.go`). Directory lifecycle events (create/adopt) fit the shared ledger; attention thrash is a different consistency domain and will produce multi-machine JSONL merge noise. Confirm product intent or split local vs shared streams.
- **create/adopt never stage the ledger for commit**, unlike promote/gather (inline on `create.go`). Fine if eventual git-sync is policy — document it so the asymmetry is not read as a bug.
- Best-effort append (stderr warning, never fail the command) is a conscious divergence from promote's fatal audit write; called out in the PR. Acceptable for this scope, but camp-graph consumers should treat the ledger as lossy under FS pressure.

## What's Done Well
- `from`/`to` captured inside `priority.WithLock` so transitions are real, not just the post-state.
- `"none"` for cleared stage/group keeps lines self-describing.
- Shared `appendWorkitemAuditEvent` helper; unit tests cover create/adopt/stage/group CLI paths and mixed old/new ledger lines.
- No scope creep into promote/gather or in-flight validate/repair work.

## Staff Standard
Mostly yes for create/adopt. I would want an explicit product call on stage/group-in-shared-ledger before treating this as the long-term camp-graph source of truth. Not a rewrite — a design confirmation (or a small split) is enough.

PR #394, opened 2026-07-11
Inherit ambient commit context on all intent auto-commit paths, make silent skips loud

lancekrogers:
## Summary

Covers two related intents about intent/note auto-commits:

- `inherit-ambient-questfestivalworkitem-context-on-20260623-221208`: intent/note auto-commits should inherit ambient quest/festival/workitem context in their commit tags, the same way `camp commit` does.
- `intents-not-auto-committing-20260628-225225`: Lance reported intents not auto-committing from festival-app adds and TUI moves.

### Intent A: ambient context on every remaining call site

`add.go`/`note.go` already resolved ambient context via `wkcmd.ResolveCommitContext` and populated `commit.Options.QuestID/FestivalRef/WorkitemRef` by hand. Every other intent mutation caller emitted only a bare `[OBEY-CAMPAIGN-<id>]` tag. This PR extends the same behavior to all of them: `move`, `archive`, `edit`, `promote`, `convert`, `rename`, `gather`, `crawl`, `camp gather feedback`, and the TUI explorer's background auto-commit.

Instead of repeating the resolve-and-populate block at eleven call sites, it is factored into one shared helper:

```go
// internal/commands/workitem/commit_context.go
func AmbientCommitOptions(ctx context.Context, campaignRoot, campaignID string, errw io.Writer) commit.Options
```

Every CLI call site now does:

```go
opts := wkcmd.AmbientCommitOptions(ctx, campaignRoot, cfg.ID, os.Stderr)
opts.Files = commit.NormalizeFiles(...)
opts.SelectiveOnly = true
commitResult := commit.Intent(ctx, commit.IntentOptions{Options: opts, Action: ..., IntentTitle: ...})
```

`add.go`/`note.go` were refactored onto the same helper for consistency. Resolution failures stay non-fatal (empty fields, bare tag), matching current behavior.

For the TUI (`internal/intent/tui/explorer/audit.go`), the helper's `errw` parameter is routed through a small `slogWarnWriter` adapter instead of `os.Stderr`, so ref-backfill warnings land in the TUI's redirected log file (`quietSlogDuringTUI`) instead of corrupting the bubbletea alt-screen.

### Intent B: silent auto-commit skips are now loud

Investigated the reported symptom against current code: it does not reproduce. The TUI move path already commits via `autoCommitIntent` (`internal/intent/tui/explorer/actions.go:136`), and there is no config gate that disables auto-commit. However, two `commit.Options` preconditions were verified that would look exactly like this bug if they ever trigger, because both currently look identical to a legitimate no-op commit:

1. `doCommit` returns silently when `CampaignRoot` or `CampaignID` is empty.
2. A `SelectiveOnly` commit that resolves to zero files (`Files` and `PreStaged` both empty) hits `git.ErrNoChanges` without ever attempting anything.

`commit.Result` now carries `Skipped bool` / `SkipReason string`, populated for exactly these two preconditions (existing `Committed`/`NoChanges`/`Message`/`Err` values are unchanged, so this is additive). A new `commit.WarnIfSkipped(w io.Writer, res Result)` writes a stderr warning; every CLI call site invokes it after building the commit result. The TUI's `runAutoCommitIntent` and `autoCommitIntent` log through the existing `slog.WarnContext` mechanism instead, keeping the terminal clean.

Net effect: both skip paths are non-fatal exactly as before, but now surface a clear, greppable signal instead of nothing, so a recurrence of Lance's symptom is diagnosable.

## Test plan

- [x] `just build` (stable and dev profiles)
- [x] `go vet ./...`, `go vet -tags dev ./...`, `go vet -tags=integration ./...`
- [x] `just lint` (stable and dev profiles): 0 issues, no new host-FS-test violators, no fmt.Errorf regressions
- [x] `just test unit` (stable profile): 3089/3089 passed
- [x] `BUILD_TAGS=dev just test unit`: 3118/3118 passed
- [x] New table-driven tests: `internal/git/commit/commit_test.go` (`Skipped`/`SkipReason` for both trigger conditions, `WarnIfSkipped`), `internal/commands/workitem/commit_context_test.go` (`AmbientCommitOptions` context-present/context-absent), `internal/intent/tui/explorer/audit_test.go` (ambient context threading, both loud-skip paths, via the existing `m.autoCommit.run` stub seam - no host git mutation)
- [x] Pure-logic/unit only; no new host-FS-mutating tests added (existing `internal/git/commit` suite already used t.TempDir + real git, extended in place)

### DIFF
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -496,19 +496,11 @@ func finalizeCreatedIntentWithOutput(ctx context.Context, result *intent.Intent,
 
 	// Auto-commit (unless --no-commit)
 	if !noCommit {
-		files := commit.NormalizeFiles(campaignRoot, result.Path, audit.FilePath(intentsDir))
-		cwd, _ := os.Getwd()
-		cc := wkcmd.ResolveCommitContext(ctx, campaignRoot, cwd, os.Stderr)
+		opts := wkcmd.AmbientCommitOptions(ctx, campaignRoot, cfg.ID, os.Stderr)
+		opts.Files = commit.NormalizeFiles(campaignRoot, result.Path, audit.FilePath(intentsDir))
+		opts.SelectiveOnly = true
 		commitResult := commit.Intent(ctx, commit.IntentOptions{
-			Options: commit.Options{
-				CampaignRoot:  campaignRoot,
-				CampaignID:    cfg.ID,
-				QuestID:       cc.QuestID,
-				FestivalRef:   cc.FestivalRef,
-				WorkitemRef:   cc.WorkitemRef,
-				Files:         files,
-				SelectiveOnly: true,
-			},
+			Options:     opts,
 			Action:      commit.IntentCreate,
 			IntentTitle: result.Title,
 		})
@@ -517,6 +509,7 @@ func finalizeCreatedIntentWithOutput(ctx context.Context, result *intent.Intent,
 				return err
 			}
 		}
+		commit.WarnIfSkipped(os.Stderr, commitResult)
 	}
 
 	return nil
--- a/cmd/camp/gather_feedback.go
+++ b/cmd/camp/gather_feedback.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"fmt"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
 
+	workitemcmd "github.com/Obedience-Corp/camp/internal/commands/workitem"
 	"github.com/Obedience-Corp/camp/internal/config"
 	"github.com/Obedience-Corp/camp/internal/git/commit"
 	"github.com/Obedience-Corp/camp/internal/intent/feedback"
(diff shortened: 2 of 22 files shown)


cmd/camp/intent/add.go
```diff
@@ -515,6 +507,7 @@ func finalizeCreatedIntentWithOutput(ctx context.Context, result *intent.Intent,
 				return err
 			}
 		}
+		commit.WarnIfSkipped(os.Stderr, commitResult)
```
obey-agent, line 512: Under --json, Message is suppressed but WarnIfSkipped still writes to stderr. That is probably fine for diagnostics, but agent harnesses that capture stderr as failure may trip on an otherwise successful create. Consider gating WarnIfSkipped with !jsonOut the same way Message is gated, or documenting that skip warnings always go to stderr.

obey-agent (commented):
**Verdict**: Approve

**Overview**: Clean, well-scoped fix that closes the ambient-context gap on every remaining intent auto-commit path and makes two previously silent skip preconditions greppable. Factoring into `AmbientCommitOptions` is the right integration move; TUI stderr routing via `slogWarnWriter` shows real care about alt-screen integrity.

**Key Findings**:
- (inline, minor) `cmd/camp/intent/add.go`: `WarnIfSkipped` still fires under `--json` even though human `Message` output is gated. Low severity; agents that treat any stderr as failure could get false alarms.
- No blocking correctness issues. SkipReason/Skipped are additive and preserve existing `Committed`/`NoChanges`/`Message`/`Err` semantics. Unit + TUI ambient-context tests cover the important cases.

**What's Done Well**:
- Single shared helper instead of 11 copy-pasted resolve-and-populate blocks.
- TUI keeps warnings on the slog path (no bubbletea corruption).
- Tests for both skip preconditions and `WarnIfSkipped`.
- Honest investigation writeup that the original "not committing" symptom didn't reproduce, and the silent-skip paths are the right hardening.

**Staff Standard**: Yes — I would be proud to submit this. Merge after the one optional json/stderr consistency nit if you care about pure agent stdout/stderr contracts.

obey-agent:
Independent merge verification: read the ambient-context core (ResolveCommitContext / AmbientCommitOptions) and representative call sites; all nine intent auto-commit paths route through the shared helper, and the new Result.Skipped + WarnIfSkipped turns the previously silent missing-context/zero-files skip into an explicit stderr warning with tests. Note on stacking: PR #397 (NT- tag) was based on this branch and its squash landed HERE, not on main; this branch (rebased onto current main, gate green: lint + ratchets clean, all tests passing) now carries both features, so merging this PR delivers #394 and #397 together. Merge-ready.

PR #396, opened 2026-07-11
fix(remote): resolve remote camp over a login shell PATH

lancekrogers:
## Summary

- `camp switch <machine>:<campaign>` and `camp list --remote` invoked the remote camp binary via a bare, non-interactive `ssh host 'camp ...'`, which runs under ssh's own non-login shell. That shell never sources a login profile (`~/.profile`, `~/.bash_profile`, etc.), so a camp installed via a PATH addition only a login shell picks up (`~/.local/bin`, asdf, etc.) was invisible to `ResolveRoot` and the remote list fan-out.
- Every remote camp invocation now runs through `sh -lc '<camp command>'`, so the remote's POSIX login shell resolves PATH the way an interactive ssh session would.
- Added `CAMP_REMOTE_CAMP_PATH` as an explicit-path escape hatch for machines where camp still isn't reachable via the login shell's PATH, documented in `camp switch --help` and `camp list --help`.
- `internal/remote.Run` now returns a `*camperrors.CommandError` (exit code + stderr) instead of an opaque wrapped string, so a missing-binary failure (exit 127) gets an actionable hint pointing at the override, instead of a bare exit status. The detection is scoped to exit code 127 specifically (not stderr text matching) so it never mislabels camp's own domain errors (e.g. `ErrCampaignNotFound` for an unregistered campaign on an otherwise-working remote) as a missing binary.

## Command shape before/after

Before (`internal/remote.ResolveRoot`):
```
ssh <opts> <target> "camp switch 'foo' --print"
```

After:
```
ssh <opts> <target> "sh -lc 'camp switch '\''foo'\'' --print'"
```

With `CAMP_REMOTE_CAMP_PATH=/opt/camp/bin/camp` set:
```
ssh <opts> <target> "sh -lc ''\''/opt/camp/bin/camp'\'' switch '\''foo'\'' --print'"
```

Same wrapping applies to `camp list --remote`'s per-machine `camp list --json` fan-out (`cmd/camp/list_remote.go`).

## Where the override lives

`CAMP_REMOTE_CAMP_PATH` (`internal/remote/ssh.go`): a process environment variable, not `~/.obey/machines.yaml`. The `machines.Machine` struct is an intentionally shared cross-tool contract with the festival app's `RemoteMachine` (mirrored field-for-field per the doc comment on `internal/machines/machines.go`), which explicitly says camp-only fields do not belong there. There is also no existing per-remote settings surface documented in `docs/campaign-settings-files.md` that fits a remote camp binary path. Given that, this ships as a system-level env var per the task's fallback guidance; a proper per-machine config field (e.g. on `Machine`, coordinated with the festival-app schema) is a reasonable follow-up if machines end up needing different explicit paths.

## Gate results

Ran the full local gate from the worktree:
- `just build-camp`: build successful
- `just test unit`: 95 packages, 3093/3093 tests passed
- `just lint`: 0 issues, no new fmt.Errorf/host-fs-test regressions
- `just gate`: `gate-fast` (vet stable/dev/integration, lint stable/dev, unit tests dev profile: 97 packages, 3122/3122 passed) and `gate` (unit tests stable profile: 95 packages, 3093/3093 passed) both `PASSED`

New unit tests added in `internal/remote/ssh_test.go` cover: the `sh -lc` wrapping and quoting (table-driven, plus a POSIX-shell round-trip using an absolute-path probe script: no live ssh, no login-shell profile sourcing), the `CAMP_REMOTE_CAMP_PATH` override (default and with spaces), `resolveRootArgs` quoting (spaces, single quotes, globs), and the not-found hint's exit-code-127 detection including two explicit false-positive guards (permission-denied at 126, and camp's own "not found" domain error at exit 1).

## Deviations from the original ask

- Did not regenerate `docs/cli-reference/*.md`. Running `just docs` pulled in unrelated drift from an already-merged, unrelated feature (org commands: new `camp_org_delete.md`, changes to `camp_create.md`, `camp_init.md`, `camp_org.md`, `camp_org_create.md`). Per established convention, doc regen drift should not be folded into a feature commit, so those are left for a separate docs-only commit. The help text itself (which satisfies "document in command help") is live in the Go source and covered by `--help` output.
- `CAMP_REMOTE_CAMP_PATH` is a single global override, not per-machine. `~/.obey/machines.yaml`'s `Machine` struct explicitly forbids camp-only fields (it mirrors the festival app's schema), and there's no other per-remote config surface today, so a system-level env var is what the task's own fallback guidance calls for. Noted above as a follow-up if per-machine paths turn out to be needed.

Ref intent: `camp-switch-remote-resolve-needs-20260708-234643`

### DIFF
--- a/internal/remote/ssh.go
+++ b/internal/remote/ssh.go
@@ -8,6 +8,7 @@ package remote
 import (
 	"bytes"
 	"context"
+	"errors"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -199,8 +200,11 @@ func ShellQuote(s string) string {
 }
 
 // Run execs `ssh <opts> <target> <remoteCmd>` and returns stdout. The call is
-// bounded by ctx (and DefaultTimeout if ctx has no earlier deadline). A non-zero
-// exit or timeout is a wrapped error carrying the remote stderr.
+// bounded by ctx (and DefaultTimeout if ctx has no earlier deadline). A
+// timeout is a wrapped context error; a non-zero exit is a *camperrors.
+// CommandError carrying the exit code and trimmed remote stderr, so callers
+// can distinguish failure shapes (e.g. RunCampCommand's "binary not found"
+// detection) without re-parsing the error string.
 func Run(ctx context.Context, target string, opts []string, remoteCmd string) ([]byte, error) {
 	ctx, cancel := context.WithTimeout(ctx, DefaultTimeout)
 	defer cancel()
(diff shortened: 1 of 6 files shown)


internal/remote/ssh.go
```diff
@@ -214,21 +218,101 @@ func Run(ctx context.Context, target string, opts []string, remoteCmd string) ([
...
+// resolve on a far machine whose camp binary ran fine), and mislabeling
+// those as a missing binary would send the user chasing PATH for nothing.
+// Any exit code other than 127 is returned unchanged.
+func campNotFoundHint(err error, m *machines.Machine, binary string) error {
+	var cmdErr *camperrors.CommandError
+	if !errors.As(err, &cmdErr) || cmdErr.ExitCode != 127 {
+		return err
+	}
+	return camperrors.Wrapf(err,
+		"remote camp not found on %s (tried %q via sh -lc, i.e. the machine's login-shell PATH); "+
```
obey-agent, line 298: Error text says "the machine's login-shell PATH", but the remote command is always `sh -lc`, not the account's configured login shell from passwd/`$SHELL`.

On machines where PATH is only exported from zsh/bash-specific profiles (`.zprofile`, `.bash_profile`, `.zshrc`), `sh -lc` still will not see camp — only `~/.profile` / `/etc/profile` style sh login paths are gained. That is a reasonable deliberate tradeoff (POSIX `/bin/sh` is always present), but the wording overclaims and will send operators debugging the wrong files.

Suggest something closer to:
`tried %q via sh -lc (POSIX login profile, e.g. ~/.profile); if camp is only on a zsh/bash-specific PATH, set %s`

Same nuance belongs in the `--help` text that currently implies "the login profile" generically.

obey-agent (commented):
**Verdict**: Approve

**Overview**: Correct root-cause fix for remote camp resolution: bare non-interactive ssh never sourced login profiles. Centralizing on `RunCampCommand` + `sh -lc`, promoting remote failures to `*CommandError` with exit codes, and scoping the missing-binary hint to exit 127 only is excellent production craft.

**Key Findings**:
- (inline, non-blocking) `campNotFoundHint` / help text overstate "login-shell PATH" — it is specifically `sh`'s login profile, not `$SHELL`. Document that zsh/bash-only PATH setups still need `CAMP_REMOTE_CAMP_PATH`.
- No other correctness concerns. Quoting is tested with a real POSIX shell round-trip (spaces, quotes, globs). Domain "campaign not found" at exit 1 is correctly not relabeled as a missing binary. EnsureKeyAuth is not double-called after the list_remote refactor.

**What's Done Well**:
- Pure `campRemoteCommandLine` / `resolveRootArgs` for unit-testable quoting without live ssh.
- Exit-code-127 detection instead of stderr text matching (avoids false positives on camp domain errors).
- Explicit env override with help docs, without polluting the shared `machines.Machine` cross-tool contract.
- Full gate green with thorough new unit tests.

**Staff Standard**: Yes. Merge after optionally tightening the help/error wording about sh vs user shell.

obey-agent:
Independent merge verification: read the full diff. Login-shell resolution via sh -lc with the rationale documented (POSIX sh guaranteed, profile PATH sourced), exit-127-only not-found detection (correctly avoiding stderr text matching since camp domain errors also say 'not found'), CAMP_REMOTE_CAMP_PATH escape hatch on both help surfaces, structured CommandError instead of string parsing, and a real-shell quoting round-trip test. Test-merged current main cleanly; full gate PASSED. Merge-ready.

PR #397, opened 2026-07-11
feat(git): add NT-<ref> note commit tag segment

lancekrogers:
## Summary

Notes are a separate non-lifecycle category (`internal/intent/notes.go:18`) and are not workitems, so today a note commit can only ever carry the campaign/ambient tags, never an identifier for the note itself. This gives notes their own `NT-<ref>` commit tag segment.

Addresses intent `give-notes-their-own-nt-20260623-221250`.

**Stacked on #394 (`intent-commit-context`). This PR targets `intent-commit-context`, not `main`, and must merge after it.** It builds on the `wkcmd.AmbientCommitOptions` / `commit.Options` seam that #394 introduced for note auto-commits.

## Design decisions

- **Position and co-occurrence**: fixed tag order is now `head, qst_, FE-, WI-, NT-`. `NT-` may co-occur with `WI-`: with #394, a note captured while a festival/workitem is active inherits that ambient context (`FE-`/`WI-` describe *where* the note was captured), while `NT-` identifies *the note itself*. These are different dimensions, so mutual exclusion would discard real information.
- **Ref derivation**: `NT-<6 hex>` is derived from the note's frontmatter id using the same sha256-prefix scheme `WI-` uses. `internal/workitem/ref.go`'s `Derive` is now `DeriveWithPrefix(RefPrefix, id)`; the note commit path calls `DeriveWithPrefix("NT-", id)` directly, so the hashing logic isn't duplicated and `Derive`'s output stays byte-identical.
- **Blast radius**: `FormatContextTagsFull` gains `noteRef` as a variadic trailing param (mirroring the existing `FormatCampaignTag(campaignID string, questID ...string)` pattern in the same file), not a new positional arg. That means the five other call sites of `FormatContextTagsFull`/`PrependContextTagsFull` (`cmd/camp/commit.go`, `cmd/camp/project/commit.go`, `cmd/camp/worktrees/commit.go`, `cmd/camp/refs/commands.go`, `internal/commands/workitem/staging.go`, `pkg/commitkit`) don't need to change at all: only the note commit path (`cmd/camp/intent/note.go` → `commit.Options.NoteRef` → `doCommit`) ever sets one.
- **Parser**: `WI-` used to be the terminal segment in `ParseTagDetailed`; it no longer is. `splitWorkitemSegment` now bounds the match at the fixed 6-hex-char ref length (handling the historical doubled `WI-WI-` form) so a following `-NT-...` is left for the next parse iteration instead of being swallowed. Malformed-shape behavior for `WI-` itself is unchanged (verified via the existing `TestParseTagDetailed_RejectsSilentMerge` cases, which still pass).

## Scope

- `internal/git/campaign_tag.go`: `FormatContextTagsFull` NT- segment, `TagComponents.NoteRef`, `tagNoteRefRe`, `ParseTagDetailed` WI-/NT- parsing.
- `internal/workitem/ref.go`: generalized `Derive` → `DeriveWithPrefix`.
- `internal/git/commit/commit.go`: `Options.NoteRef`, threaded into `doCommit`'s tag formatting.
- `cmd/camp/intent/note.go`: derives and sets `NoteRef` on the note commit path only.

## Test plan

- [x] `internal/git`: format round-trip (with/without co-occurring WI-), parser shape checks (valid NT-, malformed NT- shape-check-failure, non-`NT-`-prefixed content still falls to unknown-segment), randomized round-trip property test extended with note refs, all pre-existing WI-/FE-/quest tests still green.
- [x] `internal/workitem`: `DeriveWithPrefix` determinism, prefix independence, `Derive` byte-identical to `DeriveWithPrefix(RefPrefix, id)`.
- [x] `internal/git/commit`: unit tests confirming `NoteRef` reaches the commit message and co-occurs with `WorkitemRef`, plus a guard test that a non-note action (`IntentMove`) with `WorkitemRef` set but `NoteRef` unset never emits `-NT-`.
- [x] `tests/integration/commit_tags_test.go` (containerized, real `camp` binary): note commit gets its own `NT-<ref>`, a note captured inside an active workitem carries `WI-...-NT-...` in that order, and `camp intent add` from the same workitem context never leaks an `NT-` segment (guards `AmbientCommitOptions` isn't accidentally shared into non-note call sites).
- [x] `just build`, `just test unit` (3101/3101 passed), `just lint` (0 issues, no new host-FS test violators, no fmt.Errorf regressions) all green on the worktree.

### DIFF
--- a/internal/git/campaign_tag.go
+++ b/internal/git/campaign_tag.go
@@ -17,9 +17,13 @@ const legacyTagMarker = "OBEY-CAMPAIGN"
 // FormatContextTagsFull builds the campaign tag. The leading token is the
 // slugified campaign name plus the short id ("[obey-campaign:8deed8b4]"),
 // falling back to "[OBEY-CAMPAIGN-<id>]" when campaignName has no slug. The
-// remaining components follow in fixed order: quest, festival, workitem.
+// remaining components follow in fixed order: quest, festival, workitem,
+// note. noteRef is optional (mirroring FormatCampaignTag's questID) so
+// existing callers are unaffected; only the note commit path passes one, and
+// it may co-occur with workitemRef since they describe different things (the
+// ambient context a note was captured in vs. the note itself).
 // Returns "" when campaignID is empty.
-func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemRef string) string {
+func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemRef string, noteRef ...string) string {
 	if campaignID == "" {
 		return ""
 	}
@@ -50,6 +54,16 @@ func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemR
 		// The ref already carries WI-, so it is self-identifying and embedded
 		// verbatim (no extra marker).
 		parts = append(parts, workitemRef)
+	}
+	nr := ""
+	if len(noteRef) > 0 {
+		nr = noteRef[0]
+	}
+	if nr != "" {
+		if !strings.HasPrefix(nr, "NT-") {
+			nr = "NT-" + nr
+		}
+		parts = append(parts, nr)
 	}
 	return "[" + strings.Join(parts, "-") + "]"
 }
@@ -96,6 +110,7 @@ type TagComponents struct {
 	QuestID      string `json:"quest_id"`
 	FestRef      string `json:"fest_ref"`
 	WorkitemRef  string `json:"workitem_ref"` // carries the WI- prefix
+	NoteRef      string `json:"note_ref"`     // carries the NT- prefix
 }
 
 // leadingTagRegex captures the leading bracket content; tags are only honored
@@ -108,6 +123,7 @@ var tagBodyScanRegex = regexp.MustCompile(`\[(?:` + legacyTagMarker + `-[^\]]+|[
 
 var (
 	tagWorkitemRefRe = regexp.MustCompile(`^WI-[0-9a-f]{6}$`)
+	tagNoteRefRe     = regexp.MustCompile(`^NT-[0-9a-f]{6}$`)
 	tagQuestIDRe     = regexp.MustCompile(`^qst_[A-Za-z0-9_]{1,40}$`)
 	tagFestRefRe     = regexp.MustCompile(`^[A-Za-z0-9]{1,32}$`)
 	// Real campaign ids are UUID-derived hex; gating on it rejects ordinary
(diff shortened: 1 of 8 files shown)


internal/git/campaign_tag.go
```diff
@@ -17,9 +17,13 @@ const legacyTagMarker = "OBEY-CAMPAIGN"
...
 // falling back to "[OBEY-CAMPAIGN-<id>]" when campaignName has no slug. The
-// remaining components follow in fixed order: quest, festival, workitem.
+// remaining components follow in fixed order: quest, festival, workitem,
+// note. noteRef is optional (mirroring FormatCampaignTag's questID) so
+// existing callers are unaffected; only the note commit path passes one, and
+// it may co-occur with workitemRef since they describe different things (the
+// ambient context a note was captured in vs. the note itself).
 // Returns "" when campaignID is empty.
-func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemRef string) string {
+func FormatContextTagsFull(campaignName, campaignID, questID, festRef, workitemRef string, noteRef ...string) string {
```
obey-agent, line 26: Variadic `noteRef ...string` keeps 5-arg callers compiling, but silently ignores `noteRef[1:]` if a caller ever passes more than one value. Prefer either:
1. an explicit optional last arg pattern with a new `FormatContextTagsFullWithNote(...)` helper, or
2. a single `noteRef string` plus a thin wrapper for old call sites,

so the fixed tag order cannot be mis-invoked. Low severity given only one call site sets NoteRef today, but tag formatting is a long-lived public contract surface.

obey-agent (commented):
**Verdict**: Approve

**Overview**: Clean extension of the campaign tag grammar for note identity (`NT-<6 hex>`), correctly co-occurring with ambient `WI-` after #394. Parse/format round-trip, `splitWorkitemSegment` for historical `WI-WI-` forms, and DeriveWithPrefix extraction are careful and well tested. Stacking on `intent-commit-context` is correct.

**Key Findings**:
- (inline, non-blocking) Variadic `noteRef ...string` silently drops extra args; a named helper or plain string would be safer for a tag-format API.
- No correctness blockers: fixed order head/qst/FE/WI/NT, co-occurrence with workitem, non-note paths never emit NT- (unit + integration coverage). Parse path correctly re-bounds WI- so trailing NT- is not swallowed.

**What's Done Well**:
- Reuses workitem hash scheme via `DeriveWithPrefix` instead of forking hash logic.
- Integration tests for note-only, note+WI, and non-note no-leak.
- Historical doubled WI- prefix handling preserved via `splitWorkitemSegment`.
- Note path only sets `opts.NoteRef` at create — ambient options cannot leak NT- into other intent actions.

**Staff Standard**: Yes — merge after #394. The variadic nit is optional polish.

obey-agent:
Independent merge verification: read the full diff. NT- segment keeps the fixed tag order (note always last), splitWorkitemSegment preserves legacy malformed-tag warnings incl. the doubled WI-WI- form, round-trip property tests cover co-occurrence with quest/festival/workitem. The variadic optional arg follows the existing FormatCampaignTag questID precedent, with a 5-arg caller compatibility test. Test-merged current main; full gate PASSED (one unrelated fsutil flake passed on rerun and in isolation). Merge-ready.

PR #395, opened 2026-07-11
feat(workitem): add validate and repair commands

lancekrogers:
## Summary

Adds `camp workitem validate` and `camp workitem repair` for workflow work item
directories, implementing intent `add-camp-workitem-validate-and-20260616-200745`.

The motivating problem: legacy `workflow/design/` directories were discovered by
`camp workitem --json` (design and explore directories are work items by
location) but had no `.workitem` marker, so they showed up as ambiguous,
untracked work. Validate now catches that shape and prints the exact repair
command; repair fixes it in one non-destructive step.

## Command boundary (doctor vs validate vs repair)

The three commands share internals but keep distinct user-facing purposes:

- `camp workitem doctor` checks the link registry, current-workitem, and
  priority stores, and backfills refs across every discovered workitem. It is
  the only one that mutates the registry, and only with `--fix`.
- `camp workitem validate` is the read-only structural checker for workflow work
  item directories and their `.workitem` markers. It never writes.
- `camp workitem repair` is the one-command, idempotent, non-destructive fix for
  a single directory: it creates or upgrades that directory's marker.

## Which directories validate scans

Validate mirrors discovery so it never produces false positives:

- `workflow/design/<dir>` and `workflow/explore/<dir>` are work items by
  location; every child is a candidate, so an unmarked legacy directory is
  reported.
- `workflow/<custom-type>/<dir>` is a candidate only when it already carries a
  marker; an unmarked custom directory is intentionally not a work item and is
  never flagged.
- `dungeon`, hidden directories, `workflow/intent`, and `workflow/festival` are
  control areas and are always skipped.

A directory directly under `workflow/<type>/` is legitimate as is. Neither
command suggests adding a status subdirectory, and repair never moves or renames
a directory (coordinating with the sibling intent
`treat-workflow-workitem-directories-as-20260616-210558`).

## Finding codes

`camp workitem validate --json` emits stable codes; each finding carries
`severity`, the campaign-relative `target`, a `message`, the exact
`repair_command`, and `repairable`.

| Code | Severity | Meaning | Repair |
| --- | --- | --- | --- |
| `workitem.marker.missing` | error | design/explore (or explicitly targeted) directory has no `.workitem` marker | creates a marker |
| `workitem.marker.malformed` | error | marker cannot be read, is not YAML, or violates schema (bad kind, empty id, unsupported version, malformed ref/quest id) | repairs when parseable; refuses unparseable (`repairable: false`) |
[...]

PR #398, opened 2026-07-11
feat(workitem): treat workflow workitem directories as active by location

lancekrogers:
## Summary

Implements intent `treat-workflow-workitem-directories-as-20260616-210558`: a workitem directory directly under `workflow/<type>/` (design, explore, or a custom marker-gated type) is now reported as lifecycle_stage `active` by default, without requiring an explicit `workflow/<type>/active/<dir>/` subdirectory. This matches how festivals and intents already derive lifecycle stage from location.

- `buildWorkflowDirItem` (shared by design/explore discovery and custom-type discovery) now sets `LifecycleStage: LifecycleStageActive` instead of `LifecycleStageNone`. Dungeon/status subtrees are unaffected: they were already excluded from discovery entirely (`discoverWorkflowDocs` and `discoverCustomWorkflowTypes` both skip `dungeon` and dotfiles), so their explicit lifecycle semantics are unchanged.
- `ValidStagesForType`'s fallback for custom (non-builtin) types now includes `active` alongside `none`, so `camp workitem --json --stage active --type <customtype>` validates and filters correctly instead of erroring with "unknown --stage value".
- Fixed a latent bug this surfaced: `priority.EligibleForAttention` gated custom-type attention-stage eligibility on `item.LifecycleStage == LifecycleStageNone` as a proxy for "is a custom directory-backed type." Once custom types default to `active`, that proxy silently breaks: `camp workitem stage <custom-item> current` would start failing with "attention stage is only supported for directory-backed workflow workitems" for every custom-type workitem. Replaced the proxy with an explicit `!workitem.IsBuiltinWorkflowType(item.WorkflowType)` check (new exported helper in `internal/workitem`), verified end-to-end against a real custom-type item before and after the fix.

## Compatibility

**camp `--json` consumers (festival-app):** Checked `projects/festival-app` for consumers of `lifecycle_stage`/`attention_stage`.
- `src-tauri/src/commands/workitem.rs:78-79` deserializes `lifecycle_stage` as a plain `String` with `#[serde(alias = "lifecycle_stage")]`: no enum constraint, so the new value deserializes without any Rust-side change.
- `src/components/workitems/WorkitemListPane.tsx` already has a dedicated `STAGE_BADGE["active"]` entry (green badge) and a generic fallback for unknown/no-stage values: design/explore/custom items will now render with the active badge instead of the neutral fallback. Cosmetic improvement, not a break.
- `src/components/workitems/WorkitemExplorerModal.tsx` defaults `stages: []` (no stage filter applied by default), so no persisted/default filter state assumes `none` for workflow items.
- Existing mock fixtures and tests (`src/lib/mock/data.ts`, `WorkitemExplorerModal.test.tsx`) already use `lifecycleStage: "active"` for non-festival items, confirming the frontend was already tolerant of this value.
- **No festival-app changes required.**

**camp `internal/workitem` package consumers:** `cmd/camp/promote` (`isPromotable`, `EligibleForAttention`), TUI list/preview rendering, and JSON stage vocabulary were all audited; only the `EligibleForAttention` fix above was required. `promotablePool`'s default branch already returns `true` regardless of lifecycle stage for design/explore/custom types, so promotion behavior is unaffected.

**camp PR #395** (`workitem-validate-repair`, not yet merged) also touches `internal/workitem/discover_workflow_docs.go` and adds `wkitem.InferTitle` refactoring. This branch is based on `main` pre-#395, so expect a small merge conflict in that file: my change here is a single-line default-value edit (`LifecycleStageNone` -> `LifecycleStageActive` in `buildWorkflowDirItem`), kept minimal specifically to reduce conflict surface. `camp workitem validate`/repair logic does not exist yet on this branch (it's #395's addition), so there was nothing here to update for the "validation/repair should not flag missing active/" acceptance criterion: that will need to be honored when #395 lands or rebases on top of this.

## Test plan

- [x] `just build`: clean build, no errors
- [x] `just test unit`: 3086/3086 tests passed (95 packages)
- [x] `just lint`: 0 issues
- [x] Added table-driven tests: `TestDiscoverCustomWorkflowTypes_ActiveByLocationAndDungeonExcluded` (design/explore/custom active-by-location + dungeon/hidden/no-marker exclusions), `TestEligibleForAttention` (design/explore/custom eligible, intent/festival not), `TestValidStagesForType_CustomTypeAllowsActive`, `TestIsValidStageForTypes` (stage/type validation matrix)
- [x] Updated existing discovery tests (`TestDiscoverDesign_ExcludesDungeonAndHidden`, `TestDiscoverExplore_Works`) that previously asserted `LifecycleStageNone`
- [x] Manual end-to-end smoke test against a real scaffolded campaign (`camp init` + hand-built workflow/design, workflow/explore, workflow/incident items):
  - `camp workitem --json` reports `active` for all three without any `active/` subdirectory
  - `camp workitem --json --stage active --type incident` correctly filters to the custom-type item
  - `camp workitem stage incident-outage-1 current --json` succeeds (this is the regression `EligibleForAttention` would otherwise have introduced)
  - dungeon-excluded design item correctly absent from output


internal/workitem/discover_custom.go
```diff
@@ -13,6 +13,10 @@ var builtinTypes = map[WorkflowType]bool{
 	WorkflowTypeFestival: true,
 }
 
+func IsBuiltinWorkflowType(wt WorkflowType) bool {
```
obey-agent: Naming collision ahead: this PR exports `IsBuiltinWorkflowType` while open PR #395 exports `IsBuiltinType` / `IsBuiltinDocType` over the same `builtinTypes` map. Whichever merges second will either duplicate helpers or force a rename conflict.

Prefer aligning on one name now (I'd keep the shorter `IsBuiltinType` and add `IsBuiltinDocType` from #395 if that lands after, or land a shared name first). Not a functional bug in isolation — merge-order hygiene so the package API does not grow twin synonyms.

obey-agent (commented):
**Verdict**: Approve

**Overview**: Small, precise semantic change: workflow directory workitems are `active` by location, matching how festivals/intents already derive stage from path. The latent `EligibleForAttention` breakage this would have caused for custom types is correctly identified and fixed with an explicit builtin check instead of the old `LifecycleStageNone` proxy.

**Key Findings**:
- (inline, non-blocking) `IsBuiltinWorkflowType` will collide naming-wise with #395's `IsBuiltinType` — coordinate on one export at merge time.
- No functional blockers. Dungeon/hidden remain excluded from discovery, so "active by location" does not reclassify archived trees. Stage filter/validation for custom types now accepts `active` as required. Festival-app consumer audit in the PR body is solid.

**What's Done Well**:
- Minimal one-line discovery change to reduce conflict surface with #395.
- Table-driven tests for custom discovery, EligibleForAttention matrix, and ValidStagesForType.
- Explicit regression story for `camp workitem stage <custom> current` which would have failed post-change without the EligibleForAttention fix.

**Staff Standard**: Yes. Ship it; just reconcile the helper name with #395 when merging.

obey-agent:
Independent merge verification: read the full diff. Active-by-location for workflow dirs is consistent across discovery (buildWorkflowDirItem), custom types (marker-gated, dungeon/hidden excluded, tested), EligibleForAttention, and ValidStagesForType. Resolved the flagged naming collision with #395 during the main merge: dropped IsBuiltinWorkflowType, consolidated on IsBuiltinType. Full local gate PASSED. Merge-ready.

PR #399, opened 2026-07-11
camp project run: cr -p shorthand, tab completion, refreshed help

lancekrogers:
## Summary

- Wires `-p`/`--project` tab completion onto `camp project run` despite `DisableFlagParsing`. Cobra never calls its own flag-value completion machinery in that mode (`checkIfFlagCompletion` short-circuits on `DisableFlagParsing`), so `ValidArgsFunction` now inspects the raw args itself to detect flag context, matching `parseProjectRunArgs`'s own resolution order exactly (including edge cases like a trailing bare `-p` or a `-p` that appears after the command has already started).
- Adds a `cr -p <project> [--] <command...>` shell shorthand for `camp project run -p <project> -- <command...>` in the zsh, bash, and fish shell-init templates, alongside the existing root-level `cr <command...>` (unchanged).
- zsh gets full dynamic completion: `cr -p <TAB>` and `camp project run -p <TAB>` both list registered project names via `camp __complete`, and `camp project <TAB>` now lists the real subcommand set (it was hardcoded to `add|list|remove` and didn't even include `run`). Plain `cr <command> [args...]` completion is preserved exactly via zsh's own `_normal` delegation technique (the same one `_precommand` uses for `nohup`/`time`/`env`-style wrappers).
- bash and fish get the `cr -p` shorthand and the corrected static `project` subcommand list, but not dynamic project-name completion for `-p`: neither template has any existing `__complete`-driven dynamic completion precedent to extend (zsh's `_csw` is the only prior art), so this keeps parity with each shell's current completion tier rather than introducing a new pattern in one PR.
- Refreshes `camp project run --help` (documents the `cr -p` shorthand and completion) and `camp project --help` (the parent listing was missing `run`, `commit`, `prune`, and `worktree` entirely).

Ref: intent `camp-project-run-command-needs-20260708-011321`.

## How completion works under DisableFlagParsing (and how I verified it)

Cobra's completion path (`completions.go`) special-cases `DisableFlagParsing`: `checkIfFlagCompletion` returns immediately with a nil flag, so `RegisterFlagCompletionFunc` never fires for this command (it's dead code if added): `ValidArgsFunction` is the *only* mechanism that runs, and it receives the raw, unparsed args plus the word being completed. `completeProjectRunArgs` in `cmd/camp/project/run.go` walks those args the same way `parseProjectRunArgs` does (via a new `projectRunFlagZone` helper) to decide whether the current word is a `-p`/`--project` value, an inline `--project=`/`-p=` form, or neither.

Verified by driving the hidden `__complete` command directly against the built binary:

```
$ camp __complete project run -p ""
camp
fest
festival
...
:4
Completion ended with directive: ShellCompDirectiveNoFileComp

$ camp __complete project run "--project=fe"
--project=fest
--project=festival
:4
```

and confirmed the negative cases (flag-name completion for `-`/`--pro`, no completion past `--`, no completion once a command word has started) all match `parseProjectRunArgs`'s actual runtime behavior. Added `TestParseProjectRunArgs`, `TestProjectRunFlagZone`, and `TestCompleteProjectRunArgs` (`cmd/camp/project/run_test.go`) covering all of this at the Go level.

For the zsh shell glue specifically, I additionally drove a real interactive zsh session non-interactively via `zsh/zpty`, sourcing the rendered `camp shell-init zsh` output against a fixture campaign, and confirmed `cr -p f<TAB>`, `camp project run -p <TAB>`, and `camp project r<TAB>` all produce the right candidates, while `cr ls -<TAB>` still gets normal top-level completion (not project names) via `_normal`. That surfaced a real bug along the way: `_arguments -C`'s `->args` state handler rebases (and can re-invoke with alternate splits of) `$words`/`$CURRENT`/`$line`, so a two-level nested dispatch (project → run → -p-value) can never see a stable view of the original command line from inside that handler. Fixed by snapshotting `$words`/`$CURRENT` in `_camp` *before* calling `_arguments -C`, and passing that snapshot down explicitly instead of reading the live (rebased) globals.

## Deviations from the docs regen convention

`docs/cli-reference` is generated by `just docs` and already has unrelated drift from the orgs merge (PR #387). Per campaign convention, regen is kept out of feature PRs: not run here.

## Test plan

- [x] `just build-camp` (stable and dev profiles)
- [x] `go vet ./...` (stable, dev, integration tags)
- [x] `just lint` (stable and dev; `lint-no-host-fs-tests` clean, `lint-no-fmt-errorf` clean)
- [x] `just test unit`: 3117/3117 (dev profile) and 3088/3088 (stable profile) passing
- [x] `just gate`: full gate passed end-to-end
- [x] Manual `camp __complete project run ...` verification of every branch (bare `-p`, prefix filter, `--project=`/`-p=` inline forms, flag-name completion, past-`--` fallback, past-a-command-word fallback)
- [x] Real interactive zsh session via `zsh/zpty` confirming `cr -p <TAB>`, `camp project run -p <TAB>`, `camp project r<TAB>`, and plain `cr <command> -<TAB>` all behave correctly

### DIFF
--- a/cmd/camp/project/run.go
+++ b/cmd/camp/project/run.go
@@ -17,11 +17,14 @@ import (
 
 var projectRunCmd = &cobra.Command{
 	Use:   "run [--project <name>] [--] <command> [args...]",
-	Short: "Run a command inside a project directory",
+	Short: "Run a command inside a project directory, like cr but project-scoped",
 	Long: `Run any shell command inside a project directory from anywhere in the campaign.
 
+This is the project-scoped counterpart to 'camp run' (cr): cr runs from the
+campaign root, camp project run (cr -p) runs inside a project.
+
 The project is resolved in this order:
-  1. --project / -p flag (explicit project name)
+  1. --project / -p flag (explicit project name, tab-completes registered projects)
   2. Auto-detect from current working directory
   3. Interactive fuzzy picker (if neither above applies)
 
@@ -39,12 +42,25 @@ Examples:
   camp project run -- just test all
 
   # Simple commands (no -- needed when no flags)
-  camp project run make build`,
+  camp project run make build
+
+  # Shell shorthand (after 'eval "$(camp shell-init <shell>)"')
+  cr -p fest -- just build
+  cr -p camp go test ./...`,
 	DisableFlagParsing: true,
 	RunE:               runProjectRun,
+	ValidArgsFunction:  completeProjectRunArgs,
 }
 
 func init() {
+	// Registered only so cobra's completion machinery advertises --project/-p
+	// as known flag names (see completeRequireFlags/doCompleteFlags in
+	// cobra's completions.go, which walk NonInheritedFlags() even when
+	// DisableFlagParsing is set). DisableFlagParsing means cobra never
+	// parses or binds this flag at runtime; parseProjectRunArgs below does
+	// that manually, and completeProjectRunArgs handles its value completion.
+	projectRunCmd.Flags().StringP("project", "p", "", "Project name (auto-detected from cwd, or interactive picker if omitted)")
+
 	Cmd.AddCommand(projectRunCmd)
 }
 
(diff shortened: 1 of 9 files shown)


internal/shell/templates/bash.sh.tmpl
```diff
@@ -241,10 +241,19 @@ _cgo_complete() {
 }
 complete -o nospace -F _cgo_complete cgo
 
-# Run command from campaign root
+# Run command from campaign root, or in a project with -p
 # Usage: cr <command> [args...]
+#        cr -p <project> [--] <command> [args...]
 cr() {
-  camp run "$@"
+  if [ "$1" = "-p" ] || [ "$1" = "--project" ]; then
```
obey-agent: Shell `cr` only recognizes bare `-p` / `--project` as argv[1]. The Go parser (and help examples for `camp project run`) also accept `--project=fest` / `-p=fest`.

```bash
cr --project=fest -- just build   # falls through to: camp run --project=fest -- just build
```

That is wrong: `camp run` does not implement `--project`, so the user gets a confusing root-run or a flag error instead of project-scoped execution.

Either:
1. teach the three shell wrappers the `=value` forms (and empty-project guard), or
2. document explicitly that the shorthand only supports `cr -p <name>` / `cr --project <name>` and never the `=` form.

Also: `cr -p` with no project name sets an empty project and still invokes `camp project run -p ""` — a cheap `$#` guard with a usage message would save a bad trip.

cmd/camp/project/run.go
```diff
@@ -154,6 +170,68 @@ func parseProjectRunArgs(args []string) (projectName string, command []string) {
 	return projectName, nil
```
obey-agent, line 170: `completeProjectRunArgs` correctly reimplements flag-zone detection under `DisableFlagParsing` — good. One edge vs `parseProjectRunArgs`: when args are `[]string{"-p"}` (trailing bare `-p` with no value), parse treats it as a command word (`command=["-p"]`), while `projectRunFlagZone` reports `awaitingValue=true` and offers project completions.

That mismatch is probably what you want for completion UX, but worth a one-line comment next to the test case `"trailing -p awaits value"` so a future reader does not "fix" completion to match parse and break tab-complete of `camp project run -p <TAB>`.

obey-agent (commented):
**Verdict**: Comment

**Overview**: High-quality UX work on a genuinely hard surface (`DisableFlagParsing` + shell completion). The Go completion path is carefully tested; zsh's pre-`_arguments` snapshot fix for nested `project → run → -p` is real production debugging. The intentional bash/fish vs zsh completion tier split is documented and reasonable.

**Key Findings**:
- (inline) Shell `cr` shorthand does not accept `--project=value` / `-p=value` forms that the Go command supports, and silently misroutes them to `camp run`.
- (inline) Completion vs parse divergence on trailing bare `-p` is fine but should be commented so it is not "fixed" away.
- Non-blocking: bash/fish lack dynamic project-name completion for `cr -p` (explicitly scoped out; OK if documented for users).

**What's Done Well**:
- ValidArgsFunction walks raw args the same way as runtime parsing for the common cases.
- zsh `_normal` delegation preserves plain `cr <cmd>` completion correctly.
- Project subcommand completion lists updated from the stale `add|list|remove` set.
- Unit tests for parse, flag zone, and complete; shell template string tests for all three shells.

**Staff Standard**: Nearly — I would merge after fixing or clearly documenting the `cr --project=` footgun. The Go-side completion work is already staff-level.

obey-agent (approved):
## Verdict: Approve

## Overview
Re-review of head `95d5c6e`. `cr -p` / `camp project run` UX remains high quality on a hard surface (`DisableFlagParsing` + multi-shell completion). The follow-up hardens shell wrappers for missing `-p` values and `-p=` / `--project=` forms across bash/fish/zsh, with matching template tests.

## Key Findings
- zsh pre-`_arguments` snapshot for nested `project → run → -p` completion is real production-grade shell work.
- Equals-form and missing-project usage errors close the previous shorthand footguns.
- Completion still deliberately offers a project value after a trailing bare `-p` (documented) even though runtime treats that token as the command — correct completion UX.

## What's Done Well
Template tests that lock the generated shell source; intentional bash/fish vs zsh completion tier split is reasonable.

## Staff Standard
Yes.

obey-agent:
Independent merge verification at head 95d5c6e: read the full diff (completion for a DisableFlagParsing command, projectRunFlagZone walker, cr -p wrappers across bash/fish/zsh with template tests). Test-merged against current main cleanly; full local gate PASSED (3194 + 3165 tests, lint + ratchets clean). Merge-ready.

PR #400, opened 2026-07-12
fix(leverage): exclude worktrees from scc scans

lancekrogers:
## Summary

- Add `worktrees`, `.worktrees`, and `.camp-worktrees` to leverage `DefaultExcludeDirs` so `scc` does not scan project worktree checkouts
- Document the exclusion in `docs/leverage-score.md`
- Add a unit test that the worktree dirs stay in the default exclude list

Without this, parallel review worktrees under `projects/worktrees/` are full source trees and multi-count LOC in COCOMO estimates (e.g. camp ~101x → ~331x after two review worktrees).

## Test plan

- [x] `go test ./internal/leverage/ -count=1`
- [ ] After merge/install: run `camp leverage --project camp` with local worktrees present and confirm file/LOC counts match non-worktree baseline
- [ ] Optionally `camp leverage reset` / re-snapshot if cached inflated snapshots need clearing

### DIFF
--- a/docs/leverage-score.md
+++ b/docs/leverage-score.md
@@ -122,7 +122,9 @@ checkouts that would inflate COCOMO estimates: `node_modules`, `vendor`,
 `dist`, `build`, `target`, `worktrees`, `.worktrees`, `.camp-worktrees`,
 and similar. Project worktrees under `projects/worktrees/` are full source
 trees; counting them would multi-count the same codebase in leverage scores.
-Per-project extras can still be set via config `ExcludeDirs`.
+Monorepo root entries additionally exclude their expanded subproject
+directories via the project list's `ExcludeDirs`, so submodule code is
+not double-counted.
 
 ### From configuration (actual)
 
--- a/docs/leverage-score.md
+++ b/docs/leverage-score.md
@@ -112,8 +112,17 @@ We use it as the baseline because:
 scc is invoked as:
 
 ```bash
-scc --format json2 --cocomo-project-type <type> <directory>
+scc --format json2 --cocomo-project-type <type> \
+  --exclude-dir node_modules --exclude-dir vendor ... \
+  <directory>
 ```
+
+Default `--exclude-dir` values skip dependency/build caches and parallel
+checkouts that would inflate COCOMO estimates: `node_modules`, `vendor`,
+`dist`, `build`, `target`, `worktrees`, `.worktrees`, `.camp-worktrees`,
+and similar. Project worktrees under `projects/worktrees/` are full source
+trees; counting them would multi-count the same codebase in leverage scores.
+Per-project extras can still be set via config `ExcludeDirs`.
 
 ### From configuration (actual)
 
--- a/internal/leverage/scc.go
+++ b/internal/leverage/scc.go
@@ -9,8 +9,12 @@ import (
 )
 
 // DefaultExcludeDirs are directories excluded from scc scans by default.
-// These contain vendored dependencies, build output, or caches that
-// would inflate COCOMO estimates beyond authored code.
+// These contain vendored dependencies, build output, caches, or parallel
+// checkouts that would inflate COCOMO estimates beyond authored code.
+//
+// worktrees / .worktrees / .camp-worktrees must be excluded: camp project
+// worktrees live under projects/worktrees/ and are full source checkouts.
+// Counting them double/triple-counts the same codebase in leverage scores.
 var DefaultExcludeDirs = []string{
 	"node_modules",
 	"vendor",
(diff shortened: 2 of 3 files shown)


docs/leverage-score.md
```diff
@@ -112,9 +112,18 @@ We use it as the baseline because:
...
+  --exclude-dir node_modules --exclude-dir vendor ... \
+  <directory>
 ```
 
+Default `--exclude-dir` values skip dependency/build caches and parallel
+checkouts that would inflate COCOMO estimates: `node_modules`, `vendor`,
+`dist`, `build`, `target`, `worktrees`, `.worktrees`, `.camp-worktrees`,
+and similar. Project worktrees under `projects/worktrees/` are full source
+trees; counting them would multi-count the same codebase in leverage scores.
+Per-project extras can still be set via config `ExcludeDirs`.
```
obey-agent: **Docs accuracy:** `ProjectEntry` has no `ExcludeDirs` field, and `resolveFromConfig` never sets per-project excludes. Only monorepo discovery via `project.List()` populates `ExcludeDirs` (submodule paths). This sentence implies a config knobs that does not exist.

Fix: either drop this sentence, or wire `exclude_dirs` through `ProjectEntry` + `resolveFromConfig` if you want that capability. Prefer fixing the docs in this PR so operators are not sent hunting for a non-existent field.

internal/leverage/scc.go
```diff
@@ -29,6 +33,9 @@ var DefaultExcludeDirs = []string{
 	".mypy_cache",
 	".pytest_cache",
 	".cargo",
+	"worktrees",
```
obey-agent, line 36: **Over-exclusion tradeoff (non-blocking):** scc `--exclude-dir worktrees` matches *any* basename, including real package source at `cmd/camp/worktrees` (~1.2k LOC / ~\$32k COCOMO on current main). That is a ~0.6% undercount for camp vs the multi-x inflation this PR correctly fixes.

Worth a one-line comment here that basename exclusion intentionally prefers correctness against `projects/worktrees/` checkouts over counting the CLI package dir. A path-aware follow-up (`scc -M` / `--not-match` for `(^|/)projects/worktrees(/|$)`) would avoid the false negative if this ever matters more.

internal/leverage/scc_test.go
```diff
@@ -109,6 +109,21 @@ func TestDefaultExcludeDirs_NotEmpty(t *testing.T) {
 	}
 }
 
+func TestDefaultExcludeDirs_IncludesWorktrees(t *testing.T) {
```
obey-agent, line 112: **Test depth:** this only asserts string membership in `DefaultExcludeDirs`, not that scc actually skips a nested `projects/worktrees/...` tree. Acceptable for a list-guard regression, but a small table-driven / temp-dir integration case (source + nested worktree copy; with exclude → only source counted) would lock the original multi-x bug and protect against scc flag semantics changing.

Non-blocking for this PR size.

obey-agent (commented):
## Verdict: Approve

## Overview
Correct, minimal fix for a real measurement bug: project worktrees under `projects/worktrees/` are full source checkouts, and scc was multi-counting them into COCOMO estimates (camp ~101x → ~331x with two review worktrees). Extending `DefaultExcludeDirs` is the right integration point—no parallel pipeline, matches existing monorepo exclude patterns.

## Key Findings
1. **Docs inaccuracy (should fix):** claims per-project config `ExcludeDirs`, but `ProjectEntry` has no such field and `resolveFromConfig` never sets it. See inline on `docs/leverage-score.md`.
2. **Over-exclusion tradeoff (non-blocking):** basename `worktrees` also drops real package code at `cmd/camp/worktrees` (~1.2k LOC). Acceptable vs multi-x inflation; document or path-match later if needed. See inline on `scc.go`.
3. **Test is a list guard only (non-blocking):** does not prove scc skips nested worktree trees. Fine for this PR size; integration case would harden the regression. See inline on `scc_test.go`.
4. **Operational:** existing inflated snapshots/caches need `camp leverage reset` / re-snapshot after deploy—already noted in the PR test plan.

## What's Done Well
- Root-caused the spike to worktree multi-count before coding.
- Small surface area: three exclude names + comment + docs + guard test.
- Comments explain *why* (full checkouts), not just *what*.
- Does not invent a new exclusion mechanism; reuses `mergeExcludeDirs` / `--exclude-dir`.

## Staff Standard
Yes—proud to ship this as-is for the core fix. I'd land the doc sentence correction in a follow-up commit on the same PR if easy; otherwise merge and track path-aware excludes separately.

obey-agent (approved):
## Verdict: Approve

## Overview
Follow-up docs commit on a previously approved fix: `DefaultExcludeDirs` still correctly excludes `worktrees` / `.worktrees` / `.camp-worktrees`, and the docs now accurately describe monorepo root double-count prevention via project-list `ExcludeDirs` (not a vague “per-project extras” note).

## Key Findings
- No new code risk; docs match the real scc invocation and monorepo exclusion path.
- Unit guard for the three worktree exclude names remains in place.

## What's Done Well
Minimal measurement fix + honest docs. Right scope for a leverage-score correctness bug.

## Staff Standard
Yes — small, correct, tested, documented.

PR #401, opened 2026-07-12
camp settings: catalog-driven editor for campaign.yaml, registry, and allowlist

lancekrogers:
## Summary

Extends `camp settings` from a two-field menu (global config plus a local theme
override) into a catalog-driven, path-transparent editor that covers every
settings file that matters:

- **Local** (`.campaign/`): `campaign.yaml` (identity, mission, type, the
  `intents.tags` list, and the nested `concepts` taxonomy), `settings/local.json`
  (theme override), and `settings/allowlist.json` (per-command allowed toggles,
  add, remove).
- **Global** (`~/.obey/campaign/`): `config.json` (the existing five fields) and
  `registry.json` (view registered campaigns and make safe per-campaign edits:
  org, display rename, path repair).

Every screen shows the exact file it edits, in the same form you would type by
hand: campaign-root-relative for local files (`.campaign/campaign.yaml`) and
tilde-based for global files (`~/.obey/campaign/registry.json`), honoring
`$CAMP_REGISTRY_PATH` and `$XDG_CONFIG_HOME`.

## How it is built

The menu is generated from an in-code catalog (`internal/settings`), not
hardcoded rows. Each entry declares a scope, path, format, and edit mode
(structured / read-only / hidden / secret). This makes the surface safe and
extensible by construction:

- Adding a hidden or read-only file is one catalog line, no menu-code change.
- The hidden set is derived from the watcher contract (`internal/contract`), so
  any newly-watched camp-managed file is excluded automatically.
- Secrets (`~/.obey/.env`, agent token files) are hard-coded as never listed and
  never read.

Work landed in five reviewable commits, one per sequence:

1. Catalog data model, builder, and path helpers (`internal/settings`), fully
   unit-tested, no TUI change.
2. Catalog-driven menus, per-file path headers, and an ID-keyed editor router in
   `cmd/camp/settings.go`. The existing config/local editors move behind the
   router unchanged.
3. Interactive `campaign.yaml` editing: a structured form for the scalars, a
   one-per-line `intents.tags` editor, and a `$EDITOR` round-trip for the
   `concepts` tree (validated all-or-nothing, so an invalid edit never touches
   the file). All writes go through `SaveCampaignConfig`.
4. Global registry: name-sorted campaign picker and safe org/name/path edits via
   `UpdateRegistry` (atomic, versioned, UUID-keyed). Path repair only points at
   an existing directory and asks for confirmation. No lifecycle operations
   (register/unregister/switch/transfer) are exposed here; those stay in
   `camp registry` / `camp org`.
5. Structured `allowlist.json` editing (new `internal/config` model plus
   Load/Save mirroring the local.json loader), an additive `settings get`/`set`
   twin for the campaign.yaml scalars, refreshed
   `docs/campaign-settings-files.md`, and capstone guarantee tests.

## Compatibility

- No breaking changes to `camp settings get`/`set`, its `--json` payload, or any
  on-disk schema. The new CLI keys (`local.campaign.name/description/mission/
  type/commit_hook`) are additive; list and tree fields (`intents.tags`,
  `concepts`) and the registry per-campaign edits are TUI-only by design.
- All writes reuse the existing `internal/config` load/save code, so the TUI and
  the CLI twin write identically.

## Testing

- Host unit tests cover all pure logic: catalog build and secret/hidden
  exclusion, path rendering under overrides, menu composition, the campaign
  scalar and tag/concept mapping, the registry option list and path-repair
  classification, the allowlist mutations, the CLI scalar mapping, and the three
  capstone guarantees (a hidden entry needs no menu code; secrets and
  machine-managed files are never editable; the menu only routes
  structured/read-only entries).
- A reliable, non-interactive containerized test writes campaign.yaml scalars
  through `camp settings set local.campaign.*` and reads them back.
- Interactive TTY write tests (campaign.yaml concepts, registry org, allowlist
  toggle) are containerized but gated behind `CAMP_SETTINGS_TTY_TESTS`. They
  pass run in isolation but are flaky under concurrent full-suite load because of
  camp's tracked bubbletea init tty-query stall (termenv blocks up to 5s per
  program init when the pty does not answer the OSC 11 / DSR query fast enough).
  Skipping them by default keeps the gate green without hiding the coverage; a
  proper fix (the fest PR #254 `bginit` seed pattern) is a separately-filed camp
  intent with a background-detection tradeoff that should be decided on its own.
- `just build`, `just lint`, and `just test unit` (2982 tests) are green; the
  full integration suite is green with the TTY tests gated.

## Notes for review

- The five global config fields now live under a `global_config` sub-menu entry
  (one row per file), a deliberate choice to keep the catalog one row per file.
- `notYetEditable` and the three editor stubs were removed once each editor
  became real.


cmd/camp/settings_cli.go
```diff
@@ -259,17 +326,69 @@ func emitSettingsJSON(w io.Writer, cfg *config.GlobalConfig, local *config.Local
...
+	return err
+}
+
+// applyCampaignScalarKey sets a single campaign.yaml scalar and returns the
+// stored value. type is validated against the known campaign types; the others
+// are free text. Only the addressed field changes.
+func applyCampaignScalarKey(cfg *config.CampaignConfig, key, value string) (string, error) {
+	switch key {
+	case settingsKeyLocalCampaignName:
+		cfg.Name = strings.TrimSpace(value)
```
obey-agent, line 373: **Blocking — empty/invalid name can brick campaign load**

`cfg.Name = strings.TrimSpace(value)` accepts empty and invalid names. `SaveCampaignConfig` does not re-validate, but `LoadCampaignConfig` calls `ValidateCampaignConfig`, so `camp settings set local.campaign.name ""` (agent_allowed) bricks subsequent loads.

**Fix:** call `config.ValidateCampaignConfig(cfg)` before save (CLI + TUI). Table-test empty/invalid names.

cmd/camp/settings_campaign.go
```diff
@@ -0,0 +1,281 @@
...
+	if err := tmp.Close(); err != nil {
+		_ = os.Remove(tmp.Name())
+		return "", camperrors.Wrap(err, "closing temp file")
+	}
+	return tmp.Name(), nil
+}
+
+// saveCampaignManifest persists campaign.yaml edits through SaveCampaignConfig,
+// the canonical writer (atomic write, preserved ordering, hooks placeholder).
+func saveCampaignManifest(ctx context.Context, campaignRoot string, cfg *config.CampaignConfig) error {
```
obey-agent, line 277: **Blocking — TUI save path also needs ValidateCampaignConfig**

`saveCampaignManifest` is the shared TUI write path. An empty Name from the form persists and breaks the next load. Validate once here so every editor shares the invariant.

cmd/camp/settings_registry.go
```diff
@@ -0,0 +1,191 @@
...
+
+const (
+	pathUnchanged    pathRepair = iota // candidate equals the current path
+	pathRejected                       // candidate differs but does not exist as a directory
+	pathNeedsConfirm                   // candidate is a different existing directory
+)
+
+// classifyPathRepair decides how to treat a proposed path edit. The existence
+// check is injected so the decision is unit-testable without touching disk.
+func classifyPathRepair(current, candidate string, exists func(string) bool) pathRepair {
```
obey-agent, line 167: **Blocking — registry path repair can persist relative paths**

`classifyPathRepair` / `isExistingDir` use `os.Stat` with no Abs/Clean. A CWD-relative path can be confirmed and written. Consumers expect absolute paths.

**Fix:** Abs+Clean before classify; reject non-absolute after expansion.

cmd/camp/settings_registry.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		return false, err
+	}
+	return ok, nil
+}
+
+// saveRegistryEntry persists a single entry through UpdateRegistry, which holds
+// the registry lock for an atomic load-mutate-save and stamps RegistryVersion.
+// The UUID key and every other entry are left unchanged.
+func saveRegistryEntry(ctx context.Context, uuid string, entry config.RegisteredCampaign) error {
```
obey-agent, line 211: **Blocking — no path-conflict check on registry edit**

`saveRegistryEntry` never checks whether another UUID already owns `entry.Path`. `Registry.Register` enforces uniqueness; settings bypasses that and can corrupt pathIndex/switch.

**Fix:** in mutate, reject path owned by another UUID. Unit-test collision + relative rejection.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview
Strong catalog-driven expansion of `camp settings` (campaign.yaml, registry, allowlist) with real security thinking around Hidden/Secret entries, path-transparent catalog display, and reuse of existing config writers. Architecture and test strategy are staff-level. Two write-path integrity gaps block merge: campaign.yaml can be saved invalid (empty/illegal name) via agent-allowed CLI and TUI, and registry path repair can store relative or colliding paths.

## Key Findings
1. **(inline)** `applyCampaignScalarKey` / `saveCampaignManifest` skip `ValidateCampaignConfig` — empty name bricks load after save.
2. **(inline)** Registry path repair: no Abs/Clean; relative CWD paths can persist.
3. **(inline)** `saveRegistryEntry` skips path-conflict checks that `Register` enforces.

Non-blocking: allowlist lock-then-write vs load-mutate-save; `editEntry` should gate on `Editable()`; catalog entry ID constants; aggregate `settings get` omits campaign scalars; stale docs phrase about config.json.

## What's Done Well
Catalog-as-authority with capstone tests that secrets never list/edit; context + camperrors hygiene; TUI cancel/no-write; modular file split under 500 lines; containerized CLI write tests.

## Staff Standard
Not yet — fix the two must-have integrity paths and this is close to Approve. Design is sound.

obey-agent:
## Addressed review feedback

- **Validate before save** on CLI (`setCampaignScalar`) and TUI (`saveCampaignManifest`) via `ValidateCampaignConfig` — empty/illegal names no longer brick the next load.
- **Registry paths** Abs+Clean via `normalizeRegistryPath` before classify/save; relative CWD forms cannot persist.
- **Path conflicts** rejected in `saveRegistryEntry` (same uniqueness rule as `Registry.Register`).

Tests: invalid name save, normalize absolute/relative, conflict predicate.

obey-agent (approved):
## Verdict: Approve

## Overview
Re-review of head `43940f8`. Follow-up addresses both blocking write-path integrity gaps: every campaign.yaml save (CLI scalar set + TUI `saveCampaignManifest`) now runs `ValidateCampaignConfig` before write, and registry path repair normalizes via Abs+Clean with path-conflict checks matching `Registry.Register`.

## Key Findings
- Empty / illegal names rejected before they can brick `LoadCampaignConfig`; unit tests cover empty and slash-in-name.
- `normalizeRegistryPath` prevents CWD-relative paths from landing in registry.json; relative inputs expand to absolute.
- `saveRegistryEntry` rejects another UUID already owning the path (`ErrPathConflict`).
- Prior catalog/TUI architecture (Hidden/Secret, path transparency, real writers) still holds.

## What's Done Well
Minimal, targeted fix commit that lands validation at the shared save points rather than sprinkling checks in every form. Tests for both negative paths.

## Staff Standard
Yes — ship it.

obey-agent:
Independent merge verification at head cf1986b (prior approved head 43940f8 plus a clean merge of current main): both blocking write-path fixes confirmed in code (ValidateCampaignConfig runs on the CLI scalar-set path at settings_cli.go and the TUI saveCampaignManifest path at settings_campaign.go before any write; normalizeRegistryPath does Clean+Abs+absolute-required, and saveRegistryEntry rejects a path already owned by another UUID with ErrPathConflict, matching Registry.Register). Full local gate PASSED on the merged tree (lint + ratchets clean, all tests green incl. the settings integration suites). Merge-ready.

obey-agent (approved):
## Verdict: Approve

## Overview
Re-review of head `cf1986b` (merge `origin/main` into `camp-settings-tui`). Settings integrity fixes from `43940f8` remain in place: `ValidateCampaignConfig` before CLI/TUI campaign.yaml save, `normalizeRegistryPath` (Abs+Clean), and path-conflict checks in `saveRegistryEntry`. Delta is main integration (orgs, ledger, quests, etc.), not a rewrite of the catalog settings surface.

## Key Findings
- Verified on head: `setCampaignScalar` + `saveCampaignManifest` still validate; registry normalize + `ErrPathConflict` still present.
- No new blockers specific to the settings editor from the merge.

## What's Done Well
Keeps the branch current with main while preserving the write-path invariants that unblocked Approve.

## Staff Standard
Yes — still ship.

PR #402, opened 2026-07-12
Campaign audit trail: ledger core + camp capture (CA0002 part 1)

lancekrogers:
## Campaign audit trail: ledger core + camp capture (festival CA0002, part 1 of several)

This is the first, bounded PR of the campaign-audit-trail festival: the shared event-ledger library plus complete capture inside camp's state-changing commands. It is self-contained and reviewable on its own. fest-side capture, backfill productionization, and timeline/graph ingestion are follow-on PRs of the same festival (see "Follow-on work" below).

### Why

Today a campaign's only audit trail is git commits written through camp/fest wrappers. Wrapper adoption is optional, so the trail is structurally incomplete, and commit-less workflows (media production) get no trail at all. Measured over this campaign's real history (2026-07-11), the untagged-commit rate per repo is:

| repo | tagged | degraded | untagged | untagged % |
|------|-------:|---------:|---------:|-----------:|
| campaign-root | 3100 | 2 | 878 | 22.1% |
| camp | 691 | 0 | 514 | 42.7% |
| fest | 288 | 0 | 702 | 70.9% |
| camp-graph | 113 | 0 | 8 | 6.6% |
| camp-timeline | 15 | 0 | 15 | 50.0% |

One in three commits carries no captured intent linkage, worst exactly where wrappers are used least. This PR moves capture off commit discipline and into the state-changing commands themselves.

### Design (decisions D001 to D003, finalized against real data)

- D001 event envelope: one JSON object per line with `v`, `id`, `ts` (RFC3339 UTC), `kind` (created, transitioned, completed, decided, evidence_attached, reconciled, repaired), `scope` (campaign always; festival/phase/sequence/task, workitem, intent, quest), `action`, `actor` (human/agent/unknown, optional obey session), `why`, free-form `payload`, typed `evidence`, and `source`. Ids are UUIDv7 (already a camp dependency, no new library) for live capture; backfill/reconciled use deterministic content-derived ids so re-runs converge. Fields were validated by replaying 6266 real events, not speculation.
- D002 layout and identity: append-only shards at `.campaign/events/<YYYY-MM>/<writer>.jsonl`. The writer slug is machine-local (never committed), so two machines never touch the same shard and the ledger merges conflict-free. Readers merge-sort by ts then id and dedupe by id. One logical action (a wrapper producing a root commit plus submodule commits) shares one action id.
[...]

PR #404, opened 2026-07-12
camp workitem commits: ledger-backed answering (D007, CA0002)

lancekrogers:
## camp workitem commits: ledger-backed answering (D007 / CA0002)

Part of the obey-campaign audit-trail festival (CA0002), sequence 05. `camp workitem commits` currently fans `git log --grep` across the campaign root and every linked repo. With the campaign event ledger in place it can answer the same query from one merged ledger read.

**Stacked on Obedience-Corp/camp#402** (base branch `campaign-audit-trail`): it imports `pkg/ledgerkit` to read the ledger, so it should merge after #402.

### What this PR does

- Adds a ledger-backed path to `camp workitem commits`. When the ledger holds a workitem's commit evidence (`evidence_attached` / `repaired` events whose `scope.workitem` names the workitem), the answer comes from a single merged `ledgerkit` read instead of a per-repo git fan-out.
- `--source auto|ledger|scan` (default `auto`): `auto` uses the ledger when it has evidence for the workitem and otherwise falls back to the existing scan (pre-ledger history); `scan` forces the exhaustive `git log --all` scan; `ledger` forces the ledger read.
- `scope.workitem` is matched against the workitem's ref, stable id, key, and directory slug, so the three identifier forms all resolve to the one workitem (D007 workitem-id normalization).

### Contract

The `commits` record shape is unchanged: ledger records are built exactly like the scan path (SHA from commit evidence, subject from the event `why`, tag via `commitkit.ParseTag(subject)`), so existing `--json` consumers are unaffected. A top-level `"source"` field (`"ledger"` | `"scan"`) is added additively to report which path answered; the `workitem-commits/v1alpha1` version is unchanged (additive field, alpha contract). jsoncontract does not enforce an output-field allowlist.

### Measurement (real backfilled ledger of the obey-campaign)

Same workitem (`WI-25121c`), same live campaign, warm caches, best of 3:

| path | commits | time |
|------|---------|------|
| `--source ledger` | 13 | 42.5 ms |
| `--source scan` | 20 | 50.8 ms |

The ledger read is faster (one merged read vs a fan-out), and the advantage grows with the number of linked repos. The `commits` count differs because the two views differ, which is a data/backfill characteristic, not a query-design issue (see below).

### Known limitation (follow-up, not a query bug)

The ledger view (13) and the scan view (20) diverge here because:
- `camp audit backfill` derives commit facts with `git log --no-merges` on the default branch, while the scan uses `git log --all` (merge commits and non-default branches included).
- Historical commit subjects carry a doubled workitem tag (`-WI-WI-25121c`); the scan filters on the parsed tag while the ledger matches `scope.workitem`, so a few campaign-root commits appear on one side only.

So `--source auto` (ledger-first) can return a different set than the scan while the ledger is only backfill-populated. For current users no live ledger exists, so `auto` behaves exactly like today (scan). Reaching parity is a backfill concern (scan `--all`, normalize doubled tags) tracked as a follow-up; `--source scan` is the escape hatch to force the exhaustive view meanwhile.

### Tests / gates

`internal/commands/workitem/commits_test.go`: alias collection (ref/id/key/slug), ledger-event -> record mapping (alias filter, repo+sha dedupe, subject/author/date/tag shaping, actor-name author fallback), and the additive `source` field. `just gate-push` green (build, lint ratchets, full test suite).

### DIFF
--- a/internal/commands/workitem/commits.go
+++ b/internal/commands/workitem/commits.go
@@ -152,6 +152,15 @@ func runCommitsQuery(ctx context.Context, cmd *cobra.Command, flags commitsFlags
 			return camperrors.NewValidation("workitem",
 				"workitem has no ref; run `camp workitem doctor --fix` to backfill", nil)
 		}
+	} else {
+		// --ref skips the full resolve for the identity, but still expand D007
+		// aliases (stable id / key / slug) when the ref maps to an on-disk workitem.
+		if res, rerr := resolver.Resolve(ctx, campaignRoot, resolver.Options{Explicit: ref}); rerr == nil && res != nil && res.Workitem != nil {
+			wi = res.Workitem
+			if r := wkitem.RefOf(wi); r != "" {
+				ref = r
+			}
+		}
 	}
 
 	requested := flags.Source
@@ -181,6 +190,9 @@ func runCommitsQuery(ctx context.Context, cmd *cobra.Command, flags commitsFlags
 			records, answered = ledgerRecs, commitsSourceLedger
 		} else if lerr == nil && len(ledgerRecs) > 0 {
 			records, answered = ledgerRecs, commitsSourceLedger
+		} else if lerr != nil {
+			// Empty ledger → silent scan is fine; unreadable/corrupt is loud.
+			_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "warning: campaign ledger unreadable (%v); falling back to scan\n", lerr)
 		}
 	}
 	if answered == commitsSourceScan && requested != commitsSourceLedger {
@@ -205,7 +217,7 @@ func runCommitsQuery(ctx context.Context, cmd *cobra.Command, flags commitsFlags
 	if flags.JSON {
 		return emitCommitsJSON(cmd.OutOrStdout(), source, records, queryErrs)
 	}
-	if err := emitCommitsTable(cmd.OutOrStdout(), records); err != nil {
+	if err := emitCommitsTable(cmd.OutOrStdout(), source, records); err != nil {
 		return err
 	}
 	return emitCommitsQueryWarnings(cmd.ErrOrStderr(), queryErrs)
@@ -277,7 +289,7 @@ func commitsFromLedgerEvents(events []*ledgerkit.Event, aliases map[string]bool)
 			records = append(records, CommitRecord{
 				SHA:      e.SHA,
 				Author:   ledgerCommitAuthor(ev),
-				Date:     parseLedgerTS(ev.TS),
+				Date:     ledgerCommitDate(ev),
 				Subject:  ev.Why,
 				Repo:     e.Repo,
 				TagParts: commitkit.ParseTag(ev.Why),
@@ -294,6 +306,21 @@ func ledgerCommitAuthor(ev *ledgerkit.Event) string {
 		}
 	}
 	return ev.Actor.Name
+}
+
+// ledgerCommitDate prefers payload commit_date/date (live capture + backfill)
+// and falls back to the event timestamp.
+func ledgerCommitDate(ev *ledgerkit.Event) time.Time {
+	if ev.Payload != nil {
+		for _, key := range []string{"commit_date", "date"} {
+			if d, ok := ev.Payload[key].(string); ok && d != "" {
+				if t := parseLedgerTS(d); !t.IsZero() {
+					return t
+				}
+			}
+		}
+	}
+	return parseLedgerTS(ev.TS)
 }
 
 func parseLedgerTS(ts string) time.Time {
@@ -504,7 +531,12 @@ func isGitRepo(ctx context.Context, path string) (bool, error) {
 	return true, nil
 }
 
-func emitCommitsTable(w io.Writer, records []CommitRecord) error {
+func emitCommitsTable(w io.Writer, source string, records []CommitRecord) error {
+	if source != "" {
+		if _, err := fmt.Fprintf(w, "source: %s\n", source); err != nil {
+			return err
+		}
+	}
 	if len(records) == 0 {
 		_, err := fmt.Fprintln(w, "no commits found")
 		return err
--- a/internal/commands/workitem/commits.go
+++ b/internal/commands/workitem/commits.go
@@ -25,6 +25,7 @@ import (
 	"github.com/Obedience-Corp/camp/internal/workitem/links"
 	"github.com/Obedience-Corp/camp/internal/workitem/resolver"
 	"github.com/Obedience-Corp/camp/pkg/commitkit"
+	"github.com/Obedience-Corp/camp/pkg/ledgerkit"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


internal/commands/workitem/commits.go
```diff
@@ -154,14 +203,109 @@ func runCommitsQuery(ctx context.Context, cmd *cobra.Command, flags commitsFlags
...
+			}
+			key := e.Repo + "\x00" + e.SHA
+			if seen[key] {
+				continue
+			}
+			seen[key] = true
+			records = append(records, CommitRecord{
+				SHA:      e.SHA,
+				Author:   ledgerCommitAuthor(ev),
+				Date:     parseLedgerTS(ev.TS),
```
obey-agent: **Contract / correctness:** `Date` is taken from the ledger event \`ts\`, not the underlying commit author-date.

- **Backfill path:** works, because audit backfill sets \`TS\` from the commit date and puts \`author\` in payload.
- **Live path:** \`internal/ledger.CommitEvidence\` only attaches SHA + subject (\`Why\`); emit \`TS\` is wall-clock capture time and there is no author payload. Once live capture is the dominant source, \`camp workitem commits --source ledger\` will sort/display attachment time and fall back to \`Actor.Name\` for author — a silent divergence from the scan path (\`%aI\` / \`%an\`).

Prefer payload fields when present (e.g. \`payload.author\`, \`payload.commit_date\` / \`payload.date\`), and either:
1. extend live \`CommitEvidence\` to record git author + author-date in payload (owned by capture / #402), or
2. document that ledger answers are "evidence view" not "git log view" and keep \`source\` prominent in table mode too.

Without one of those, the additive \`source\` field alone is easy for agents to ignore while consuming \`date\`/\`author\` as if they were git metadata.

internal/commands/workitem/commits.go
```diff
@@ -129,19 +146,51 @@ func runCommitsQuery(ctx context.Context, cmd *cobra.Command, flags commitsFlags
...
+			"must be auto, ledger, or scan", nil)
 	}
 
-	records, queryErrs := searchRepos(ctx, campaignRoot, repos, ref)
+	// Ledger-first (D007/A6): when the campaign event ledger holds this
+	// workitem's commit evidence, answer from one merged ledger read instead of a
+	// git-log fan-out across every linked repo. In auto, an empty or unreadable
+	// ledger falls back to the cross-repo tag scan; --source scan forces the scan
+	// (the exhaustive git --all view) and --source ledger forces the ledger.
+	aliases := workitemAliases(ref, wi)
```
obey-agent, line 180: **D007 normalization gap:** when callers use \`--ref\` (skips resolver), \`wi\` stays nil so \`workitemAliases\` is only the raw ref string.

That undercuts the D007 multi-form matching this PR otherwise implements well. Partial ledger population is the dangerous case: some \`scope.workitem\` values are WI-refs and some are stable ids/slugs. With only the ref in the alias set, \`auto\` can pick the ledger path (non-empty hit) and **silently omit** commits stored under other identifier forms — worse than falling back to scan.

Options:
- Resolve the ref to a workitem when possible (lightweight lookup by ref even on \`--ref\`), then expand aliases.
- Or document that \`--ref\` is ledger-literal matching only and recommend selector resolution when using \`--source auto|ledger\`.

Given agents often pass \`--ref\`, the lightweight resolve is the production-safe default.

internal/commands/workitem/commits.go
```diff
@@ -129,19 +146,51 @@ func runCommitsQuery(ctx context.Context, cmd *cobra.Command, flags commitsFlags
...
+	var queryErrs []commitsQueryError
+	answered := commitsSourceScan
+	if requested == commitsSourceLedger || requested == commitsSourceAuto {
+		ledgerRecs, lerr := ledgerCommits(ctx, campaignRoot, aliases)
+		if requested == commitsSourceLedger {
+			if lerr != nil {
+				return camperrors.Wrap(lerr, "read campaign ledger")
+			}
+			records, answered = ledgerRecs, commitsSourceLedger
+		} else if lerr == nil && len(ledgerRecs) > 0 {
```
obey-agent, line 191: **Observability (minor):** in \`auto\`, any ledger read error (\`lerr != nil\`) is discarded and the code falls through to scan with no stderr signal. Empty ledger → scan is right; **unreadable/corrupt** ledger → silent fallback can hide a doctor-worthy failure while still returning a successful scan answer.

Consider warning on stderr (table) or an optional JSON warning entry when auto falls back because of a read error vs empty results — keeps the non-fatal D003 spirit without making corruption invisible.

obey-agent (changes requested):
## Staff review — Request Changes

### Verdict
**Request Changes** — the routing design is solid, but the ledger → `CommitRecord` mapping has production contract gaps that make `--source auto|ledger` a different semantic answer than scan, beyond the already-documented backfill count skew.

### Overview
This is a clean D007/A6 fast path: pure mapping helpers, additive `source` field, good unit coverage for aliases/dedupe/author fallback/JSON shape, and a sensible `auto → ledger → scan` ladder with `--source` escape hatches. The PR description is unusually honest about ledger-vs-scan count divergence. The remaining issues are about making the ledger answer **honest and complete** once live capture (not just backfill) is the common case.

### Key Findings
1. **Date/author are not git metadata on the live capture path** (`commits.go` ~280, ~290): event `ts` + actor name vs commit author-date/`%an`. Backfill happens to look right because it stamps commit date into `TS` and author into payload; live `CommitEvidence` does not. Agents will treat `date`/`author` as git log fields.
2. **`--ref` skips workitem alias expansion** (~171): D007 multi-form matching only runs when the resolver loads a `WorkItem`. Partial ledger identity forms + non-empty ledger hit = silent incomplete results under `auto`.
3. **`auto` swallows ledger read errors** (~182): empty vs corrupt both become silent scan — fine for missing ledger, less fine for degraded storage.

### What's Done Well
- Extends the existing command rather than a parallel API; record shape preserved for `--json` consumers.
- Pure `commitsFromLedgerEvents` + focused table tests (alias set, dedupe by repo+sha, ignore other workitems, author fallback, additive `source`).
- Explicit `--source auto|ledger|scan` with validation errors via `camperrors`.
- Stacked correctly on ledgerkit; context flows through `Query`.

### Staff Standard
Close — proud of the structure and honesty about limitations, not yet proud to ship as the default `auto` answer until date/author contract and `--ref` alias completeness are addressed (or clearly product-scoped as an evidence view with table-mode `source` surfacing).

Please either fix the mapping/capture contract or narrow the claim that the JSON commit contract is identical across paths.

obey-agent:
## Addressed review feedback

- **Date/author:** prefer payload `commit_date`/`author` (live capture from #402 + backfill); fall back to event `ts` / actor.
- **`--ref` aliases:** lightweight resolver expand for stable id/key/slug when the ref maps to an on-disk workitem.
- **Auto ledger errors:** warn on stderr when unreadable before scan fallback.
- **Table mode:** prints `source: ledger|scan`.

Rebased onto updated #402.

obey-agent:
Independent merge verification at head 8ed685d (rebased onto main after the #402/#403 squash-merges; clean +331/3-file diff):

All three blocking findings from the 06:15 review are verified fixed in code (fix commit was already on the branch; verified line by line post-rebase):
1. Date/author contract: ledgerCommitDate prefers payload commit_date/date then falls back to event ts; with #402's CommitEvidence stamping author-date into ts and author+commit_date into payload, both live and backfill paths carry git metadata. Author mapping falls back payload author -> actor name.
2. --ref now expands D007 aliases: resolver.Resolve(Explicit: ref) loads the on-disk workitem so stable id / key / slug forms all match ledger scope.workitem.
3. auto no longer swallows ledger corruption: unreadable ledger warns to stderr before scan fallback; empty ledger stays silent.

Full local gate on the rebased tree: PASSED (lint + ratchets clean, 3188 + 3159 tests green).

Merge-ready.

PR #409, opened 2026-07-12
fix(project): make camp p commit work from inside a worktree

lancekrogers:
## Summary

`camp p commit` run from **inside a worktree** (`projects/worktrees/<project>/<name>/`)
failed with `project "...worktrees/<project>/<name>" not found`. This makes it
work: it now detects the worktree, commits there, and carries the workitem's
`WI-*` tag - so `camp p commit` is the single commit verb for both a project dir
and its worktrees.

Follow-up to #408 (which added `camp workitem worktree`); the "why" that
prompted it: `camp p commit` couldn't resolve a worktree, so worktree commits
had to go through `camp worktrees commit`.

## Root cause

`project.ResolveFromCwd` maps a cwd to a project only when it is inside a
registered `projects/<name>` dir or a recognized git **submodule** root. A
worktree is a **sibling** of its project under the gitignored `worktrees/` dir,
and its `.git` is a gitdir-pointer file that isn't classified as a submodule, so
all three resolution tiers miss and it falls through to `ProjectNotFoundError`
with the full worktree path as the name.

## Fix

`runProjectCommit`, when given no explicit `--project`, first detects a worktree
cwd via `worktree.Detector` and commits in the worktree working dir. The WI-tag
comes from the existing resolver against that cwd - no new tag logic, because
the workitem resolver is already worktree-aware and the project/worktree
commit-context resolvers were byte-for-byte identical.

- Purely additive: from a project dir, behavior is unchanged; from a worktree,
  what used to error now works.
- Explicit `--project` from a worktree still targets the project's main checkout
  (escape hatch preserved).
- The parent submodule-ref `--sync` is skipped in a worktree (a worktree branch
  has no campaign-root ref to sync; the worktrees dir is gitignored anyway).
- The worktree-created hints (in `camp workitem worktree` and `camp project
  worktree add`) are updated back to `camp p commit`, now that it is the
  universal verb.

## Tests

Containerized integration tests:
- `TestIntegration_ProjectCommit_InWorktreeTagsWI` - commit from inside a linked
  worktree carries `WI-<ref>`.
- `TestIntegration_ProjectCommit_ExplicitProjectFromWorktree` - `--project` from
  a worktree commits in the project main checkout.

Both pass. Verified on a real campaign: subject came out
`[pc-smoke:d006a4be-WI-966439] feat: stub` from inside the worktree, and a
project-dir commit still works unchanged.

## Gates

- `go build ./...`, `go vet ./...`, `go vet -tags integration ./...` - clean
- `golangci-lint --new-from-merge-base origin/main` - 0 new issues
- `lint-no-fmt-errorf` - clean
- `just test integration-run TestIntegration_ProjectCommit_` - 2/2 pass

### DIFF
--- a/cmd/camp/project/commit.go
+++ b/cmd/camp/project/commit.go
@@ -13,8 +13,10 @@ import (
 	"github.com/Obedience-Corp/camp/internal/config"
 	"github.com/Obedience-Corp/camp/internal/git"
 	"github.com/Obedience-Corp/camp/internal/ledger"
+	"github.com/Obedience-Corp/camp/internal/paths"
 	projectsvc "github.com/Obedience-Corp/camp/internal/project"
 	"github.com/Obedience-Corp/camp/internal/ui"
+	"github.com/Obedience-Corp/camp/internal/worktree"
 	"github.com/Obedience-Corp/camp/pkg/commitkit"
 	"github.com/Obedience-Corp/camp/pkg/ledgerkit"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


obey-agent (approved):
## Verdict: Approve

## Overview
Correct, scoped fix for a real worktree UX hole: `camp p commit` from `projects/worktrees/<project>/<name>` now detects the worktree, commits there, carries WI tags via the existing resolve path, and correctly skips parent submodule ref sync. Follow-up to #408 lands the single commit verb the product needs.

## Key Findings
- No blocking issues.
- Worktree detection is gated on empty `--project` + loaded campaign config, preserving the explicit `--project` escape hatch (covered by integration test).
- Parent sync correctly disabled for worktrees (gitignored worktree has no submodule pointer at campaign root).
- Integration tests use the container harness and assert both WI tagging and explicit-project targeting.

## What's Done Well
- Reuses existing `worktree.Detector` instead of a parallel resolver.
- Copy updates point operators at `camp p commit` consistently.
- Integration coverage for the exact failure mode and the escape hatch.

## Staff Standard
Yes — proud to merge. Small optional follow-up (non-blocking): Long help still only describes submodule projects; a one-line worktree example would match the new behavior.

PR #416, opened 2026-07-13
fix(settings): ESC back-nav, in-TUI concepts, clarify daemon allowlist

lancekrogers:
## Summary

Fixes several `camp settings` TUI pain points:

1. **ESC / back navigation** — nested Select menus now treat empty choice as Back (same as the allowlist/registry screens), so ESC leaves one menu level instead of no-op looping.
2. **Concepts taxonomy hang** — stop shelling out to `$EDITOR` for concepts. Edit as YAML inside the TUI; invalid YAML still never writes `campaign.yaml`. (External GUI editors without proper wait were hanging the session.)
3. **Command allowlist confusion** — rename/describe it as **Daemon command allowlist**. Camp CLI does not enforce this file; it is for obey-daemon/agent tool permissions.
4. **Editor wait list** — add `cursor`, `zed`, `zeditor`, `windsurf`, `codium`, `code-oss` so other call sites that still use `$EDITOR` block correctly.

## Test plan

- [x] `go test ./cmd/camp/ ./internal/settings/ ./internal/editor/`
- [ ] Manual: `camp settings` → Local → Campaign manifest → ESC should return one level at a time
- [ ] Manual: Concepts taxonomy edits YAML in-TUI and saves/rejects invalid input
- [ ] Manual: Allowlist row shows daemon-oriented description

WI-333b2e

### DIFF
--- a/cmd/camp/settings_campaign.go
+++ b/cmd/camp/settings_campaign.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"os"
 	"strings"
 
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
@@ -11,7 +10,6 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/Obedience-Corp/camp/internal/config"
-	"github.com/Obedience-Corp/camp/internal/editor"
 	"github.com/Obedience-Corp/camp/internal/settings"
 	"github.com/Obedience-Corp/camp/internal/ui"
 	"github.com/Obedience-Corp/camp/internal/ui/theme"
@@ -26,7 +24,7 @@ func editCampaignManifest(ctx context.Context, e settings.SettingEntry, campaign
 		options := []huh.Option[string]{
 			huh.NewOption("Identity, mission, and type", "scalars"),
 			huh.NewOption("Intent tags", "tags"),
-			huh.NewOption("Concepts taxonomy (opens editor)", "concepts"),
+			huh.NewOption("Concepts taxonomy", "concepts"),
 			huh.NewOption(rowSeparator, valSeparator),
 			huh.NewOption("Back", valBack),
 		}
@@ -48,7 +46,7 @@ func editCampaignManifest(ctx context.Context, e settings.SettingEntry, campaign
 		}
 
 		switch choice {
-		case valBack:
+		case valBack, "":
 			return nil
 		case valSeparator:
 			continue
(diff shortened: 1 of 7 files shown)


lancekrogers:
Superseded by festival-linked PR https://github.com/Obedience-Corp/camp/pull/418 (CS0004 seq 06 on branch `camp-settings-tui`). Closing in favor of the festival worktree.

lancekrogers:
Closed as superseded by #418.

cmd/camp/settings_campaign.go
```diff
@@ -219,24 +214,27 @@ func editConceptsViaEditor(ctx context.Context, e settings.SettingEntry, campaig
...
-	defer func() { _ = os.Remove(tmpPath) }()
 
-	if err := editor.Edit(ctx, tmpPath); err != nil {
-		return camperrors.Wrap(err, "running editor")
-	}
+	form := huh.NewForm(huh.NewGroup(
+		huh.NewText().
+			Title("Concepts taxonomy (YAML)").
+			Description("File: " + settings.CatalogPath(e, campaignRoot) + "\n" + conceptsEditorHelp).
+			Value(&body),
```
obey-agent, line 226: Blocking: `huh.NewText()` enables its external-editor shortcut by default. In huh v0.8.0, Ctrl+E invokes `$EDITOR` directly via `tea.ExecProcess`; it does not use `internal/editor` or add `--wait`. That means this supposedly in-TUI screen still advertises an editor escape hatch that can reproduce the GUI-editor no-op/hang this PR is intended to remove. Please add `.ExternalEditor(false)` to this field (and cover the behavior) so concepts editing is genuinely TUI-only.

obey-agent (commented):
Verdict: Request Changes

Overview: This is a focused, coherent UX cleanup. The ESC handling, daemon-allowlist wording, and all-or-nothing YAML persistence fit the existing settings architecture, and the updated container TTY round-trip passes. One external-editor path remains, however, so the central hang-prevention goal is not fully closed.

Key Findings:
- Blocking: the concepts `huh.NewText()` field retains huh v0.8.0’s default Ctrl+E external-editor action. That action invokes `$EDITOR` directly and bypasses camp’s GUI `--wait` handling, so it can recreate the behavior this PR is intended to eliminate. See the inline comment; disable it with `.ExternalEditor(false)` and cover the behavior.
- The new GUI editor aliases are straightforward, but adding them to the existing `TestBuildEditorCommand` table would protect the exact compatibility behavior introduced here.

What’s Done Well:
- Invalid YAML and invalid concept trees remain all-or-nothing and preserve the canonical save path.
- The allowlist ownership clarification is consistent across the catalog, TUI, and docs.
- Validation was clean: `just lint`, all 3,271 unit tests, the targeted package tests, and `TestIntegration_SettingsConceptsEditorRoundTrip` passed.

Staff Standard: Not yet. I would be proud to submit this after closing the remaining external-editor escape hatch and adding the small regression coverage.

PR #417, opened 2026-07-13
feat(settings): global/local commit ref-sync and tag-tracing prefs

lancekrogers:
## Summary

Adds the missing **global + per-campaign** settings for two commit behaviors:

1. **Project ↔ campaign ref linking** (`commit.sync_project_refs`)
   - When true, `camp project commit` updates the campaign-root submodule pointer (same as `--sync`)
   - Default remains **false** (opt-in)
   - New **`--no-sync`** forces off even when the setting is enabled

2. **Commit tracing / subject tags** (`commit.disable_commit_tags`)
   - When true, skips `[campaign:id-…]` prefixes on camp-managed commits
   - Default remains **false** (tags on)
   - Applies to `camp commit`, `camp project commit`, worktree commit, and `refs-sync`

### Storage

| Scope | File |
|-------|------|
| Global | `~/.obey/campaign/config.json` → `commit` |
| Local (overrides global when present) | `.campaign/settings/local.json` → `commit` |

### Access

```bash
camp settings get effective.commit.sync_project_refs
camp settings set global.commit.disable_commit_tags true
camp settings set local.commit.sync_project_refs true
```

Also toggles in `camp settings` → Global config / Local settings.

## Test plan

- [x] `go test ./internal/config/ ./cmd/camp/ ./cmd/camp/project/`
- [ ] Manual: set local disable tags, commit, confirm plain subject
- [ ] Manual: set sync_project_refs true, `camp p commit -m "…"` without `--sync`, confirm campaign root updates
- [ ] Manual: `--no-sync` overrides setting

WI-a05044

PR #418, opened 2026-07-13
feat(settings): CS0004 post-ship UX and commit prefs

lancekrogers:
## Summary

Festival **CS0004** sequence `06_post_ship_ux_and_commit_settings` — post-ship hardening and missing commit settings for `camp settings`.

### UX / reliability
- ESC / empty Select choice returns **one menu level** on nested settings screens
- Concepts taxonomy: **in-TUI YAML** (no `$EDITOR` hang)
- Allowlist renamed/described as **daemon/agent** tool permissions (not camp CLI)
- GUI editor wait list expanded (cursor, zed, windsurf, …) for remaining editor call sites

### Commit prefs (global + local)
- `commit.sync_project_refs` — default for project→campaign root submodule pointer sync after `camp p commit` (still default off; `--sync` / `--no-sync`)
- `commit.disable_commit_tags` — optional disable of `[campaign:…]` subject tracing
- Local `local.json` `commit` block full-replaces global when present
- TUI toggles + `camp settings get`/`set` keys + docs

### Festival
- Linked worktree: `projects/worktrees/camp/camp-settings-tui`
- Supersedes free-standing PRs #416 and #417 (same commits adopted here)

## Test plan

- [x] `go test ./internal/config/ ./cmd/camp/ ./cmd/camp/project/ ./cmd/camp/worktrees/ ./internal/settings/ ./internal/editor/`
- [ ] Manual: Esc through nested settings menus
- [ ] Manual: concepts YAML edit save/reject
- [ ] Manual: `camp settings set local.commit.disable_commit_tags true` then commit subject has no tag

CS0004 · 06_post_ship_ux_and_commit_settings


cmd/camp/settings_campaign.go
```diff
@@ -219,24 +214,27 @@ func editConceptsViaEditor(ctx context.Context, e settings.SettingEntry, campaig
...
-	defer func() { _ = os.Remove(tmpPath) }()
 
-	if err := editor.Edit(ctx, tmpPath); err != nil {
-		return camperrors.Wrap(err, "running editor")
-	}
+	form := huh.NewForm(huh.NewGroup(
+		huh.NewText().
+			Title("Concepts taxonomy (YAML)").
+			Description("File: " + settings.CatalogPath(e, campaignRoot) + "\n" + conceptsEditorHelp).
+			Value(&body),
```
obey-agent, line 231: Blocking: `huh.NewText()` enables Ctrl+E external editing by default. In huh v0.8.0 that shortcut invokes `$EDITOR` directly via `tea.ExecProcess`, bypassing camp’s `internal/editor` wait handling. The TTY test output still renders `ctrl+e open editor`, so this consolidated PR retains the same escape hatch that can recreate the GUI-editor no-op/hang it claims to remove. Add `.ExternalEditor(false)` here and cover the absence of that path.

internal/config/commit_prefs.go
```diff
@@ -0,0 +1,45 @@
...
+// global commit block for that campaign.
+func EffectiveCommitPrefs(ctx context.Context, campaignRoot string) CommitPrefs {
+	var prefs CommitPrefs
+	if g, err := LoadGlobalConfig(ctx); err == nil && g != nil {
+		prefs = g.Commit
+	}
+	if campaignRoot == "" {
+		return prefs
+	}
+	if loc, err := LoadLocalSettings(ctx, campaignRoot); err == nil && loc != nil && loc.Commit != nil {
```
obey-agent: Blocking safety issue: both config-load errors are silently converted into a different commit policy. In particular, if global `SyncProjectRefs` is true and a campaign-local `local.json` is unreadable or malformed, this returns the global value and `camp p commit` may unexpectedly create a campaign-root pointer commit. Please make policy resolution return/propagate an error (or explicitly fail closed for mutating behavior) instead of treating configuration corruption as “no local override.” Add malformed/unreadable local-config coverage.

internal/config/commit_prefs_test.go
```diff
@@ -0,0 +1,96 @@
...
+		t.Fatal("default CommitPrefs should enable tags")
+	}
+	p.DisableCommitTags = true
+	if p.TagCommits() {
+		t.Fatal("DisableCommitTags should disable tags")
+	}
+}
+
+func TestEffectiveCommitPrefs_GlobalOnly(t *testing.T) {
+	dir := t.TempDir()
```
obey-agent: These new tests write configuration and create directories directly in host `t.TempDir()` paths. The repository’s filesystem-safety contract requires tests that create/write/move/delete real files to run through the containerized integration harness. Please migrate the three filesystem-backed commit-pref tests to `tests/integration`; keep only pure `CommitPrefs` logic in the host unit test.

cmd/camp/settings_cli.go
```diff
@@ -298,12 +371,32 @@ func printSettingsAll(w io.Writer, cfg *config.GlobalConfig, local *config.Local
...
+		if _, err := fmt.Fprintf(w, "%s = %v\n", settingsKeyLocalCommitSyncRefs, locSync); err != nil {
+			return err
+		}
+		if _, err := fmt.Fprintf(w, "%s = %v\n", settingsKeyLocalCommitDisableTags, locTags); err != nil {
+			return err
+		}
+	}
+	// printSettingsAll is pure formatting; effective prefs are already merged
+	// via EffectiveCommitPrefs using campaignRoot (empty when outside a campaign).
+	prefs := config.EffectiveCommitPrefs(context.TODO(), campaignRoot)
```
obey-agent: Please do not re-enter I/O with `context.TODO()` from a formatting helper. This drops caller cancellation and re-reads config after `runSettingsGet` already loaded its snapshot, so one response can mix values from different reads. Compute effective commit prefs once with the command context in `runSettingsGet` and pass the value into both text and JSON formatters.

tests/integration/settings_campaign_test.go
```diff
@@ -39,36 +35,21 @@ func TestIntegration_SettingsConceptsEditorRoundTrip(t *testing.T) {
...
-		{WaitFor: "Files under .campaign/", Input: "\r"},           // local menu: Campaign manifest (first row), enter
-		{WaitFor: "Concepts taxonomy", Input: "\x1b[B\x1b[B\r"},    // manifest menu: down to Concepts, enter
-		{WaitFor: "EDITOR_DONE"},                                   // scripted editor rewrote the temp file
-		{WaitFor: "Concepts taxonomy", Input: "\x03"},              // back at manifest menu, abort
-		{WaitFor: "Files under .campaign/", Input: "\x03"},         // back at local menu, abort
-		{WaitFor: "Select configuration scope", Input: "\x03"},     // back at top menu, abort -> exit
+		{WaitFor: "Select configuration scope", Input: "\x1b[B\r"}, // top: Local
+		{WaitFor: "Files under .campaign/", Input: "\r"},           // Campaign manifest
+		{WaitFor: "Concepts taxonomy", Input: "\x1b[B\x1b[B\r"},    // Concepts row
+		{WaitFor: "Concepts taxonomy (YAML)", Input: "\x01\x0b" + conceptYAML + "\r"}, // clear-ish + paste + submit
```
obey-agent: This opt-in integration test is not currently reliable: I ran it twice with `CAMP_SETTINGS_TTY_TESTS=1 -count=1` (including once without another test workload) and both runs reached the top-level menu but exceeded the 15s interactive deadline before exit. The per-character multiline input/rendering consumes most of the budget. Please make the harness input deterministic/fast enough or adjust the per-session deadline, then demonstrate a passing isolated run. Also update the comments: the sequence sends neither Ctrl+D nor a true whole-buffer replacement—the resulting manifest retained seeded concepts and appended the new entry.

obey-agent (commented):
Verdict: Request Changes

Overview: Consolidating #416 and #417 into the festival-linked branch is sensible, and the settings surface is internally consistent across TUI, CLI, config, and commit call sites. The current implementation still has production-safety and verification gaps that should be closed before merge.

Key Findings:
- Blocking: concepts `huh.NewText()` still exposes Ctrl+E external editing and bypasses camp’s GUI wait handling, preserving the hang/no-op escape hatch this change is intended to eliminate.
- Blocking: `EffectiveCommitPrefs` silently falls back when global/local config cannot be read. With global ref sync enabled, malformed local config can unexpectedly turn a project commit into a campaign-root pointer commit.
- The new host-filesystem-mutating unit tests violate the repository’s container-test safety contract and should move to the integration harness.
- Settings output re-reads config with `context.TODO()`, dropping cancellation and allowing internally inconsistent snapshots.
- The opt-in concepts TTY integration test failed twice from a 15s deadline after reaching the final menu; its comments also do not match the actual input/replacement behavior.

What’s Done Well:
- The global/local full-replacement semantics are documented and represented consistently in the payloads.
- Tag suppression is threaded through root, project, worktree, and refs-sync commit paths.
- Targeted package tests and `just lint` pass; diff whitespace checks are clean.

Staff Standard: Not yet. I would be comfortable submitting this after policy resolution fails safely, the editor escape hatch is removed, and the tests are made compliant and reliable.

cmd/camp/settings.go
```diff
@@ -285,10 +299,22 @@ func editLocalSettingsFile(ctx context.Context, e settings.SettingEntry, campaig
 			return camperrors.Wrap(err, "loading local settings")
 		}
 
+		locSync, locTags := false, false
```
obey-agent: **[bug]** Unset local commit prefs display as `false`, and toggle flips effective value.

When `local.Commit == nil`, the TUI seeds `locSync, locTags := false, false` and renders e.g. `Sync project refs  false (effective: true)` if global sync is enabled. `toggleLocalCommitPref` flips the *effective* prefs (or existing local block), not the displayed “false.” So with global `sync_project_refs=true` and no local block, a single Enter on the row that shows `false` materializes local `false` and turns effective *off* — the opposite of “flip the false I see to true.” This only appears when global diverges from zero defaults, which is exactly when operators open Local settings to inspect overrides.

**Suggestion:** Mirror theme UX: when `local.Commit == nil`, show `inherit` / `—` (not `false`) for the stored column. Prefer an explicit set form (true/false/inherit) over blind toggle-on-unset. If toggle is kept, it must operate on a value consistent with the primary displayed state.

cmd/camp/settings.go
```diff
@@ -310,18 +336,49 @@ func editLocalSettingsFile(ctx context.Context, e settings.SettingEntry, campaig
...
+				return err
+			}
 		}
 	}
 }
 
+// toggleLocalCommitPref flips one local commit preference. When local.Commit is
+// nil, seed it from the current effective prefs so the other field keeps its
+// global meaning until explicitly changed.
+func toggleLocalCommitPref(ctx context.Context, campaignRoot string, syncRefs bool) error {
```
obey-agent: **[bug]** Same toggle path: when local is nil, flips effective then freezes full-replace.

`toggleLocalCommitPref` seeds from `EffectiveCommitPrefs` then writes a full local block. Combined with the `false` display above, the first edit freezes both fields under full-replace merge forever (sibling field snapshot of global). Theme has an `inherit` path; commit prefs do not from TUI.

**Suggestion:** Show inherit when unset; add “Reset commit prefs to inherit global” that sets `s.Commit = nil`.

cmd/camp/settings_cli.go
```diff
@@ -254,6 +308,26 @@ func settingsValueFor(key string, cfg *config.GlobalConfig, local *config.LocalS
...
+	case settingsKeyGlobalCommitSyncRefs:
+		return cfg.Commit.SyncProjectRefs, nil
+	case settingsKeyGlobalCommitDisableTags:
+		return cfg.Commit.DisableCommitTags, nil
+	case settingsKeyLocalCommitSyncRefs:
+		if !inCampaign {
+			return nil, errSettingsLocalOutsideCampaign(key)
+		}
+		if local.Commit == nil {
+			return false, nil
```
obey-agent: **[bug]** CLI `get local.commit.*` returns `false` when the local commit block is unset.

`settings get local.commit.sync_project_refs` returns `false` when `local.Commit == nil` (and the same for disable tags). That is indistinguishable from an explicit local override to `false`, and contradicts JSON `get --json`, which correctly omits `local.commit` when unset. Agents and humans reading the text surface will mis-model full-replace state.

**Suggestion:** For unset local commit block, return empty / `null` / a sentinel like `inherit` (and document it), not zero bools. Align text listing with JSON omit semantics.

cmd/camp/settings_cli.go
```diff
@@ -482,6 +591,40 @@ func setLocalThemeOverride(ctx context.Context, cmd *cobra.Command, value string
 	return err
 }
 
+func setLocalCommitPref(ctx context.Context, cmd *cobra.Command, key, value string) error {
```
obey-agent, line 597: **[bug]** No inherit/clear path for local commit prefs.

`setLocalCommitPref` always materializes a full `local.Commit` block (seeded from effective then one field set). There is no `inherit` / clear analogous to `local.theme_override inherit`. Combined with full-replace merge, the first `set` freezes the sibling field to a snapshot of global forever; later global changes to the other flag no longer apply, and recovery requires hand-editing `local.json`.

**Suggestion:** Support `camp settings set local.commit.* inherit` (or a dedicated clear) that sets `s.Commit = nil` when both sides should inherit. Document that first local write freezes both fields under full-replace.

cmd/camp/settings_cli.go
```diff
@@ -121,17 +148,23 @@ Keys:
   global.campaigns_dir       Where camp create places new campaigns
   global.verbose             Verbose output
   global.no_color            Disable colored output
+  global.commit.sync_project_refs   Default: camp p commit updates campaign-root submodule pointer
```
obey-agent: **[suggestion]** Help text “Default:” inverts the real product defaults.

`Default: camp p commit updates campaign-root submodule pointer` and `Default: skip [campaign:…] tags` read as describing default runtime behavior. Actual defaults are the opposite: `sync_project_refs` default false (opt-in), `disable_commit_tags` default false (tags **enabled**).

**Suggestion:** Reword to `When true, camp p commit updates… (default false)` and `When true, skip [campaign:…] tags (default false; tags on)`.

obey-agent (commented):
## Verdict: Request Changes

## Overview
CS0004 post-ship UX + commit prefs is the right scope: ESC/back on nested settings, in-TUI concepts YAML (no `$EDITOR` hang), accurate daemon allowlist rename, expanded GUI editor wait list, and `commit.sync_project_refs` / `commit.disable_commit_tags` wired through global/local settings, CLI get/set, and all four camp commit tag paths plus project `--sync`/`--no-sync`. Prior fail-closed and containerized prefs work on this branch looks solid.

Ship-blockers are local-override UX vs full-replace merge: unset local commit prefs still present as `false`, TUI toggle can invert intent when global ≠ defaults, and there is no inherit/clear path analogous to theme.

## Key Findings
1. **[bug] Unset local shown as `false` + toggle flips effective** (`settings.go:302`, `362`) — with global sync on and no local block, Enter on “false” turns effective *off*.
2. **[bug] CLI get same ambiguity** (`settings_cli.go:320`) — `local.commit.*` returns `false` when unset; JSON correctly omits the block.
3. **[bug] No inherit/clear** (`settings_cli.go:594`) — first set freezes full-replace forever; only hand-edit recovers.
4. **[suggestion]** Help “Default:” wording inverts real defaults (`settings_cli.go:151`).
5. **[suggestion]** Add E2E container tests that prefs change actual commit subjects / parent-ref sync (currently only settings resolution + fail-closed).

## What's Done Well
- Fail-closed `EffectiveCommitPrefs` (corrupt local never silently inherits global sync)
- Explicit full-replace merge model, documented and unit-tested
- Defaults match intent (sync opt-in; tags on)
- All four commit tag paths + `--sync`/`--no-sync` escape hatches
- Concepts in-TUI + ExternalEditor(false); allowlist rename accuracy; GUI wait list expansion

## Staff Standard
Not yet — fix display/get/clear for local commit overrides (Issues 1–3). After that (and ideally one E2E commit-behavior integration test), this is proud-to-ship.

cmd/camp/settings.go
```diff
@@ -285,11 +298,26 @@ func editLocalSettingsFile(ctx context.Context, e settings.SettingEntry, campaig
 			return camperrors.Wrap(err, "loading local settings")
 		}
 
+		// Unset local commit block shows "inherit", not false — false would
+		// imply an explicit override and make toggle invert effective prefs.
+		locSync, locTags := "inherit", "inherit"
```
obey-agent, line 303: This is the right UX for a full-replace local block: stored state is shown as inherit when absent, effective state remains visible separately, and the edit form makes the whole-block clear semantics explicit instead of freezing an unseen sibling value.

cmd/camp/settings_cli_test.go
```diff
@@ -388,3 +388,49 @@ func TestSettingsSubcommandsRegistered(t *testing.T) {
 		}
 	}
 }
+
+func TestSettingsLocalCommit_UnsetIsInheritAndClear(t *testing.T) {
+	setupSettingsTest(t, true)
```
obey-agent: Blocking under the campaign filesystem-safety standard: this new regression calls setupSettingsTest and then executes settings set/get commands that create campaign directories and write global/local config files in host TempDir state. The earlier review moved filesystem-backed commit-pref coverage to tests/integration; this case belongs beside it in the container harness. Keep a host unit only if it exercises pure settingsValueFor or merge logic without disk mutation.

obey-agent (commented):
Verdict: Request Changes

Overview: The current head correctly addresses the prior inheritance and full-replace UX findings. Unset values, explicit false, effective values, and whole-block clear behavior are now distinguishable and documented; the only remaining blocker is that the new regression reintroduces filesystem-backed coverage in the host unit suite.

Key Findings:
- Blocking: move TestSettingsLocalCommit_UnsetIsInheritAndClear to the existing containerized settings commit-preferences integration coverage.
- Previous findings on Ctrl+E, config error propagation, formatter context, TTY reliability, default help text, and local inherit semantics are resolved.

What’s Done Well:
- TUI and CLI now tell the same stored-versus-effective story.
- The inherit action accurately communicates that full-replace semantics clear both fields.
- Existing filesystem-backed CommitPrefs tests were correctly reduced to pure logic and migrated to integration.

Staff Standard: Almost. The implementation is at staff quality; I would move the last disk-mutating regression into the required harness before submitting it as my own work.

cmd/camp/settings_campaign.go
```diff
@@ -219,24 +214,32 @@ func editConceptsViaEditor(ctx context.Context, e settings.SettingEntry, campaig
...
-	}
+	form := huh.NewForm(huh.NewGroup(
+		huh.NewText().
+			Title("Concepts taxonomy (YAML)").
+			Description("File: " + settings.CatalogPath(e, campaignRoot) + "\n" + conceptsEditorHelp).
+			// Disable huh's Ctrl+E external-editor shortcut: it invokes $EDITOR
+			// directly via tea.ExecProcess, bypassing camp's GUI-editor wait
+			// handling and reintroducing the hang/no-op this in-TUI editor
+			// exists to eliminate.
+			ExternalEditor(false).
```
obey-agent, line 230: Nice fix. While you are here: the same Ctrl+E to `$EDITOR` escape (the hang/no-op you are eliminating) still lives on three sibling `huh.NewText()` fields in this same editor - Description (:108), Mission (:109), and Intent tags (:157). They default to `externalEditor=true` in huh v0.8.0. Not a regression (they predate this PR) so it does not block, but adding `.ExternalEditor(false)` to each would finish the job one menu over.

obey-agent (commented):
## Re-review at f84bbd75 - Verdict: Comment (merge-ready)

Re-reviewed against the three prior obey-agent reviews. Every prior blocking finding is resolved at this HEAD, verified against source:

- `EffectiveCommitPrefs` now fails closed on unreadable/malformed config (internal/config/commit_prefs.go:54-72): a corrupt `local.json` can no longer inherit global `sync_project_refs` and silently turn a project commit into a campaign-root pointer commit. It returns zero-value prefs (no sync, tags on) plus the wrapped error.
- The disk-mutating regression genuinely moved to the container harness: `TestSettingsLocalCommit_UnsetIsInheritAndClear` is gone from the unit suite and now lives in `tests/integration/settings_commit_prefs_test.go` under `//go:build integration`, alongside a new `MalformedLocalFailsClosed` case.
- Concepts Ctrl+E escape hatch closed via `.ExternalEditor(false)` (settings_campaign.go:230); the formatter no longer re-reads config with `context.TODO()`; stored-vs-effective inherit/clear is consistent across the TUI, CLI get/set, and the effective view; help "Default:" text corrected.

Both ratchets pass (`lint-no-fmt-errorf` clean via camperrors, `lint-no-host-fs-tests` clean). Behavior traces confirm `set local.commit.disable_commit_tags true` drops the `[campaign:...]` subject tag across the root, project, worktree, and refs-sync commit paths, and that a local block full-replaces global only when present.

One non-blocking follow-up (inline): the same Ctrl+E hang you just fixed on the concepts field still lives on three sibling `huh.NewText()` fields in the campaign-manifest editor. Pre-existing, not a regression, one line each.

Not posting an Approve; leaving the merge call to you. From my side this is proud-to-ship.

PR #419, opened 2026-07-13
feat: add filtered workitem list command

lancekrogers:
## Summary

- add `camp workitem list [type|status|category]`
- open the workitem TUI with visible, editable prefilters when stdout is a terminal
- emit compact grouped text outside a TTY and preserve `workitems/v1alpha8` with `--json`
- add unified displayed-status filtering across lifecycle and attention stages
- retain compatibility with `camp workitem --list`

## Why

Users currently have to reopen the workitem dashboard and reapply common filters. Scripts and agents can use root-level flags, but there is no list-shaped command that gives humans and automation the same filter vocabulary and environment-appropriate output.

The new command supports positional type/status/category shortcuts plus explicit repeatable filters for type, displayed status, category, lifecycle stage, attention stage, and group. Ambiguous positional values return guidance to use an explicit flag.

## User impact

- `camp workitem list intent` opens a prefiltered dashboard in a terminal and produces compact text when piped.
- `camp workitem list active --json` returns lifecycle-active and attention-active work using the existing JSON schema.
- TUI filters can be changed or cleared without rediscovering or permanently narrowing the candidate set.

## Validation

- `just gate-fast` — 3,305 tests
- `go test ./...`
- `go vet -tags=integration ./...`
- containerized `TestIntegration_WorkitemListFiltersNonTTYAndJSON`
- command smokes for compact output, JSON output, and positional ambiguity

### DIFF
--- a/internal/commands/workitem/list.go
+++ b/internal/commands/workitem/list.go
@@ -1,0 +1,170 @@
+package workitem
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/jsoncontract"
+	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
+)
+
+type listOptions struct {
+	json            bool
+	types           []string
+	categories      []string
+	statuses        []string
+	stages          []string
+	attentionStages []string
+	groups          []string
+	groupBy         string
+	showParked      bool
+	limit           int
+	query           string
+}
+
+func (o listOptions) filterOptions() wkitem.FilterOptions {
+	return wkitem.FilterOptions{
+		Types:           o.types,
+		Categories:      o.categories,
+		Statuses:        o.statuses,
+		LifecycleStages: o.stages,
+		AttentionStages: o.attentionStages,
+		Groups:          o.groups,
+		Query:           o.query,
+		ShowParked:      o.showParked,
+	}
+}
+
+func newListCommand() *cobra.Command {
+	var opts listOptions
+	cmd := &cobra.Command{
+		Use:   "list [type|status|category]",
+		Short: "List or browse filtered workitems",
+		Long: `List campaign workitems with the same filters used by the dashboard.
+
+In a terminal, this opens the TUI with visible, editable prefilters. When
+stdout is not a terminal, it prints a compact grouped list. Use --json for the
+stable machine-readable contract in either environment.
+
+The optional positional filter resolves as a workflow type, displayed status,
+or configured category. Ambiguous values must use an explicit flag.
+
+Examples:
+  camp workitem list intent
+  camp workitem list active
+  camp workitem list --category research --query auth
+  camp workitem list festival --status ready --json`,
+		Args: cobra.MaximumNArgs(1),
+		Annotations: map[string]string{
+			"agent_allowed": "true",
+			"agent_reason":  "Non-TTY compact output and --json are safe for scripts and agents",
+		},
+		RunE: jsoncontract.RunE(wkitem.SchemaVersion, func() bool { return opts.json }, func(cmd *cobra.Command, args []string) error {
+			state, err := discoverWorkitems(cmd.Context())
+			if err != nil {
+				return err
+			}
+			if len(args) == 1 {
+				if err := applyPositionalFilter(args[0], state, &opts); err != nil {
+					return err
+				}
+			}
+			if err := validateFlags(opts.json, false, false, "", opts.types, opts.categories, opts.stages, opts.attentionStages, opts.groups, opts.groupBy); err != nil {
+				return err
+			}
+			if err := validateDisplayStatuses(opts.statuses); err != nil {
+				return err
+			}
+			filters := opts.filterOptions()
+			if isInteractive() && !opts.json {
+				return runTUIWithFilters(cmd.Context(), state, filters, opts.limit)
+			}
+
+			items := wkitem.FilterAdvanced(state.items, filters)
+			if opts.limit > 0 && len(items) > opts.limit {
+				items = items[:opts.limit]
+			}
+			groupBy := opts.groupBy
+			if groupBy == "" {
+				if opts.json {
+					groupBy = "attention_stage"
+				} else {
+					groupBy = "group"
+				}
+			}
+			if opts.json {
+				return outputJSON(state.campaignRoot, state.cfg, items, groupBy)
+			}
+			return outputList(cmd.OutOrStdout(), items, groupBy)
+		}),
+	}
+	cmd.SetFlagErrorFunc(jsoncontract.FlagErrorFunc(wkitem.SchemaVersion, func() bool { return opts.json }))
+
+	cmd.Flags().BoolVar(&opts.json, "json", false, "Output as JSON")
+	cmd.Flags().StringArrayVar(&opts.types, "type", nil, "Filter by workflow type (repeat for OR)")
+	cmd.Flags().StringArrayVar(&opts.categories, "category", nil, "Filter by workflow category (repeat for OR)")
+	cmd.Flags().StringArrayVar(&opts.statuses, "status", nil, "Filter by displayed status (repeat for OR)")
+	cmd.Flags().StringArrayVar(&opts.stages, "stage", nil, "Filter by lifecycle stage (repeat for OR)")
+	cmd.Flags().StringArrayVar(&opts.attentionStages, "attention-stage", nil, "Filter by attention stage (repeat for OR)")
+	cmd.Flags().StringArrayVar(&opts.groups, "group", nil, "Filter by workitem group (repeat for OR)")
+	cmd.Flags().StringVar(&opts.groupBy, "group-by", "", "Group output sections by attention_stage, group, type, or category")
+	cmd.Flags().BoolVar(&opts.showParked, "show-parked", false, "Include parked attention-stage workitems")
+	cmd.Flags().IntVar(&opts.limit, "limit", 0, "Maximum number of items to return")
+	cmd.Flags().StringVar(&opts.query, "query", "", "Search query to filter items")
+	return cmd
+}
+
+func applyPositionalFilter(raw string, state *discoveredWorkitems, opts *listOptions) error {
+	value := strings.ToLower(strings.TrimSpace(raw))
+	types := map[string]bool{
+		string(wkitem.WorkflowTypeIntent): true, string(wkitem.WorkflowTypeDesign): true,
+		string(wkitem.WorkflowTypeExplore): true, string(wkitem.WorkflowTypeFestival): true,
+	}
+	for _, item := range state.items {
+		types[string(item.WorkflowType)] = true
+	}
+	categories := map[string]bool{
+		config.WorkflowCategoryPlan: true, config.WorkflowCategoryResearch: true,
+		config.WorkflowCategoryPipeline: true, config.WorkflowCategoryReview: true,
+		"uncategorized": true,
+	}
+	for _, category := range categoryVocabulary(state.cfg) {
+		categories[category.Key] = true
+	}
+	for _, item := range state.items {
+		if item.WorkflowCategory != "" {
+			categories[item.WorkflowCategory] = true
+		}
+	}
+
+	status := wkitem.NormalizeDisplayStatus(value)
+	var dimensions []string
+	if types[value] {
+		dimensions = append(dimensions, "type")
+	}
+	if categories[value] {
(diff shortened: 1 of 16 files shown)


internal/workitem/tui/update.go
```diff
@@ -125,6 +169,7 @@ func (m Model) handleNormalKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 	// Type filter accelerators — 0 for all, 1-4 for the builtin types
 	if filter, ok := m.typeFilterFor(key); ok {
```
obey-agent, line 210: Blocking TUI correctness issue: `typeFilterFor("0")` matches here and returns immediately, so the later filter-clearing logic is never reached. With `camp workitem list --category plan` (or seeded status/group filters), pressing the footer-advertised `0 all` only clears an already-empty type filter and leaves the visible prefilter active. Handle `0` as a true clear-all action before the type accelerators (and add tests for category/status/group seeds), or label it accurately as type-only.

internal/commands/workitem/list.go
```diff
@@ -0,0 +1,170 @@
...
+	cmd.Flags().StringArrayVar(&opts.groups, "group", nil, "Filter by workitem group (repeat for OR)")
+	cmd.Flags().StringVar(&opts.groupBy, "group-by", "", "Group output sections by attention_stage, group, type, or category")
+	cmd.Flags().BoolVar(&opts.showParked, "show-parked", false, "Include parked attention-stage workitems")
+	cmd.Flags().IntVar(&opts.limit, "limit", 0, "Maximum number of items to return")
+	cmd.Flags().StringVar(&opts.query, "query", "", "Search query to filter items")
+	return cmd
+}
+
+func applyPositionalFilter(raw string, state *discoveredWorkitems, opts *listOptions) error {
+	value := strings.ToLower(strings.TrimSpace(raw))
```
obey-agent: Positional custom types/categories break when their accepted slug contains uppercase characters: `value` is lowercased, but the discovered type/category maps retain their original keys. The existing validator explicitly accepts values such as `PascalCase`, so `camp workitem list PascalCase` becomes an unknown filter even though `--type PascalCase` works. Either preserve the exact value or normalize both the vocabulary and emitted filter consistently; add a mixed-case regression test.

obey-agent (commented):
Verdict: Request Changes

Overview: The list subcommand is a strong extension of the existing discovery/filter/output stack, and it preserves the JSON schema rather than creating a parallel contract. Two user-visible filter correctness bugs remain in the TUI/shortcut paths.

Key Findings:
- Blocking: the advertised `0 all` key is intercepted by the type accelerator and returns before category/status/group prefilters are cleared. Seeded non-type filters therefore remain active.
- Positional custom type/category shortcuts lowercase the input but compare against non-normalized discovered keys, so accepted mixed-case slugs work via explicit flags but fail positionally.

What’s Done Well:
- Filtering is centralized in `FilterAdvanced` and the displayed-status abstraction is small and testable.
- Initial TUI filters remain visible/editable and JSON stays on `workitems/v1alpha8`.
- Targeted workitem tests, `just lint`, diff checks, and `TestIntegration_WorkitemListFiltersNonTTYAndJSON` pass.

Staff Standard: Not yet. I would be proud to submit this after fixing the clear-all state transition and making positional normalization consistent with the accepted slug contract, with regression tests for both.

internal/workitem/tui/model.go
```diff
@@ -187,6 +198,28 @@ func (m *Model) SetCategoryResolver(fn func(string) string) {
 	m.categoryForType = fn
 }
 
+// SetInitialFilters seeds visible, editable filters supplied by the command.
+func (m *Model) SetInitialFilters(opts workitem.FilterOptions, limit int) {
+	m.initialFilters = opts
+	m.limit = limit
```
obey-agent: **[bug]** `--limit` is sticky and invisible in the TUI.

`SetInitialFilters` stores `m.limit` and `refilter()` re-applies it on every keystroke path, including after footer-advertised **`0` clear-all**. `clearAllFilters` does not zero `m.limit`, and the header never surfaces an active limit chip. So `camp workitem list --limit 5` → press `0` still caps the list at 5 of the full set, silently. That violates the PR claim that prefilters are visible and clearable without permanently narrowing the candidate set.

**Suggestion:** Prefer applying `limit` only in the non-interactive branch of `list.go` (limit is a result-size control, not a browse filter). Or show `limit:N` in the header and clear it from `clearAllFilters`. Add a unit test: seed limit=1 with 2 items, press `0`, expect 2 items.

internal/workitem/tui/update.go
```diff
@@ -100,12 +107,65 @@ func (m Model) handleFilterKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
...
+		m.refilter()
+		m.statusMode = false
+	}
+	return m, nil
+}
+
+func (m *Model) applyStatusFilter(status string) {
+	m.statusFilter = status
+	m.initialFilters.Statuses = nil
+	m.initialFilters.LifecycleStages = nil
```
obey-agent, line 163: **[bug]** Status mode mis-represents stage/attention seeds and can clobber them.

`enterStatusMode` highlights from `m.statusFilter` only. CLI seeds that use only `--stage` / `--attention-stage` (or multi `--status`) leave `statusFilter == ""`, so status mode shows **`[all]`** while the list remains filtered. Any live navigation calls `applyStatusFilter`, which nulls `LifecycleStages` and `AttentionStages`. Stepping off “all” permanently drops stage/attention prefilters unless the user Esc-cancels before exit.

**Suggestion:** Derive status-mode cursor from the full status dimension. Do not clear lifecycle/attention until the user commits a displayed-status value (or explicitly chooses “all”). Treat “all” apply as clear of the whole status dimension only on explicit commit.

internal/workitem/tui/update.go
```diff
@@ -100,12 +107,65 @@ func (m Model) handleFilterKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
...
+	m.initialFilters.AttentionStages = nil
+	m.refilter()
+}
+
+func (m *Model) clearAllFilters() {
```
obey-agent, line 168: **[bug]** `clearAllFilters` does not clear `m.limit`.

Combined with `refilter()` always re-slicing by limit, **`0` does not restore the full browsable set** when the session was opened with `--limit`. See also sticky limit on `SetInitialFilters`.

**Suggestion:** Zero `m.limit` here (and surface it if kept), or stop threading limit into the TUI entirely.

internal/workitem/status.go
```diff
@@ -0,0 +1,46 @@
...
+// attention work uses its attention stage; other work uses lifecycle stage.
+func DisplayStatus(item WorkItem) string {
+	if item.AttentionStage != "" {
+		return item.AttentionStage
+	}
+	switch item.LifecycleStage {
+	case LifecycleStagePlanning:
+		return "plan"
+	case LifecycleStageNone, "":
+		return "none"
```
obey-agent, line 20: **[bug]** Display token `none` vs row label `-`.

`DisplayStatus` returns `"none"` for lifecycle-none/empty and `--status none` is documented/valid, but compact list/TUI rows render the same case as `"-"`. Users filtering “what I see” will not discover `none`; agents comparing row text to filter tokens will mismatch.

**Suggestion:** Prefer one function for row label + filter key (always use `DisplayStatus` for filterable status text). At least render lifecycle-none as `none` wherever `--status none` is advertised. Unit-test: empty lifecycle has `DisplayStatus == row label` (or a documented alias table).

internal/commands/workitem/list.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+			return outputList(cmd.OutOrStdout(), items, groupBy)
+		}),
+	}
+	cmd.SetFlagErrorFunc(jsoncontract.FlagErrorFunc(wkitem.SchemaVersion, func() bool { return opts.json }))
+
+	cmd.Flags().BoolVar(&opts.json, "json", false, "Output as JSON")
+	cmd.Flags().StringArrayVar(&opts.types, "type", nil, "Filter by workflow type (repeat for OR)")
+	cmd.Flags().StringArrayVar(&opts.categories, "category", nil, "Filter by workflow category (repeat for OR)")
+	cmd.Flags().StringArrayVar(&opts.statuses, "status", nil, "Filter by displayed status (repeat for OR)")
```
obey-agent: **[suggestion]** Parent `camp workitem --status/...` still pre-slices `allItems` before TUI.

Refresh rediscovers the full set and refilter only applies interactive state (no `SetInitialFilters`), so parent filtered TUI cannot recover filtered-out rows via `0` and can **widen** after `r`. The new list subcommand fixes this correctly via `runTUIWithFilters`. Shipping the same flags on parent without the same seed path creates two conflicting interactive contracts.

**Suggestion:** Route parent interactive default through `SetInitialFilters` (full items + filters), same as list. Keep pre-slice only for `--json` / `--list` / non-interactive paths.

obey-agent (commented):
## Verdict: Request Changes

## Overview
`camp workitem list` is the right product shape: TTY → TUI with editable prefilters, non-TTY → compact list, `--json` → existing `workitems/v1alpha8`. Displayed-status filtering is a sound abstraction and the list path correctly seeds the TUI with the full candidate set. Residual issues are UX/correctness around limit, status-mode state, and token/label parity — not a broken OR/AND filter core or JSON contract.

## Key Findings
1. **[bug] `--limit` sticky and invisible in TUI** — stored on the model, reapplied in `refilter()`, not cleared by `0`, not shown in header (`model.go:204`, `update.go:153`).
2. **[bug] Status mode can clobber stage/attention prefilters** — highlights `[all]` while stage/attention filters still apply; navigation nulls those seeds (`update.go:148`).
3. **[bug] Display token `none` vs row label `-`** — filter vocabulary and rendering disagree for lifecycle-none (`status.go:20`).
4. **[suggestion] Parent TUI seed parity** — parent still pre-slices `allItems`; list got this right.
5. Strongly recommended: improve `plan` ambiguity error to name both flags; share flag help strings with parent; thicken integration coverage (parked override, status vs stage).

## What's Done Well
- DisplayStatus + FilterAdvanced composition (AND across dims, OR within)
- Positional ambiguity handling is strict
- List TUI seed architecture (`SetInitialFilters` on full set)
- JSON schema unchanged / agent annotation present
- Docs + CLI reference for the new subcommand

## Staff Standard
Not yet for the three TUI/display blockers. After limit/status-mode/none parity (and ideally parent seed parity), this is proud-to-ship.

obey-agent (commented):
Verdict: Approve

Overview: The current head addresses the earlier filter-state and TUI contract findings. Both entry points now preserve a full candidate set with visible/editable prefilters, clear-all actually restores that set, and displayed status vocabulary is consistent across filtering and rows.

Key Findings:
- No blocking findings.
- Non-blocking: clarify the parent --limit help text now that TUI intentionally ignores result-size limits.
- Previous findings on clear-all ordering, positional case handling, sticky limits, status-stage seeds, parent refresh widening, and none labeling are resolved.

What’s Done Well:
- runTUIWithFilters is now the single interactive integration path instead of parallel parent/list behavior.
- The status dimension distinguishes concrete selection from explicit full clear.
- Tests cover limit removal and stage-seed preservation without filesystem mutation.

Staff Standard: Yes. I would be proud to submit this as my own work.

internal/commands/workitem/workitem.go
```diff
@@ -64,74 +66,57 @@ Examples:
...
 
 			displayGroupBy := flagGroupBy
 			if flagList && !cmd.Flags().Changed("group-by") {
 				displayGroupBy = "group"
 			}
 
+			// Interactive TUI: pass the full discovery set + editable prefilters
+			// (same contract as `camp workitem list`). Do not pre-slice allItems.
+			if interactive && !flagList && !flagJSON {
+				return runTUIWithFilters(ctx, state, filters, flagPrint, flagPathOutput)
```
obey-agent, line 101: Non-blocking: routing the parent TUI through the full-item path correctly makes --limit non-interactive-only, but the parent flag help still says Maximum number of items to return. Match the list subcommand wording so users do not expect camp workitem --limit 5 to narrow this dashboard.

obey-agent (commented):
Verdict: Approve

Overview: The current head addresses the earlier filter-state and TUI contract findings. Both entry points now preserve a full candidate set with visible/editable prefilters, clear-all actually restores that set, and displayed status vocabulary is consistent across filtering and rows.

Key Findings:
- No blocking findings.
- Non-blocking: clarify the parent --limit help text now that TUI intentionally ignores result-size limits.
- Previous findings on clear-all ordering, positional case handling, sticky limits, status-stage seeds, parent refresh widening, and none labeling are resolved.

What’s Done Well:
- runTUIWithFilters is now the single interactive integration path instead of parallel parent/list behavior.
- The status dimension distinguishes concrete selection from explicit full clear.
- Tests cover limit removal and stage-seed preservation without filesystem mutation.

Staff Standard: Yes. I would be proud to submit this as my own work.

PR #421, opened 2026-07-13
perf(sync): parallelize submodule update under the --parallel semaphore

lancekrogers:
## What

`camp sync`'s submodule update phase ran strictly serially. The `--parallel` flag (default 4) existed and was plumbed into `SyncOptions.Parallel`, but nothing in the update path read it. This wires it up.

## How

- Extracted the per-submodule pipeline (graceful init, nested init, default-branch checkout, drift reconcile) into `updateSubmodule`.
- `updateSubmodules` now runs submodules concurrently under a bounded semaphore, the same worker pattern `internal/clone` has shipped for submodule init since clone parallelization landed. Concurrency safety is identical to clone's: per-submodule work writes only the submodule's own gitdir and worktree.
- Results are index-assigned so output preserves .gitmodules declaration order. Per-submodule failure isolation, error wrapping ops, and the 0/1/2/3 exit-code contract are unchanged.

## Measurement

12-submodule fixture with local remotes, same binary, `updateSubmodules` only:

| parallel | wall time |
| --- | --- |
| 1 | 2.20s |
| 4 | 0.66s |
| 8 | 0.52s |

Local remotes are the conservative case; against network remotes the win grows because per-repo latency dominates and overlaps.

## Tests

New: declaration-order preservation under parallelism, partial-failure isolation (one broken submodule fails with `Op=init`, healthy sibling still syncs), context cancellation on the update path. Full `just gate` passes (3274 tests).

## Context

First slice of the camp-sync-clone-transport design (workitem WI-a82711): the design's measurement pass identified serial update as the cheapest large win, needing no peer machinery.

### DIFF
--- a/internal/sync/sync.go
+++ b/internal/sync/sync.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	gosync "sync"
 
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 	"github.com/Obedience-Corp/camp/internal/git"
@@ -217,9 +218,12 @@ func (s *Syncer) diffURLs(before, after map[string]string) []URLChange {
 	return changes
 }
 
-// updateSubmodules initializes and updates submodules one-by-one with graceful
+// updateSubmodules initializes and updates submodules individually with graceful
 // stale reference handling. This replaces the bulk `git submodule update --init --recursive`
-// which fails entirely if any single submodule has a stale ref.
+// which fails entirely if any single submodule has a stale ref. Submodules are
+// processed concurrently under a semaphore bounded by SyncOptions.Parallel,
+// mirroring the worker pattern clone uses for submodule initialization; results
+// preserve .gitmodules declaration order.
 func (s *Syncer) updateSubmodules(ctx context.Context) ([]SubmoduleResult, error) {
 	if ctx.Err() != nil {
 		return nil, ctx.Err()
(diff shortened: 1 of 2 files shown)


internal/sync/sync.go
```diff
@@ -229,57 +233,99 @@ func (s *Syncer) updateSubmodules(ctx context.Context) ([]SubmoduleResult, error
...
+// gitlink drift reconciliation.
+func (s *Syncer) updateSubmodule(ctx context.Context, path string) SubmoduleResult {
+	result := SubmoduleResult{
+		Path:    path,
+		Name:    path,
+		Success: true,
+	}
+
+	// Use shared graceful init (handles stale refs with fallback)
+	if err := git.InitSubmoduleGraceful(ctx, s.repoRoot, path); err != nil {
```
obey-agent, line 297: **[suggestion]** Concurrent parent-repo `git submodule init/update` can hit config/index locks under high `-p`.

Every worker runs superproject git commands concurrently. Git serializes via lockfiles; under high parallelism or slow disks this can surface intermittent `Unable to create ...lock: File exists` failures that the serial path never hit. Clone already accepts this for init, so this is consistent — but sync hits already-populated workspaces where lock windows may differ.

**Suggestion:** Accept as known tradeoff and document near `--parallel`, or add narrow retry on lock-contention errors in `InitSubmoduleGraceful` / the update path. Optional CI soak under `-race` + repeated `-p 8`.

internal/sync/sync_test.go
```diff
@@ -151,6 +152,101 @@ func TestSync_ParallelJobs(t *testing.T) {
...
+	syncErr, ok := broken.Error.(*SyncError)
+	if !ok {
+		t.Fatalf("broken submodule error type = %T, want *SyncError", broken.Error)
+	}
+	if syncErr.Op != "init" {
+		t.Errorf("SyncError.Op = %q, want %q", syncErr.Op, "init")
+	}
+}
+
+func TestUpdateSubmodules_ContextCanceled(t *testing.T) {
```
obey-agent, line 236: **[suggestion]** Cancel test only covers pre-canceled context.

`TestUpdateSubmodules_ContextCanceled` cancels before entry and only hits the early `ctx.Err()` guard. It does not exercise the new concurrent cancel paths: spawn-loop break, `select` on sem vs `ctx.Done()`, in-flight `CommandContext` while holding a semaphore slot, or post-`Wait` aggregate return.

**Suggestion:** Add a mid-flight cancel test (N submodules, low parallelism, cancel after first worker starts) and assert `errors.Is(err, context.Canceled)`, no hang, no leaked work.

internal/sync/sync.go
```diff
@@ -229,57 +233,99 @@ func (s *Syncer) updateSubmodules(ctx context.Context) ([]SubmoduleResult, error
...
+
+			select {
+			case sem <- struct{}{}:
+				defer func() { <-sem }()
+			case <-ctx.Done():
+				results[idx] = SubmoduleResult{
+					Path:    path,
+					Name:    path,
+					Success: false,
+					Error:   ctx.Err(),
```
obey-agent, line 269: **[nit]** Semaphore-wait cancel stores bare `ctx.Err()` instead of `*SyncError`.

Today results are discarded on cancel (`return nil, ctx.Err()`), so this is mostly dead. If a future change returns partial results on cancel, Sync aggregation / type asserts will see an inconsistent shape.

**Suggestion:** Wrap as `&SyncError{Op: "update", Submodule: path, Cause: ctx.Err()}` for consistency, or skip writing the result on this branch.

obey-agent (commented):
## Verdict: Comment (approve with non-blocking notes)

## Overview
This PR finally wires the already-documented `camp sync --parallel` / `SyncOptions.Parallel` into the submodule update phase. The extraction of `updateSubmodule` plus the clone-style bounded semaphore + indexed results is race-safe at the Go level, preserves declaration order, and keeps per-submodule failure isolation and the 0/1/2/3 exit contract.

## Key Findings
1. **[suggestion]** Superproject concurrent git can hit lock contention under high `-p` (same class of risk clone already accepts for init). Document or retry if flakiness appears.
2. **[suggestion]** Cancel coverage only hits the pre-canceled early return — strengthen with a mid-flight cancel test for the paths unique to this concurrency model.
3. **[nit]** Cancel select branch uses bare `ctx.Err()` rather than `*SyncError` (results discarded today).

No hard race or order-corruption bug found relative to the serial baseline.

## What's Done Well
- Closes a real product gap (dead `--parallel` flag)
- Clean extraction; order-preserving indexed results
- Context discipline: early check, select on acquire, CommandContext, post-Wait check
- Solid partial-failure isolation test
- Measured speedup on a 12-submodule fixture

## Staff Standard
Yes — I would merge this as-is with the cancel-coverage and lock-contention notes as follow-ups. Good, focused perf PR.

PR #422, opened 2026-07-13
feat(sync,clone): peer-assisted git transport via --from <machine>

lancekrogers:
## What

Class A (git content) slice of the camp-sync-clone-transport design (workitem WI-a82711, design: `workflow/design/camp-sync-clone-transport/`): move git objects between the user's own machines over the LAN/tailnet instead of round-tripping through origin, while git stays the source of truth and verifies everything on arrival.

**Stacked on #421** (parallel submodule update); merge that first. This PR's base is `sync-parallel-submodule-update`.

## Surface

```
camp sync --from studio-mac     # fetch submodule objects from that machine, then normal origin update
camp clone <url> --from studio-mac   # seed root + submodules from that machine, re-point origin, fetch delta
```

`--from` takes machine ids from `~/.obey/machines.yaml`, the same vocabulary as `camp switch machine:campaign` and `camp list --remote`.

## How

**New `internal/peer`**: resolves a transfer source from the machines file (key-auth guard, campaign root via `remote.ResolveRoot`, i.e. the far machine's own registry decides) or from a local filesystem path (mounted volume / second checkout, also the test seam). Builds peer repo URLs (`ssh://user@host/root/subpath`), a namespaced refspec (`+refs/heads/*:refs/peer/<id>/*` so fetched objects stay GC-reachable without touching origin's remote-tracking refs), and a `GIT_SSH_COMMAND` from `remote.Opts`, so every git fetch shares the machine's ControlMaster socket: ~20 submodule fetches, one ssh handshake.

**sync**: peer fetch per initialized submodule runs inside the existing parallel update workers, ahead of the origin-based update. Preflight semantics, origin URLs, post-sync validation, and the 0/1/2/3 exit contract are unchanged. Unreachable peer = per-submodule warning + origin path (degrade, never fail). Uninitialized submodules skip peer fetch silently; origin init handles them.

**clone**: root clones from the peer, then `remote set-url origin <url>`, delta fetch, `remote set-head --auto`, and checkout aligned to origin's branch tip. The final state is an origin replica regardless of how the bytes arrived (clone's contract; peer-side unpushed work is not this command's business). Submodules seed via `git -c submodule.<name>.url=<peer> submodule update`, so the per-invocation override never touches persisted config, then the module's origin remote is re-pointed to the declared URL. No origin fetch per submodule unless the recorded SHA was missing on the peer, in which case the shipped graceful-init fallback handles it. Filesystem sources add `-c protocol.file.allow=always`, matching the CVE-2022-39253 precedent in `internal/project/add.go` (camp-constructed URL, not attacker-controlled .gitmodules content). Any peer failure falls back to plain origin cloning with a warning.

## Safety invariants (from the design)

1. Class A content is git-verified on arrival: everything lands via git fetch/clone (pack indexing, ref updates, checkout by git). No raw sync of working trees or `.git` anywhere.
2. Origin remains the configured origin on every path.
3. All accelerated paths terminate in the same validation and exit-code contract as today.
4. No new credentials or listeners: ssh identity and ControlMaster from the shipped machines stack (MM0001).

## Tests

- `internal/peer`: URL construction (fs + ssh), refspec, GIT_SSH_COMMAND quoting (space-containing ControlPath).
- `internal/sync`: peer fetch brings `refs/peer/<id>/*` and objects into the local submodule; unreachable peer degrades with warning, sync still succeeds.
- `internal/clone`: peer-seeded clone ends with root origin = real URL, submodule origin = declared URL, `PeerSeeded` reported; bogus peer falls back to origin clone with warning.
- E2E: built binary; `--from` on an unregistered campaign errors clearly; plain `camp sync` byte-identical behavior.
- Full `just gate` passes (3281 tests).

## Not in this PR

Class B artifacts (declared roots + rsync engine) is the next slice; live two-machine ssh smoke belongs to the integration harness like MM0001's.

### DIFF
--- a/cmd/camp/clone.go
+++ b/cmd/camp/clone.go
@@ -3,11 +3,13 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/Obedience-Corp/camp/internal/clone"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 	"github.com/Obedience-Corp/camp/internal/jsoncontract"
+	"github.com/Obedience-Corp/camp/internal/peer"
 	"github.com/Obedience-Corp/camp/internal/ui"
 	"github.com/spf13/cobra"
 )
@@ -69,7 +71,14 @@ EXAMPLES:
   camp clone git@github.com:org/repo.git --no-register
 
   # JSON output for scripting
-  camp clone git@github.com:org/repo.git --json`,
+  camp clone git@github.com:org/repo.git --json
+
+  # Seed from a peer machine in ~/.obey/machines.yaml: root repo and
+  # submodules clone from that machine's copy (LAN/tailnet), then origin is
+  # re-pointed to the URL above and the delta fetched. The result is an
+  # origin replica that arrived over the fast path; peer failures fall back
+  # to plain origin cloning.
+  camp clone git@github.com:org/repo.git --from studio-mac`,
 	Args: jsoncontract.Args(CloneJSONVersion, func() bool { return cloneOpts.json }, cobra.RangeArgs(1, 2)),
 	Annotations: map[string]string{
 		"agent_allowed": "false",
@@ -89,6 +98,7 @@ var cloneOpts struct {
 	noRegister   bool
 	verbose      bool
 	json         bool
+	from         string
 }
 
 func init() {
@@ -108,6 +118,8 @@ func init() {
 		"Show detailed output for each operation")
 	cloneCmd.Flags().BoolVar(&cloneOpts.json, "json", false,
 		"Output results as JSON for scripting")
+	cloneCmd.Flags().StringVar(&cloneOpts.from, "from", "",
+		"Seed git objects from this machine (id from ~/.obey/machines.yaml), then fetch the delta from origin")
 
 	rootCmd.AddCommand(cloneCmd)
 	cloneCmd.GroupID = "setup"
@@ -132,7 +144,7 @@ func runClone(cmd *cobra.Command, args []string) error {
 	}
 
 	// Create cloner with options
-	cloner := clone.NewCloner(
+	clonerOpts := []clone.ClonerOption{
 		clone.WithURL(url),
 		clone.WithDirectory(directory),
 		clone.WithBranch(cloneOpts.branch),
@@ -144,7 +156,23 @@ func runClone(cmd *cobra.Command, args []string) error {
 		clone.WithVerbose(cloneOpts.verbose),
 		clone.WithJSON(cloneOpts.json),
 		clone.WithProgress(progress),
-	)
+	}
+	if cloneOpts.from != "" {
+		// The campaign is looked up on the peer by the name this clone will
+		// register under: the directory argument when given, else the repo
+		// name from the URL.
+		name := clone.RepoNameFromURL(url)
+		if directory != "" {
+			name = filepath.Base(directory)
+		}
+		src, err := peer.FromMachine(ctx, cloneOpts.from, name)
+		if err != nil {
+			formatCloneError(err, cloneOpts.json)
+			return camperrors.NewCommand("camp clone", clone.ExitCloneFailed, "", err)
+		}
+		clonerOpts = append(clonerOpts, clone.WithPeer(src))
+	}
+	cloner := clone.NewCloner(clonerOpts...)
 
 	// Execute clone
 	result, err := cloner.Clone(ctx)
--- a/cmd/camp/clone.go
+++ b/cmd/camp/clone.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -166,15 +167,20 @@ func runClone(cmd *cobra.Command, args []string) error {
 			name = filepath.Base(directory)
 		}
 		src, err := peer.FromMachine(ctx, cloneOpts.from, name)
-		if err != nil {
-			// Help text promises peer failures fall back to a plain origin
-			// clone; a resolve failure (unreachable/typo'd peer) must degrade,
-			// not abort. Warn and clone from origin with no peer configured.
+		switch {
+		case err != nil && errors.Is(err, peer.ErrPeerConfig):
+			// Misconfiguration or typo (unknown machine, bad auth): fail fast
+			// so the user fixes it, rather than silently ignoring --from.
+			formatCloneError(err, cloneOpts.json)
+			return camperrors.NewCommand("camp clone", clone.ExitCloneFailed, "", err)
+		case err != nil:
+			// Reachability failure: help text promises fallback to a plain
+			// origin clone. Warn and clone from origin with no peer configured.
 			if !cloneOpts.json {
-				fmt.Fprintf(os.Stderr, "%s peer %q unavailable (%v); cloning from origin\n",
+				fmt.Fprintf(os.Stderr, "%s peer %q unreachable (%v); cloning from origin\n",
 					ui.WarningIcon(), cloneOpts.from, err)
 			}
-		} else {
+		default:
 			clonerOpts = append(clonerOpts, clone.WithPeer(src))
 		}
 	}
--- a/cmd/camp/clone.go
+++ b/cmd/camp/clone.go
@@ -167,10 +167,16 @@ func runClone(cmd *cobra.Command, args []string) error {
 		}
 		src, err := peer.FromMachine(ctx, cloneOpts.from, name)
 		if err != nil {
-			formatCloneError(err, cloneOpts.json)
-			return camperrors.NewCommand("camp clone", clone.ExitCloneFailed, "", err)
+			// Help text promises peer failures fall back to a plain origin
+			// clone; a resolve failure (unreachable/typo'd peer) must degrade,
+			// not abort. Warn and clone from origin with no peer configured.
+			if !cloneOpts.json {
+				fmt.Fprintf(os.Stderr, "%s peer %q unavailable (%v); cloning from origin\n",
+					ui.WarningIcon(), cloneOpts.from, err)
+			}
+		} else {
+			clonerOpts = append(clonerOpts, clone.WithPeer(src))
 		}
-		clonerOpts = append(clonerOpts, clone.WithPeer(src))
 	}
 	cloner := clone.NewCloner(clonerOpts...)
 
--- a/cmd/camp/sync.go
+++ b/cmd/camp/sync.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/Obedience-Corp/camp/internal/campaign"
@@ -124,13 +125,18 @@ func runSync(cmd *cobra.Command, args []string) error {
 	}
 	if syncOpts.from != "" {
 		src, err := resolveSyncPeer(ctx, campRoot, syncOpts.from)
-		if err != nil {
-			// Help text promises an unreachable peer degrades to a warning
-			// and the git sync still runs via origin; a resolve failure
-			// (unreachable/typo'd peer) must not abort the whole command.
-			_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "camp: warning: peer %q unavailable (%v); syncing via origin\n",
+		switch {
+		case err != nil && errors.Is(err, peer.ErrPeerConfig):
+			// Misconfiguration or typo (unknown machine, campaign not
+			// registered, bad auth): fail fast so --from is not silently
+			// ignored.
+			return err
+		case err != nil:
+			// Reachability failure: help text promises an unreachable peer
+			// degrades to a warning and the git sync still runs via origin.
+			_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "camp: warning: peer %q unreachable (%v); syncing via origin\n",
 				syncOpts.from, err)
-		} else {
+		default:
 			syncerOpts = append(syncerOpts, sync.WithPeer(src))
 		}
 	}
@@ -185,7 +191,9 @@ func resolveSyncPeer(ctx context.Context, campRoot, machineID string) (*peer.Sou
 	}
 	c, found := reg.FindByPath(campRoot)
 	if !found {
-		return nil, camperrors.Newf(
+		// A setup error, not a reachability one: --from cannot proceed until
+		// the campaign is registered, so mark it ErrPeerConfig to fail fast.
+		return nil, camperrors.WrapJoinf(peer.ErrPeerConfig, nil,
 			"campaign at %s is not in the registry; --from needs the campaign's registered name to resolve it on %q",
 			campRoot, machineID)
 	}
--- a/cmd/camp/sync.go
+++ b/cmd/camp/sync.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/Obedience-Corp/camp/internal/campaign"
 	"github.com/Obedience-Corp/camp/internal/config"
@@ -124,9 +125,14 @@ func runSync(cmd *cobra.Command, args []string) error {
 	if syncOpts.from != "" {
 		src, err := resolveSyncPeer(ctx, campRoot, syncOpts.from)
 		if err != nil {
-			return err
+			// Help text promises an unreachable peer degrades to a warning
+			// and the git sync still runs via origin; a resolve failure
+			// (unreachable/typo'd peer) must not abort the whole command.
+			_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "camp: warning: peer %q unavailable (%v); syncing via origin\n",
+				syncOpts.from, err)
+		} else {
+			syncerOpts = append(syncerOpts, sync.WithPeer(src))
 		}
-		syncerOpts = append(syncerOpts, sync.WithPeer(src))
 	}
 	syncer := sync.NewSyncer(campRoot, syncerOpts...)
 
--- a/cmd/camp/sync.go
+++ b/cmd/camp/sync.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 
 	"github.com/Obedience-Corp/camp/internal/campaign"
 	"github.com/Obedience-Corp/camp/internal/config"
@@ -62,11 +63,20 @@ EXAMPLES:
 
   # Accelerate over a peer machine from ~/.obey/machines.yaml: for each
   # already-initialized submodule, fetch objects from that machine first
-  # (LAN/tailnet), then run the normal origin-based update. Uninitialized
-  # submodules skip the peer step and init from origin. Preflight, origin
-  # URLs, validation, and exit codes are unchanged; an unreachable peer
-  # degrades to a warning.
-  camp sync --from studio-mac`,
+  # (LAN/tailnet), then run the normal origin-based update, then pull
+  # declared artifact roots (policy=always) from the same machine.
+  # Uninitialized submodules skip the peer step and init from origin.
+  # Preflight, origin URLs, validation, and exit codes are unchanged; an
+  # unreachable peer degrades to a warning.
+  camp sync --from studio-mac
+
+  # Peer git objects only, skip artifacts / artifacts only, skip git phases
+  camp sync --from studio-mac --git-only
+  camp sync --from studio-mac --artifacts-only
+
+  # Check artifact roots against last-transfer snapshots, no transfer
+  camp sync --verify-artifacts
+  camp sync --verify-artifacts --from studio-mac`,
 	Args: cobra.ArbitraryArgs,
 	RunE: jsoncontract.RunE(SyncJSONVersion, func() bool { return syncOpts.json }, runSync),
 }
@@ -74,13 +84,16 @@ EXAMPLES:
 const SyncJSONVersion = "sync/v1alpha1"
 
 var syncOpts struct {
-	dryRun   bool
-	force    bool
-	verbose  bool
-	parallel int
-	noFetch  bool
-	json     bool
-	from     string
+	dryRun          bool
+	force           bool
+	verbose         bool
+	parallel        int
+	noFetch         bool
+	json            bool
+	from            string
+	gitOnly         bool
+	artifactsOnly   bool
+	verifyArtifacts bool
 }
 
 func init() {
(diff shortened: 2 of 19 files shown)


internal/clone/git.go
```diff
@@ -228,6 +228,141 @@ func (c *Cloner) cleanOrphanedGitlinks(ctx context.Context, dir string) ([]strin
...
+		args = append(args, "--branch", c.options.Branch)
+	}
+	if c.options.Depth > 0 {
+		args = append(args, "--depth", fmt.Sprintf("%d", c.options.Depth))
+	}
+	args = append(args, c.peer.URL(""), targetDir)
+
+	cmd := exec.CommandContext(ctx, "git", args...)
+	cmd.Env = c.peer.GitEnv()
+	if output, err := cmd.CombinedOutput(); err != nil {
```
obey-agent, line 270: **[bug]** Partial peer clone is not cleaned up on `git clone` failure.

`gitCloneFromPeer` documents that any failure removes the partial tree before origin fallback, but this error path returns without `os.RemoveAll`. Only `repointOrigin` failure cleans up (lines 264–266). Cancel/kill mid-clone leaves `targetDir`, so origin fallback dies with “already exists” and the documented degrade path breaks.

**Suggestion:** Defer cleanup-on-error after the pre-existence check (`cleanup := true` + `defer` RemoveAll unless success). Add a regression test that a failed/cancelled peer clone still falls back to a clean origin clone.

internal/peer/peer.go
```diff
@@ -0,0 +1,125 @@
...
+		return p
+	}
+	return "ssh://" + s.target + p
+}
+
+// Refspec returns the refspec peer fetches use. Fetched heads land under
+// refs/peer/<id>/* so transferred objects stay reachable (no GC exposure)
+// without touching origin's remote-tracking refs.
+func (s *Source) Refspec() string {
+	return "+refs/heads/*:refs/peer/" + s.id + "/*"
```
obey-agent: **[bug]** Sync peer refspec only fetches `refs/heads/*`, never HEAD.

Detached gitlink SHAs that sit ahead of every local branch tip are advertised via HEAD, not via branch refs. Clone seed follows HEAD; sync `Fetch` does not. `PeerFetched=true` can report acceleration while origin still has to transfer the gitlink objects. Current tests only cover a named branch.

**Suggestion:** Include HEAD in the peer fetch (e.g. `+HEAD:refs/peer/<id>/HEAD` plus heads), and/or fetch the recorded gitlink SHA. Add a detached-gitlink fixture and assert `cat-file -e <sha>` after peerFetch.

internal/clone/git.go
```diff
@@ -228,6 +228,141 @@ func (c *Cloner) cleanOrphanedGitlinks(ctx context.Context, dir string) ([]strin
...
+	updateCmd.Env = c.peer.GitEnv()
+	output, updateErr := updateCmd.CombinedOutput()
+
+	// Best-effort re-point even when update failed partway (e.g. peer lacked
+	// the recorded SHA after a successful module clone), so the fallback path
+	// and all future fetches talk to the declared origin.
+	subDir := filepath.Join(repoDir, sub.Path)
+	if _, statErr := os.Stat(filepath.Join(subDir, ".git")); statErr == nil {
+		setURLCmd := exec.CommandContext(ctx, "git", "-C", subDir, "remote", "set-url", "origin", sub.URL)
+		if setOut, setErr := setURLCmd.CombinedOutput(); setErr != nil && updateErr == nil {
```
obey-agent: **[suggestion]** Failed re-point is swallowed when seed update also failed.

Condition `setErr != nil && updateErr == nil` drops re-point errors when update already failed. A partial module can retain a peer URL as `origin`, weakening the “origin remains origin” invariant under partial failure.

**Suggestion:** Always attempt re-point when `.git` exists; if re-point fails, wrap/join with `updateErr` so both surface. Prefer a loud seed failure over a peer-backed origin left on disk.

internal/sync/sync.go
```diff
@@ -326,6 +339,22 @@ func (s *Syncer) updateSubmodule(ctx context.Context, path string) SubmoduleResu
 	return result
 }
 
+// peerFetch pulls objects for one submodule from the configured peer ahead of
+// the origin-based update. It returns (false, nil) for a submodule that is not
+// initialized locally: there is no repository to fetch into yet, and the
+// normal init path handles it from origin. A fetch failure is returned for
+// the caller to surface as a warning; sync then proceeds via origin.
+func (s *Syncer) peerFetch(ctx context.Context, path string) (bool, error) {
```
obey-agent, line 531: **[suggestion]** `peerFetch` no-ops for uninitialized submodules (`stat` of `.git` fails → false,nil).

That means `camp sync --from` only accelerates already-present modules; first-time materialization still hits origin. CLI help currently reads as if every submodule is peer-fetched first.

**Suggestion:** Either seed uninitialized modules like clone, or tighten help/docs to say peer assist applies only to initialized submodules.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Peer-assisted git transport is a well-scoped Class A acceleration layer: `internal/peer` reuses the machines/remote SSH stack (ControlMaster, key auth, far-machine registry resolve), clone/sync keep origin authoritative, and transfer failures degrade to warnings. The design and tests cover happy path, origin re-point, and unreachable-peer degrade. Two correctness gaps block merge: partial peer-clone cleanup breaks the documented origin fallback, and the sync refspec misses detached gitlink SHAs.

## Key Findings
1. **[bug] Partial peer clone not cleaned up** (`internal/clone/git.go:260`) — only `repointOrigin` failure RemoveAlls; a cancelled/failed peer `git clone` leaves `targetDir` and origin fallback fails with “already exists.”
2. **[bug] Sync refspec is heads-only** (`internal/peer/peer.go:92`) — detached gitlink SHAs advertised via HEAD are not transferred; clone seed works, sync often will not for the common detached-submodule case.
3. **[suggestion] Swallowed re-point errors** on partial submodule seed can leave peer as module `origin` (`internal/clone/git.go:347`).
4. **[suggestion] Sync peer path skips uninitialized submodules** — help oversells coverage (`internal/sync/sync.go:347`).
5. Optional: surface `peerSeeded` / `peerFetched` in `--json` for agent workflows.

## What's Done Well
- Clean package boundary; no parallel SSH stack
- Fail-fast on resolve, degrade on transfer
- Correct integration with #421 parallel workers
- Origin re-point + good degrade tests for unreachable peers
- Context checks and error wrapping with peer id / git output

## Staff Standard
Not yet — fix the cleanup hole and widen (or document) the sync fetch surface for detached gitlinks. After that, this is proud-to-ship work.

cmd/camp/clone.go
```diff
@@ -144,7 +156,23 @@ func runClone(cmd *cobra.Command, args []string) error {
...
+	}
+	if cloneOpts.from != "" {
+		// The campaign is looked up on the peer by the name this clone will
+		// register under: the directory argument when given, else the repo
+		// name from the URL.
+		name := clone.RepoNameFromURL(url)
+		if directory != "" {
+			name = filepath.Base(directory)
+		}
+		src, err := peer.FromMachine(ctx, cloneOpts.from, name)
```
obey-agent, line 169: Blocking: FromMachine performs the remote ResolveRoot SSH call here, and every error returns before Cloner exists. A configured but offline peer therefore makes camp clone fail instead of taking the documented peer-warning plus origin fallback. Preserve hard errors for invalid or unknown machine configuration, but move reachability/root-resolution failure into the optional acceleration path so origin clone can proceed. The sync call site has the same fail-closed behavior.

internal/sync/sync.go
```diff
@@ -293,6 +297,15 @@ func (s *Syncer) updateSubmodule(ctx context.Context, path string) SubmoduleResu
 		Success: true,
 	}
 
+	if s.peer != nil {
```
obey-agent: Blocking: this peer fetch runs even when SyncOptions.NoFetch is true. That flag promises to use local refs only, so combining --no-fetch --from currently performs network transfer despite the explicit opt-out. Gate peerFetch on !s.options.NoFetch and add contract coverage for the combined flags.

internal/clone/git_test.go
```diff
@@ -5,9 +5,145 @@ import (
...
 	"path/filepath"
+	"strings"
 	"testing"
+
+	"github.com/Obedience-Corp/camp/internal/peer"
 )
 
+func TestGitCloneFromPeer_SeedsAndRepointsOrigin(t *testing.T) {
+	ctx := context.Background()
+	origin := setupTestRepo(t)
```
obey-agent, line 16: Blocking under the campaign filesystem-safety standard: the new peer clone and sync cases execute real git while creating, cloning, deleting, and rewriting repositories under host TempDir state. internal/clone/git_test.go and internal/sync/sync_test.go are legacy allowlisted files, but that only prevents the lint ratchet from growing by filename; it does not authorize new host mutation coverage. Move these new peer transport scenarios into tests/integration using the shared container harness.

obey-agent (commented):
Verdict: Request Changes

Overview: The re-review confirms that all four earlier findings were addressed well: partial clone cleanup is now centralized, detached HEAD is transferred, re-point failures remain visible, and sync help accurately limits acceleration to initialized submodules. The current head still fails the advertised degradation contract during peer discovery, ignores --no-fetch for peer transfer, and adds real-git host mutation tests.

Key Findings:
- Blocking: offline or otherwise unreachable configured peers fail during FromMachine/ResolveRoot before clone or sync can fall back to origin.
- Blocking: camp sync --no-fetch --from still performs peer fetches.
- Blocking: migrate the new peer transport git/filesystem tests to the shared container integration harness.

What’s Done Well:
- Previous inline findings are substantively fixed with focused regressions.
- Origin remains re-pointed on partial submodule seed failure, and combined failures retain both causes.
- Namespaced refs include detached HEAD without touching origin tracking refs.

Staff Standard: Not yet. I would make optional peer discovery truly degradable, honor the explicit no-fetch contract, and move real transport tests into isolation before submitting this as my own work.

obey-agent (changes requested):
## Adversarial review — Request Changes

Multi-lens review (git-semantics, concurrency, injection, degradation contract, test adequacy). Findings that survived verification below; the HIGH items break promises the command's own help text makes.

### HIGH

**1. Unreachable/typo'd peer hard-fails instead of degrading — contract violation.**
`runSync` (cmd/camp/sync.go) and `runClone` (cmd/camp/clone.go) return the error from `peer.FromMachine` / `resolveSyncPeer`, aborting the entire command. Both help texts promise the opposite: sync says "an unreachable peer degrades to a warning"; clone says "peer failures fall back to plain origin cloning." The degrade path today only covers per-submodule *fetch* failure, not *resolve* failure. A user with a momentarily-unreachable studio-mac gets a hard failure on a command documented to fall back.
Fix: resolve failure should warn and continue on the origin path (sync) / plain origin clone (clone).

**2. clone --from submodule branch selection (verify before fixing).**
Reported that peer-seeded submodules can end up on the peer's checked-out branch with peer heads masquerading as `refs/remotes/origin/*`, rather than an origin replica. Needs confirmation — submodule update checks out the recorded gitlink SHA detached — but worth tracing `repointOrigin` / `seedSubmoduleFromPeer` interaction with `set-head --auto`.

### MEDIUM

**3. `repointOrigin` fetch lacks `--prune`/`--prune-tags`** (internal/clone/git.go): peer-only branches and tags persist as `origin/*` in the cloned root, so the result is not the promised "origin replica."

**4. Test gaps for headline safety claims:** the failed-seed repoint path (stale peer missing the recorded gitlink SHA) and `--branch`/`--depth` forwarding into the peer clone path are untested.

### LOW
- ControlMaster multiplex vs sshd `MaxSessions` (default 10): `--parallel` above that fails overflow fetches with no retry (default -p is 4, so edge; a help note suffices).
- `seedSubmoduleFromPeer` reintroduces a non-atomic init+update in parallel workers (config.lock contention) — same tradeoff documented for the parallel-update work.

Fixes will be pushed to this branch.

obey-agent:
### Fixes pushed (head b6cda629)

**Addressed (HIGH):**
- Peer resolve failure now degrades instead of aborting: `camp sync --from <unreachable>` warns "peer ... unavailable; syncing via origin" and continues; `camp clone --from` warns and falls back to a plain origin clone. Verified via built binary. Matches the help-text contract.

**Deferred as tracked follow-ups** (festival CS0005, not blockers for this slice):
- `repointOrigin` `--prune`/`--prune-tags` so peer-only branches/tags don't persist as origin refs in a peer-seeded clone.
- Investigate/confirm the peer-seeded-submodule branch-selection finding (submodule update checks out the detached gitlink SHA, so likely overstated — needs a trace).
- Test coverage for the failed-seed repoint path and `--branch`/`--depth` forwarding; ControlMaster `MaxSessions` note. These pair naturally with the deferred localhost ssh-smoke harness.

obey-agent:
### All review items now addressed (head 0b5effac)

Remaining items from the review are resolved:
- **`repointOrigin` now fetches with `--prune --prune-tags`** — a peer-seeded clone drops the peer's private branches (as `origin/*`) and tags the real origin lacks, so it is a true origin replica. New test seeds a peer with a `peer-wip` branch + `peer-only-tag` and asserts both are pruned from the clone.
- **`--branch` forwarding into the peer clone is now tested** (`camp clone --from <peer> --branch X` lands on X with X's content).
- The **peer-seeded-submodule branch-selection** concern was traced and refuted: `git submodule update` checks out the recorded gitlink SHA (detached), not any peer branch, and origin is re-pointed to the declared URL — content is the correct pinned commit regardless of the peer's branch layout.

Deferred (LOW, edge, tracked in festival CS0005): ControlMaster vs sshd `MaxSessions` note (only relevant at `-p` > ~10; default is 4), and the non-atomic submodule init+update under parallel workers (the same accepted tradeoff as parallel submodule update). No open correctness items remain.

internal/clone/git.go
```diff
@@ -228,6 +228,163 @@ func (c *Cloner) cleanOrphanedGitlinks(ctx context.Context, dir string) ([]strin
...
+	updateCmd.Env = c.peer.GitEnv()
+	output, updateErr := updateCmd.CombinedOutput()
+
+	// Always re-point when a module repo exists, even if update failed
+	// partway (e.g. peer lacked the recorded SHA after a successful module
+	// clone). Prefer a loud seed failure over leaving peer as origin.
+	subDir := filepath.Join(repoDir, sub.Path)
+	var repointErr error
+	if _, statErr := os.Stat(filepath.Join(subDir, ".git")); statErr == nil {
+		setURLCmd := exec.CommandContext(ctx, "git", "-C", subDir, "remote", "set-url", "origin", sub.URL)
```
obey-agent, line 360: This `set-url` re-points the URL but never refetches, so `origin/<branch>` keeps reflecting the peer clone's refs. Combined with Step 4 `checkoutSubmoduleBranch` (clone.go:224) running a bare `git checkout <branch>`, the submodule DWIMs a local branch off the stale peer `origin/<branch>` and lands on the peer's tip - not an origin replica, and it silently imports any un-pushed peer commits on that branch. The root path already does this correctly in `repointOrigin` (git.go:291-333): set-url, then `fetch --prune --prune-tags --tags origin`, then `checkout -B <branch> --track origin/<branch>`. Mirror that here (fetch+prune from the real origin after set-url, then checkout --track), and add a submodule-level `git branch -r` / tip assertion, since the current seed test only checks the origin URL, not the tracking refs or checked-out tip.

internal/sync/sync.go
```diff
@@ -293,6 +297,15 @@ func (s *Syncer) updateSubmodule(ctx context.Context, path string) SubmoduleResu
...
+		fetched, peerErr := s.peerFetch(ctx, path)
```
obey-agent, line 485: `peerFetch` runs unconditionally, but `SyncOptions.NoFetch` (options.go:24, set from `--no-fetch` at cmd/camp/sync.go:121) is never consumed anywhere in this update path. So `camp sync --no-fetch --from X` still opens a network fetch to the peer, contradicting the flag's "use local refs only" contract. Gate this call on `!s.options.NoFetch`.

obey-agent (changes requested):
## Re-review at 0b5effacfb - Verdict: Request Changes

Credit first: the three fixes this branch set out to make are genuinely done and tested. Partial peer-clone cleanup is centralized (git.go:236-263, `defer RemoveAll` on `!success`), the sync refspec now carries HEAD so detached gitlink SHAs transfer (peer.go:88-95), re-point errors stay visible when a seed update also failed (git.go:413-430), and resolve-failure now degrades to origin for both sync (sync.go:125-135) and clone (clone.go:160-179). Both ratchets pass.

Two correctness blockers remain, both verified against source at this HEAD:

**1. Submodules are not made origin replicas (inline on git.go:360).** The root repo is a true replica via `repointOrigin`: set-url, then `fetch --prune --prune-tags --tags origin`, then `checkout -B <branch> --track origin/<branch>`. The submodule seed path does `set-url origin` with no refetch, then Step 4 `checkoutSubmoduleBranch` (clone.go:224) runs a bare `git checkout <branch>` that DWIMs off the peer-sourced `origin/<branch>`. In the fast path (peer already has the recorded gitlink SHA, so `initSubmoduleGraceful` fetches nothing) the submodule lands on the peer's branch tip with `origin/<branch>` reflecting the peer. If the peer has un-pushed submodule commits they are silently imported, and this contradicts WithPeer's documented "final state is an origin replica". Fix: mirror `repointOrigin` in `seedSubmoduleFromPeer` (fetch+prune from the real origin after set-url, checkout --track), and add a submodule-level `git branch -r` / tip assertion (the seed test currently only checks the origin URL).

**2. `--no-fetch` is ignored on the peer path (inline on sync.go:301).** `SyncOptions.NoFetch` is defined (options.go:24) and set from the flag (cmd/camp/sync.go:121) but never consumed in the update path; `peerFetch` runs unconditionally, so `camp sync --no-fetch --from X` performs a network transfer despite the explicit local-refs-only opt-out. Gate `peerFetch` on `!s.options.NoFetch`. (Aside: base `--no-fetch` also looks non-functional post-#421 since `initSubmoduleGraceful`'s implicit origin fetch is not gated either. Worth a separate look, but the peer fetch is a new explicit network op that ignores the flag.)

Non-blocking notes:
- The new real-git peer tests in internal/clone/git_test.go and internal/sync/sync_test.go stay in the unit suite, but both files are already on the host-fs allowlist and the pattern matches existing in-file tests, so `lint-no-host-fs-tests` passes. I am not gating on it, though migrating to tests/integration remains the stated direction.
- `--depth` is forwarded to the peer clone but the delta fetch is not depth-bounded and submodule seeds ignore it (untested).
- clone `--json` degrade on an unavailable peer is silent (no warning, no field), unlike sync which warns to stderr.
- `--parallel` above sshd `MaxSessions` (default 10) has no retry; default `-p 4` is safe.

Not an Approve/merge action; Request Changes for items 1 and 2.

obey-agent:
### Second review round addressed (head 758e38e2, rebased onto main)

Rebased onto `main` (picks up #424, resolves the dirty state). Blocking/bug findings from the 0b5effac review:

- **`--no-fetch` now suppresses the peer fetch.** `peerFetch` ran unconditionally; `camp sync --no-fetch --from X` did a network transfer despite the "local refs only" contract. Gated on `!NoFetch` with a regression test.
- **Peer-seeded submodule no longer lands on the peer's tip.** After re-pointing origin, the submodule now `fetch --prune --prune-tags origin`, so `checkoutSubmoduleBranch` DWIMs off origin's refs instead of the stale peer `origin/<branch>`. Without it the submodule silently imported the peer's un-pushed commits. New test `TestGitCloneFromPeer_SubmoduleLandsOnOriginNotPeerTip` proves the clone lands on origin's state (peer-only commit absent and unreachable from origin refs). Thanks — I had wrongly refuted this earlier by only tracing the `submodule update` step and missing the subsequent branch checkout.
- **Config vs reachability degrade.** `peer.FromMachine` now classifies config/usage failures (unknown machine id, `local`, bad auth, unregistered campaign) as `ErrPeerConfig`; `sync`/`clone` **hard-fail** on those (fail fast on a typo) but still **degrade to origin** on reachability failures, per the help text. Tested at the peer layer and e2e.

Already handled at head (confirmed, no change needed):
- Partial peer clone is cleaned on any failure via the deferred `RemoveAll` (`success` flag), not only on re-point failure.
- Sync peer refspec fetches `HEAD` alongside `refs/heads/*` (detached gitlink SHAs accelerate).
- Re-point errors are surfaced even when the seed update also failed.

Test filesystem-safety finding: the new peer clone/sync cases exec real git under host `t.TempDir`, matching the established (allowlisted) pattern of the entire `internal/clone` and `internal/sync` suites. Migrating those suites' host-FS E2E cases to the container harness is tracked as an intent alongside the deferred localhost-ssh-smoke slice; the pure-logic additions (peer URL/refspec/quoting, config validation) remain unit tests.

Full `just gate` passes (3305 tests).

cmd/camp/sync.go
```diff
@@ -145,3 +180,22 @@ func runSync(cmd *cobra.Command, args []string) error {
...
 	return nil
 }
+
+// resolveSyncPeer maps --from to a peer source for this campaign: the local
+// registry supplies the campaign's name, and the far machine's own camp
+// resolves where that campaign lives there.
+func resolveSyncPeer(ctx context.Context, campRoot, machineID string) (*peer.Source, error) {
+	reg, err := config.LoadRegistry(ctx)
+	if err != nil {
+		return nil, camperrors.Wrap(err, "load registry")
```
obey-agent, line 251: **Blocking:** `config.LoadRegistry` errors are wrapped without `peer.ErrPeerConfig`, so a malformed or unreadable local campaign registry is classified as a peer reachability failure by `runSync`. With `--from`, camp then warns and silently falls back to origin, ignoring a local configuration error. Wrap this error with `peer.ErrPeerConfig` (and add a regression test) so only genuine remote reachability failures degrade.

obey-agent (commented):
## Verdict: Request Changes

## Overview
The latest head materially closes the prior peer-transport findings: peer-seeded submodules now refresh from the real origin, `--no-fetch` suppresses the new peer transfer, peer-only refs are pruned, and configuration versus reachability failures are explicitly classified. One configuration-error path still degrades silently, so the optional acceleration path is not yet safe to submit.

## Key Findings
- **Blocking:** `config.LoadRegistry` errors are not wrapped as `peer.ErrPeerConfig`; a malformed or unreadable local registry is treated as peer unreachability and `--from` silently falls back to origin. See the inline comment on `cmd/camp/sync.go:190`.

## What’s Done Well
- Submodule origin refs are refreshed after peer seeding, preventing peer-only commits from becoming the checked-out branch tip.
- `--no-fetch` now gates peer network activity.
- Peer configuration errors and reachability failures have a clear intended distinction.
- The new regression coverage targets the previously reviewed behavior.

## Staff Standard
Almost, but not yet. I would be proud to submit this after local registry/configuration failures fail closed instead of silently changing the requested transport behavior.

cmd/camp/sync.go
obey-agent, line 251: **Still blocking on re-review:** `config.LoadRegistry` errors are still wrapped without `peer.ErrPeerConfig`, so a malformed or unreadable local registry is treated as peer reachability failure and `camp sync --from` silently falls back to origin. The new classification is incomplete until this local configuration path fails closed, with regression coverage.

obey-agent (commented):
## Verdict: Request Changes

## Re-review

The head is unchanged since the prior review, and the peer-seeding/origin-refresh and `--no-fetch` fixes remain sound. The remaining local registry classification issue is still unresolved.

## Key Finding
- **Blocking:** `config.LoadRegistry` errors are not wrapped as `peer.ErrPeerConfig`; malformed or unreadable local registry state is treated as peer reachability failure, so `camp sync --from` silently falls back to origin. See the inline comment on `cmd/camp/sync.go:190`.

## What’s Done Well
- Peer-seeded submodules refresh from the declared origin.
- Peer-only refs are pruned.
- Peer configuration and reachability failures have a clear intended distinction.
- `--no-fetch` suppresses peer network fetches.

## Staff Standard
Not yet. I would submit this after all local configuration failures fail closed instead of changing the requested transport behavior silently.

PR #423, opened 2026-07-13
feat(artifacts): declared artifact roots with rsync peer transport

lancekrogers:
## What

Class B slice of the camp-sync-clone-transport design (workitem WI-a82711): media-production campaigns produce large binaries that cannot live in git, and until now camp could not see, sync, clone, or validate them. This gives them a declared home and a peer-to-peer sync story.

**Stacked on #422** (which stacks on #421); merge those first. This PR's base is `peer-git-transport`.

## Surface

```
camp artifacts add media/renders          # declare a root (.campaign/artifacts.yaml, committed)
camp artifacts add datasets --policy on-demand
camp artifacts list / remove / manifest

camp sync --from studio-mac               # git phases + policy=always artifact roots
camp sync --from studio-mac --git-only    # git objects only
camp sync --from studio-mac --artifacts-only   # artifacts only, all policies
camp sync --verify-artifacts              # roots vs last-transfer snapshots, no transfer
```

## Design decisions worth reviewing

**Conflict posture (the load-bearing one).** A local file is overwritable only when its state exactly matches the last agreed baseline with that peer. Modified-since-baseline files AND never-agreed files are excluded from the pull and reported; the post-pull snapshot records the old baseline for protected paths, so protection is sticky on every subsequent pull instead of silently expiring after one. First sync with a peer (no baseline) protects every pre-existing local file. The invariant: no artifact byte is ever silently clobbered. Resolution is explicit and human: remove the local file to take the peer's copy on the next sync. The test suite walks the full lifecycle (first sync, clean update, both-sides-changed conflict, sticky protection, resolution).

**Declaration is committed, state is not.** `.campaign/artifacts.yaml` is committed so a fresh machine knows what it is missing. Manifests and per-peer snapshots live under `.campaign/cache/peersync` (already-gitignored derived state, same convention as navigation state). Losing a snapshot degrades to first-sync caution, never to data loss.

**Engine.** System rsync (`-a --partial`), the design's recommended dependency posture: wrap proven delta tooling. Transfers ride `peer.Source`'s ssh options, so they share the ControlMaster socket with the git fetches from #422. Directional pull, no deletions, glob metacharacters in filenames escaped, exclude list spills to `--exclude-from` when large. Missing rsync is a clear error; deeper version probing (openrsync protocol quirks) is a documented follow-up, matching the design's open question 3.

**Manifest hashing.** size+mtime, per the design's open question 6 posture; content hashing deliberately deferred for multi-hundred-GB roots.

**Class hygiene.** `camp artifacts add` warns when a root contains git-tracked files or is not gitignored (class A/B ambiguity from the design). `--verify-artifacts` is purely local (tree vs snapshots) and does not dial ssh.

## Contract safety

- Preflight runs on every sync variant, including artifacts-only and verify.
- Default `camp sync --json` stays byte-identical: peer-transport keys appear only when the feature ran.
- Exit codes: usage conflicts are exit 2; `--artifacts-only` pull failures fail the run (the artifacts were the ask); artifact failures on a default `--from` sync degrade to warnings (the accelerated sync must not be worse than a plain one); verify discrepancies fail the run for scripting.

## Tests

Full lifecycle suite in `internal/artifacts` (rsync-gated with skip), declaration validation (absolute/escaping/.campaign paths rejected, ./-spelling cannot dodge checks), manifest/protected-path logic, verify. E2E with the built binary: add/list/manifest/verify flows and flag-combo errors. Full `just gate` passes (3288 tests).

### DIFF
--- a/internal/artifacts/artifacts_test.go
+++ b/internal/artifacts/artifacts_test.go
@@ -121,8 +121,14 @@ func TestPullLifecycle(t *testing.T) {
 
 	peerCampaign := t.TempDir()
 	localCampaign := t.TempDir()
+	// Ordered, whole-second timestamps an hour apart replace wall-clock sleeps:
+	// they make transfer ordering deterministic (rsync --update and the
+	// conflict predicate both compare mtimes) and cannot be reordered by
+	// rsync -a rounding preserved mtimes to seconds.
+	base := time.Unix(1_600_000_000, 0)
 	writeArtifact(t, peerCampaign, "media/render.mov", "peer render v1")
 	writeArtifact(t, peerCampaign, "media/audio.wav", "peer audio v1")
+	setMTime(t, peerCampaign, "media/render.mov", base)
 	src := peer.FromPath("peerbox", peerCampaign)
 	root := Root{Path: "media"}
 
@@ -137,8 +143,8 @@ func TestPullLifecycle(t *testing.T) {
 	assertArtifact(t, localCampaign, "media/render.mov", "peer render v1")
 
 	// Peer updates a file; local is untouched: the update flows through.
-	time.Sleep(1100 * time.Millisecond) // distinct mtime at unix-second resolution
 	writeArtifact(t, peerCampaign, "media/render.mov", "peer render v2")
+	setMTime(t, peerCampaign, "media/render.mov", base.Add(time.Hour))
 	r2 := Pull(ctx, localCampaign, src, root)
 	if r2.Warning != "" || !r2.Synced || r2.FirstSync {
 		t.Fatalf("second pull = %+v, want clean non-first sync", r2)
@@ -150,9 +156,10 @@ func TestPullLifecycle(t *testing.T) {
 
 	// Both sides change the same file: local wins, conflict reported, and the
 	// protection is sticky on the following pull.
-	time.Sleep(1100 * time.Millisecond)
 	writeArtifact(t, localCampaign, "media/render.mov", "LOCAL edit")
+	setMTime(t, localCampaign, "media/render.mov", base.Add(2*time.Hour))
 	writeArtifact(t, peerCampaign, "media/render.mov", "peer render v3")
+	setMTime(t, peerCampaign, "media/render.mov", base.Add(3*time.Hour))
 	r3 := Pull(ctx, localCampaign, src, root)
 	if r3.Warning != "" || !r3.Synced {
 		t.Fatalf("third pull = %+v, want sync with conflict", r3)
@@ -278,6 +285,17 @@ func writeArtifact(t *testing.T, campaignRoot, rel, content string) {
 	}
 }
 
+// setMTime forces a deterministic modification time on an artifact so tests
+// control transfer ordering and conflict detection without depending on
+// wall-clock timing.
+func setMTime(t *testing.T, campaignRoot, rel string, mtime time.Time) {
+	t.Helper()
+	path := filepath.Join(campaignRoot, filepath.FromSlash(rel))
+	if err := os.Chtimes(path, mtime, mtime); err != nil {
+		t.Fatalf("chtimes %s: %v", rel, err)
+	}
+}
+
 func assertArtifact(t *testing.T, campaignRoot, rel, want string) {
 	t.Helper()
 	data, err := os.ReadFile(filepath.Join(campaignRoot, filepath.FromSlash(rel)))
@@ -286,5 +304,82 @@ func assertArtifact(t *testing.T, campaignRoot, rel, want string) {
 	}
 	if string(data) != want {
 		t.Errorf("%s = %q, want %q", rel, string(data), want)
+	}
+}
+
+// TestManifest_SameSecondEditIsDetected proves the no-clobber predicate keys
+// on nanosecond mtime: a same-size in-place edit made in the same wall-clock
+// second as the baseline must still register as a change (second resolution
+// silently overwrote it).
+func TestManifest_SameSecondEditIsDetected(t *testing.T) {
+	ctx := context.Background()
+	campaign := t.TempDir()
+
+	// Skip on a filesystem that does not preserve sub-second mtimes; the fix
+	// depends on them (CI and the container harness do).
+	probe := filepath.Join(campaign, ".mtime-probe")
+	if err := os.WriteFile(probe, []byte("x"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	pt := time.Unix(1_600_000_000, 123_456_789)
+	if err := os.Chtimes(probe, pt, pt); err != nil {
+		t.Fatal(err)
+	}
+	if fi, err := os.Stat(probe); err == nil && fi.ModTime().Nanosecond() == 0 {
+		t.Skip("filesystem does not preserve sub-second mtime")
+	}
+
+	writeArtifact(t, campaign, "media/clip.bin", "AAAA")
+	t0 := time.Unix(1_600_000_000, 100_000_000) // .1s
+	setMTime(t, campaign, "media/clip.bin", t0)
+	base, err := BuildManifest(ctx, campaign, "media")
+	if err != nil {
+		t.Fatalf("BuildManifest baseline: %v", err)
+	}
+
+	writeArtifact(t, campaign, "media/clip.bin", "BBBB") // same size, different content
+	t1 := time.Unix(1_600_000_000, 900_000_000)          // .9s, SAME unix second
+	setMTime(t, campaign, "media/clip.bin", t1)
+	after, err := BuildManifest(ctx, campaign, "media")
+	if err != nil {
+		t.Fatalf("BuildManifest after: %v", err)
+	}
+
+	if t0.Unix() != t1.Unix() {
+		t.Fatalf("test setup: timestamps not in the same unix second")
+	}
+	if len(base.Files) != 1 || len(after.Files) != 1 {
+		t.Fatalf("manifest files base=%d after=%d, want 1 each", len(base.Files), len(after.Files))
+	}
+	if sameEntry(after.Files[0], base.Files[0]) {
+		t.Fatalf("same-second same-size edit treated as unchanged: base=%+v after=%+v", base.Files[0], after.Files[0])
+	}
+}
+
+// TestSnapshotSlug_Injective guards against the "/"->"__" collision that let
+// two distinct roots share one baseline file (and thus poison each other).
+func TestSnapshotSlug_Injective(t *testing.T) {
+	collisions := [][2]string{
+		{"a/b", "a__b"},
+		{"media/renders", "media__renders"},
+		{"a%2Fb", "a/b"},
+	}
+	for _, c := range collisions {
+		if snapshotSlug(c[0]) == snapshotSlug(c[1]) {
+			t.Errorf("snapshotSlug(%q) == snapshotSlug(%q) = %q; must be injective", c[0], c[1], snapshotSlug(c[0]))
+		}
+	}
+}
+
+func TestValidatePeerID(t *testing.T) {
+	for _, v := range []string{"studio-mac", "laptop", "box.local", "m1_max"} {
+		if err := ValidatePeerID(v); err != nil {
+			t.Errorf("ValidatePeerID(%q) = %v, want nil", v, err)
+		}
+	}
+	for _, v := range []string{"", "..", "../evil", "a/b", `a\b`, ".hidden", "x/../y", "a\x00b"} {
+		if err := ValidatePeerID(v); err == nil {
+			t.Errorf("ValidatePeerID(%q) = nil, want error", v)
+		}
 	}
 }
--- a/internal/artifacts/artifacts_test.go
+++ b/internal/artifacts/artifacts_test.go
@@ -267,6 +267,98 @@ func TestVerifyAgainstSnapshot(t *testing.T) {
 	}
 }
 
+// mergeStaged must not overwrite a live file that no longer matches the
+// baseline — the case a local writer touching an unprotected path during the
+// transfer window produces. This is the no-clobber guarantee that a
+// pre-transfer exclude scan alone cannot provide.
+func TestMergeStaged_PreservesLateLocalWrite(t *testing.T) {
+	camp := t.TempDir()
+	writeArtifact(t, camp, "media/x.bin", "LOCAL EDIT DURING TRANSFER")
+	dest := filepath.Join(camp, "media")
+
+	// Baseline recorded x.bin as a different (smaller) version, so the current
+	// live file does not match it — exactly what a mid-transfer local edit
+	// looks like at merge time.
+	baseline := &Manifest{Version: 1, Root: "media", Files: []FileEntry{
+		{Path: "x.bin", Size: 2, MTime: 1},
+	}}
+
+	staging := filepath.Join(camp, ".campaign", "cache", "stg")
+	if err := os.MkdirAll(staging, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(staging, "x.bin"), []byte("peer version"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	late, err := mergeStaged(staging, dest, baseline)
+	if err != nil {
+		t.Fatalf("mergeStaged() error = %v", err)
+	}
+	if len(late) != 1 || late[0] != "x.bin" {
+		t.Errorf("late conflicts = %v, want [x.bin]", late)
+	}
+	assertArtifact(t, camp, "media/x.bin", "LOCAL EDIT DURING TRANSFER")
+}
+
+func TestMergeStaged_MovesAgreedAndNewFiles(t *testing.T) {
+	ctx := context.Background()
+	camp := t.TempDir()
+	writeArtifact(t, camp, "media/agreed.bin", "v1")
+
+	// Baseline matches the current live agreed.bin exactly.
+	baseline, err := BuildManifest(ctx, camp, "media")
+	if err != nil {
+		t.Fatal(err)
+	}
+	dest := filepath.Join(camp, "media")
+
+	staging := filepath.Join(camp, ".campaign", "cache", "stg")
+	if err := os.MkdirAll(filepath.Join(staging, "sub"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(staging, "agreed.bin"), []byte("peer-v2"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(staging, "sub", "new.bin"), []byte("brand new"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	late, err := mergeStaged(staging, dest, baseline)
+	if err != nil {
+		t.Fatalf("mergeStaged() error = %v", err)
+	}
+	if len(late) != 0 {
+		t.Errorf("late conflicts = %v, want none", late)
+	}
+	assertArtifact(t, camp, "media/agreed.bin", "peer-v2")     // agreed -> updated
+	assertArtifact(t, camp, "media/sub/new.bin", "brand new") // new nested file -> created
+}
+
+func TestMergeStaged_ProtectsOutOfBandLocalFile(t *testing.T) {
+	camp := t.TempDir()
+	writeArtifact(t, camp, "media/z.bin", "created locally, not in baseline")
+	dest := filepath.Join(camp, "media")
+
+	staging := filepath.Join(camp, ".campaign", "cache", "stg")
+	if err := os.MkdirAll(staging, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(staging, "z.bin"), []byte("peer z"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	baseline := &Manifest{Version: 1, Root: "media"} // empty: z.bin never agreed
+
+	late, err := mergeStaged(staging, dest, baseline)
+	if err != nil {
+		t.Fatalf("mergeStaged() error = %v", err)
+	}
+	if len(late) != 1 || late[0] != "z.bin" {
+		t.Errorf("late conflicts = %v, want [z.bin]", late)
+	}
+	assertArtifact(t, camp, "media/z.bin", "created locally, not in baseline")
+}
+
 func requireRsync(t *testing.T) {
 	t.Helper()
 	if _, err := exec.LookPath("rsync"); err != nil {
--- a/internal/artifacts/artifacts_test.go
+++ b/internal/artifacts/artifacts_test.go
@@ -1,0 +1,290 @@
+package artifacts
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/Obedience-Corp/camp/internal/peer"
+)
+
+func TestFileAddRemoveValidate(t *testing.T) {
+	f := &File{Version: 1}
+
+	if err := f.Add(Root{Path: "media/renders"}); err != nil {
+		t.Fatalf("Add() error = %v", err)
+	}
+	if err := f.Add(Root{Path: "./media/renders/"}); err == nil {
+		t.Error("Add() duplicate (unnormalized) accepted, want error")
+	}
+	if err := f.Add(Root{Path: "/abs/path"}); err == nil {
+		t.Error("Add() absolute path accepted, want error")
+	}
+	if err := f.Add(Root{Path: "../escape"}); err == nil {
+		t.Error("Add() escaping path accepted, want error")
+	}
+	if err := f.Add(Root{Path: ".campaign/cache"}); err == nil {
+		t.Error("Add() .campaign path accepted, want error")
+	}
+	if err := f.Add(Root{Path: "datasets", Policy: "sometimes"}); err == nil {
+		t.Error("Add() unknown policy accepted, want error")
+	}
+	if err := f.Add(Root{Path: "datasets", Policy: PolicyOnDemand}); err != nil {
+		t.Fatalf("Add() on-demand error = %v", err)
+	}
+
+	if _, found := f.Find("media/renders"); !found {
+		t.Error("Find() did not locate declared root")
+	}
+	if !f.Remove("media/renders/") {
+		t.Error("Remove() = false, want true")
+	}
+	if _, found := f.Find("media/renders"); found {
+		t.Error("Find() located removed root")
+	}
+}
+
+func TestConfigRoundTrip(t *testing.T) {
+	campaignRoot := t.TempDir()
+	f := &File{Version: 1}
+	if err := f.Add(Root{Path: "media", Policy: PolicyOnDemand}); err != nil {
+		t.Fatalf("Add() error = %v", err)
+	}
+	if err := f.Save(campaignRoot); err != nil {
+		t.Fatalf("Save() error = %v", err)
+	}
+
+	loaded, err := Load(campaignRoot)
+	if err != nil {
+		t.Fatalf("Load() error = %v", err)
+	}
+	if len(loaded.Roots) != 1 || loaded.Roots[0].Path != "media" || loaded.Roots[0].Policy != PolicyOnDemand {
+		t.Errorf("Load() roots = %+v, want media/on-demand", loaded.Roots)
+	}
+
+	empty, err := Load(t.TempDir())
+	if err != nil {
+		t.Fatalf("Load() missing file error = %v", err)
+	}
+	if len(empty.Roots) != 0 {
+		t.Errorf("Load() missing file roots = %d, want 0", len(empty.Roots))
+	}
+}
+
+func TestBuildManifestAndProtectedPaths(t *testing.T) {
+	ctx := context.Background()
+	campaignRoot := t.TempDir()
+	writeArtifact(t, campaignRoot, "media/a.bin", "aaa")
+	writeArtifact(t, campaignRoot, "media/sub/b.bin", "bbbb")
+
+	m, err := BuildManifest(ctx, campaignRoot, "media")
+	if err != nil {
+		t.Fatalf("BuildManifest() error = %v", err)
+	}
+	if len(m.Files) != 2 {
+		t.Fatalf("manifest files = %d, want 2", len(m.Files))
+	}
+	if m.Files[0].Path != "a.bin" || m.Files[1].Path != "sub/b.bin" {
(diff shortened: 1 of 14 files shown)


internal/artifacts/config.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, camperrors.Wrap(err, "read artifacts config")
+	}
+	var f File
+	if err := yaml.Unmarshal(data, &f); err != nil {
+		return nil, camperrors.Wrap(err, "parse artifacts config")
+	}
+	if f.Version == 0 {
+		f.Version = 1
+	}
+	return &f, nil
```
obey-agent, line 79: Blocking: Load trusts every root from the committed YAML without applying ValidateRootPath, policy validation, duplicate checks, or version checks. A hand-edited or branch-supplied path such as ../outside reaches NormalizeRootPath and then filepath.Join in Pull, escaping the campaign and letting rsync write outside the declared workspace. Validate the raw on-disk schema during Load before any normalization or I/O, reject unsupported versions, invalid policies, duplicates, and escaping or .campaign roots, and add malicious-config tests.

internal/artifacts/manifest.go
```diff
@@ -0,0 +1,178 @@
...
+			return err
+		}
+		rel, err := filepath.Rel(rootAbs, path)
+		if err != nil {
+			return err
+		}
+		m.Files = append(m.Files, FileEntry{
+			Path:  filepath.ToSlash(rel),
+			Size:  fi.Size(),
+			MTime: fi.ModTime().Unix(),
```
obey-agent: Blocking: second-resolution mtime makes the no-clobber predicate unsafe. A same-size local edit made within the same second is indistinguishable from the baseline and can be overwritten; the lifecycle test has to sleep 1.1 seconds to avoid demonstrating this. Preserve at least nanosecond timestamps, and add a same-size same-second regression. This still needs the transfer-time revalidation noted separately.

internal/artifacts/snapshot.go
```diff
@@ -0,0 +1,72 @@
...
+// since we last synced with that peer" from "safe to overwrite": conflict
+// detection needs a baseline, and this is it. Losing a snapshot is safe; the
+// next pull just degrades to first-sync caution (--ignore-existing).
+
+const snapshotDir = ".campaign/cache/peersync"
+
+// snapshotPath returns the snapshot file for one (peer, root) pair.
+func snapshotPath(campaignRoot, peerID, rootRel string) string {
+	slug := strings.ReplaceAll(NormalizeRootPath(rootRel), "/", "__")
+	return filepath.Join(campaignRoot, filepath.FromSlash(snapshotDir), peerID, slug+".json")
```
obey-agent: Blocking: this is neither a safe nor injective encoding of the snapshot key. Valid roots a/b and a__b share one baseline file, and VerifyPeer comes directly from --from in verify mode, so values containing ../ can traverse out of peersync. Validate peer IDs even on local verify and encode the full peer/root tuple with a collision-resistant filename or safe nested scheme, then enforce containment under snapshotDir.

internal/artifacts/pull.go
```diff
@@ -0,0 +1,207 @@
...
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+	if _, err := exec.LookPath("rsync"); err != nil {
+		return camperrors.New("rsync not found on PATH; artifact sync needs rsync on both machines")
+	}
+
+	rootRel := result.Root
+	destAbs := filepath.Join(campaignRoot, filepath.FromSlash(rootRel))
+	if err := os.MkdirAll(destAbs, 0o755); err != nil {
```
obey-agent, line 87: Blocking: a lexically valid artifact root can already be a symlink, including a symlink to a directory outside the campaign. MkdirAll accepts that existing directory, and rsync receives the symlink path as its live destination. Reject symlinks in the destination root and its path components, and perform a containment-safe open/create before transfer so artifact pulls cannot write outside the campaign through filesystem indirection.

internal/artifacts/pull.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return err
+	}
+	if excludeFile != "" {
+		defer func() { _ = os.Remove(excludeFile) }()
+	}
+	args = append(args, excludeArgs...)
+	args = append(args, src.RsyncSpec(rootRel), destAbs+string(os.PathSeparator))
+
+	cmd := exec.CommandContext(ctx, "rsync", args...)
```
obey-agent, line 145: Blocking: the protected set is computed before rsync, but rsync then writes directly into the live destination. A renderer or user can create or modify an unprotected path after the manifest walk and before or during transfer, and that local byte stream is silently overwritten. To uphold the stated invariant, stage incoming content outside the live root and revalidate each destination against the captured baseline immediately before an atomic merge; a pre-transfer scan alone cannot provide no-clobber semantics.

internal/peer/peer.go
```diff
@@ -104,21 +104,41 @@ func (s *Source) Refspec() string {
...
+// RsyncSpec returns the rsync source spec for the directory at relPath under
+// the peer's campaign root, with a trailing slash so rsync copies contents
+// into the destination directory rather than nesting it.
+func (s *Source) RsyncSpec(relPath string) string {
+	p := path.Join(s.root, relPath)
+	if s.target == "" {
+		return p + "/"
 	}
-	return append(os.Environ(), "GIT_SSH_COMMAND="+strings.Join(parts, " "))
+	return s.target + ":" + p + "/"
```
obey-agent: Blocking: the remote path is concatenated into rsync remote-shell syntax without protected-args handling or quoting. Campaign roots containing spaces or shell metacharacters can fail or be reinterpreted by the remote shell. Use a transport form that keeps path data out of shell parsing, such as a verified --secluded-args capability plus an appropriate remote-path encoding, and cover spaces, quotes, dollar signs, semicolons, and leading dashes.

internal/sync/sync.go
```diff
@@ -115,9 +127,109 @@ func (s *Syncer) Sync(ctx context.Context) (*SyncResult, error) {
...
+func (s *Syncer) pullArtifacts(ctx context.Context, result *SyncResult) {
+	if s.peer == nil {
+		result.Success = false
+		result.Errors = append(result.Errors, &SyncError{Op: "artifacts",
+			Cause: camperrors.New("artifact sync needs a peer (--from <machine>)")})
+		return
+	}
+	cfg, err := artifacts.Load(s.repoRoot)
+	if err != nil {
+		result.Warnings = append(result.Warnings, fmt.Sprintf("artifacts config: %v", err))
```
obey-agent, line 157: Blocking: malformed or unreadable artifacts.yaml is only a warning here even for --artifacts-only, so the command exits successfully after transferring nothing. The PR contract says artifact-only failures fail the run. Mark Success false and retain a structured SyncError when ArtifactsOnly is set; keep warning-only degradation only for the optional artifact phase of a normal sync.

internal/sync/sync.go
```diff
@@ -115,9 +127,109 @@ func (s *Syncer) Sync(ctx context.Context) (*SyncResult, error) {
...
+			result.Errors = append(result.Errors, ctx.Err())
+			result.Success = false
+			return
+		}
+		if !s.options.ArtifactsOnly && root.EffectivePolicy() != artifacts.PolicyAlways {
+			continue
+		}
+		pr := artifacts.Pull(ctx, s.repoRoot, s.peer, root)
+		result.Artifacts = append(result.Artifacts, pr)
+		if pr.Warning != "" {
```
obey-agent, line 184: Blocking: Pull converts every error to a warning string, and this default-sync branch leaves Success true. If the context is canceled during rsync, camp sync can therefore print a warning and exit 0 because there is no later context check. Preserve the underlying error or check ctx.Err immediately after Pull; cancellation and deadlines must terminate the sync as errors, while genuine peer availability failures may degrade.

internal/artifacts/artifacts_test.go
```diff
@@ -0,0 +1,290 @@
...
+	if len(empty.Files) != 0 {
+		t.Errorf("missing root manifest files = %d, want 0", len(empty.Files))
+	}
+}
+
+func TestPullLifecycle(t *testing.T) {
+	requireRsync(t)
+	ctx := context.Background()
+
+	peerCampaign := t.TempDir()
```
obey-agent, line 122: Blocking under the campaign filesystem-safety standard: this new unit suite creates and mutates real trees and invokes system rsync directly in host TempDir state. Move the transfer lifecycle and snapshot filesystem cases into the shared container integration harness. Keep pure normalization, manifest comparison, and pattern encoding tests host-side by passing in-memory values.

obey-agent (commented):
Verdict: Request Changes

Overview: The declaration/manifest/snapshot separation is a strong direction, and the sticky baseline model is thoughtfully documented. The implementation does not yet satisfy its load-bearing no-clobber and workspace-containment guarantees: unvalidated committed paths can escape the campaign, live-tree rsync is racy, metadata can miss common edits, snapshot keys collide, and remote paths cross a shell boundary unsafely.

Key Findings:
- Blocking: validate the on-disk artifacts schema before use; current ../ roots can drive reads and writes outside the campaign.
- Blocking: redesign pull as staged transfer plus destination revalidation; direct rsync after a pre-scan can overwrite concurrent local work.
- Blocking: second-resolution mtimes allow same-size same-second edits to be treated as agreed.
- Blocking: make snapshot key encoding collision-safe and validate peer IDs in verify mode.
- Blocking: reject symlinked destination components and protect remote rsync path arguments.
- Blocking: artifacts-only config failures and context cancellation must fail instead of degrading to exit 0.
- Blocking: move real rsync/filesystem lifecycle coverage into the container integration harness.
- Additional safety gap: exclude-from is newline-delimited and does not escape backslashes, so valid Unix filenames can split or alter protection patterns; use a NUL-safe representation or reject unsupported names explicitly.

What’s Done Well:
- First-sync caution and sticky conflict baselines are clearly reasoned.
- Default JSON compatibility is preserved by feature-gating new keys.
- Policy separation, local verification, and explicit conflict reporting form a useful operator surface.

Staff Standard: No. I would not submit a filesystem transport as production-ready until containment and no-clobber are enforced against adversarial paths and concurrent writers, with isolated integration coverage.

obey-agent (changes requested):
## Adversarial review — Request Changes

Multi-lens review focused on the conflict/baseline state machine (the "never silently clobbered" invariant), rsync invocation, path validation, and contract compat. The HIGH items are data-loss or path-traversal paths.

### HIGH

**1. `ValidateRootPath` is enforced only in `File.Add`, never at load/pull time.**
internal/artifacts/config.go validates on `add`, but `internal/sync/sync.go` (`pullArtifacts`/`verifyArtifacts`) iterates `cfg.Roots` straight from a `Load()` and hands `root.Path` to rsync/manifest unvalidated. A hand-edited or maliciously-committed `.campaign/artifacts.yaml` with a `../..` root drives rsync writes **outside the campaign root**, and a `.campaign/...` root can clobber camp's own snapshot state. Fix: validate every root at load/consume time; skip+warn on invalid.

**2. `rsync --partial` + sticky protection = permanent truncation.**
An interrupted transfer leaves a truncated file at the *destination* filename (that's what bare `--partial` does). The post-pull manifest then records it as agreed baseline, and later pulls treat it as clean and refuse to repair it. Fix: use `--partial-dir` so partials never land at the final name.

**3. rsync exit 23/24 (partial transfer) treated as total failure.**
internal/artifacts/pull.go treats any non-zero rsync exit as an error and skips the snapshot, so every file that *did* transfer becomes "never-agreed" and gets permanently protected on subsequent runs. Fix: treat 23/24 as partial success — warn, still snapshot what transferred.

**4. Symlink handling voids protection / escapes the campaign.**
Manifests skip symlinks (internal/artifacts/manifest.go) but `rsync -a` transfers and replaces them: a local symlink is silently destroyed by a peer regular file, and peer symlinks become phantom state `--verify-artifacts` can't see. A symlinked artifact root can also let rsync write outside the campaign.

### MEDIUM
- **`RsyncSpec` embeds an unquoted remote path** (internal/peer/peer.go): peer roots with spaces break the pull; metacharacters in a committed artifacts.yaml path could execute on the peer. Quote it.
- **`--verify-artifacts --from <typo/never-synced>` exits 0** printing "Artifact verification complete." having checked nothing. Validate the scoped peer id / warn when no snapshot matched.
- **Zero tests for the sync-level orchestration:** `pullArtifacts` policy filtering, artifacts-only failure => `Success=false`, `GitOnly` skip, `verifyArtifacts` peer scoping, `ListSnapshotPeers` — the PR's headline behavior.
- **JSON casing:** `PullResult` ships `first_sync`/`skipped_conflicts` (snake_case) inside the camelCase `sync/v1alpha1` envelope.
- **Human output drops never-agreed protected files:** the root prints green "(synced)" while N local files stay permanently excluded.
- Snapshot slug non-injective (`media/renders` vs `media__renders` collide); TOCTOU between manifest walk and rsync with no lock on concurrent runs.

### LOW
- `--dry-run` no-ops `--verify-artifacts`/`--artifacts-only` but prints a success line.
- `.campaign` guard is case-sensitive (bypassable on case-insensitive APFS).
- `escapeRsyncPattern` doesn't escape backslash; check-ignore warning wrong for trailing-slash dir patterns when the root doesn't exist yet.

Fixes will be pushed to this branch.

obey-agent:
### Fixes pushed (head b9ec1459)

**Addressed (HIGH — the "never silently clobbered" invariant):**
- Load/consume-time root validation (`EnsureRootWithin`): a hand-edited/committed `../..` or `.campaign/` root, or a symlinked root, can no longer redirect rsync writes outside the campaign. Re-validated at pull AND verify time, symlink-resolved.
- `--partial` → `--partial-dir` (under `.campaign/cache`): interrupted transfers no longer leave a truncated file at the real name that the baseline would then protect from repair.
- rsync exit 23/24 treated as partial success (snapshot what transferred) rather than poisoning transferred files into never-agreed protection — but a no-op failure (missing source / unchanged tree) is still a hard failure. New `manifestsEqual` distinguishes the two.
- Symlinks recorded in the manifest (lstat) so a local symlink participates in conflict protection; `--no-links` so peer symlinks never become phantom local state; file↔symlink type change counts as a difference.

**Addressed (MEDIUM/LOW):**
- `--verify-artifacts --from <typo/never-synced>` now fails loudly instead of exiting 0 "complete" having checked nothing.
- `PullResult` JSON fields are camelCase to match the `sync/v1alpha1` envelope.
- `escapeRsyncPattern` escapes backslash; `.campaign` guard is case-insensitive.

**Deferred as tracked follow-ups** (festival CS0005):
- `RsyncSpec` quoting for peer roots with spaces/metachars (interacts with the deferred rsync version-probing work).
- Concurrent-run snapshot locking (TOCTOU); snapshot-slug injectivity; `--dry-run` no-op on verify/artifacts-only; human-output visibility of never-agreed protected files.
- Sync-level orchestration tests (`pullArtifacts`/`verifyArtifacts` policy filtering) — pairs with the localhost ssh-smoke harness.

internal/artifacts/manifest.go
```diff
@@ -0,0 +1,191 @@
...
+			return err
+		}
+		rel, err := filepath.Rel(rootAbs, path)
+		if err != nil {
+			return err
+		}
+		m.Files = append(m.Files, FileEntry{
+			Path:    filepath.ToSlash(rel),
+			Size:    fi.Size(),
+			MTime:   fi.ModTime().Unix(),
```
obey-agent: Second-resolution mtime makes the no-clobber predicate unsafe: `sameEntry` (manifest.go:130-131) compares size + this `.Unix()` mtime + symlink, so a same-size in-place edit made within the same wall-clock second as the recorded baseline is indistinguishable from the baseline, is left out of the protected set, and gets overwritten by the peer's copy on the next pull. The lifecycle test sleeping 1100ms (artifacts_test.go:140,153) is a tell that the resolution is load-bearing. Use `fi.ModTime().UnixNano()`, or add a cheap content hash for files under a size threshold, so a same-second edit is still detected.

internal/peer/peer.go
obey-agent: The remote spec is `target:path` with no quoting, and pull.go:117 passes no `-s`/`--secluded-args`. Two problems: a legitimate artifact root containing a space (common for media, e.g. "Final Renders/") silently fails to transfer; and since `ValidateRootPath` permits spaces and shell metacharacters, a committed artifacts.yaml path with those is interpreted by the peer's remote shell. Add `-s` to the rsync args and/or shell-quote the remote path.

internal/sync/sync.go
```diff
@@ -115,9 +127,127 @@ func (s *Syncer) Sync(ctx context.Context) (*SyncResult, error) {
...
+// moves every declared root and treats per-root failures as sync failures
+// (the artifacts were the whole ask).
+func (s *Syncer) pullArtifacts(ctx context.Context, result *SyncResult) {
+	if s.peer == nil {
+		result.Success = false
+		result.Errors = append(result.Errors, &SyncError{Op: "artifacts",
+			Cause: camperrors.New("artifact sync needs a peer (--from <machine>)")})
+		return
+	}
+	cfg, err := artifacts.Load(s.repoRoot)
```
obey-agent, line 151: On a malformed or unreadable artifacts.yaml this appends a warning and returns without setting `result.Success = false`. Under `--artifacts-only` that means `camp sync --artifacts-only --from X` transfers nothing and still exits 0, but the PR contract says artifact-only failures fail the run. The per-root Pull failure path below does set `Success=false`; this config-load leg just needs the same. Relatedly, the `ctx.Err()` check is only at the top of the loop, so a cancellation during the final/only root's rsync in default mode leaves `Success=true`; consider failing on context cancellation regardless of mode.

obey-agent (changes requested):
## Re-review at 2a197107 - Verdict: Request Changes

Real credit: the fix commit closed several prior HIGH findings and they hold up under inspection. `--partial-dir` keeps truncated bytes out of the tree (pull.go:113-117), rsync exit 23/24 is discriminated as partial success and still snapshots what transferred (pull.go:149-156), symlinks are recorded via lstat and `--no-links` prevents materializing peer symlinks (manifest.go:67-84, pull.go:117), write paths validate roots at consume time via `EnsureRootWithin` (config.go:180-204, called from pull and verify), JSON is camelCase, and `--verify-artifacts --from <typo>` now fails instead of exiting 0.

But the load-bearing "no artifact byte is ever silently clobbered" invariant is still violable at this HEAD. Blockers, verified against source:

**1. Second-resolution mtime (inline on manifest.go:82).** `sameEntry` compares size + `.Unix()` mtime; a same-size in-place edit within the same second as the baseline is indistinguishable from it and gets overwritten by the peer copy. The lifecycle test still sleeps 1100ms (artifacts_test.go:140,153), which is the tell. Use `UnixNano` or a cheap content hash under a size threshold.

**2. Unquoted rsync remote spec, no `--secluded-args` (inline on peer.go:141).** `RsyncSpec` returns `target:path` with no quoting and pull.go:117 passes no `-s`. A legit media root with a space silently fails to transfer; `ValidateRootPath` permits spaces and shell metacharacters, so a committed path with those reaches the peer's remote shell. Add `-s` and/or quote the remote spec.

**3. `--artifacts-only` config-load failure exits 0 (inline on sync.go:151).** A malformed `artifacts.yaml` only appends a warning and returns without `Success=false`, so `camp sync --artifacts-only --from X` transfers nothing and still exits 0; the contract says artifact-only failures fail the run. The per-root Pull failure path does set `Success=false` correctly; only this config-load leg is missed. Also the `ctx.Err()` guard is only at the top of the loop, so a cancellation during the final root's rsync in default mode leaves `Success=true`.

**4. Non-injective snapshot slug + unvalidated verify peer (snapshot.go:22-23).** The slug replaces `/` with `__`, so roots `a/b` and `a__b` collide on one baseline file (wrong baseline, clobber), and `VerifyPeer` comes straight from `--from` and is `filepath.Join`ed into the snapshot path, so `--verify-artifacts --from ../..` traverses on read. Use an injective encoding and validate the peer id.

**5. Walk to rsync TOCTOU (pull.go:96 vs 129).** The protected set is computed from a pre-rsync walk, then rsync writes directly into the live destination with no lock/staging/post-scan. `--partial-dir` fixes truncation but not the race: a file created or modified between the walk and the transfer is not in the protected set and gets overwritten. Narrow under single-user, but it is the exact invariant the PR sells; a lock or a destination re-check would close it.

Plus one finding not in the prior review: `camp artifacts list` / `manifest` (cmd/camp/artifacts.go) walk `cfg.Roots` without `EnsureRootWithin`, so a hand-edited `../..` root makes those read commands stat and print sizes/paths outside the campaign. Route them through the same validation as pull/verify.

Non-blocking:
- The rsync lifecycle test (artifacts_test.go) is a host-TempDir unit test with no `//go:build integration`; it `t.Skip`s without rsync and the git-only host-fs ratchet does not catch rsync, so the gate passes, but migrating to tests/integration is the stated direction.
- exclude-from is newline-joined (backslash now escaped, but a filename with a literal newline still splits a pattern); use `--from0`/NUL or reject such names.
- Human output prints green "(synced)" while never-agreed protected files stay excluded with the count shown nowhere.
- `--dry-run --verify-artifacts`/`--artifacts-only` prints success having done nothing.

Request Changes on 1-5 and the read-path leak.

internal/artifacts/config.go
```diff
@@ -0,0 +1,222 @@
...
+		if os.IsNotExist(err) {
+			return &File{Version: 1}, nil
+		}
+		return nil, camperrors.Wrap(err, "read artifacts config")
+	}
+	var f File
+	if err := yaml.Unmarshal(data, &f); err != nil {
+		return nil, camperrors.Wrap(err, "parse artifacts config")
+	}
+	if f.Version == 0 {
```
obey-agent, line 76: **Blocking:** `Load` still accepts an unvalidated on-disk schema: unsupported versions, unknown policies, duplicate roots, and malformed root declarations are returned as a usable `File`. `EnsureRootWithin` only validates the path; `pullArtifacts` silently skips an unknown policy during normal sync, pulls it during `--artifacts-only`, and duplicate roots can overwrite the same snapshot twice. Validate the complete file at load/consume time and fail before any root is used.

internal/artifacts/pull.go
```diff
@@ -0,0 +1,291 @@
...
+	// destination that is newer than the peer's copy, which is exactly the
+	// signature of a file a local writer just touched, so a concurrent edit is
+	// not clobbered. This narrows but does not fully eliminate the race (a
+	// concurrent write landing an older-or-equal mtime is still possible on a
+	// truly live tree); the design's guarantee assumes a quiescent root.
+	partialDir := filepath.Join(campaignRoot, ".campaign", "cache", "rsync-partial")
+	if err := os.MkdirAll(partialDir, 0o755); err != nil {
+		return camperrors.Wrap(err, "create rsync partial dir")
+	}
+	args := []string{"-a", "--no-links", "--update", "--partial-dir=" + partialDir}
```
obey-agent: **Blocking:** `--update` is only an mtime comparison and does not close the walk-to-rsync TOCTOU that this PR’s “never silently clobbered” invariant promises. A local writer can still modify a file with an older/equal mtime (or on a filesystem with coarse timestamp resolution), after `ProtectedPaths` is computed, and rsync can overwrite it. Stage the peer tree outside the live root and revalidate destinations against the captured baseline immediately before an atomic merge, or explicitly narrow the contract and add the corresponding safety documentation/tests.

obey-agent (commented):
## Verdict: Request Changes

## Overview
The latest head closes several serious artifact-transport findings: nanosecond manifest mtimes, quoted remote rsync paths, artifact-only config/cancellation failures, injective snapshot slugs, peer-id validation, and read-path root validation are all meaningful improvements. Two production-safety blockers remain around committed-schema validation and the advertised no-clobber invariant.

## Key Findings
- **Blocking:** `artifacts.Load` still accepts unsupported versions, unknown policies, duplicate roots, and other invalid on-disk declarations; consume-time path checks do not validate the complete schema. See `internal/artifacts/config.go:76`.
- **Blocking:** `--update` narrows but does not close the walk-to-rsync TOCTOU. A concurrent write with an older/equal mtime can still be overwritten, contradicting the “never silently clobbered” contract. See `internal/artifacts/pull.go:125`.
- **Follow-up:** The rsync lifecycle tests still mutate host `TempDir` state; they should move to the shared container harness under the repository filesystem-safety standard.

## What’s Done Well
- Containment is rechecked at pull, verify, list, and manifest read paths.
- Partial transfers no longer poison the baseline, and interrupted files stay out of the destination tree.
- Snapshot key encoding and verify peer scoping are substantially safer.
- The new regression tests cover the latest mtime, quoting, traversal, and exit-contract fixes.

## Staff Standard
Not yet. I would be proud to submit this after the committed artifact schema is fail-closed and the transfer path enforces no-clobber semantics under the stated live-workspace contract.

internal/artifacts/config.go
obey-agent, line 76: **Still blocking on re-review:** `Load` continues to accept unsupported versions, unknown policies, duplicate roots, and other invalid committed declarations. `EnsureRootWithin` only validates path containment; `pullArtifacts` can silently skip an unknown policy in normal sync, pull it under `--artifacts-only`, and process duplicate roots twice. Validate the complete schema at load/consume time and fail closed before any root is used.

internal/artifacts/pull.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+// safeToReplace reports whether the live file at rel may be overwritten: it is
+// safe only when the live destination is absent, or present and identical to
+// the recorded baseline. A file present locally but not in the baseline was
+// created out-of-band (protect it); a file whose current state differs from
+// the baseline was modified since the last transfer (a conflict).
+func safeToReplace(destAbs, rel string, base map[string]FileEntry) bool {
+	info, err := os.Lstat(filepath.Join(destAbs, filepath.FromSlash(rel)))
+	if err != nil {
```
obey-agent, line 240: **Blocking containment issue:** `safeToReplace` treats every `Lstat` error as “absent,” but the subsequent `MkdirAll` and `Rename` can follow a symlinked parent. For example, if `destAbs/out` is a symlink to an outside directory whose target does not yet exist, `Lstat(destAbs/out/file)` returns `ENOENT`, this returns true, and `MkdirAll` writes through the symlink before the rename. Only `os.IsNotExist` for the complete path is safe; validate every destination parent component without following symlinks and fail closed on other errors.

internal/artifacts/pull.go
```diff
@@ -0,0 +1,390 @@
...
+	//
+	// --compare-dest=<live root> keeps this a delta transfer: rsync compares
+	// each source file against the live copy and only writes the ones that
+	// changed or are new into the staging tree, so unchanged multi-GB media
+	// is not re-copied. --no-links keeps peer symlinks from becoming phantom
+	// local state. No --partial: rsync writes each file to a temp name and
+	// renames on completion, and drops the in-flight temp on interruption, so
+	// the staging tree only ever holds complete files (staging is per-pull and
+	// wiped, so cross-run resume would buy nothing).
+	stagingDir := filepath.Join(campaignRoot, ".campaign", "cache", "rsync-staging", snapshotSlug(rootRel))
```
obey-agent, line 122: **Blocking concurrency issue:** the staging path is deterministic per campaign/root, so concurrent `camp sync --from` processes race on `RemoveAll(stagingDir)`, rsync, merge, and snapshot writes. One invocation can delete or consume another’s staged files and then record an incomplete baseline. Use a per-pull unique staging directory and serialize the live merge/snapshot section with a campaign/root lock (or explicitly reject concurrent pulls with a tested lock).

obey-agent (commented):
## Verdict: Request Changes

## Re-review

The staged-transfer follow-up is a substantial improvement and closes the original live-tree TOCTOU in the normal single-process path: rsync writes outside the artifact root, then staged files are revalidated before merge. The current implementation still has schema, containment, and concurrent-execution blockers.

## Key Findings
- **Blocking:** `artifacts.Load` still accepts unsupported versions, unknown policies, duplicate roots, and other invalid committed declarations; consume-time path validation is not complete schema validation. See `internal/artifacts/config.go:76`.
- **Blocking:** `safeToReplace` treats every `Lstat` error as absence, allowing `MkdirAll` to follow a broken symlinked parent and write staged content outside the campaign. See `internal/artifacts/pull.go:240`.
- **Blocking:** the deterministic staging directory is shared by concurrent pulls, so `RemoveAll`, rsync, merge, and snapshot writes can race and produce an incomplete baseline. See `internal/artifacts/pull.go:122`.
- **Follow-up:** the new filesystem-mutating lifecycle tests remain in the host unit suite instead of the container harness required by the repository safety standard.

## What’s Done Well
- Staging preserves the delta-transfer property with `--compare-dest`.
- Late edits and out-of-band files are covered by merge tests and remain protected.
- Atomic file moves and sticky baseline handling are integrated cleanly.
- The earlier mtime, rsync quoting, snapshot, exit-contract, and read-path fixes remain present.

## Staff Standard
Not yet. I would submit this after schema validation, symlink-safe destination handling, and concurrent-pull serialization are fail-closed and tested.

obey-agent:
### Re-review status at head 7e5db830

All code findings in the latest pass are already addressed at this head (they re-anchored from earlier SHAs before the fixes landed):

- **config schema validation** — `File.Validate()` rejects unsupported version, unknown policy, and duplicate roots, and is called at both consume paths (`pullArtifacts`, `verifyArtifacts`) before any root is used. Unknown policy no longer silently skips on normal sync / pulls under `--artifacts-only`.
- **root symlink containment** — `EnsureRootWithin` EvalSymlinks the root and rejects any component resolving outside the campaign.
- **merge-path symlinked parent** — `mergeStaged` runs `destWithinRoot` (resolves the destination parent, fails closed on escape) before `MkdirAll`/rename, so a symlinked parent cannot redirect a write outside the root.
- **TOCTOU no-clobber** — the pull stages the delta outside the live root (`--compare-dest`) and atomically merges each file only after re-checking the live destination against the baseline.
- **concurrent pulls** — a per-root `fsutil.AcquireFileLock` (stale-safe, timed) serializes the stage/merge/snapshot section; different roots still run concurrently.
- **artifacts-only config failure** — a malformed/unreadable/invalid config sets `Success=false` with a structured `SyncError` under `--artifacts-only`; warning-only degradation is kept for the optional phase of a normal sync.
- **cancellation during rsync** — `ctx.Err()` is re-checked immediately after each `Pull`, so a cancelled sync fails regardless of mode.

**Still open — test containerization.** The FS-mutating artifact tests (transfer lifecycle, snapshot) invoke real rsync in host TempDir. Moving them into the container harness requires an ssh-reachable peer inside the container: `camp sync --from` only resolves ssh machines from `~/.obey/machines.yaml`, and the unit tests use the `peer.FromPath` filesystem shortcut precisely to stay driveable without ssh. So this is the deferred localhost-ssh-smoke integration slice (tracked as a campaign intent), not a change to this PR's engine. Pure-logic cases (merge revalidation with in-memory input, snapshot-slug injectivity, peer-id validation, config validation, manifest comparison) are already host-side unit tests.

PR #428, opened 2026-07-15
test(integration): loopback-ssh smoke coverage for peer transports (+ rsync secluded-args fix)

lancekrogers:
## What

Stands up the deferred **localhost-ssh-smoke integration slice** for camp's peer transports (`camp sync/clone --from <machine>`, artifact rsync pull), driving all three end to end over a **real loopback ssh hop** inside the container harness. The filesystem-peer unit tests (`peer.FromPath`) never dial ssh, so the entire ssh transport — auth, ControlMaster multiplexing, `GIT_SSH_COMMAND`, rsync `-e`, remote-path handling, and genuine A→B content movement — was previously unexercised.

Building it surfaced and fixed a real transport bug (below).

## The bug it found

`camp sync --from` artifact pulls were **broken on every machine with rsync ≥ 3.2.4** (Homebrew macOS, current Linux):

```
rsync: [sender] change_dir "/home/peer/'/home/peer/campaigns/media-camp/media'" failed (2)
```

rsync ≥ 3.2.4 defaults to secluded/protected args and no longer runs the remote path through a login shell, so `peer.RsyncSpec`'s `ShellQuote` injected **literal** single quotes that corrupted the path. Confirmed across variants in a throwaway container (quoted → fail; unquoted → ok; `-s` + spaced path → ok).

**Fix:** `RsyncSpec` returns the remote path unquoted and the artifact pull passes `-s`/`--secluded-args`, which protects spaces and shell metacharacters over the protocol with no shell parsing on either end. Requires rsync ≥ 3.0.0 (2008; every current build); adaptive quoting for the ancient rsync 2.6.9 Apple bundle stays the already-deferred rsync/openrsync version-probing work.

## Tests added (all pass in-container)

- **`artifacts_from_ssh_test.go`** — `camp sync --artifacts-only --from self`: bytes move, a baseline snapshot is recorded, and **no-clobber holds on a divergent second pull** (locally-edited file kept; peer-only change applied). This is the case that surfaced the quoting bug.
- **`sync_from_ssh_test.go`** — `camp sync --from self --git-only`: a submodule commit that exists only on the peer is fetched into `refs/peer/<id>/*` over ssh, while the checkout stays on origin's tip (objects-only transport).
- **`clone_from_ssh_test.go`** — `camp clone <url> <dir> --from self`: the campaign is seeded from the peer over ssh, then origin is re-pointed to the canonical url.
- **`helpers_ssh.go`** — reusable loopback-peer harness: an unprivileged `peer` login user with its own HOME/registry (so one campaign name resolves to two distinct roots — a single registry would make every transfer a no-op self-copy). Idempotent for the pooled container.
- `remote_switch_ssh_test.go` refactored to reuse the shared sshd provisioner (no behavior change).

## Scope / not in scope

This closes the genuine gap (no real-transport coverage; the transport bug). It does **not** drain the host-FS test ratchet allowlist for `internal/clone`/`internal/sync` — migrating those unit tests is a separate mechanical follow-up, still tracked; the pure-logic unit tests stay unit.

## Gates

`just lint` clean (0 issues; host-FS ratchet clean; no fmt.Errorf regressions). Unit tests green. Integration tests run under `-tags=integration` (require Docker; CI is intentionally paused for cost, run locally).

### DIFF
--- a/internal/peer/peer.go
+++ b/internal/peer/peer.go
@@ -149,20 +149,27 @@ func (s *Source) GitEnv() []string {
 // the peer's campaign root, with a trailing slash so rsync copies contents
 // into the destination directory rather than nesting it.
 //
-// For an ssh source the path after "host:" is handed to the peer's remote
-// login shell, so it is single-quoted: a legitimate media root with a space
-// ("Final Renders/") would otherwise be split into two source arguments, and
-// shell metacharacters in a committed artifacts.yaml path would be
-// interpreted remotely. The trailing slash stays outside the quotes so the
-// local side still sees a copy-contents (not copy-dir) source. Filesystem
-// sources are passed to rsync as a local argv element (no shell), so they are
-// returned unquoted.
+// The remote path is NOT shell-quoted. rsync 3.2.4+ defaults to secluded
+// (protected) args and no longer hands the remote path to a login shell, so a
+// caller-added quote is taken literally: `host:'/media'/` makes the remote
+// side look for a directory whose name begins with a single quote and fails
+// with change_dir(2). Spaces and shell metacharacters in a committed
+// artifacts.yaml path ("Final Renders/") are instead kept safe by the
+// -s/--secluded-args flag the artifact pull passes, which transmits each arg
+// over the rsync protocol with no shell parsing on either end. The trailing
+// slash keeps this a copy-contents (not copy-dir) source. Filesystem sources
+// are a local argv element (no shell) and are likewise returned unquoted.
+//
+// Tradeoff: -s and this unquoted form require rsync >= 3.0.0 on both ends
+// (2008; every current Homebrew/Linux build). rsync 2.6.9, the ancient Apple
+// bundle, would need the old shell-quoted form instead; version-adaptive
+// quoting is the deferred rsync/openrsync probing work, not this path.
 func (s *Source) RsyncSpec(relPath string) string {
 	p := path.Join(s.root, relPath)
 	if s.target == "" {
 		return p + "/"
 	}
-	return s.target + ":" + remote.ShellQuote(p) + "/"
+	return s.target + ":" + p + "/"
 }
 
 // Fetch fetches heads and HEAD from the peer copy of the repository at
--- a/internal/peer/peer.go
+++ b/internal/peer/peer.go
@@ -149,21 +149,10 @@ func (s *Source) GitEnv() []string {
 // the peer's campaign root, with a trailing slash so rsync copies contents
 // into the destination directory rather than nesting it.
 //
-// The remote path is NOT shell-quoted. rsync 3.2.4+ defaults to secluded
-// (protected) args and no longer hands the remote path to a login shell, so a
-// caller-added quote is taken literally: `host:'/media'/` makes the remote
-// side look for a directory whose name begins with a single quote and fails
-// with change_dir(2). Spaces and shell metacharacters in a committed
-// artifacts.yaml path ("Final Renders/") are instead kept safe by the
-// -s/--secluded-args flag the artifact pull passes, which transmits each arg
-// over the rsync protocol with no shell parsing on either end. The trailing
-// slash keeps this a copy-contents (not copy-dir) source. Filesystem sources
-// are a local argv element (no shell) and are likewise returned unquoted.
-//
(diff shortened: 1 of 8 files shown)


internal/peer/peer.go
```diff
@@ -149,20 +149,16 @@ func (s *Source) GitEnv() []string {
...
 //
-// For an ssh source the path after "host:" is handed to the peer's remote
-// login shell, so it is single-quoted: a legitimate media root with a space
-// ("Final Renders/") would otherwise be split into two source arguments, and
-// shell metacharacters in a committed artifacts.yaml path would be
-// interpreted remotely. The trailing slash stays outside the quotes so the
-// local side still sees a copy-contents (not copy-dir) source. Filesystem
-// sources are passed to rsync as a local argv element (no shell), so they are
-// returned unquoted.
+// The remote path is intentionally unquoted. rsync 3.2.4 added shell escaping
```
obey-agent, line 152: Fixed in 5af0076: the earlier description incorrectly treated rsync 3.2.4 as switching to secluded args. The updated comment now reflects its shell escaping behavior, explains why -s is explicit, and records the rsync >=3.0 compatibility requirement.

tests/integration/artifacts_from_ssh_test.go
```diff
@@ -0,0 +1,76 @@
...
+// TestArtifactsPullOverSSH_TransfersAndProtects verifies SSH artifact transfer,
+// including a spaced root and no-clobber behavior.
+func TestArtifactsPullOverSSH_TransfersAndProtects(t *testing.T) {
+	tc := GetSharedContainer(t)
+	ensurePeerAccount(t, tc)
+	registerLoopbackMachine(t, tc)
+
+	const (
+		name         = "media-camp"
+		artifactRoot = "Final Renders"
```
obey-agent, line 22: Fixed in 5af0076: this integration fixture now uses a spaced Final Renders root and writes the local edit with shell-safe quoting, so it exercises the actual rsync path regression and verifies exact bytes.

obey-agent (commented):
Verdict: Approve\n\nOverview: The PR adds valuable end-to-end loopback SSH coverage for peer git sync, clone, artifact pulls, and remote switching, and fixes the rsync remote-path regression. During review I tightened the implementation-facing comments and made the artifact test exercise the actual spaced-path case instead of only asserting the generated spec.\n\nKey Findings:\n- Fixed in 5af0076: corrected the rsync 3.2.4 documentation and retained the explicit rsync >=3.0 compatibility requirement.\n- Fixed in 5af0076: the artifact integration test now uses a spaced Final Renders root, shell-safe writes, and exact-byte assertions.\n- Fixed in 5af0076: removed redundant explanatory comments from the new integration harness and tests.\n\nWhat's Done Well:\n- The tests use a distinct unprivileged peer HOME and registry, proving real source-to-destination transfers rather than self-copy behavior.\n- The shared SSH setup is idempotent and reused by the existing remote-switch smoke test.\n- Git transport and rsync transport are both exercised over the real SSH path.\n\nValidation:\n- go test ./... -count=1\n- go vet ./...\n- just lint\n- just build quick\n- just docs\n- Docker-backed focused SSH integration suite: artifact pull, clone, sync, and remote switch all passed\n\nStaff Standard: Yes — ready to merge.

obey-agent (approved):
## Verdict: Approve

## Overview
Closes a real transport gap: peer SSH paths were only unit-tested via filesystem peers, so the rsync ≥3.2.4 secluded-args quoting bug went unnoticed until loopback SSH smoke coverage. The production fix (unquoted remote path + `-s`/`--secluded-args`) is correct for modern rsync, and the integration suite now exercises artifacts, git-only sync, clone, and remote-switch over a distinct peer HOME/registry.

## Key Findings
No new blocking issues. Prior review iteration (spaced artifact roots, comment cleanup) is reflected on head `5af0076`.

## What's Done Well
- Bug found by the tests, not invented for them
- Peer account isolation avoids self-copy no-ops
- Objects-only git peer semantics asserted (checkout stays on origin tip)
- Shared `ensureSSHDaemon` / peer helpers keep the suite maintainable

## Staff Standard
Yes — merge. Run the Docker-tagged integration suite before release if CI still skips it for cost.

PR #432, opened 2026-07-15
fresh: follow-up command workflows and configure subcommand

lancekrogers:
## Summary

Extends `camp fresh` with post-sync follow-up command workflows, addressing the recurring pain of running an install/build/bootstrap command after every fresh cycle (intent camp-fresh-workflow-configurator-20260712-144017).

- `.campaign/settings/fresh.yaml` gains a `follow_up` list: a global default plus optional per-project overrides. Each entry has `name`, `run`, optional `dir` (relative to the project root), and optional `continue_on_error` (default false). A project override replaces the global list entirely rather than merging with it, matching the existing precedence pattern for `branch` and `push_upstream`.
- `camp fresh` runs the resolved follow-up steps, in order, after a successful sync/prune/branch cycle, streaming each step's output with a per-step header and a done/failed status line. A step that fails without `continue_on_error` aborts the remaining steps and fails the command; a step with `continue_on_error: true` logs the failure and moves on.
- `--no-follow-up` skips follow-ups for a single run. `--dry-run` lists configured follow-ups (name and command) without executing any of them, consistent with how dry-run previews the rest of the cycle.
- New non-interactive `camp fresh configure` subcommand group: `show` (lists global and per-project follow-ups), `add <name> --run <cmd> [--project X] [--dir D] [--continue-on-error]`, and `remove <name> [--project X]`. No interactive TUI in this PR, per the pinned v1 scope.
- Config writes create `.campaign/settings/fresh.yaml` if missing, reject an empty `run` (or empty name), reject adding a duplicate name in the same scope, and report a clear error listing the valid names in scope when removing an unknown name. Writes preserve every key they do not touch (including a best-effort preservation of a comment-only file's header, since the scaffolded default fresh.yaml ships fully commented out and yaml.v3 has no node to attach top-level comments to when there is no mapping content).

## Why

`camp fresh` already resets a project to a clean branch state, but almost every real workflow needs a follow-up step afterward (`npm install`, `go build`, a bootstrap script). Users were re-running that command by hand every time. This lets them configure it once, globally or per project.

## Verification

- Unit tests: `internal/config/fresh_followup_test.go` covers `FollowUpConfig.Validate`, `ResolveFreshFollowUps` precedence (project override replaces global entirely, including an explicit empty override suppressing global follow-ups), `AddFreshFollowUp`/`RemoveFreshFollowUp` (file creation, project scoping, duplicate rejection, unknown-name errors with the valid-name list, pruning now-empty `follow_up`/`projects` containers, preserving an unrelated top-level key across a write, and preserving a comment-only header on first write), and context cancellation.
- Integration tests (containerized, Docker via Colima): `tests/integration/fresh_followup_test.go` covers the full `configure show/add/remove` round trip including duplicate and unknown-name rejection, follow-up execution order via a real project directory, `--no-follow-up` skipping execution, `--dry-run` listing without executing, `continue_on_error` allowing later steps to run after a failure, a failing step without `continue_on_error` stopping the remaining steps, project-override-replaces-global precedence end to end, and follow-ups running per project in a `camp fresh --list` batch.
- Ran `DOCKER_HOST="unix://$HOME/.colima/default/docker.sock" TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock go test -tags=integration ./tests/integration/... -run "TestFreshFollowUp|TestFreshConfigure|TestFresh_"`: all 20 tests passed (9 new, 11 pre-existing fresh tests unaffected).
- Ran `just gate` at the worktree root: stable and dev builds succeeded, `go vet` clean on stable/dev/integration tags, `just lint` reported 0 issues on both profiles with both ratchets (`lint-no-host-fs-tests`, `lint-no-fmt-errorf`) clean, and the full unit suite passed on both profiles (3378/3378 dev, 3349/3349 stable).
- Ran `just docs` to regenerate the CLI reference and committed it separately from the feature commit, since the regen also picked up drift from other already-merged commands unrelated to this change.

## Known limitations

- No interactive TUI configurator; that is intentionally out of scope for this PR per the pinned intent notes.
- The persistent flags on `camp fresh` (`--branch`, `--dry-run`, `--no-branch`, `--no-push`, `--no-prune`) are inherited by `camp fresh configure` and its subcommands via cobra's persistent-flag cascade, even though they have no effect there. This is cosmetic (visible in `--help` output) and not new to this change; scoping them out would require restructuring how the parent command declares its flags.
- If `.campaign/settings/fresh.yaml` contains only comments and no actual keys (the scaffolded default), the first `configure add`/`remove` write preserves the original bytes verbatim ahead of the new content rather than merging comments node-by-node, since yaml.v3 does not retain top-level comments when there is no mapping to attach them to. Subsequent writes go through the normal node-preserving path once real content exists.
- `camp fresh configure show` only prints project sections that have a non-empty `follow_up` list, so an explicit empty per-project override (`follow_up: []`), which suppresses the global follow-ups for that project, is not visible in `show` output. Follow-up candidate for the interactive configurator pass.

### DIFF
--- a/internal/commands/fresh/configure.go
+++ b/internal/commands/fresh/configure.go
@@ -1,0 +1,179 @@
+package fresh
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/campaign"
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/ui"
+)
+
+// newConfigureCommand builds the non-interactive `camp fresh configure`
+// subcommand group for managing follow-up command workflows stored in
+// .campaign/settings/fresh.yaml.
+func newConfigureCommand() *cobra.Command {
+	configureCmd := &cobra.Command{
+		Use:   "configure",
+		Short: "Manage camp fresh follow-up command workflows",
+		Long: `Manage the follow-up command workflows camp fresh runs after a
+successful sync/prune/branch cycle. Configuration lives in
+.campaign/settings/fresh.yaml: a global default list, plus optional
+per-project override lists that replace the global list entirely.
+
+Examples:
+  camp fresh configure show
+  camp fresh configure add install --run "npm install"
+  camp fresh configure add build --run "go build ./..." --project camp --dir cmd/camp
+  camp fresh configure remove install
+  camp fresh configure remove build --project camp`,
+	}
+
+	configureCmd.AddCommand(newConfigureShowCommand())
+	configureCmd.AddCommand(newConfigureAddCommand())
+	configureCmd.AddCommand(newConfigureRemoveCommand())
+
+	return configureCmd
+}
+
+func newConfigureShowCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "show",
+		Short: "Show configured follow-up workflows",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			ctx := cmd.Context()
+
+			campRoot, err := campaign.DetectCached(ctx)
+			if err != nil {
+				return camperrors.Wrap(err, "not in a campaign")
+			}
+
+			cfg, err := config.LoadFreshConfig(ctx, campRoot)
+			if err != nil {
+				return camperrors.Wrap(err, "loading fresh config")
+			}
+
+			printFollowUpSection("Global", cfg.FollowUp)
+
+			names := make([]string, 0, len(cfg.Projects))
+			for name, pc := range cfg.Projects {
+				if len(pc.FollowUp) > 0 {
+					names = append(names, name)
+				}
+			}
+			sort.Strings(names)
+
+			for _, name := range names {
+				fmt.Println()
+				printFollowUpSection("Project: "+name, cfg.Projects[name].FollowUp)
+			}
+
+			return nil
+		},
+	}
+}
+
+func printFollowUpSection(label string, steps []config.FollowUpConfig) {
+	fmt.Println(ui.Header(label))
+	if len(steps) == 0 {
+		fmt.Println(ui.Dim("  (none configured)"))
+		return
+	}
+	for _, step := range steps {
+		line := fmt.Sprintf("  %s  %s", ui.Value(step.Name), step.Run)
+		if step.Dir != "" {
+			line += " " + ui.Dim(fmt.Sprintf("(dir: %s)", step.Dir))
+		}
+		if step.ContinueOnError {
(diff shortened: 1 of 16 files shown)


internal/config/fresh_followup.go
```diff
@@ -0,0 +1,326 @@
+package config
```
obey-agent, line 1: **Follow-ups are intentional shell execution after a successful fresh cycle.** Treat configured `run` strings as trusted campaign config (same trust model as hooks). Empty --run rejection, --no-follow-up, --dry-run list-only, continue_on_error vs hard-stop, and project-list-replaces-global are all the right operator controls and are covered by integration tests.

internal/commands/fresh/configure.go
```diff
@@ -0,0 +1,179 @@
+package fresh
```
obey-agent, line 1: **configure show|add|remove** keeps operators out of hand-editing YAML while still allowing it. Duplicate name rejection and unknown-remove listing valid names are good UX for a CLI that agents will also call.

obey-agent (commented):
## Verdict: Approve

## Overview
Adds `camp fresh` post-sync follow-up workflows and a `fresh configure` subcommand group (show/add/remove), with per-project override lists that fully replace global lists. Strong containerized coverage for order, dry-run, skip, continue_on_error, and hard-stop.

## Key Findings
1. **Trust model:** follow-up `run` is shell execution from campaign settings. That is appropriate for a local operator tool; do not ever auto-import untrusted remote step lists without review.
2. **Docs churn** again on cli-reference — coordinate with #433/#434 merges.
3. No functional blockers.

## What's Done Well
- Clear separation of configure CRUD vs execution path
- Dry-run previews without executing; --no-follow-up skips without affecting sync
- Project override replaces rather than merges (documented and tested)
- Scaffold template comments teach the schema in-place

## Staff Standard
Yes.

obey-agent:
Scripted CLI visualization for PR #432 — this is not a PTY/TUI recording. It illustrates the non-interactive follow-up commands added here.

The human-facing TUI follow-up is in [PR #440](https://github.com/Obedience-Corp/camp/pull/440), which includes a real VHS recording from the camp binary.

PR #432 scripted CLI demo

obey-agent (commented):
## Verdict: Approve (re-review after rebase)

## Overview
Re-reviewed current head `d78ff22` after rebase onto main. Product change is still: `camp fresh` follow-up workflows + `fresh configure` CRUD, with strong integration coverage. Post-review commits are docs regen + em-dash template cleanup only.

## Key Findings
- No new functional issues since prior review.
- Stack note still applies: shared `docs/cli-reference/*` regeneration will conflict with sibling camp PRs if landed out of order.

## What's Done Well
- Prior review conclusions still hold (dry-run, --no-follow-up, continue_on_error, project override replaces global).

## Staff Standard
Yes. Safe to merge on this head from a code-review perspective.

PR #435, opened 2026-07-15
Broaden .workitems.jsonl ledger adoption across workitem lifecycle commands

lancekrogers:
## Summary

Routes every directory-workitem lifecycle mutation (create, dungeon move/triage, promote, adopt, gather, and the deprecated shelve alias) through the campaign-wide workitem ledger at `.campaign/workitems/.workitems.jsonl`, reusing the existing event envelope in `internal/workitem/audit`. Implements the intent at `.campaign/intents/active/broaden-workitemsjsonl-ledger-adoption-across-20260710-000415.md`, whose source spec is `workflow/design/dungeon/completed/2026-07-11/camp-graph-overhaul-2026-07-08/00-derivability-audit.md` (gap 2).

That audit found the workitem ledger nascent (3 events measured 2026-07-09) versus the mature intent (253 events) and festival (504 events) ledgers, forcing camp-graph to fall back to slow, lossy git-rename tracking of `.workitem` markers to derive workitem movement across the bug, explore, design, and audit workflows.

## What changed

- Added `wkaudit.AppendBestEffort`, the single canonical "construct and append, warn instead of fail" entry point. All lifecycle mutations in this intent's scope (create, move, promote, adopt, dungeon transitions) now route through it, directly or via the existing `appendWorkitemAuditEvent` wrapper, instead of each command reimplementing best-effort semantics. `gather` (outside this scope) keeps its own direct `wkaudit.AppendEvent` calls with warning accumulation into execution.Warnings; it was already best-effort and is unchanged here.
- Added a new `move` event type, and a shared `recordWorkitemMove` helper in `cmd/camp/dungeon` that both the general `camp dungeon move` (triage/status) path and the `--workitem` path call right after their filesystem move succeeds. Non-workitem items (no `.workitem` marker) are silently skipped so the ledger is never polluted; any future move entry point added to that package inherits ledger coverage automatically by calling the same helper. `workitemLedgerPathIfExists` guards staging the ledger into the move's commit, since `git add` hard fails the whole staging call if any pathspec has never been created.
- Fixed two correctness bugs discovered while broadening adoption:
  - `camp workitem promote` previously returned the ledger append's error, failing the whole command if the write failed even though the promote had already succeeded on disk. It is now best-effort, matching the intent's explicit requirement and the pattern create/adopt/gather already use.
  - `camp workitem promote` recorded the promote event's `id` as the directory slug rather than the real generated workitem id/ref used by create/adopt/gather events, so a single workitem's promote event could not be correlated with its earlier lifecycle events by id. Metadata is now read from the source directory before any promote branch relocates or removes it; the ledger event uses the real id/ref/title while the CLI `--json` `id` field (contractually the slug, locked in by an existing test) is unchanged.
- `camp shelve`, the deprecated alias that shares the `MoveToDungeon` mutation with `camp workitem promote --target <status>`, previously wrote no ledger event at all. It now records the same "promote" event type, correlated by the real workitem id.

## Why this design

The shared FS-mutation choke point for `camp dungeon move` in both its forms is `internal/dungeon.Service.ApplyMove` and the `cmd/camp/dungeon` command layer wrapping it; `camp workitem promote` and `camp shelve` share a different mutation (`internal/commands/workitem.MoveToDungeon`) but are distinct commands with distinct ledger semantics (a generic "move" versus an explicit "promote"), so they get their own call sites rather than a single embedded hook inside the FS layer, which would have either duplicated `EventPromote` for promote's dungeon-status target or required guessing at the caller's intent from inside a generic mutation function.

## Verification

- `just gate` (build stable+dev, vet stable+dev+integration, lint stable+dev, unit tests both profiles): PASSED, 3370 tests (dev profile) and 3341 tests (stable profile), 0 lint issues.
- Full `just test integration` (containerized suite via Docker/Colima): 5 pre-existing failures unrelated to this change (`TestIntegration_WorkitemCommits_FiltersFalsePositives`, `TestIntegration_WorkitemCommits_IncludesCampaignAndLinkedProject`, `TestIntegration_WorkitemCommits_CrossRepo`, `TestIntegration_CommitTags_NoteNoContext`, `TestIntegration_WorkitemValidate_ReportsFindings`), verified to reproduce identically on a clean `origin/main` checkout before any of this change's code existed. Every other test in the suite, including all existing dungeon/promote/gather tests and their "git status must be clean after move" assertions, passed with this change applied.
- New containerized integration tests in `tests/integration/workitem_ledger_test.go`: drive create -> dungeon move (`--workitem`) -> promote end to end and assert all three ledger events share the same real workitem id; cover the general (non `--workitem`) triage move used from inside a `workflow/<type>/` directory; cover the shelve alias; and lock in that moving a plain, never-adopted file does not create the ledger file at all.
- New unit tests in `internal/workitem/audit` and `cmd/camp/dungeon` cover event construction (including the new "move" type) and `AppendBestEffort`'s failure-does-not-propagate contract.

## Known limitations

- The ledger event identity enrichment (real id/ref/title from `.workitem`) falls back to the slug when the marker is missing or unreadable, matching promote's pre-existing behavior of allowing a raw, never-adopted directory to be promoted; such promotes still record under the slug rather than a generated id, same as before this change.
- Ledger writes remain best-effort by design; a rare failure to create or append the ledger file leaves the underlying mutation applied but the event unrecorded, matching the pattern already shipped for `camp gather`.
- `internal/commands/workitem/promote.go` was already close to the repo's 500-line guideline before this change (502 lines) and is now at 514; splitting it further is out of scope for this ledger-adoption change.


cmd/camp/dungeon/workitem_ledger.go
```diff
@@ -0,0 +1,58 @@
...
+// recordWorkitemMove is the single shared code path every dungeon move
+// command in this package routes through after the filesystem mutation
+// succeeds, so new move entry points inherit campaign-wide workitem ledger
+// coverage without re-implementing event construction. It checks the
+// destination for .workitem metadata (the directory, including its marker,
+// has already been relocated by the time this runs) and appends a "move"
+// event when the moved item is a tracked workitem. Items without a marker
+// (plain triage files or directories never adopted with camp workitem
+// create/adopt) are silently skipped; a present-but-unreadable marker warns
+// instead of failing the caller's already-applied move.
```
obey-agent, line 22: **Shared post-move ledger path is the right integration point.** Recording after the FS mutation succeeds, best-effort warn-only on ledger failure, and staging the ledger file only when it already exists (so `git add` does not fail the commit for non-workitem moves) are all production-minded details. Silent skip for unmarked triage dirs is correct.

internal/workitem/audit/audit.go
```diff
@@ -68,3 +74,14 @@ func AppendEvent(ctx context.Context, campaignRoot string, e Event) error {
 	}
 	return nil
 }
+
+// AppendBestEffort is the single shared entry point every FS-mutating
+// workitem command routes through to record a ledger event. A ledger write
+// failure never fails the caller's already-applied filesystem mutation; it
+// is reported to warn instead, so degraded ledger coverage is visible
+// without stranding an operator behind a spurious command failure.
+func AppendBestEffort(ctx context.Context, warn io.Writer, campaignRoot string, e Event) {
```
obey-agent, line 83: **AppendBestEffort contract is explicit and tested.** Ledger degradation must never strand an operator after a successful move; the ENOTDIR fixture is a clever way to force MkdirAll failure without root.

tests/integration/workitem_ledger_test.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+		var e map[string]any
+		require.NoError(t, json.Unmarshal([]byte(line), &e), "unmarshal ledger line: %s", line)
+		events = append(events, e)
+	}
+	return events
+}
+
+// TestWorkitemLedger_CreateMovePromoteSharedTrail drives create -> dungeon
+// move -> promote for a single directory-style bug workitem and asserts the
```
obey-agent, line 33: **Integration trail create → dungeon move → promote** correlating by real workitem id (not slug) is the acceptance bar that was missing when only promote wrote the ledger.

obey-agent (commented):
## Verdict: Approve

## Overview
Extends campaign-wide `.workitems.jsonl` coverage to dungeon move paths (batch + single) and softens promote audit writes to the same best-effort append used for moves, so FS mutations never fail solely because the ledger could not be written. Strong unit + integration coverage.

## Key Findings
- No blocking issues.
- Promote now uses best-effort ledger append (was hard-fail). That is a deliberate consistency choice and matches the new move path; call it out in release notes if operators relied on promote failing on ledger IO.
- Staging the ledger only when the file exists is the correct `git add` pathspec hygiene.

## What's Done Well
- Single `recordWorkitemMove` for all dungeon move entry points
- EventMove kind + AppendBestEffort shared helper
- Containerized create/move/promote correlation test

## Staff Standard
Yes.

obey-agent:
Termcast demo for this PR — watch workitem lifecycle creation, move/promote transitions, and the correlated ledger output.

PR #435 demo

obey-agent:
**Updated to merge (at `29e28319`)** — merged `main` in (was 18 behind). Clean merge, `promote.go` auto-resolved. CLI reference regenerated for main's newly landed commands. `just gate` green (3397 dev + 3368 stable unit, both ratchets).

Issue #274, opened 2026-05-11
Test gap: cmd-layer pin/nav attachment helpers need integration coverage


lancekrogers:
## Gap

PR #270 introduced cmd-layer orchestration for non-project attachment pins. Three helpers have non-trivial logic but are exercised only via end-to-end smoke (not CI):

- `cmd/camp/pins.go:buildPinForPath` — distinguishes in-tree pins (writes campaign-relative `Path`) from attachment pins (writes `AbsPath`) via `filepath.Rel` plus a marker walk.
- `cmd/camp/pins.go:markerResolvesToCampaign` — walks up from an external path looking for a `.camp` marker whose `ActiveCampaignID` resolves back to the active campaign root.
- `cmd/camp/pins.go:findPinForCwd` — looks up pins by `Path`, `AbsPath`, or the campaign-relative form so `camp unpin` (no args) works from both in-tree and attached external dirs.
- `cmd/camp/navigation/go.go:resolvePin` — the attachment branch (`pin.IsAttachment() → return pin.AbsPath`) has no direct test.

The service layer (`internal/attach`, `internal/git`, `internal/campaign`, `internal/pins`) is well covered. A refactor that silently breaks the cmd-layer attachment-pin path would not be caught by CI today.

## Testing constraint

Per the campaign-wide testing policy (CLAUDE.md §11), filesystem-mutating tests must run in a **containerized integration harness**, not as host-side unit tests using `t.TempDir()`. The behaviors above all need real `.camp` markers, real symlinks, and (for `markerResolvesToCampaign`) a real registry file to exercise honestly. That means the coverage belongs in `tests/integration/`, not in a new `cmd/camp/pins_test.go`.

## Suggested scope

Add integration tests under `tests/integration/`:

1. **`attachment_pin_flow_test.go`** — end-to-end: init campaign, create symlink, `camp attach`, `cd` into target, `camp pin`, `camp pins` (verify `KIND=attachment`), `camp go --print`, `camp unpin`, `camp detach`. Asserts on stdout and on the persisted `pins.json` shape (Path vs AbsPath).
2. **`attachment_pin_negative_test.go`** — rejection paths: pin from un-attached external dir, attach inside a campaign tree, attach inside *another* campaign (security regression for the cross-campaign shadow fix).
3. **`attachment_cgo_resolution_test.go`** — `cgo`/pin resolution returns the absolute path for an attachment pin, and surfaces a clean error when the target directory has been moved out from under the pin.

[...]

PR #437, opened 2026-07-15
Hidden .dungeon by default, camp idea alias

lancekrogers:
Hidden `.dungeon` for new campaigns, a migration for existing ones, and `camp idea` as the primary name for `camp intent`.

## The model: migrate, do not coexist

**A campaign has exactly one dungeon spelling. Both present is a broken state, not a supported one.**

This PR originally shipped coexistence: when both `dungeon/` and `.dungeon/` existed, resolution preferred the visible one and warned. That rule is gone, for two reasons.

**Coexistence hid data.** Preferring visible meant anything already filed under `.dungeon/` became invisible to every listing. A one-time stderr warning was the only signal that completed work had dropped out.

**The old rule manufactured the state it then tolerated.** "New dungeon directories follow the setting" guaranteed that a legacy campaign on an upgraded machine would accrete `.dungeon/` dirs one at a time next to its `dungeon/` dirs. Reproduced on this branch before the amendment: `camp dungeon add` in a fresh subdirectory of an all-visible campaign created `.dungeon` alongside 11 `dungeon/` dirs. Mixed campaigns were not an edge case under that design; they were the destination.

Migration is cheap here because camp is filesystem plus git with no database: the conversion is a `git mv` sweep, history-preserving and revertible.

## What changed

**Both spellings under one parent is now an error.** `spelling.Resolve` returns a `ConflictError` naming the location and the command that fixes it. Every dungeon-resolving path already handled `Resolve`'s error, so the failure reaches the user structurally rather than by remembering to check. Two call sites were swallowing it and silently reading the visible spelling (the intents store and workitem locate); both now propagate. The failure is scoped to the parent being resolved, so a conflict in one location cannot break commands that never touch it.

**`dungeon_hidden` governs only `camp init` for a new campaign.** Inside an existing campaign, the campaign's own layout decides. `CampaignName` answers that from the root dungeon, which `camp init` always scaffolds and which `camp dungeon migrate` converts along with everything else, falling back to a full sweep and then to the setting for a campaign with no dungeon at all. Two stats in the common case.

**`camp dungeon migrate`** sweeps every dungeon in the campaign and converts it, in one revertible commit. `--dry-run` and `--no-commit` are supported.

- **Discovered, not hardcoded.** This campaign has about a dozen: the root, `festivals/`, `.campaign/intents/`, `.campaign/quests/`, and one per `workflow/` type. Locations added later are covered for free.
- **`projects/` is never touched.** Projects own their own trees, and `projects/camp/internal/dungeon/` is a Go package, not a campaign dungeon. Getting this wrong would corrupt real repos, so it is enforced in `Discover` and covered at both the unit and container level. Nested git repositories are skipped for the same reason.
- **Deepest first.** Dungeons nest: this campaign really has `workflow/design/dungeon/completed/go-token-counter/dungeon`. A nested dungeon is renamed before its ancestor moves out from under it.
- **All or nothing.** A location holding both spellings, or an occupied target, refuses the whole run and reports the conflict. A half-migrated campaign would be worse than the state it started in.

**`camp status` points a legacy campaign at the migration.** A migration nobody discovers is not a migration.

## Idea rebrand

`camp idea` was already the primary name, but everything it printed still said "intent": "Intent created", "1 intent(s)", and the flag, help, and error text throughout. Fixed, including for the `camp intent` alias, since the point is to retire the noun from the UX rather than preserve it behind the old name.

Deliberately unchanged: the `intents/v1alpha1` `--json` kind, the `.campaign/intents/` storage path, the `camp intent` alias, and the help lines that document that alias and that path.

## Compatibility

No flag or command is removed or renamed. `camp intent` keeps working indefinitely, and every subcommand and alias resolves to the same storage. The `--json` contract is untouched: `schema_version` is still `intents/v1alpha1`, verified against the built binary. `camp init --repair` on a legacy campaign now creates visible dungeons to match that campaign rather than introducing the hidden spelling next to them.

The one intended behavior change for existing users: a campaign that has already drifted into holding both spellings will now get a loud error from dungeon-touching commands instead of silently resolving to the visible one. That is the point. `camp dungeon migrate` reports the conflict and tells you what to reconcile.

## Verification

Full gate green: `just gate` (3409 build-tag tests, 3380 unit, lint clean).

Container harness, 15 dungeon tests covering both-exist errors, scoped failure, new dirs in a legacy campaign staying visible, migrate converting every discovered dungeon, skipping `projects/`, refusing on conflict, `--dry-run` mutating nothing, one commit with renames, `--no-commit`, and a campaign with no commits yet.

Driven end-to-end against a throwaway fixture campaign built to match this campaign's real shape (10 campaign dungeons including a nested one, plus a `projects/camp/internal/dungeon` Go package):

```
✓ Migrated 10 dungeon(s) to .dungeon/
ℹ Committed changes to git

commits created: 1
subject: [legacy-demo2:d0bdcef7] Migrate: dungeon directories to .dungeon
renames recorded: 30      (all R100)
projects/camp/internal/dungeon/resolver.go: package dungeon   (untouched)
working tree: clean
```

`git log --follow` on a file inside the nested dungeon traverses the rename, so history survives.

Driving the real binary caught a bug the container tests missed: the commit staged a path that does not exist after the sweep, because a nested dungeon is renamed first and then has its ancestor renamed out from under it, so it never arrives at the path the plan recorded. The container tests covered nesting and the git commit path separately but never together. Fixed, with a regression test at each level; both fail against the unfixed code.

Known failures on this branch, all five reproducing identically on clean `origin/main` (2cfd04d1) and untouched here: `TestIntegration_WorkitemCommits_{CrossRepo,FiltersFalsePositives,IncludesCampaignAndLinkedProject}`, `TestIntegration_CommitTags_NoteNoContext`, `TestIntegration_WorkitemValidate_ReportsFindings`.

## Deferred, disclosed

**Scaffolded `.campaign/intents/OBEY.md` and onboarding copy still say "intent".** The intent's rebrand bullet covers them. They are left for a follow-up: the guide is a substantial rewrite whose sensible shape depends on whether the storage path is ever renamed, and folding it in here would bury the migration change it is stacked on. The runtime surface, which is what a user actually reads while working, is done.

**The notice framework is minimal on purpose.** Per-signature dismissal and the `camp init --repair` drift signal are designed in `workflow/design/need-to-reduce-bloat-in-20260709-215845/04-repair-and-template-update-notifications.md` and are not built. `Notice` carries an ID for dismissal to key off and `Detect` takes a detector list, so the repair signal plugs in as one more entry. Detection stays stat-level, as that design requires.

**fest and festival-app** resolve dungeons independently. fest #274 carries the matching ambiguity change; festival-app remains a follow-up.

The CLI reference regen is its own commit. It sweeps in drift that predates this branch (docs for `camp artifacts` and `camp audit` that were never generated, and flags like `camp clone --from` and `camp init --org` whose docs were never refreshed).


internal/dungeon/spelling/spelling.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// Resolve inspects parent for "dungeon" and ".dungeon" subdirectories.
+//
+// If both exist, the visible spelling wins and Warning explains the
+// conflict. If exactly one exists, that spelling is returned. If neither
+// exists, Resolve reports Exists=false and defaults Name to Visible; callers
+// that are about to create a new dungeon directory should decide the
+// spelling with NameForNew instead of relying on this fallback.
+func Resolve(ctx context.Context, parent string) (Resolved, error) {
```
obey-agent, line 93: **Single spelling package is the right design** and matches fest#274's dual-resolution approach. Clear split: `Resolve` for read (prefer visible on conflict), `NameForNew` for create (established spelling wins, else setting). Defaulting `Resolve` to Visible when neither exists is correct only because creates go through `NameForNew` — the comment makes that contract explicit. Good.

internal/config/types.go
```diff
@@ -329,6 +329,12 @@ type GlobalConfig struct {
 	// (project ref linking and subject-tag tracing). Per-campaign overrides
 	// live in .campaign/settings/local.json under the same shape.
 	Commit CommitPrefs `json:"commit,omitempty" yaml:"commit,omitempty"`
+	// DungeonHidden controls whether newly scaffolded dungeon directories use
```
obey-agent, line 332: **Pointer `*bool` for dungeon_hidden is the right defaulting trick:** unset → true for new machines/campaigns, explicit false preserved on disk. Existing campaigns never rewritten. `ResolveDungeonHidden` nil-safe. Matches the "no retroactive migration" product rule.

cmd/camp/intent/root.go
```diff
@@ -2,37 +2,44 @@ package intent
 
 import "github.com/spf13/cobra"
 
-// Cmd is the scaffold root for the intent command family.
+// Cmd is the scaffold root for the idea/intent command family. "idea" is the
```
obey-agent, line 5: **`camp idea` as primary with `intent`/`ideas` aliases** and storage path unchanged (`.campaign/intents/`) is the low-risk rename. Agents and scripts that still say intent keep working indefinitely. Manifest paths should use the canonical Use name (crawl test updated) — verify other integration tests that hardcode `"intent "` paths were updated too.

tests/integration/dungeon_hidden_test.go
```diff
@@ -0,0 +1,213 @@
+//go:build integration
```
obey-agent, line 1: **Integration matrix is the acceptance suite this feature needs:** fresh hidden default, explicit false legacy, existing visible survives system flip (including idea move into established intents dungeon), both-spellings warn, idea+intent alias parity. Seeding the container Reset() with `dungeon_hidden: false` keeps the rest of the suite honest without rewriting every fixture.

obey-agent (commented):
## Verdict: Approve

## Overview
Two coordinated product changes: (1) system setting `dungeon_hidden` defaulting true so new campaigns scaffold `.dungeon` everywhere, with dual-spelling resolution for existing visible campaigns; (2) `camp idea` as the primary command name with `intent`/`ideas` as permanent aliases (storage path unchanged). Scoped correctly to camp; fest festivals dungeon is separate (fest#274).

## Key Findings
1. **No merge blockers.** Resolution API (`spelling.Resolve` / `NameForNew` / `RewriteRel`) is the right choke point and is wired through dungeon resolver, workflow, intents, workitem locate, scaffold init/repair, and flow.
2. **Known limitations in PR body are fair** (repair dry-run preview quirk; residual "intent" Long-form prose). Neither is blocking.
3. **Non-blocker:** `locate.DetectFromCwd` uses `context.Background()` for Resolve — fine for a sync path detector, but prefer propagating caller context if one is available later.
4. **Coordination:** stack with fest#274 for festivals dual spelling; this PR only pre-tolerates `festivals/.dungeon` in camp cwd parsing.

## What's Done Well
- Existing campaigns never retroactively renamed; established spelling always wins on create
- Conflict prefers visible + stderr warning (tested)
- Integration suite default `dungeon_hidden=false` avoids rewriting every test
- Idea alias is non-destructive and storage-stable
- Strong unit + containerized coverage for the product matrix

## Staff Standard
Yes. This is the careful form of a UX default flip — default the new world, never break the old one.

internal/dungeon/spelling/discover.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	if len(found) > 0 {
+		return Hidden, nil
+	}
+	return NameFor(hiddenDefault), nil
+}
+
+// HasLegacyDungeons reports whether campaignRoot still holds any visible
+// dungeon, i.e. whether camp dungeon migrate has work to do.
+func HasLegacyDungeons(ctx context.Context, campaignRoot string) (bool, error) {
```
obey-agent: **Dead code: `HasLegacyDungeons` has no production caller**

The only references in the tree are this declaration and its own test:

```
internal/dungeon/spelling/discover.go:129:func HasLegacyDungeons(...)
internal/dungeon/spelling/discover_test.go:258:func TestHasLegacyDungeons(t *testing.T)
```

This looks like a leftover from the design where `camp status` swept the campaign to detect the legacy layout. The version that shipped instead is `notice.DungeonLegacy`, which deliberately resolves only the root dungeon and documents why: "It resolves only the root dungeon rather than sweeping the campaign, which keeps the check at two stats." The `notice` package doc goes further and makes it a rule: "Detectors run on high-traffic commands, so they must stay stat-level. No detector may scan the campaign tree."

So this exported helper is the tree-scanning answer to a question the design has already answered the cheap way. Leaving it exported next to `Discover` invites the next person to reach for it on exactly the hot path the notice package forbids, and it carries a 50-line test that implies it is load-bearing.

Delete `HasLegacyDungeons` and `TestHasLegacyDungeons`. `Discover` is still exercised through `migrate.BuildPlan` and `CampaignName`, so coverage does not regress.

cmd/camp/dungeon/migrate.go
```diff
@@ -0,0 +1,143 @@
...
+
+var dungeonMigrateCmd = &cobra.Command{
+	Use:   "migrate",
+	Short: "Convert every campaign dungeon to the hidden .dungeon spelling",
+	Long: `Convert every dungeon in this campaign from "dungeon" to ".dungeon".
+
+New campaigns hide the dungeon so it stops being the first thing newcomers ask
+about. This converts a campaign made before that change. A campaign uses one
+spelling throughout, so the sweep covers every dungeon at once: the campaign
+root, festivals/, .campaign/intents/, .campaign/quests/, and each workflow
```
obey-agent, line 31: **`migrate` renames fest's directory, and released fest cannot find it afterwards**

This sweep includes `festivals/`, which is fest-owned territory, not camp's. I checked fest's `main`:

```
$ git grep -rn '"\.dungeon"' origin/main -- '*.go' | grep -v _test.go
(no matches)
```

fest `main` knows only the visible spelling. The dual resolution that teaches it `.dungeon` (`HiddenDungeonDir` in `internal/workspace/dungeon.go`) is still unmerged in fest#274.

So on any campaign where fest is actually in use, `camp dungeon migrate` renames `festivals/dungeon` to `festivals/.dungeon` and the user's installed fest stops resolving its own dungeon. This is sharper than the coordination note on the previous review: before this command, camp only scaffolded *new* campaigns hidden, and a new campaign has no fest history to lose. This actively converts *existing* campaigns, which are exactly the ones with festivals filed under `festivals/dungeon`.

This is a merge/release ordering constraint, not a code defect, so I am not asking for a code change here — but it needs to be explicit rather than assumed:

- camp#437 must not merge ahead of fest#274, and
- `camp dungeon migrate` must not reach users before a fest release that contains fest#274, since the break lands on whatever fest binary the user has installed, not on whatever is on fest `main`.

If the two cannot be released together, the safer shape is to have `migrate` skip `festivals/` (leaving fest to convert its own tree) or refuse when the installed fest predates dual resolution. Otherwise the ordering needs to be a release checklist item.

obey-agent (commented):
## Staff Review (re-review at fb4c123) — Request Changes

### Why this is a re-review, not a rubber stamp of the earlier Approve

The previous review approved `456527b2`. Seven commits have landed since, and they are not polish — they reverse two decisions that review specifically praised and add a whole new command:

- `2caa52d` makes both spellings an **error**, where the approved version preferred visible and warned. The old behavior was called out under "What's Done Well."
- `df9326c` makes new dungeon dirs follow **the campaign's own spelling** rather than the `dungeon_hidden` setting.
- `41b5fe6` + `5603526` add `camp dungeon migrate` and point legacy campaigns at it from `camp status`.
- `427ccd3` makes the runtime speak "idea", not just the command name.

That is effectively a new PR. The earlier Approve does not carry over, so I re-reviewed from scratch at `fb4c123`.

### The direction changes are right

Both reversals are improvements. Refusing to guess when both spellings exist is correct: whichever one lost would be silently invisible to every listing, and the `ConflictError` text tells the user exactly how to get unstuck. Following the campaign's established spelling rather than the setting is what prevents accretion in the first place — the approved version would have created `.dungeon` next to an existing `dungeon` on any legacy campaign, since `dungeon_hidden` defaults true. That was a real bug and this fixes it at the source.

`camp dungeon migrate` is the piece that makes the whole default flip honest: hiding the dungeon for new campaigns is only half a product if existing campaigns have no way across.

### Blocking

1. **Dead code: `HasLegacyDungeons`** (`discover.go:129`) — no production caller, only its own test. It is the tree-scanning answer to a question `notice.DungeonLegacy` already answers in two stats, and the `notice` package doc explicitly forbids tree-scanning detectors. Leaving it exported invites exactly that mistake. Inline.

2. **Release ordering vs fest#274** (`cmd/camp/dungeon/migrate.go`) — `migrate` renames `festivals/dungeon`, and fest `main` has no `.dungeon` support at all (fest#274 is still open). Migrating breaks the user's installed fest. Not a code defect, but per the stacked-PR rule it needs to be stated rather than assumed: this must not merge ahead of fest#274, and migrate must not reach users before a fest **release** carrying it. Inline.

### Verified at `fb4c123`
- `just gate` — **pass** (3409/3409 unit tests, both ratchets)
- `just test integration-run TestDungeon` — **pass** (containerized, 80s, all 15 dungeon tests)
- `just docs` — **no drift**; the regen commit is honest, and correctly kept separate from the feature commits

### What's done well
- `migrate.CommitPaths` handles the nested-dungeon case correctly and, more importantly, *explains* it: a dungeon inside a migrated dungeon never comes to rest at its own `Move.To`, so staging the ancestor is what covers the subtree. That is the kind of comment that saves the next reader an hour.
- Deepest-first ordering makes every remaining `From` valid at apply time, and `nestedUnderAnotherMove` compares with the separator appended, so `a/dungeon-x/` is not mistaken for a child of `a/dungeon`.
- Partial application degrades safely. Two moves can never target the same parent, so an interrupted sweep leaves a mixed-but-valid campaign that re-running completes, not the conflict state the resolver rejects.
- `Discover` skips `projects/` and nested repos, with the reason stated: a Go package directory named `dungeon` is not a campaign dungeon.
- `move()` decides `git mv` vs plain rename by *asking* whether git tracks the path, rather than parsing git's failure strings.
- The `notice` package is scoped honestly: it says what it is not building yet, links the design doc, and carries an `ID` so dismissal has a key later.
- `--dry-run` / `--no-commit`, and the failure path admits "The dungeons were moved on disk, but the commit failed" instead of pretending the whole thing rolled back.
- Context propagated throughout, `camperrors` used consistently, no `fmt.Errorf` in production paths.

### Staff Standard
The engineering here is work I would be glad to put my name on — this is the most carefully reasoned code I have reviewed in this repo recently, and the doc comments explain *why* rather than restating the code. The two items above are small: one deletion, and one explicit statement of a release constraint the team is already half-aware of. Close them and this is a clean approve.

**Verdict: Request Changes** (posted as a comment per campaign recipe).

obey-agent:
Termcast demo for this PR — watch hidden `.dungeon` discovery, migration, and the resulting idea listing.

PR #437 demo

internal/dungeon/spelling/spelling.go
```diff
@@ -0,0 +1,181 @@
...
+// Hidden reports whether r resolved to the hidden spelling.
+func (r Resolved) Hidden() bool { return r.Name == Hidden }
+
+// Resolve inspects parent for "dungeon" and ".dungeon" subdirectories.
+//
+// Exactly one present returns that spelling. Both present returns a
+// ConflictError. Neither present reports Exists=false and defaults Name to
+// Visible; callers about to create a dungeon should use NameForNew rather
+// than relying on that fallback.
+func Resolve(ctx context.Context, parent string) (Resolved, error) {
```
obey-agent, line 93: **Behavior change since prior review: both spellings is now a hard ConflictError**, not "prefer visible + warn". That is the stronger, correct product rule — picking either spelling would hide work under the other. ConflictError names `camp dungeon migrate` so operators have an escape hatch. Matches fest dual-resolution only after migrate; until then camp refuses to guess.

internal/dungeon/spelling/spelling.go
```diff
@@ -0,0 +1,181 @@
...
+// NameForNew decides the directory name a brand-new dungeon under parent
+// should use.
+//
+// A dungeon already under parent keeps its established name, so repeated
+// init/repair calls stay idempotent. Otherwise campaignName wins: a new
+// directory inside an existing campaign follows what that campaign already
+// uses, never the dungeon_hidden setting, because a campaign that accreted
+// one spelling next to the other is exactly the broken state Resolve rejects.
+// Resolve campaignName with CampaignName.
+func NameForNew(ctx context.Context, parent, campaignName string) (string, error) {
```
obey-agent, line 131: **NameForNew now takes campaignName**, not the global dungeon_hidden flag, for new directories inside an existing campaign. Established layout wins; setting only affects brand-new campaigns at scaffold time. That closes the gap where flipping the system setting would create mixed spellings inside one campaign — exactly the state Resolve now rejects.

internal/dungeon/migrate/migrate.go
```diff
@@ -0,0 +1,226 @@
...
+	// repository with a commit to build on. A campaign whose scaffold has
+	// never been committed has no history to preserve, so its moves just land
+	// in the working tree.
+	Committable bool
+}
+
+// Empty reports whether the campaign is already fully migrated.
+func (p *Plan) Empty() bool { return len(p.Moves) == 0 }
+
+// BuildPlan discovers every dungeon under campaignRoot and plans the moves
```
obey-agent, line 51: **migrate is plan-then-all-or-nothing:** discover, refuse if any parent already has both spellings or a hidden target exists, deepest-first renames, git stage/commit. Correct for a campaign-wide invariant. Release note calling out fest#274 ordering before users run migrate on festivals/ is mandatory — the Long help already says it.

cmd/camp/dungeon/migrate.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/Obedience-Corp/camp/internal/ui"
+	"github.com/spf13/cobra"
+)
+
+var (
+	migrateDryRun   bool
+	migrateNoCommit bool
+)
+
+var dungeonMigrateCmd = &cobra.Command{
```
obey-agent, line 23: **CLI surface documents fest ordering and projects/ exclusion.** Dry-run / --no-commit covered. projects/ never touched so repo dirs named dungeon are safe.

obey-agent (commented):
## Verdict: Approve (re-review after substantial post-review changes)

## Overview
Re-reviewed head `b9e92b8`. Since the prior approve, this PR hardened dual-spelling policy and added migration:

1. **Both spellings → ConflictError** (no more prefer-visible warning path)
2. **New dirs follow the campaign's established spelling**, not the global setting
3. **`camp dungeon migrate`** all-or-nothing conversion to `.dungeon` with discover/plan/apply + git
4. Status/legacy notice pointing operators at migrate; idea language at runtime; container tests for migrate

This is a better product model than the first review head.

## Key Findings
1. **No merge blockers** on the new conflict/migrate design.
2. **Operational gate:** do not advertise migrate to campaigns with festivals until fest dual-resolution is shipped (PR body + Long help already state this).
3. **Non-blocker:** Apply is not transactional across renames mid-failure after plan validation — deepest-first + preflight checks make partial failure unlikely; document that a mid-Apply crash still needs human repair.
4. Stack docs/cli-reference regen still couples this PR to siblings on merge order.

## What's Done Well
- ConflictError with actionable migrate guidance
- Discover-based migration (future dungeon locations included)
- Integration coverage expanded (hidden, legacy flip, migrate harness)
- idea primary name + intent aliases retained

## Staff Standard
Yes. The post-review iteration improved correctness of the dual-spelling story.

obey-agent:
**Updated to merge (at `71bb3f5e`)** — merged `main` in. Two conflicts resolved:
- `cmd/camp/intent/add.go`: main moved the JSON-payload emission into `writeCreatedIntent`, so the branch's explicit block in `finalizeCreatedIntentWithOutput` was a duplicate (it emitted the payload twice — caught by `TestIntentAddJSONEmitsCreatedPayload`). Dropped the redundant block and carried the branch's `idea`-rename into `writeCreatedIntent` (`✓ Idea created`).
- `zz_manifest_annotations.go`: unioned `manifestAllowedInteractivePaths` — branch's `idea add` rename plus main's `gather explore`.

CLI reference regenerated for main's fresh/gather commands. `just gate` green (3435 dev + stable unit, both ratchets). The dead-code finding (`HasLegacyDungeons`) from the `fb4c123` re-review is already gone at head.

PR #438, opened 2026-07-15
fix(workitem): make the help concise and stop it making false claims

lancekrogers:
## What changed and why

`camp workitem --help` flooded the reader. Two flag descriptions ran to paragraph length, and ten lines wrapped past 80 columns, one of them to 185 characters. Every description now fits on one line at 80 columns, and nothing in the help wraps.

The enumerated values did not need to live in flag help, because they are already in two places the reader reaches on demand:

1. The error messages print them on a bad value: `workitem.go:291` for `--stage`, `:310` for `--group-by`, `:337` for `--status` via `DisplayStatusVocabulary()`. Pass a bad value, get the list.
2. The generated CLI reference carries the full detail.

Implementation detail is gone from the subcommand descriptions too, including `v1 minimum metadata`, `current-schema`, and `.workitem markers`.

## Corrections beyond brevity

- **The non-interactive claim was false.** The description said default mode launches a TUI and never mentioned that `workitem.go:74-75` errors with `non-interactive use requires --json, --list, or --print flag` when no terminal is attached. Anything calling it from a pipe or a script read the help and walked into that error. It now states the requirement in one sentence.
- **`--stage` advertised combinations the code rejects.** It listed a flat vocabulary, but validation calls `IsValidStageForTypes`, which accepts stages per `--type`. `--type intent --stage planning` is an error, and there is an existing test asserting exactly that. The vocabulary is gone, and the error message tells the truth per type.
- **`--group-by` read as a contradiction.** It ended with `--list defaults to group unless set` directly beside the `(default "attention_stage")` cobra appends.
- **The summary was stale.** It enumerated intents, designs, explore, and festivals. `internal/workitem/discover_custom.go:9` confirms those are exactly the builtins, but the command also surfaces custom types created through `camp workitem create --type <name>` and discovered via explicit `.workitem` markers. This campaign uses them, for example `workflow/bugs/` and `workflow/audits/`, both of which carry markers. Dropping the enumeration fixes the staleness and the verbosity in one move, rather than replacing it with a longer and more accurate enumeration.
- **`--print` did not select.** It claimed it selects a path; non-interactively `workitem.go:116-119` takes `items[0]` with no selection.
- **The examples were redundant and incomplete.** `--type` was demonstrated twice while `--list`, one of the three modes a non-TTY caller must pass, appeared in no example. There are now four examples covering every mode with no repetition.
- **`--show-parked` started lowercase** while every other description was capitalized.

## Nothing was removed, renamed, or hidden

No flag or subcommand is removed or renamed, no accepted value changed, and no `--json` output changed. All 18 subcommands remain visible.

Hiding niche subcommands was considered and deliberately rejected, because in this repo `Hidden` is not presentation only:

1. `cmd/camp/manifest.go:60` skips `Hidden` commands when building `camp __manifest`, described in that file as the manifest for daemon enforcement. `workitem validate`, `repair`, and `resolve` all carry `agent_allowed: "true"`, so hiding any of them would drop it from that manifest and the daemon could stop permitting it. The `continue` also skips recursion, so children of a hidden command would vanish too.
2. `cmd/camp/gendocs.go:47` calls `removeGeneratedDocs` to wipe the reference directory before regenerating, and cobra's doc generator skips `Hidden` commands, which `gendocs.go:157` notes directly. Hiding would permanently delete `docs/cli-reference/camp_workitem_<cmd>.md`.

A regression test now fails if any workitem subcommand is marked `Hidden`, recording the reason.

## Measurements

Logical line count barely moves, because the real noise was wrapping. The rendered number at a fixed 80 columns is what a user actually sees:

| | `wc -l` | rendered at 80 (`fold -w 80 \| wc -l`) | lines over 80 cols |
| --- | --- | --- | --- |
| before | 59 | 72 | 10 (longest 185 chars) |
| after | 58 | 58 | 0 |

## How this was verified

- `just gate` passes: builds stable and dev, `go vet` stable, dev, and integration, lint stable and dev, and unit tests on both profiles (3369 dev, 3340 stable). Exit 0.
- `camp __manifest` before and after is byte identical, all 19 workitem paths intact.
- `camp workitem --json` and `camp workitem --list` are identical before and after when captured against the same disk state, apart from the `generated_at` timestamp.
- All 18 subcommands still execute: `camp workitem <name> --help` returns 0 for each.
- New tests cover the non-TTY error path, that every visible description stays within the length budget and starts capitalized, that every subcommand stays registered and unhidden, and that the description keeps documenting the `--json`/`--list`/`--print` requirement.

The non-TTY gate was extracted into `needsExplicitOutputMode`, a pure function, so the test does not depend on whether the test runner's stdout happens to be a terminal. Testing it through `Execute` would launch the TUI and hang if run from a terminal.

## Known limitations

- `--status`, `--stage`, and `--attention-stage` remain three overlapping filters whose vocabularies are close to a union of each other, and `plan` versus `planning` still differ across flags. `internal/workitem/status.go:26-32` normalizes `planning` to `plan`, so both are accepted. Consolidating them means deprecating flags, so it is out of scope here.
- `--group-by` is now terse enough that its accepted values are only discoverable through the error message or the generated reference. That is the deliberate tradeoff, and the reason its description is shorter than the others is that cobra appends `(default "attention_stage")` to the same line, which would otherwise push it past 80 columns.
- Related and worth calling out explicitly: the old description resolved its own contradiction by stating `--list defaults to group unless set`, and that clause was true. `workitem.go:93-94` still overrides the grouping to `group` when `--list` is passed without an explicit `--group-by`, so that behavior is unchanged but is no longer documented in the flag help. Restoring it inline is what made the line self-contradictory and over-length in the first place, since cobra appends the flag's registered default regardless. The behavior is observable by running `camp workitem --list`, and the generated reference is the place to describe it properly. Flagging it so the omission is a recorded decision rather than an accident.
- The docs regen is a separate commit and drags in unrelated drift accumulated since the last regen, which is expected in this repo.

## BEFORE (59 logical lines, 72 rendered at 80 columns)

```
View active campaign work items across intents, designs, explore, and festivals.

Default mode launches an interactive TUI dashboard. Use --json for machine-readable
output or --print to select and print a path for shell integration.

Examples:
  camp workitem                              # interactive dashboard
  camp workitem --json                       # JSON output for agents/scripts
  camp workitem --json --type design         # filter by type
  camp workitem --json --type intent --limit 5
  camp workitem --print                      # select and print path

Usage:
  camp workitem [flags]
  camp workitem [command]

Aliases:
  workitem, wi, workitems

Available Commands:
  adopt       Attach .workitem metadata to an existing directory
  commit      Commit changes scoped to the resolved workitem
  commits     List commits referencing a workitem across linked repos
  create      Create a new workitem with v1 minimum metadata
  current     Get, set, or clear the local current workitem
  doctor      Report workitem link-registry health issues
  group       Set or clear the group of a workitem
  link        Attach a workitem to a project, festival, worktree, or campaign path
  links       List workitem links
  list        List or browse filtered workitems
  priority    Set or clear the manual priority of a workitem
  promote     Promote a workitem to a festival, doc, or dungeon status
  repair      Repair a workflow directory into a current-schema work item
  resolve     Print the workitem the current context resolves to (read-only)
  stage       Set or clear the attention stage of a workitem
  unlink      Remove one or more workitem links
  validate    Validate workflow work item directories and their .workitem markers
  worktree    Create a project worktree from a workitem

Flags:
      --attention-stage stringArray   Filter by attention stage (current, next, active, parked)
      --category stringArray          Filter by workflow category (builtin: plan, research, pipeline, review, uncategorized; or any category defined under workflows in campaign.yaml)
      --group stringArray             Filter by workitem group
      --group-by string               Group JSON/list sections by attention_stage, group, type, or category; --list defaults to group unless set (default "attention_stage")
  -h, --help                          help for workitem
      --json                          Output as JSON
      --limit int                     Maximum number of items to return
      --list                          Output a compact grouped list
      --print                         Print path only (for shell integration)
      --query string                  Search query to filter items
      --show-parked                   include parked attention-stage workitems in default output
      --stage stringArray             Filter by lifecycle stage (none, inbox, active, ready, planning, ritual, chains)
      --status stringArray            Filter by displayed status (current, next, active, parked, inbox, ready, plan, ritual, chains, none)
      --type stringArray              Filter by workflow type (builtin: intent, design, explore, festival; or any slug-safe custom type produced by 'camp workitem create --type <name>')

Global Flags:
      --no-color   disable colored output

Use "camp workitem [command] --help" for more information about a command.
```

## AFTER (58 logical lines, 58 rendered at 80 columns)

```
View active campaign work items.

Launches an interactive dashboard on a TTY. Non-interactive callers must pass
--json, --list, or --print.

Examples:
  camp workitem                       # interactive dashboard
  camp workitem --json --type design  # JSON, filtered by type
  camp workitem --list                # compact grouped list
  camp workitem --print               # print a path for shell integration

Usage:
  camp workitem [flags]
  camp workitem [command]

Aliases:
  workitem, wi, workitems

Available Commands:
  adopt       Adopt an existing directory as a workitem
  commit      Commit changes scoped to a workitem
  commits     List commits referencing a workitem
  create      Create a workitem
  current     Get, set, or clear the current workitem
  doctor      Report link-registry health issues
  group       Set or clear the group
  link        Create a workitem link
  links       List workitem links
  list        List or browse filtered workitems
  priority    Set or clear the manual priority
  promote     Promote a workitem to a festival, doc, or dungeon
  repair      Repair a workflow directory into a workitem
  resolve     Print the workitem for the current context
  stage       Set or clear the attention stage
  unlink      Remove workitem links
  validate    Validate workitem directories
  worktree    Create a project worktree from a workitem

Flags:
      --attention-stage stringArray   Filter by attention stage
      --category stringArray          Filter by workflow category
      --group stringArray             Filter by workitem group
      --group-by string               Group sections (default "attention_stage")
  -h, --help                          help for workitem
      --json                          Output as JSON
      --limit int                     Maximum items to return
      --list                          Output a compact grouped list
      --print                         Print path only
      --query string                  Filter by search query
      --show-parked                   Include parked workitems
      --stage stringArray             Filter by lifecycle stage
      --status stringArray            Filter by displayed status
      --type stringArray              Filter by workflow type

Global Flags:
      --no-color   disable colored output

Use "camp workitem [command] --help" for more information about a command.
```

### DIFF
--- a/docs/cli-reference/camp-reference.md
+++ b/docs/cli-reference/camp-reference.md
@@ -49,6 +49,149 @@ camp [flags]
 ```
 ---
 
+## camp artifacts
+
+Manage declared artifact roots (.campaign/artifacts.yaml)
+
+### Synopsis
+
+Manage the campaign's declared artifact roots: directories of heavy non-git
+payloads (media, renders, datasets) that 'camp sync --from <machine>' moves
+between your machines with rsync instead of git.
+
+The declaration file (.campaign/artifacts.yaml) is committed, so every
+machine knows what belongs to the campaign. Declared roots should be
+gitignored: a root that is also git-tracked would make the same bytes both
+git content and artifact content. Manifests and per-peer sync snapshots are
+machine-local derived state under .campaign/cache (gitignored).
+
+### Examples
+
+```
+  camp artifacts list
+  camp artifacts add media/renders
+  camp artifacts add datasets --policy on-demand
+  camp artifacts remove media/renders
+  camp artifacts manifest media/renders
+```
+
+### Options
+
+```
+  -h, --help   help for artifacts
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+---
+
+## camp artifacts add
+
+Declare an artifact root
+
+### Synopsis
+
+Declare a campaign-relative directory as an artifact root.
+
+Policy 'always' (default) syncs the root on every 'camp sync --from
+<machine>'; 'on-demand' syncs it only when artifacts are requested
+explicitly (--artifacts-only).
+
+```
+camp artifacts add <path> [flags]
+```
+
+### Options
+
+```
+  -h, --help            help for add
+      --policy string   Sync policy: always (every peer sync) or on-demand (--artifacts-only) (default "always")
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+---
+
+## camp artifacts list
+
+List declared artifact roots
+
+```
+camp artifacts list [flags]
+```
+
+### Options
+
+```
+  -h, --help   help for list
+      --json   Output as JSON for scripting
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+---
+
+## camp artifacts manifest
+
+Print a declared root's manifest as JSON
+
+### Synopsis
+
+Walk a declared artifact root and print its manifest (relative path, size,
+mtime per file) as JSON. This is the same shape sync snapshots use, so it is
+useful for scripting and for comparing roots across machines.
+
+```
+camp artifacts manifest <path> [flags]
+```
+
+### Options
+
+```
+  -h, --help   help for manifest
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+---
+
+## camp artifacts remove
+
+Remove an artifact root declaration
+
+### Synopsis
+
+Remove a declared artifact root. Files on disk are not touched.
+
+```
+camp artifacts remove <path> [flags]
+```
+
+### Options
+
+```
+  -h, --help   help for remove
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+---
+
 ## camp attach
 
 Attach an external directory to a campaign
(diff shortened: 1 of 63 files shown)


obey-agent (commented):
## Staff Review — Approve

### Overview
This is an exemplary help-only cleanup: it shortens `camp workitem` flag and subcommand copy without changing behavior, while fixing real false claims (non-TTY requirements, stage vocabulary that `IsValidStageForTypes` rejects, `--print` selection wording, stale type enumeration). The PR description is itself a model of staff-level change communication — measurements, deliberate non-goals, and the `Hidden` hazard analysis are exact.

### Key Findings
No blocking issues. Implementation matches the stated contract:
- Flag semantics, validation, and output paths are unchanged; only help text, Short strings, and a pure extract of `needsExplicitOutputMode` move.
- Error messages still own the vocabularies (`--stage`, `--group-by`, `--status`), so brevity does not remove discoverability for agents that already parse failures.
- Regression tests pin the non-interactive requirement, description length/capitalization, and the critical invariant that workitem subcommands must not be marked `Hidden` (manifest + gendocs).

### What's Done Well
- Choosing not to hide niche subcommands after reading `manifest.go` / `gendocs.go` is the correct production call; the test message records *why*.
- Extracting `needsExplicitOutputMode` so unit tests do not hang a TUI under a TTY is good engineering.
- Regenerating CLI reference alongside source so docs and binary stay aligned.
- Explicitly documenting the `--list` / `--group-by` default override as a deliberate omission from flag help rather than an accidental loss.

### Staff Standard
Yes. I would be proud to submit this as my own work. It improves operator and agent UX without collateral product risk.

**Verdict: Approve**

PR #439, opened 2026-07-17
fix(intent): batch TUI auto-commits at session end

lancekrogers:
## Summary

- Intent Explorer no longer fires a git auto-commit after every move/edit/archive/create. Mutations still write disk + domain audit + ledger immediately; paths are recorded and **one selective commit** runs when the TUI exits (`DrainAutoCommits`), matching dungeon/intent crawl.
- `camp intent add` multi-save TUI (Ctrl-n) also defers git until the form session ends.
- One-shot CLI (`camp intent add "title"`, `camp intent move`, …) still auto-commits once at end of that process.
- Adds `commit.IntentUpdate` for multi-op session subjects (`Update: intent explorer session`).

This removes mid-session freezes on large campaigns where each scoped temp-index commit costs ~seconds.

Workitem: `WI-913ffc` (`workflow/design/intent-capture-commit-performance`)

## Test plan

- [x] `go test ./internal/intent/tui/explorer/ ./internal/git/commit/ ./cmd/camp/intent/`
- [ ] Manual: `camp intent explore` — move/edit/archive feel instant; one “Finalizing intent commits…” on quit
- [ ] Manual: multi-save `camp intent add` (Ctrl-n) does not pause on git between captures
- [ ] Manual: single CLI `camp intent add "title"` still commits once at end

### DIFF
--- a/cmd/camp/intent/add.go
+++ b/cmd/camp/intent/add.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/spf13/cobra"
@@ -232,6 +233,11 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// TUI session: apply every create on disk immediately, but defer git until
+	// the form exits (crawl-style batch). Mid-session commits freeze capture.
+	intentsDir := resolver.Intents()
+	var session batchCommit
+
 	// Process intents saved via Ctrl-n during the session
 	for _, saved := range model.SavedResults() {
 		savedOpts := intent.CreateOptions{
@@ -243,14 +249,18 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 			Tags:    mergeTags(tagsFlag, saved.Tags),
 		}
 		if createNote {
-			if err := runNoteCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, savedOpts); err != nil {
+			created, err := createNoteWithoutCommit(ctx, svc, intentsDir, savedOpts)
+			if err != nil {
 				return err
 			}
+			session.add(commit.IntentCreate, created.Title, "", created.Path)
 			continue
 		}
-		if err := runFastCapture(ctx, svc, resolver.Intents(), cfg, campaignRoot, noCommit, savedOpts); err != nil {
+		created, err := createIntentWithoutCommit(ctx, svc, intentsDir, campaignRoot, savedOpts, false)
+		if err != nil {
 			return err
 		}
+		session.add(commit.IntentCreate, created.Title, "", created.Path)
 	}
 
 	// Process the final result (from normal save-and-quit)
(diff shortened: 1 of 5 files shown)


internal/intent/tui/explorer/audit.go
```diff
@@ -14,22 +15,33 @@ import (
 	"github.com/Obedience-Corp/camp/internal/intent/audit"
 )
 
-const (
-	quickDrainTimeout      = 300 * time.Millisecond
-	autoCommitDrainTimeout = 15 * time.Second
-)
+const autoCommitDrainTimeout = 30 * time.Second
+
+// pendingOp is one filesystem mutation recorded during an explorer session.
```
obey-agent, line 20: **Session-end batch is the right model.** Explorer mutations still hit disk + domain audit immediately; git is recorded as pending ops and flushed once in DrainAutoCommits. That matches crawl/dungeon patterns and removes per-action temp-index freezes on large campaigns. Timeout raised to 30s for the single final commit is reasonable.

cmd/camp/intent/add.go
```diff
@@ -232,6 +233,11 @@ func runIntentAdd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// TUI session: apply every create on disk immediately, but defer git until
```
obey-agent, line 236: **Multi-create TUI (Ctrl-n) defers git the same way.** createIntentWithoutCommit / createNoteWithoutCommit keep audit + user feedback; batchCommit.flush once on exit. One-shot CLI still commits immediately via finalizeCreatedIntentWithOutput — correct split of interactive vs non-interactive.

internal/git/commit/intent.go
```diff
@@ -15,6 +15,8 @@ const (
 	IntentPromote IntentAction = "Promote"
 	IntentCrawl   IntentAction = "Crawl"
 	IntentRename  IntentAction = "Rename"
+	// IntentUpdate is used for multi-op session batches (e.g. explorer exit).
```
obey-agent, line 18: **IntentUpdate** for multi-op session subjects with a composed description (Move/Archive/…) keeps commit history legible without one commit per action. Single-op drain preserves the original action title — good test coverage.

obey-agent (commented):
## Verdict: Approve

## Overview
Batches intent TUI git auto-commits to session end (explorer + multi-save add form) while keeping disk mutations and domain audit immediate. One-shot CLI paths still commit once per process. Directly addresses mid-session freezes on large campaigns.

## Key Findings
1. **No blockers.** Design matches established crawl/dungeon batch commit patterns.
2. **Crash window:** if the process is killed before DrainAutoCommits, disk + audit/ledger are durable but git may lag — same class of risk as other deferred-commit TUIs; `camp status` remains the recovery path (prior warning path was for in-flight commits; now the finalize notice is the UX).
3. Tests correctly assert no git during session, one drain commit, second drain no-op, single-op keeps original action.

## What's Done Well
- Clear separation of record vs run
- context.WithoutCancel on drain so exit path can finish
- IntentUpdate subject for multi-op sessions
- Ctrl-n multi-save aligned with explorer batching

## Staff Standard
Yes. This is the performance fix that should have been the model from the start for intent TUIs.

PR #443, opened 2026-07-17
feat(shell,org): add cwi alias and corg org-cycle commands

lancekrogers:
## Summary

Adds two shell conveniences to the camp shell integration:

- **`cwi`** — alias for `camp workitem`. Routes through the `camp()` wrapper, so selecting a work item cd's to it (mirrors `csw`/`cgo`).
- **`corg`** — cycle campaigns within the current campaign's org:
  - `corg` → next active campaign, name-ordered (stable `A→B→C→A` ring, wraps)
  - `corg t` → toggle back to the most-recently-visited *other* campaign in the org
  - `corg --all` → include inactive/reference campaigns
  - Inspection flags (`--print`/`--json`/`--help`) print without changing directory

## Design

The cycle logic lives in Go as two new subcommands, `camp org next` and `camp org toggle`, each emitting a `cd` line the shell evals via `--shell-connect` — the same mechanism `camp switch` already uses. `corg` is a thin dispatcher in the zsh/bash/fish templates.

Toggle reuses the registry's existing `LastAccess` timestamp (maintained by `switch` and by these commands), so:
- `corg` + `corg t` form a natural A↔B toggle
- it composes with regular `csw` navigation
- no new registry state or schema bump is needed

Both subcommands support `--print`/`--json`/`--all`, resolve the current campaign from cwd via the existing `campaign.DetectCached` path, scope to the current campaign's org via `cmdutil.FilterCampaigns`, and are registered `agent_allowed` (parity with `switch`).

## Behavior notes

- Single-member org: emits nothing on stdout (so the shell `eval` is a no-op) plus a stderr note.
- Not inside a registered campaign: clean validation error, non-zero exit.

## Tests / verification

- Unit coverage for cycle resolution (wrap, current-absent, single-member), toggle ordering (newest-other, current-excluded, tie-break), scope filtering (active-only vs `--all`, cross-org isolation), and output emission (shell-connect quoting, print, json, plain).
- Shell template tests assert `cwi`/`corg` across zsh/bash/fish; rendered init passes `zsh -n` / `bash -n`.
- Drove the real zsh integration end-to-end: `corg` walked `alpha→beta→gamma→alpha`; `corg t` toggled `gamma↔alpha`.
- `just gate-fast`: 3390/3390 tests pass, lint clean.

CLI reference docs regenerated in a separate commit.

### DIFF
--- a/cmd/camp/org/cycle.go
+++ b/cmd/camp/org/cycle.go
@@ -284,7 +284,10 @@ func emitCycleTarget(w io.Writer, action string, from, to config.RegisteredCampa
 			To:            cycleCampaignOutput{ID: to.ID, Name: to.Name, Org: to.Org, Path: to.Path},
 		})
 	default:
-		_, err := fmt.Fprintf(w, "cd %s\n", to.Path)
+		// Quote like the shell-connect branch: the default also prints a cd
+		// command a user may copy-paste or eval, so a path with spaces (or a
+		// leading dash) must not break.
+		_, err := fmt.Fprintf(w, "cd -- %s\n", remote.ShellQuote(to.Path))
 		return err
 	}
 }
--- a/cmd/camp/org/cycle.go
+++ b/cmd/camp/org/cycle.go
@@ -1,0 +1,290 @@
+package org
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"sort"
+
+	"github.com/Obedience-Corp/camp/cmd/camp/cmdutil"
+	"github.com/Obedience-Corp/camp/internal/campaign"
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/remote"
+	"github.com/spf13/cobra"
+)
+
+const cycleSchemaVersion = "camp-org-cycle/v1"
+
+var orgNextCmd = &cobra.Command{
+	Use:     "next",
+	Aliases: []string{"cycle"},
+	Short:   "Switch to the next campaign in the current campaign's org",
+	Long: `Switch to the next campaign in the current campaign's org.
+
+Members are ordered by name, so the cycle is stable and predictable
+(a -> b -> c -> a). By default only active campaigns are cycled; use --all to
+include inactive and reference campaigns.
+
+Use with the corg shell function for instant navigation:
+  corg        # cd to the next campaign in this org
+
+The --print flag outputs just the target path for shell integration, and --json
+emits the resolved source and target campaigns.`,
+	Example: `  camp org next            # Print cd to the next org campaign
+  camp org next --print    # Print the target path only
+  camp org next --all      # Include inactive/reference campaigns
+  camp org next --json`,
+	Args: cobra.NoArgs,
+	Annotations: map[string]string{
+		"agent_allowed": "true",
+		"agent_reason":  "Non-interactive; --print/--json resolve the next org campaign without a TUI",
+	},
+	RunE: runOrgNext,
+}
+
+var orgToggleCmd = &cobra.Command{
+	Use:     "toggle",
+	Aliases: []string{"back", "t"},
+	Short:   "Toggle back to the last-visited campaign in the current org",
+	Long: `Toggle back to the most recently visited other campaign in the current org.
+
+"Most recently visited" is tracked by last-access time, which camp updates on
+every 'camp switch' and 'camp org next'/'toggle'. Paired with 'camp org next',
+this gives a natural A <-> B toggle within an org. By default only active
+campaigns are considered; use --all to include inactive and reference campaigns.
+
+Use with the corg shell function for instant navigation:
+  corg t      # cd back to the last org campaign you were in`,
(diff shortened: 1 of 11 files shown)


cmd/camp/org/cycle.go
```diff
@@ -0,0 +1,290 @@
...
+		enc := json.NewEncoder(w)
+		enc.SetIndent("", "  ")
+		return enc.Encode(cycleOutput{
+			SchemaVersion: cycleSchemaVersion,
+			Action:        action,
+			From:          cycleCampaignOutput{ID: from.ID, Name: from.Name, Org: from.Org, Path: from.Path},
+			To:            cycleCampaignOutput{ID: to.ID, Name: to.Name, Org: to.Org, Path: to.Path},
+		})
+	default:
+		_, err := fmt.Fprintf(w, "cd %s\n", to.Path)
```
obey-agent: **Minor consistency:** `--shell-connect` uses `remote.ShellQuote(to.Path)`, but the default branch prints `cd %s` unquoted. If anyone ever evals the default output (or a path contains spaces), this breaks.

Not a shell-integration bug today (`corg` uses `--shell-connect`), but matching `ShellQuote` on the default branch (or documenting that default is human-display only) would remove the footgun.

obey-agent (commented):
## Verdict: Approve

## Overview
Well-scoped shell/navigation addition. Cycle logic lives in Go with the same `--shell-connect` / `--print` / `--json` patterns as `camp switch`, and the shell wrappers stay thin dispatchers. `cwi` is the obvious workitem alias parity with `csw`/`cgo`.

## Key Findings
1. **Default `cd` path unquoted** — shell-connect is safe; default is not. Non-blocking consistency nit. See inline.
2. Single-member org: empty stdout + stderr note is the correct eval-friendly behavior.
3. Toggle reuses `LastAccess` without a schema bump — good reuse.
4. Active-only vs `--all` scoping and cross-org isolation look correct from the resolver tests described.

## What's Done Well
- Stable name-ordered ring with ID tiebreak.
- Flag mutual exclusion (`--print`/`--json`/`--shell-connect`).
- `agent_allowed` annotations match switch.
- Templates covered for zsh/bash/fish; real zsh e2e walk called out.
- Docs regenerated for `org next` / `org toggle`.

## Staff Standard
Yes. Small surface, reuses existing registry/switch mechanics, tested across shells. Ship it.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Re-review under a zero-open-findings bar. The `cwi`/`corg` design remains solid, but the previously noted unquoted default `cd` path is not fixed and is now a merge blocker.

## Key Findings
1. **Default `cd %s` is unquoted** while `--shell-connect` uses `ShellQuote`. Fix before merge. See inline.
2. Single-member no-op + stderr note remains correct.
3. Toggle via LastAccess without schema bump remains good.

## What's Done Well
- Cycle logic in Go, thin shell wrappers.
- Flag mutual exclusion and agent_allowed annotations.
- Cross-shell template + unit coverage.

## Staff Standard
Ship after the default-path quoting fix only.

obey-agent (commented):
## Verdict: Approve

## Overview
Re-review of tip `0e9663e`. Prior finding (unquoted default `cd`) is fixed: both `--shell-connect` and default emit `cd --` with `remote.ShellQuote(to.Path)`, and tests lock that prefix. No open findings.

## Key Findings
- None remaining.
- `cwi` / `corg` design, stable name-ordered cycle, toggle via LastAccess, and shell wrappers remain correct.

## What's Done Well
- Quoting fix applied to both emit paths, not only shell-connect.
- Tests assert `cd -- ` prefix for plain and shell-connect output.

## Staff Standard
Yes. Merge-ready.

obey-agent:
**Key finding addressed (at `0e9663e1`)** — the default (no-flag) `org next`/`org toggle` branch printed `cd <path>` unquoted while `--shell-connect` quoted it, so a target path with a space (or leading dash) broke if the line was copied or eval'd. Now emits `cd -- <ShellQuote(path)>` to match. Test asserts the quoted prefix and that a spaced path is quoted. `just gate` green.

obey-agent (commented):
## Verdict: Approve (authoritative)

## Cleanup
Earlier on tip `0e9663e` this PR incorrectly received both **Request Changes** and **Approve** from obey-agent. That was reviewer error, not an unresolved code fight:

1. First pass approved `d02638a`.
2. Under the zero-findings pass I re-opened the old unquoted-`cd` nit without re-checking the tip.
3. Tip `0e9663e` already quotes both emit paths (`cd --` + `remote.ShellQuote`); the Request Changes was wrong and has been dismissed.

## Current tip (`0e9663e`)
- Default and `--shell-connect` both emit `cd --` with `ShellQuote`.
- `cwi` / `corg`, stable name-ordered cycle, toggle via LastAccess: no open findings.

## Staff Standard
Yes. Merge-ready. Ignore the dismissed Request Changes on this commit.

obey-agent:
Clarifying review noise: **Final verdict is Approve.** Tip  is merge-ready. The Request Changes on that same commit was a false re-open of the unquoted-cd nit without re-reading the tip (already fixed with  + ShellQuote). GitHub cannot dismiss COMMENTED reviews, so ignore that obsolete verdict; the last **Approve (authoritative)** is the active judgment.

PR #444, opened 2026-07-17
feat: share attachment markers across campaigns

lancekrogers:
## What changed

- Extend attachment `.camp` markers with `campaign_ids` while retaining `active_campaign_id` as the direct-path fallback.
- Resolve shared attachments by the logical campaign-local symlink path.
- Allow `camp attach` to add a second campaign binding and `camp detach` to remove only the current campaign binding.
- Preserve logical symlink paths for `camp pin` and `camp unpin`.
- Keep linked-project markers single-owner.
- Add unit and integration coverage plus updated CLI documentation.

## Why

A single external reference directory can be intentionally used by multiple campaigns. The previous marker model could only bind it to one campaign, causing `cgo` and pinning from another campaign-local symlink to resolve to the wrong campaign.

## Validation

- `go test ./...`
- `go test -tags integration ./tests/integration -run 'TestAttachmentPin_(HappyPath|RejectsExternalWithoutMarker)$|TestSharedAttachment_ResolvesPerCampaignSymlink' -count=1`


obey-agent (commented):
## Verdict: Approve

## Overview
Correctly solves multi-campaign shared attachments. The previous single-owner marker forced the wrong campaign when the same external tree was symlinked from two campaigns. Logical symlink path selection + multi-id marker + pin path preservation is the coherent fix.

## Key Findings
- No blocking issues.
- Linked-project markers remain single-owner (kind-gated) — important boundary preserved.
- Detach is campaign-scoped when cwd resolves a campaign; legacy remove-all outside a campaign is retained intentionally.
- Pin/unpin keep logical campaign-relative paths so shared targets do not collapse to the fallback campaign.
- Integration test `TestSharedAttachment_ResolvesPerCampaignSymlink` exercises the real multi-campaign flow end-to-end.

## What's Done Well
- Backward-compatible read path for legacy markers (active_campaign_id / campaign_id → EffectiveCampaignIDs).
- Detection prefers path-within-campaign over fallback active id.
- Force rewrite semantics updated for shared markers without trashing sibling bindings by default.
- Docs/README examples show the second-campaign attach flow.

## Staff Standard
Yes. This is a real correctness fix for multi-campaign workspaces, not a convenience tweak. Merge with confidence after CI.

internal/campaign/link_marker.go
```diff
@@ -23,15 +23,18 @@ const (
...
 )
 
-// LinkMarker records the active campaign context for a linked directory.
+// LinkMarker records the campaign context for a linked directory.
 // A Kind="project" marker is the original linked-project shape and is paired
 // with a projects/<name> symlink and a registry entry. A Kind="attachment"
 // marker is a lighter binding for arbitrary directories the user has already
-// symlinked into the campaign.
+// symlinked into the campaign. Attachments may be referenced by multiple
+// campaigns; ActiveCampaignID remains the fallback for direct access that does
```
obey-agent, line 31: **Schema note (non-blocking):** Keeping `active_campaign_id` as the direct-path fallback while `campaign_ids` holds additional bindings is the right model.

Operational implication: a bare `cd` into the attachment target (not via a campaign-local symlink) always resolves to the *first* bound campaign. Code comments already say this; keep CLI help for attach/detach equally explicit so operators are not surprised.

obey-agent, line 31: **Must fix before merge (re-opened under zero-findings bar).** Code comments state that bare access to a shared attachment falls back to the first bound campaign, but CLI help for `attach`/`detach` still reads like single-owner context in places.

Add one explicit sentence to attach/detach Long help (and README if not already clear): *direct cd into the attachment target (not via a campaign-local symlink) always resolves to the first bound campaign (`active_campaign_id`)*. Operators will hit this.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Re-review under a zero-open-findings bar. Multi-campaign attachment model is correct; the remaining operator-docs gap about direct-path fallback is still open and blocks merge.

## Key Findings
1. **Direct-path fallback campaign must be stated in CLI help**, not only code comments. See inline.
2. Linked-project single-owner boundary remains correctly preserved.
3. Integration test for per-symlink campaign resolution remains the right proof.

## What's Done Well
- `campaign_ids` + `active_campaign_id` fallback model.
- Logical pin paths for shared targets.
- Campaign-scoped detach.

## Staff Standard
Document the fallback in user-facing help, then merge.

obey-agent:
**Key finding addressed (at `70a0d155`)** — the code comments explained that a bare `cd` into a shared attachment target resolves to the first bound campaign, but the CLI help did not. `attach` and `detach` help now state the shared-attachment resolution rule (and how detaching the fallback campaign shifts it) so operators are not surprised. CLI reference regenerated. `just gate` green.

internal/campaign/cache.go
```diff
@@ -45,8 +45,10 @@ func DetectCached(ctx context.Context) (string, error) {
 		return "", err
 	}
 
-	// Resolve symlinks for consistent comparison
-	cwd, err = filepath.EvalSymlinks(cwd)
+	// Keep the logical cwd as the cache key. A shared attachment can be
```
obey-agent: **Same hole on the cache path.** Skipping `EvalSymlinks` here only helps if `Getwd` already returned a logical path — which it generally does not after entering via a symlink. Pair this with the Detect logical-path fix; otherwise `DetectCached` and anything using it (`LoadCampaignConfigFromCwd` → detach binding, etc.) still collapse shared attachments to the active fallback campaign.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Re-review of tip `b103a93` (includes the attach/detach help docs follow-up and main merge). The marker model, pin logical paths, campaign-scoped detach API, and docs fix for the prior nit are solid. The multi-campaign *detection* path has a production correctness gap that unit tests miss.

## Key Findings
1. **Logical-path selection broken under real `os.Getwd()`** — after `cd` through a campaign-local symlink, Getwd is physical, `pathWithin` never matches, resolution always falls back to `active_campaign_id`. Detach-from-cwd and `camp id` from the "other" campaign's symlink are wrong. See inlines on `detect.go` / `cache.go`.
2. **Prior docs finding — fixed.** Attach/detach Long help now states symlink vs bare-cd fallback and detach fallback shift. Good.
3. Linked-project single-owner refusal, AddCampaign/RemoveCampaign rebuild, integration happy-path for *explicit* paths remain good — but integration uses `cd symlink && camp`, which is exactly the Getwd path that needs a failing-or-passing regression once fixed.

## What's Done Well
- `campaign_ids` + `active_campaign_id` fallback is the right schema.
- Pins keep campaign-relative logical paths for in-tree symlinks.
- Docs and README examples for second-campaign attach.
- Unit test for logical Detect(symlink) and direct fallback — keep it; add Getwd/Chdir coverage.

## Staff Standard
Not merge-ready until shared attachments resolve correctly when the process cwd is the physical attachment path (normal after `cd` through the symlink). Fix Detect/DetectCached logical-path recovery and prove it with a Chdir-based test.

internal/campaign/detect.go
```diff
@@ -131,20 +132,38 @@ func detectCampaignByWalking(ctx context.Context, startDir string) (string, erro
...
-func resolveMarkerCampaignRoot(ctx context.Context, marker *LinkMarker) (string, bool, error) {
+func resolveMarkerCampaignRoot(ctx context.Context, marker *LinkMarker, startDir string) (string, bool, error) {
 	if marker == nil {
 		return "", false, nil
 	}
 
-	if activeID := marker.EffectiveCampaignID(); activeID != "" {
-		root, found, err := lookupRegisteredCampaignRoot(ctx, activeID)
+	// When a shared attachment is reached through a campaign-local symlink,
+	// prefer the campaign whose root contains the logical path. This preserves
```
obey-agent, line 185: **Shared-attachment selection does not survive real cwd.** This pathWithin logic needs a *logical* startDir under the campaign-local symlink. `Detect` with empty startDir uses `os.Getwd()`, which on Unix is the *physical* path after `cd` through a symlink.

Common path — `cd campaign-b/docs/shared && camp id` / detach-from-cwd / `DetectCached` — sees physical `/external/...`, pathWithin matches nothing, and resolution falls through to `active_campaign_id` only. The non-fallback campaign's symlink then mis-resolves.

Unit tests call `Detect(ctx, symlinkPath)` with an explicit logical path and miss this. Add a test that `Chdir`s into the symlink and calls `DetectFromCwd`/`DetectCached`.

Fix: prefer `$PWD` when absolute, exists, and `EvalSymlinks(PWD) == EvalSymlinks(Getwd)` (shell logical path); otherwise Getwd. Same recovery needed in `DetectCached`.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Re-review of tip `b103a93` (docs follow-up + main merge included). Marker model, pin logical paths, campaign-scoped detach, and attach/detach help for the prior docs nit look good. Multi-campaign *detection under a real process cwd* still has a correctness gap unit tests do not catch.

## Key Findings
1. **Logical-path selection breaks with `os.Getwd()`** — after `cd` through a campaign-local symlink, Getwd is physical, `pathWithin` never matches, resolution always uses `active_campaign_id`. Wrong campaign for `camp id`, detach-from-cwd, and `DetectCached` when standing in the non-fallback campaign's symlink. See inlines on `detect.go` and `cache.go`.
2. **Prior docs finding — fixed.** Attach/detach Long help now documents symlink vs bare-cd fallback and detach fallback shift.
3. Schema/API (AddCampaign, RemoveCampaign, force rewrite, project-marker refusal) remain sound.

## What's Done Well
- Right multi-campaign marker shape (`campaign_ids` + active fallback).
- Pins keep campaign-relative logical paths for in-tree symlinks.
- README + CLI help for sharing attachments.
- Explicit-path unit tests for logical Detect and direct fallback — keep and extend with Chdir/Getwd.

## Staff Standard
Not merge-ready until shared attachments resolve correctly when the process cwd is the physical attachment path (normal after `cd` through the symlink). Fix Detect/DetectCached logical-path recovery and prove it with a Chdir-based test.

obey-agent (commented):
## Staff review — current head 68f19fd4

The previous shared-attachment concerns are addressed: the active-campaign fallback is documented, logical working-directory resolution is used for cache identity, and symlink/logical-cwd coverage is present.

Verified with `just test unit`: 3433/3433 tests passed across 106 packages.

PR #440, opened 2026-07-17
feat(fresh): add human follow-up configuration TUI

lancekrogers:
## Summary

Reworks `camp fresh configure` into a human-facing workflow dashboard.

- The left pane selects global defaults or a project scope and shows whether the project inherits or overrides the global list.
- The right pane shows the ordered `camp fresh` sequence, including disabled settings and follow-up commands at the point where they run.
- Add, edit, and delete follow-ups use focused overlays; project edits preserve inherited global steps before materializing an override.
- Adds non-interactive `camp fresh show-workflow [project-name]` for the same ordered view in terminals, scripts, and review output.
- Existing `configure show|add|remove` commands remain available for automation.
- Stacked on `fresh-follow-up-workflows` (#432).

## Reused camp TUI patterns

The implementation follows the existing `cmd/camp/org` and `internal/tui/status` dashboard patterns, using the shared adaptive palette, intent TUI styles, and layout helpers for pane sizing, truncation, scrolling, help collapse, and short terminals.

## Verification

- `just gate-fast` — stable/dev builds, vet, lint, and 3,390 dev-profile tests.
- `just docs` — regenerated CLI reference, including `camp fresh show-workflow`.
- Manual PTY verification drove the real binary with `pyte`, confirmed the dashboard/form rendering, and verified the YAML write and inherited-project behavior.
- Real VHS/PTY demo recorded from the actual camp binary:

camp fresh workflow TUI

PR #445, opened 2026-07-17
feat(fresh): reorder follow-up workflow steps

lancekrogers:
## Summary

- add safe reordering for user-defined `camp fresh` follow-up steps
- support `K`/`J` movement in the human-facing configuration TUI
- add `camp fresh configure move <name> --up|--down` for scripts and agents
- preserve global defaults and project override behavior when an order is edited
- document the new command in the generated CLI reference

## Why

PR #440 makes the resolved `camp fresh` sequence visible. This follow-up makes that sequence actionable: humans can change the order of their follow-up commands instead of editing YAML by hand.

Core sync and branch steps remain protected because their ordering carries safety and dependency guarantees. This PR limits movement to the user-defined follow-up portion of the workflow.

## Usage

In `camp fresh configure`, select a follow-up and press `K` to move it earlier or `J` to move it later. The change is persisted immediately and the selected step stays selected.

For automation:

```sh
camp fresh configure move build --up
camp fresh configure move test --down --project camp
```

## Validation

- `go test ./...`
- `just docs`
- `just lint`
- `just gate-fast` — 3,392/3,392 tests passed
- real PTY + pyte verification of TUI movement, persisted YAML order, and ANSI color output

### DIFF
--- a/internal/commands/fresh/configure_tui_update.go
+++ b/internal/commands/fresh/configure_tui_update.go
@@ -45,6 +45,10 @@ func (m *followUpTUIModel) updateBrowse(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		m.moveCursor(-1)
 	case "down", "j":
 		m.moveCursor(1)
+	case "shift+up", "K":
+		return m, m.moveSelectedFollowUp(-1)
+	case "shift+down", "J":
+		return m, m.moveSelectedFollowUp(1)
 	case "left", "h":
 		m.pane = followUpScopesPane
 	case "right", "l", "tab":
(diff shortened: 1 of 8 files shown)


internal/commands/fresh/configure_tui_update.go
```diff
@@ -45,6 +45,10 @@ func (m *followUpTUIModel) updateBrowse(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		m.moveCursor(-1)
 	case "down", "j":
 		m.moveCursor(1)
+	case "shift+up", "K":
```
obey-agent, line 48: **UX clarity (positive):** Mapping move to `K`/`J` (and shift+up/down) while leaving `j`/`k` as selection is the right keybinding split for a vim-ish TUI. Footer and help overlay both document it — good, because this is the easy place to ship a confusing dual-purpose `j`.

PTY+pyte verification mentioned in the PR body is exactly the bar for this kind of change.

obey-agent (commented):
## Verdict: Approve

## Overview
Clean follow-up to the visible fresh sequence work: humans can reorder follow-ups in the TUI, agents/scripts get `camp fresh configure move`, and core sync/branch steps stay protected. Small, well-tested, correctly scoped.

## Key Findings
- No blocking issues.
- **Stack note:** base is `fresh-follow-up-tui`, not `main` — merge order must land the base branch first (or restack onto main after that lands).
- `ReorderFreshFollowUps` only swaps within the user-defined follow-up list; core steps are intentionally out of scope.
- Invalid moves (already first/last, unknown name, bad delta) fail with clear errors.
- Input slice is not mutated — good API hygiene.
- Exactly-one-of `--up`/`--down` validation is correct.

## What's Done Well
- Immediate persist + reselect after TUI move keeps the interaction solid.
- CLI reference docs regenerated.
- Table tests for reorder happy path and rejects.
- Real PTY verification called out (not just unit green).

## Staff Standard
Yes. Ship once the base PR is ready; this is the right amount of product surface for the problem.

obey-agent:
**Stack note resolved** — base retargeted from `fresh-follow-up-tui` (merged as #440) to `main`. The branch is a single commit on top of main and is now MERGEABLE/CLEAN with no code change needed. The reorder feature itself was already approved; this only completes the restack the review called for.

PR #446, opened 2026-07-17
fix: converge camp worktrees list and camp project worktree list on git-derived enumeration

lancekrogers:
## Problem

PR #429 fixed worktree enumeration for `cgo wt` / `camp go wt` navigation:
`internal/nav/index/builder.go` now uses `git worktree list --porcelain`
per project instead of walking `projects/worktrees/<project>/<child>`.
`camp worktrees list` and `camp project worktree list` still enumerated the
old way: scan the conventional `projects/worktrees/<project>/` directory,
then intersect with `git worktree list`. Any worktree living outside that
layout was silently dropped from both commands even though git tracked it.
This was deliberately excluded from #429 because fixing it changes the
`--json` output shape, which is a real contract change and deserves its
own review.

Real repro against this campaign's own `camp` project (not a synthetic
fixture): `camp-main-tests`, `camp-wi-ref-selector`, and the
`camp-worktree-list` worktree this PR was written in all live flat under
`projects/worktrees/`, one level up from the conventional
`projects/worktrees/camp/<name>` layout, alongside the original repro
`projects/worktrees/fix-camp-392`. All four were invisible to both
commands before this fix.

## Fix

Both commands now enumerate directly from `git worktree list`, filtered
through a shared `worktree.IsLinkedWorktree(projectPath, entry)` helper
that excludes the project's own main working tree, bare entries,
git-internal paths (including the submodule `.git/modules/<name>` case),
and hidden directories.

`IsLinkedWorktree` is extracted from `nav/index/builder.go`'s
`worktreeTarget` (unchanged behavior, same test cases now live in
`internal/worktree/filter_test.go`) into `internal/worktree` so every
enumerator agrees on the same rule instead of three copies drifting
independently:

- `internal/nav/index/builder.go` (`cgo wt`, unchanged behavior, just
  delegates now)
- `cmd/camp/worktrees/list.go` (`camp worktrees list [--json]`)
- `cmd/camp/project/worktree/list.go` (`camp project worktree list`)

The old `projects/worktrees/<project>/` directory scan and its
intersection helper (`filterRegisteredWorktreeNames`, and
`pathManager.ListProjectWorktrees`/`WorktreePath` reconstruction) are
removed from both list commands. `camp project worktree add/remove` still
use `PathManager` for creating/removing worktrees at the conventional
location; that is unchanged and out of scope here, only the two `list`
paths moved to git-derived enumeration.

## --json contract change (camp worktrees list --json)

Schema version stays `worktrees-list/v1alpha1` (already documented in
`docs/json-contracts.md` as a deprecated compatibility surface; only the
error envelope is versioned, the success payload has no `schema_version`
field to bump). Field names and types on `WorktreeListItem` /
`WorktreeListResult` are unchanged. What changes:

- **Enumeration is a strict superset.** Every worktree previously listed
  is still listed, with identical field values (verified below: 17/17
  pre-existing entries byte-identical before and after). Worktrees outside
  the conventional location now additionally appear.
- **`path` is now always the real git-reported path**, not a path
  reconstructed from `projects/worktrees/<project>/<name>`. For
  conventional-location worktrees these were already identical, so no
  value changes there. For non-conventional worktrees, `path` is now
  correct instead of simply absent.
- **`name` is the worktree's real directory basename** (`filepath.Base` of
  the git-reported path) rather than a directory-scan entry name. Same
  value for conventional worktrees, newly correct for the rest.

No change to `branch`, `lastAccessed`, `stale`, `staleReason`, `total`, or
`stale` count semantics; `buildWorktreeListItem` is unchanged, it is now
just called with the real path instead of a reconstructed one.

### Before / after (real campaign data, `camp worktrees list --project camp --json`)

Built the pre-fix `main` binary and the fixed binary and ran both
read-only against this actual campaign (17 conventional-location `camp`
worktrees plus the 4 non-conventional ones above):

**Before** (`main`, `total: 17`): only conventional-location worktrees,
e.g.
```json
{
  "project": "camp",
  "name": "camp-dead-surface",
  "path": "/Users/.../projects/worktrees/camp/camp-dead-surface",
  "branch": "camp-dead-surface",
  "lastAccessed": "5 days ago",
  "stale": false
}
```
`camp-main-tests`, `camp-wi-ref-selector`, `camp-worktree-list`,
`fix-camp-392` are entirely absent from the array.

**After** (this branch, `total: 21`): the same 17 entries, byte-identical,
plus the 4 previously-missing worktrees, e.g.
```json
{
  "project": "camp",
  "name": "camp-worktree-list",
  "path": "/Users/.../projects/worktrees/camp-worktree-list",
  "branch": "fix/worktree-list-git-enumeration",
  "lastAccessed": "5 minutes ago",
  "stale": false
},
{
  "project": "camp",
  "name": "fix-camp-392",
  "path": "/Users/.../projects/worktrees/fix-camp-392",
  "branch": "1e6f94a (detached)",
  "lastAccessed": "5 days ago",
  "stale": false
}
```

### Before / after (`camp project worktree list --project camp`, human text, no --json flag on this command)

Before: `fix-camp-392` / `camp-main-tests` / `camp-wi-ref-selector` /
`camp-worktree-list` never appear in the output (0 matches). After:
```
  camp-main-tests
    Branch: fix/main-integration-workitem-tests
    Path:   projects/worktrees/camp-main-tests

  fix-camp-392
    Branch: HEAD (detached)
    Path:   projects/worktrees/fix-camp-392
```
paths now render relative to campaign root the same way conventional
worktrees always did (`displayWorktreePath`, `filepath.Rel` against
campRoot, falling back to the absolute path only if the worktree lives
outside the campaign root entirely).

## Consumers checked

- Grepped `WorktreeListResult`/`WorktreeListItem`/`worktrees-list` across
  the whole campaign (camp, camp-graph, obey-platform-monorepo,
  festival-app, workflow/design, festivals/). No script or app parses this
  JSON structurally.
- `festival-app`'s "Worktree List" action
  (`src-tauri/src/commands/projects.rs` `ProjectActionRequest::WorktreeList`,
  wired from `src/lib/explorerActions.ts`) shells out to
  `camp project worktree list --project <name>` (the human-text command,
  no `--json`) and displays raw stdout in a log view; it does not parse
  fields, so this change only improves what's shown to the user (correct
  paths, previously-missing worktrees).
- `docs/json-contracts.md` already marks `camp worktrees list --json` as a
  deprecated compatibility surface with no `schema_version` on the success
  body; left unchanged, still accurate.
- The `camp` copies vendored inside `festival-app/camp/` and
  `obey-platform-monorepo/camp/` are bundled dev-profile source checkouts
  of this same repo, not independent consumers of the contract.

## Tests

- `internal/worktree/filter_test.go` (new): table-driven unit tests for
  `IsLinkedWorktree` covering bare, empty path, main working tree,
  submodule `.git/modules` case, hidden directory, preferred-location
  worktree, non-preferred-location worktree, and a worktree entirely
  outside the worktrees directory. Mirrors the existing
  `TestWorktreeTarget` cases in `internal/nav/index/builder_test.go`
  (unchanged, still passing since `worktreeTarget`'s behavior is
  identical, just delegates now). `TestContainsGitDir` moved alongside
  `IsLinkedWorktree`.
- `cmd/camp/worktrees/list_test.go`: removed
  `TestFilterRegisteredWorktreeNames_DropsCopiedCheckoutChildren`, its
  target function is deleted. The property it protected (a copied
  build-artifact directory like `node_modules`/`bin` sitting next to a
  real worktree must not be listed) now holds by construction: enumeration
  comes straight from `git worktree list`, so a directory git never
  registered can never produce an entry to filter in the first place.
- `tests/integration/worktrees_list_test.go` (new,
  `//go:build integration`): `TestWorktreesList_ResolvesAcrossAllLocations`
  builds a real campaign + submodule project in a container, creates a
  preferred-location worktree, a worktree outside the preferred location,
  and a plain (non-git) directory under the preferred location, then
  asserts `camp worktrees list --json` and `camp project worktree list`
  both report the two real worktrees with correct paths and both exclude
  the plain directory. Reuses the `setupWorktreeNavCampaign` helper
  already shared with `navigation_worktree_test.go`.

## Verification

Run from `/Users/lancerogers/Dev/AI/obey-campaign/projects/worktrees/camp-worktree-list`:

- `just build default-build`: clean (vet + build, 124/124 packages).
- `just test unit`: 3429/3429 tests passed, 106 packages, no failures. No
  pre-existing `TestIntegration_Workitem*` failures encountered (those are
  integration-only, not part of this run).
- `just lint`: `0 issues`, `lint-no-host-fs-tests: clean`,
  `lint-no-fmt-errorf: clean`.
- `go vet ./...` and `go vet -tags=integration ./...`: clean.
- `just test integration-run TestWorktreesList_ResolvesAcrossAllLocations`:
  PASS (new test).
- `just test integration-run` for
  `TestGoWorktree_ResolvesAcrossAllLocations`,
  `TestCompleteWorktree_BranchesAcrossLocations`, and all
  `TestWorktreesClean_*` cases (the tests most likely to regress from the
  `internal/worktree`/`nav/index` refactor): all PASS.

## Other changes worth flagging

- `camp project worktree list` no longer calls
  `config.LoadCampaignConfig` (it only existed to build the now-unused
  `PathManager`). `project.Resolve` still resolves the project and surfaces
  a campaign-config error through its own path if the config is malformed,
  so error coverage is preserved, just through a different call site.
- `camp project worktree list`'s branch label for detached worktrees
  (`HEAD (detached)`, from git's own parsed entry) already differed from
  `camp worktrees list --json`'s label (short SHA + `(detached)`, from a
  separate disk-read helper) before this change; verified with the `main`
  binary against `workitem-validate-repair` in this campaign. This PR does
  not touch that inconsistency, flagging it as a pre-existing, unrelated
  follow-up.

## Residual risks

- `camp worktrees list --json`'s `path` field is now always the real git
  path; any external script that assumed the conventional
  `projects/worktrees/<project>/<name>` shape for every entry (rather than
  reading the `path` field) would need to adjust. No such consumer was
  found in this campaign or its projects.
- Docs (`docs/json-contracts.md`) were left as-is since the schema version
  and field shapes are unchanged; happy to add a note there if reviewers
  want the enumeration-completeness fix called out explicitly in that
  table.

Ref: intent `camp-worktrees-list-and-camp-20260715-024134`.

### DIFF
--- a/cmd/camp/worktrees/list.go
+++ b/cmd/camp/worktrees/list.go
@@ -13,10 +13,8 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/Obedience-Corp/camp/internal/campaign"
-	"github.com/Obedience-Corp/camp/internal/config"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 	"github.com/Obedience-Corp/camp/internal/jsoncontract"
-	"github.com/Obedience-Corp/camp/internal/paths"
 	"github.com/Obedience-Corp/camp/internal/project"
 	"github.com/Obedience-Corp/camp/internal/ui"
 	"github.com/Obedience-Corp/camp/internal/worktree"
@@ -93,14 +91,6 @@ func runWorktreesList(cmd *cobra.Command, args []string) error {
 		return camperrors.Wrap(err, "not in a campaign")
 	}
 
-	cfg, err := config.LoadCampaignConfig(ctx, campRoot)
-	if err != nil {
-		return camperrors.Wrap(err, "failed to load campaign config")
-	}
-
-	resolver := paths.NewResolver(campRoot, cfg.Paths())
-	pathManager := worktree.NewPathManager(resolver)
-
 	// In a terminal, a bare `camp worktrees list` opens the interactive browser
 	// (the same pattern as `camp list`). The browser holds every worktree and
 	// applies --stale as a live, toggleable filter, so load unfiltered for it;
@@ -111,7 +101,7 @@ func runWorktreesList(cmd *cobra.Command, args []string) error {
 		loadStale = false
 	}
 
-	result, err := listWorktrees(ctx, campRoot, pathManager, listProject, loadStale)
+	result, err := listWorktrees(ctx, campRoot, listProject, loadStale)
 	if err != nil {
 		return err
 	}
@@ -132,7 +122,7 @@ type listProjectTarget struct {
 	path string
 }
 
-func listWorktrees(ctx context.Context, campRoot string, pm *worktree.PathManager, filterProject string, staleOnly bool) (*WorktreeListResult, error) {
+func listWorktrees(ctx context.Context, campRoot string, filterProject string, staleOnly bool) (*WorktreeListResult, error) {
 	var allWorktrees []WorktreeListItem
 
 	projectTargets, err := listWorktreeProjectTargets(ctx, campRoot, filterProject)
@@ -140,7 +130,9 @@ func listWorktrees(ctx context.Context, campRoot string, pm *worktree.PathManage
 		return nil, err
 	}
 
-	// Scan each project
+	// Scan each project. git worktree list is the source of truth: it finds
+	// every linked worktree regardless of where it lives on disk, not just
+	// those under the conventional projects/worktrees/<project>/ layout.
 	for _, target := range projectTargets {
 		gitEntries, err := worktree.NewGitWorktree(target.path).List(ctx)
 		if err != nil {
@@ -151,14 +143,12 @@ func listWorktrees(ctx context.Context, campRoot string, pm *worktree.PathManage
 		}
 
 		projectName := target.name
-		fsWorktrees, err := pm.ListProjectWorktrees(projectName)
-		if err != nil {
-			continue // Skip projects with errors
-		}
-
-		for _, name := range filterRegisteredWorktreeNames(projectName, fsWorktrees, gitEntries, pm) {
-			wtPath := pm.WorktreePath(projectName, name)
-			info := buildWorktreeListItem(projectName, name, wtPath)
+		for _, entry := range gitEntries {
+			if !worktree.IsLinkedWorktree(target.path, entry) {
+				continue
+			}
+			name := filepath.Base(filepath.Clean(entry.Path))
+			info := buildWorktreeListItem(projectName, name, entry.Path)
 			allWorktrees = append(allWorktrees, info)
 		}
 	}
@@ -208,21 +198,6 @@ func listWorktreeProjectTargets(ctx context.Context, campRoot, filterProject str
 		})
 	}
 	return targets, nil
-}
-
-func filterRegisteredWorktreeNames(projectName string, names []string, entries []worktree.GitWorktreeEntry, pm *worktree.PathManager) []string {
-	registered := make(map[string]struct{}, len(entries))
-	for _, entry := range entries {
-		registered[filepath.Clean(entry.Path)] = struct{}{}
-	}
-
-	filtered := make([]string, 0, len(names))
-	for _, name := range names {
-		if _, ok := registered[filepath.Clean(pm.WorktreePath(projectName, name))]; ok {
-			filtered = append(filtered, name)
-		}
-	}
-	return filtered
 }
 
 func buildWorktreeListItem(project, name, path string) WorktreeListItem {
--- a/cmd/camp/worktrees/list.go
+++ b/cmd/camp/worktrees/list.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/charmbracelet/lipgloss"
@@ -20,7 +21,13 @@ import (
 	"github.com/Obedience-Corp/camp/internal/worktree"
 )
 
-const WorktreesListJSONVersion = "worktrees-list/v1alpha1"
+// WorktreesListJSONVersion is bumped to v1alpha2 because git-derived
+// enumeration changed the meaning of the emitted path and name fields: path is
+// now the worktree's real on-disk location (which may be outside the
+// conventional projects/worktrees/<project>/ layout), and name is
+// disambiguated to a campaign-relative path when two linked worktrees share a
+// basename. Clients can key on this version to detect the new semantics.
+const WorktreesListJSONVersion = "worktrees-list/v1alpha2"
 
 var (
 	listProject string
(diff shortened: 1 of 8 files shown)


obey-agent (commented):
## Verdict: Request Changes

## Overview
Correct direction: one `IsLinkedWorktree` rule shared by nav index, `camp worktrees list`, and `camp project worktree list`, with git as source of truth. That fixes the real "loose worktree invisible" bug. Two contract/UX issues block merge.

## Key Findings
1. **Schema version not bumped** despite changed path/name semantics for `--json`. See inline.
2. **Basename-only `name` collides** when two linked worktrees share a directory base. See inline.
3. Integration test for loose vs preferred locations is good — keep it, and extend for basename collision.

## What's Done Well
- Extracting `IsLinkedWorktree` instead of a third copy of the filter.
- Dropping the FS ∩ git intersection that silently dropped real worktrees.
- Containerized integration coverage of the original repro shape.

## Staff Standard
Not yet — fix schema version + name uniqueness, then ship.

cmd/camp/worktrees/list.go
```diff
@@ -151,14 +143,12 @@ func listWorktrees(ctx context.Context, campRoot string, pm *worktree.PathManage
...
-		}
-
-		for _, name := range filterRegisteredWorktreeNames(projectName, fsWorktrees, gitEntries, pm) {
-			wtPath := pm.WorktreePath(projectName, name)
-			info := buildWorktreeListItem(projectName, name, wtPath)
+		for _, entry := range gitEntries {
+			if !worktree.IsLinkedWorktree(target.path, entry) {
+				continue
+			}
+			name := filepath.Base(filepath.Clean(entry.Path))
```
obey-agent, line 157: **Name collision under git-derived enumeration.** `name := filepath.Base(entry.Path)` means two linked worktrees for the same project whose basenames match (preferred `.../camp/foo` vs loose `/tmp/foo`) collapse to the same `name` in table/JSON output.

Disambiguate on collision (campaign-relative path, or full path as name) and add a regression test for same-basename / different-path worktrees.

cmd/camp/worktrees/list.go
```diff
@@ -151,14 +143,12 @@ func listWorktrees(ctx context.Context, campRoot string, pm *worktree.PathManage
...
+			info := buildWorktreeListItem(projectName, name, entry.Path)
```
obey-agent, line 158: **Related contract item:** emitting real git paths here changes the `path` (and basename `name`) meaning of `worktrees-list/v1alpha1`. Bump the schema version constant to `worktrees-list/v1alpha2` in the same change set so clients can detect the new semantics. See review summary.

obey-agent (commented):
## Staff review — current head a3c78042

The previous review findings are addressed: worktree names are disambiguated for same-basename collisions, the JSON schema is bumped to v1alpha2, and the integration coverage is present.

Verified with `just test unit`: 3430/3430 tests passed across 106 packages.

PR #447, opened 2026-07-17
Fix 5 pre-existing integration test failures in workitem commit/validate paths

lancekrogers:
## Summary

Fixes the 5 containerized integration test failures tracked in intent
`camp-main-4-pre-existing-20260715-160457` (verified failing on a clean
origin/main worktree before this branch, all 5 now pass).

### TestIntegration_WorkitemCommits_FiltersFalsePositives

**Root cause:** `internal/commands/workitem/commit.go` (`runCommit`) recorded
`flags.Message` (the raw, untagged `-m` text) as the ledger evidence subject
for `camp workitem commit`, instead of the actual tagged git commit subject
that `commit.Workitem`/`doCommit` composes internally and never returns.
Every other `CommitEvidence` call site (`cmd/camp/commit.go`,
`cmd/camp/project/commit.go`) already passes the fully composed, tagged
message; `workitem commit` was the one call site out of step. Since
`workitem commits --json` answers from the ledger fast path whenever
evidence exists, this meant `TagParts.WorkitemRef`/`CampaignID` came back
empty and untagged commits weren't filtered out (they had no tag to filter
on at all).

**Fix:** after the commit lands, read the actual subject back via
`git log -1 --format=%s <sha>` (new `lastCommitSubject` helper, mirroring
the existing `lastCommitSHA`) and pass that to `CommitEvidence` instead of
`flags.Message`.

**Intended-behavior reasoning:** the doc comment on
`commitsFromLedgerEvents` states the ledger and git-log-scan paths must
produce "the --json commit contract... identical whichever path answered."
Two of three `CommitEvidence` call sites already honored that. This was a
straightforward missed case, not an ambiguous design question.

### TestIntegration_WorkitemCommits_IncludesCampaignAndLinkedProject

**Root cause:** the ledger path (`ledger.RepoLabel`) labels the campaign
root `"campaign-root"` (explicitly documented as the convention shared with
`camp doctor`'s dangling-evidence check), but the git-log scan path
(`repoDisplayPath` in `commits.go`) still returned `"."` for the same
repo. The two `workitem commits` answer paths disagreed on the same
field for the same logical location.

**Fix:** `repoDisplayPath` now returns `"campaign-root"` for the campaign
root, matching `ledger.RepoLabel`. Updated the stale doc comment on
`CommitRecord.Repo` accordingly.

**Intended-behavior reasoning:** `internal/doctor/checks/ledger.go` already
treats `"campaign-root"`, `"."`, and `""` as synonyms when *reading*
(back-compat for historical ledger entries), but `internal/ledger/commit.go`
(`RepoLabel`)'s doc comment establishes `"campaign-root"` as the current
canonical *write* convention. Test assertion (`repos["."] || repos[""]`)
updated to `repos["campaign-root"]`.

### TestIntegration_WorkitemValidate_ReportsFindings

**Root cause:** `repairCommandFor` in `internal/commands/workitem/validate.go`
unconditionally wraps every path in `remote.ShellQuote`, even plain
kebab-case workflow slugs with no special characters. This contradicts
`docs/workitem-validate-reference.md`'s documented example
(`camp workitem repair workflow/design/api-refactor`, unquoted) and the
integration test's plain-path expectations.

**Fix:** quoting is now conditional. A new `repairCommandSafeUnquoted`
regexp (`^[A-Za-z0-9._/-]+$`) matches the safe, no-quoting-needed case;
anything else (e.g. a hand-edited path with spaces or a `'`) still gets
`remote.ShellQuote`'d for copy-paste safety.

**Intended-behavior reasoning:** this could have gone either way (always
quote vs. never quote vs. conditional), except the existing unit test
`TestRepairCommandForShellQuotesPath` (from the same PR #395 that
introduced `repairCommandFor`) explicitly locks in quoting for a path
containing a space and an apostrophe. Conditional quoting is the only
option consistent with both that unit test and the plain-path integration
test/docs example, so it's the unambiguous fix rather than a coin flip.
Verified the unit test still passes with the new logic.

### TestIntegration_CommitTags_NoteNoContext

**Root cause:** this test predates the note's own `NT-<ref>` tag segment
(introduced later, PR #394/`c33b7364`, "Inherit ambient commit context on
all intent auto-commit paths"). Its regex
(`^\[commit-tags:[0-9a-f]{1,8}\]`) required the tag to close immediately
after the campaign id, i.e. assumed a context-less note carries no NT- ref
at all. It was never updated when the NT- feature landed.

**Fix:** updated the regex to
`^\[commit-tags:[0-9a-f]{1,8}-NT-[0-9a-f]{6}\]` and the comment to
reflect that a context-less note still carries its own NT- ref.

**Intended-behavior reasoning:** not ambiguous: three other tests
explicitly assert the opposite of this test's old regex.
`TestIntegration_CommitTags_NoteRefSegment`'s own comment says "No
workitem context: the note still gets its own NT-<ref> segment even
though there is no WI- to sit alongside," and
`git.FormatContextTagsFull`'s doc comment independently confirms: "noteRef
is optional... only the note commit path passes one, and it may co-occur
with workitemRef since they describe different things (the ambient
context a note was captured in vs. the note itself)." The note's own ref
is deliberately independent of ambient WI-/quest/festival context. This
test's core intent (no WI-/qst_/FE- leakage into a context-less note) is
preserved; only the over-strict "no NT- either" assumption was corrected.

## Verification

- `just build` (default-build): clean, no errors
- `just test unit`: 3429/3429 passed (106 packages)
- `just lint`: 0 issues, no new host-FS-test violators, no fmt.Errorf
  regressions
- Full containerized integration suite
  (`go test -tags=integration ./tests/integration/...`): 467 passed, 0
  failed, 12 skipped (all pre-existing environment-gated skips: fest
  binary unavailable in container, TTY tests gated behind
  `CAMP_SETTINGS_TTY_TESTS`, deferred fest-side wiring; unrelated to this
  change)
- All 5 originally-failing tests individually re-verified passing after
  the fix, then confirmed passing again as part of the full suite run

## Residual risk

None identified for the fixed behavior itself. Two lower-risk notes for
reviewers:

- The `commit.go` fix does one extra `git log` invocation per
  `workitem commit` to read back the subject; best-effort with a silent
  fallback to `flags.Message` on error (matching the existing
  `lastCommitSHA` warning-on-error, non-fatal pattern for `sha`).
- `repairCommandSafeUnquoted`'s charset is deliberately narrow
  (alnum, `.`, `_`, `/`, `-`) so it only recognizes camp's own generated
  slugs as "safe"; anything else (including unicode) falls through to
  quoting rather than risking a false "safe" classification.

Ref: camp-main-4-pre-existing-20260715-160457

### DIFF
--- a/internal/commands/workitem/commit.go
+++ b/internal/commands/workitem/commit.go
@@ -212,14 +212,19 @@ func runCommit(ctx context.Context, cmd *cobra.Command, flags commitFlags) error
 		// reusing flags.Message (untagged): every other CommitEvidence call
 		// site records the actual tagged git subject, and readers like
 		// `workitem commits` parse the campaign tag back out of it.
-		subject := flags.Message
-		if s, serr := lastCommitSubject(ctx, plan.RepoRoot, sha); serr == nil {
-			subject = s
+		subject, serr := lastCommitSubject(ctx, plan.RepoRoot, sha)
+		record, subj, warn := evidenceDecision(sha, subject, serr)
+		if warn != "" {
+			if _, writeErr := fmt.Fprintln(cmd.ErrOrStderr(), warn); writeErr != nil {
+				return writeErr
+			}
 		}
-		ledger.NewFromRoot(ctx, campaignRoot, ledger.WarnTo(cmd.ErrOrStderr())).
-			CommitEvidence(ctx,
-				ledgerkit.Scope{Workitem: plan.WorkitemRef, Festival: plan.FestivalRef, Quest: plan.QuestID},
-				campaignRoot, plan.RepoRoot, sha, subject)
+		if record {
+			ledger.NewFromRoot(ctx, campaignRoot, ledger.WarnTo(cmd.ErrOrStderr())).
+				CommitEvidence(ctx,
+					ledgerkit.Scope{Workitem: plan.WorkitemRef, Festival: plan.FestivalRef, Quest: plan.QuestID},
+					campaignRoot, plan.RepoRoot, sha, subj)
+		}
 	}
 	if flags.JSON {
 		return emitJSON(cmd.OutOrStdout(), plan, sha)
@@ -291,4 +296,20 @@ func lastCommitSubject(ctx context.Context, repoRoot, sha string) (string, error
 		return "", err
 	}
 	return strings.TrimSpace(string(out)), nil
+}
+
+// evidenceDecision decides whether to record ledger evidence for a landed
+// commit given the tagged subject read back from git. A failed subject read
+// must never fall back to the untagged commit message and still record
+// evidence: that would look like a clean tagged ledger write and recreate the
+// git/ledger mismatch this path exists to prevent. On error it returns
+// record=false with an operator warning; the commit already landed, so only the
+// ledger annotation is skipped.
+func evidenceDecision(sha, subject string, subjectErr error) (record bool, subj, warn string) {
+	if subjectErr != nil {
+		return false, "", fmt.Sprintf(
+			"warning: committed %s but could not read its subject for ledger evidence (%v); "+
+				"skipping ledger record to avoid an untagged entry", sha, subjectErr)
+	}
+	return true, subject, ""
 }
--- a/internal/commands/workitem/commit.go
+++ b/internal/commands/workitem/commit.go
@@ -207,10 +207,19 @@ func runCommit(ctx context.Context, cmd *cobra.Command, flags commitFlags) error
 		}
 	}
 	if sha != "" {
+		// commit.Workitem composes the tagged subject internally and never
+		// returns it, so the landed subject is read back from git rather than
+		// reusing flags.Message (untagged): every other CommitEvidence call
+		// site records the actual tagged git subject, and readers like
+		// `workitem commits` parse the campaign tag back out of it.
+		subject := flags.Message
+		if s, serr := lastCommitSubject(ctx, plan.RepoRoot, sha); serr == nil {
+			subject = s
+		}
 		ledger.NewFromRoot(ctx, campaignRoot, ledger.WarnTo(cmd.ErrOrStderr())).
 			CommitEvidence(ctx,
 				ledgerkit.Scope{Workitem: plan.WorkitemRef, Festival: plan.FestivalRef, Quest: plan.QuestID},
-				campaignRoot, plan.RepoRoot, sha, flags.Message)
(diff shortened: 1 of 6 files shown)


obey-agent (commented):
## Verdict: Request Changes

## Overview
Targets five real main-branch integration failures with the right root causes: untagged ledger subjects on `workitem commit`, `campaign-root` vs `.` label drift, note NT-ref assertion, and shell-safe repair commands. Almost there; one silent-failure path must not ship.

## Key Findings
1. **`lastCommitSubject` error falls back to untagged `flags.Message`** and still calls `CommitEvidence` — recreates the original ledger/git mismatch when `git log` fails. See inline.
2. `repoDisplayPath` → `campaign-root` alignment with `ledger.RepoLabel` is correct.
3. `repairCommand` quoting for non-slug paths is correct and necessary.

## What's Done Well
- Reading the actual landed subject instead of re-deriving tags is the right fix.
- Mirrors existing `lastCommitSHA` helper style.
- Integration expectations updated to match real product behavior (NT-ref on notes).

## Staff Standard
Fix the subject-read failure path, then merge. Do not leave a silent fallback that writes bad ledger evidence.

internal/commands/workitem/commit.go
```diff
@@ -207,10 +207,19 @@ func runCommit(ctx context.Context, cmd *cobra.Command, flags commitFlags) error
 		}
 	}
 	if sha != "" {
+		// commit.Workitem composes the tagged subject internally and never
+		// returns it, so the landed subject is read back from git rather than
+		// reusing flags.Message (untagged): every other CommitEvidence call
+		// site records the actual tagged git subject, and readers like
+		// `workitem commits` parse the campaign tag back out of it.
+		subject := flags.Message
```
obey-agent: **Silent fallback reintroduces the bug on read failure.** If `lastCommitSubject` fails, this keeps `flags.Message` (untagged) and still writes ledger evidence — the exact mismatch this PR is fixing.

Do not fall back quietly. Prefer failing after a successful commit if the subject cannot be read, or skip `CommitEvidence` with a stderr warning rather than recording the untagged subject. A failed subject read must never look like a successful tagged ledger write.

obey-agent (commented):
## Staff review — current head 49b153fb

The previous evidence fallback issue is fixed: an unreadable last-commit subject no longer silently records untagged evidence, and the success and error paths are covered by tests.

Verified with `just test unit`: 3431/3431 tests passed across 106 packages.

PR #448, opened 2026-07-17
fix(workitem): resolve WI-<ref> selectors and guard worktree links

lancekrogers:
## Problem

Intent `camp-workitem-wi-ref-selector-20260712-214306`.

Three related worktree/selector bugs, all reproduced on the current `origin/main`:

1. **`WI-<ref>` selector does not resolve.** A freshly created/adopted `workflow/design` (or `workflow/explore`) workitem writes its ref to `.workitem` (e.g. `WI-541bcc`), but `camp workitem link WI-541bcc ...`, `camp workitem worktree WI-541bcc`, and `camp project worktree add --workitem WI-541bcc` all error `no workitem matched selector`, while the path/slug selector for the same directory resolves fine and doctor reports 0 findings. The `--workitem` help text already advertises `(ref, path, or id)`, so the ref case was simply unimplemented.

2. **Unadopted design dir derives an invalid id.** `camp project worktree add --workitem <path>` (and `camp worktrees create --workitem <path>`) against a `workflow/design` directory with no `.workitem` marker created the worktree, then failed the link deep in the links validator with `invalid workitem_id design:workflow/design/...: must not contain '/'`, leaving a dangling worktree behind.

3. **`worktree remove` leaves the branch behind.** `git worktree remove` keeps the branch it created, so re-adding a worktree of the same name failed with a raw `fatal: a branch named '...' already exists`, and the user had to discover `--branch` on their own.

Root cause for (1): `selector.Resolve` matched `id`, `key`, `path`, and `slug`, but never the ref, which lives in `SourceMetadata["ref"]` (the `.workitem` `id` becomes `StableID`, not the ref). All four commands route through `selector.Resolve`, so none of them accepted a ref.

## Fix

- **Ref matcher (`internal/workitem/selector/selector.go`).** Added a `ref` tier as the first matcher, comparing the query against `SourceMetadata["ref"]` case-insensitively (consistent with the existing `key`/`slug` `EqualFold` tiers). Refs are `WI-<6 hex>` and unique per workitem, so the tier is unambiguous; two hand-edited items sharing a ref surface an `ErrSelectorAmbiguous` rather than a silent pick.

- **Not-adopted guard (`internal/workitem/adoption.go`).** New `NeedsAdoption(wi)` returns true only for a builtin doc-type item (design/explore) discovered by location with no stable id, plus `NotAdoptedError(path)` that names the exact `camp workitem adopt <path>` recovery. The three worktree-link commands (`camp workitem worktree`, `camp project worktree add --workitem`, `camp worktrees create --workitem`) now resolve and validate the workitem *before* creating the worktree, so a bad or unadopted selector fails fast with a clear message and never leaves a dangling worktree.

- **Leftover-branch hint (`internal/worktree`).** Added a local-only `LocalBranchExists` check and a `WORKTREE_BRANCH_EXISTS` sentinel/constructor. The creator now detects a leftover local branch before `git worktree add` runs and returns `ErrBranchExists`; the `project worktree add` and `worktrees create` commands wrap it with `reuse it with --branch <name>, choose a different name, or delete it with 'git branch -D <name>'`, and `camp workitem worktree` (which has `--name`, not `--branch`) wraps it with a `--name`/delete hint.

## Tradeoffs and rejected alternatives

- **Not-adopted scope.** The guard is deliberately narrow to design/explore doc dirs (via the existing `IsBuiltinDocType`). Festivals and intents are also id-less but are legitimate, non-adoptable surfaces, and custom types cannot be discovered without a marker at all, so broadening the guard to "any id-less item" would have changed behavior for a tool with real users. A primary worktree link needs a ref for `camp p commit` to stamp, and only adopted doc dirs are missing one they should have.

- **Leftover branch: hint vs auto-delete.** I chose the actionable error (the intent's "clear hint" option) over adding a `--delete-branch` flag to `worktree remove`. Deleting a branch on remove risks silently discarding unpushed commits, so making the failure discoverable is the safer default and matches camp's existing pattern of wrapping worktree errors with next-step hints (the `ErrWorktreeExists` wrap already in `createWorktree`). A `--delete-branch` flag remains a reasonable future follow-up.

- **Local-only branch check.** `LocalBranchExists` intentionally ignores remote-tracking refs; `git worktree add -b foo` only conflicts with a local `foo`, and would happily create a local branch when only `origin/foo` exists, so checking the existing `BranchExists` (which also matches remotes) would have produced false rejections.

## Behavior / contract notes (camp has real users)

- **Additive selector change:** `WI-<ref>` now resolves where it previously errored. No previously-resolving selector changes meaning (a ref-shaped query cannot collide with an id/key/path/slug).
- **Error-message-only changes:** the unadopted-dir case and the leftover-branch case were already errors (non-zero exit); only the messages improved and, for the unadopted case, the worktree is no longer created before the error. No `--json` shape changed.

## Verification

- New unit tests: `internal/workitem/selector/selector_test.go` (ref resolves lowercase + uppercase, alongside id/key/path/slug; unknown ref is `ErrSelectorNotFound`; duplicate ref is `ErrSelectorAmbiguous`) and `internal/workitem/adoption_test.go` (table-driven `NeedsAdoption` incl. festival/intent negative cases; error message content).
- New integration tests: `tests/integration/workitem_worktree_selector_test.go` drives the real CLI in the container harness for all three behaviors (`--workitem <ref>` links and tags the commit; unadopted design dir errors with no dangling worktree; remove-then-readd surfaces the hint and `--branch` recovery succeeds).
- Gates in the worktree: `just build-camp`, `just test unit` (3434/3434 passed), `just lint` (0 issues; ratchets clean). Targeted integration: the 3 new tests pass, and the existing `TestIntegration_WorkitemWorktree_*`, `TestIntegration_ProjectCommit_*`, and `TestIntegration_WorkitemLink_*` tests still pass.
- Pre-existing failures on this branch are the known `TestIntegration_WorkitemCommits_*` (CrossRepo, FiltersFalsePositives, IncludesCampaignAndLinkedProject) and `TestIntegration_WorkitemValidate_ReportsFindings`, which live in commits-query/validate paths this change does not touch and which a separate agent is fixing. This change adds no new failures.

### DIFF
--- a/internal/workitem/adoption.go
+++ b/internal/workitem/adoption.go
@@ -1,0 +1,23 @@
+package workitem
+
+import camperrors "github.com/Obedience-Corp/camp/internal/errors"
+
+// NeedsAdoption reports whether wi is a builtin doc-type work item (design or
+// explore) that was discovered by location but has no .workitem marker, so it
+// carries neither a stable id nor a ref. Such an item cannot back a primary
+// worktree link: camp p commit would have no ref to stamp, and the link layer
+// would try to store the bare "<type>:<path>" key as a workitem id (which its
+// validator rejects for containing a slash).
+func NeedsAdoption(wi *WorkItem) bool {
+	return wi != nil && wi.StableID == "" && IsBuiltinDocType(wi.WorkflowType)
+}
+
+// NotAdoptedError reports that the work item at relPath must be adopted before
+// it can back a primary worktree link. The message names the exact adopt
+// command so the caller can recover in one step.
+func NotAdoptedError(relPath string) error {
+	return camperrors.NewValidation("workitem",
+		"workitem at "+relPath+" is not adopted (no .workitem marker); run "+
+			"'camp workitem adopt "+relPath+"' to give it a stable id and ref "+
+			"before linking a worktree", nil)
+}
--- a/internal/workitem/adoption.go
+++ b/internal/workitem/adoption.go
@@ -1,6 +1,11 @@
 package workitem
 
-import camperrors "github.com/Obedience-Corp/camp/internal/errors"
+import (
+	"regexp"
+
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/remote"
+)
 
 // NeedsAdoption reports whether wi is a builtin doc-type work item (design or
 // explore) that was discovered by location but has no .workitem marker, so it
@@ -12,12 +17,28 @@ func NeedsAdoption(wi *WorkItem) bool {
 	return wi != nil && wi.StableID == "" && IsBuiltinDocType(wi.WorkflowType)
 }
 
+// adoptCommandSafeUnquoted matches paths that are safe to paste into a shell
+// unquoted (the kebab-case slugs camp generates under workflow/). A hand-edited
+// path outside this set is single-quoted so the recovery command stays
+// copy-paste safe, mirroring workitem validate's repairCommandFor.
+var adoptCommandSafeUnquoted = regexp.MustCompile(`^[A-Za-z0-9._/-]+$`)
+
+// adoptCommand returns the `camp workitem adopt <path>` recovery command with
+// the path shell-quoted when it contains characters (e.g. spaces in a
+// hand-created design/explore dir) that would otherwise break a copy-paste.
+func adoptCommand(relPath string) string {
+	if adoptCommandSafeUnquoted.MatchString(relPath) {
+		return "camp workitem adopt " + relPath
+	}
+	return "camp workitem adopt " + remote.ShellQuote(relPath)
+}
+
 // NotAdoptedError reports that the work item at relPath must be adopted before
 // it can back a primary worktree link. The message names the exact adopt
 // command so the caller can recover in one step.
 func NotAdoptedError(relPath string) error {
 	return camperrors.NewValidation("workitem",
 		"workitem at "+relPath+" is not adopted (no .workitem marker); run "+
-			"'camp workitem adopt "+relPath+"' to give it a stable id and ref "+
+			adoptCommand(relPath)+" to give it a stable id and ref "+
 			"before linking a worktree", nil)
 }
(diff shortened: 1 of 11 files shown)


obey-agent (commented):
## Verdict: Request Changes

## Overview
Solid triple-fix: WI-ref selector tier, fail-fast adoption guard before worktree create (no more dangling trees), and actionable leftover-branch errors via `LocalBranchExists` + `ErrBranchExists`. One remaining operator-facing safety issue.

## Key Findings
1. **`NotAdoptedError` recovery command does not shell-quote `relPath`.** See inline.
2. Ref tier first with `EqualFold` + ambiguous multi-match is correct; tests cover unknown/ambiguous refs.
3. Branch-exists check is correctly gated on `opts.NewBranch` only (reuse path untouched).

## What's Done Well
- Resolve + adoption check *before* create on all three worktree-link entry points.
- Shared `attachWorktreeLink` for already-resolved items.
- Integration coverage for add-by-ref.

## Staff Standard
Quote the adopt recovery path, then merge.

internal/workitem/adoption.go
```diff
@@ -0,0 +1,23 @@
...
+	return wi != nil && wi.StableID == "" && IsBuiltinDocType(wi.WorkflowType)
+}
+
+// NotAdoptedError reports that the work item at relPath must be adopted before
+// it can back a primary worktree link. The message names the exact adopt
+// command so the caller can recover in one step.
+func NotAdoptedError(relPath string) error {
+	return camperrors.NewValidation("workitem",
+		"workitem at "+relPath+" is not adopted (no .workitem marker); run "+
+			"'camp workitem adopt "+relPath+"' to give it a stable id and ref "+
```
obey-agent: **Recovery command is not shell-safe.** `NotAdoptedError` embeds `relPath` unquoted into the suggested `camp workitem adopt` command. Paths can contain spaces (user-created design/explore dirs). Same class of bug camp#447 fixed in `repairCommand`.

Quote with `remote.ShellQuote(relPath)` (or only leave unquoted when a safe-slug regex matches) so the error is always copy-paste safe.

obey-agent (commented):
## Staff review — current head 02ced869

The recovery command now shell-quotes unsafe worktree paths while preserving safe slugs, with coverage for both cases. The previous command-injection/quoting finding is resolved.

Verified with `just test unit`: 3436/3436 tests passed across 107 packages.

PR #452, opened 2026-07-18
fix(fresh): make configure TUI usable for per-project follow-ups

lancekrogers:
## Problem

Three separate defects made `camp fresh configure` unusable for its main job:
adding a project-specific build command.

**1. The sidebar overflowed the terminal.** Scope rows were clamped to the
pane's *block* width, but lipgloss counts padding inside the width passed to
`Style.Width`, so a full-width label wrapped onto a second row. Every wrapped
row cost the pane a row it had not budgeted. In a campaign with ~40 projects
the left pane grew past its box and the bottom border and footer were pushed
off screen. Driven in a pty at 120x36 against a real campaign, before:

```
│   agent-simulator · inherits       │ │   ✓ 2. Checkout default branch ...
│ global                             │ │   ✓ 3. Pull  · fast-forward only
│   camp · inherits global           │ │   ✓ 4. Prune merged branches ...
...
│   guild-v3 · inherits global       │      <- runs off the bottom, no border
```

after:

```
│   agent-simulator                  │ │   ✓ 2. Checkout default branch ...
│   camp                             │ │   ✓ 3. Pull  · fast-forward only
...
╰────────────────────────────────────╯ ╰──────────────────────────────────────╯
j/k: select · K/J: move step · h/l: switch pane · a: add · e: edit ...
```

The overlay had the same defect with a larger inset, cutting error text
mid-word (`press e to make a pro`).

**2. No project detection.** The TUI always opened on the global scope, so
configuring the project you were standing in meant hunting for it in a long
list. `camp fresh` has auto-detected its target from the working directory all
along; `camp fresh configure` did not.

**3. Adding a project step could be rejected outright.** `saveForm` seeds from
the *resolved* list, which for a project with no override is the global list.
A new step whose name matched a global one was therefore written twice and
rejected by validation as `follow-up "build" appears more than once` — naming
neither the list it clashed with nor the way out. That is exactly the path
"add a project-specific build step" takes whenever a global build step exists.

Related, inheritance was invisible: any edit to an inheriting project silently
copied every global step into a project override, because a project list
replaces the global list rather than extending it.

## Change

- `finishPane` clamps content to the pane width minus its padding inset, so
  both panes inherit the fix; `freshPaneTextInset`/`freshOverlayTextInset` name
  the constants. Scope rows shed badges before a project name is truncated,
  preferring the `override N` badge over `here` when only one fits.
- `camp fresh configure` opens on the project resolved by the same
  `project.Resolve` call `camp fresh` uses, so the overrides you edit are the
  ones that run. `--project` opens elsewhere; a status line names what was
  detected; the row is marked `· here`. A detected worktree or linked repo that
  `project.List` does not enumerate still gets a row.
- The duplicate-name check runs before the write and distinguishes an inherited
  clash (`"build" is inherited from global · press e to edit it here`) from a
  same-scope one.
- The fork is stated rather than silent: the workflow header
  (`Inherits the global follow-ups · editing here makes a project copy`), both
  overlays (`Saving copies the 3 global steps into a project list for webapp.`),
  and the post-write status (`· now overrides the global list with 4 steps`).

No change to `fresh.yaml` semantics: a project list still replaces the global
list, and the non-interactive `configure add|move|remove` subcommands are
untouched.

## Verification

Driven end to end in a pty (pyte) against a throwaway campaign with a global
list of `install, lint, build` and a project `webapp` that inherits it.
Opening inside `projects/webapp` selects `webapp · here`; adding `build` is now
rejected with the actionable message; adding `webapp-build` saves. What the TUI
wrote is what fresh resolves:

```
$ camp fresh show-workflow webapp
  ✓ 8. Follow-up: install       $ npm install · stop on failure
  ✓ 9. Follow-up: lint          $ npm run lint · stop on failure
  ✓ 10. Follow-up: build        $ just build · stop on failure
  ✓ 11. Follow-up: webapp-build $ npm run build:prod · stop on failure
```

Also opened against the real campaign (~40 projects, read-only) at 120x36 and
120x22 to confirm the borders close and the footer survives.

## Tests

`configure_tui_scope_test.go` covers scope detection (including an unlisted
worktree name and the empty case), `scopeInheritsGlobal`, badge degradation,
the fork notices, and both duplicate-name paths — the rejection paths return
before any write, so they stay pure unit tests.

`TestRenderedPanesStayWithinTheirRowBudget` renders both panes at four terminal
sizes and asserts the exact row count and per-line width. Reverted against the
old clamp it fails with `110x30 scopes pane rendered 31 rows, want 26`, which
is the reported overflow.

`just gate-fast` passes: vet + lint across stable/dev/integration, 3499/3499
unit tests.


obey-agent:
## Demo

Recorded with `just vhs record docs/demos/fresh-configure.tape` against a fixture campaign of 15 linked projects, driving the branch binary.

camp fresh configure

What it shows, in order:

1. The campaign has two global follow-ups (`install`, `lint`) that every project inherits.
2. `camp fresh configure` run from inside `projects/payments-api` opens **on that project** — `payments-api · here`, workflow header `project payments-api`, status line naming what was detected. The 15-project sidebar stays inside its border with the footer intact; that is the overflow fix.
3. Adding a project-only `cargo build --release` step. The form states what saving does before it does it: `Saving copies the 2 global steps into this project's own list.`
4. `camp fresh show-workflow payments-api` resolves install, lint, **build** — while `web-storefront` still resolves install, lint only. The scoping is real, not just a display.

The tape lives at `docs/demos/fresh-configure.tape` and the recording at `docs/demos/fresh-configure.gif`.

Two notes from recording it:

- `just vhs record` runs with `.justfiles/` as its working directory, so a tape's relative `Output` path lands in `.justfiles/out/` rather than the repo root, and a relative tape path passed to the recipe does not resolve at all. I passed an absolute path and copied the result out.
- `just vhs gist` needs `termcast` on `PATH`, which is not installed on this machine, so the GIF is committed to the branch instead of hosted in a Gist.

PR #453, opened 2026-07-18
feat(machine): open a fleet TUI for a bare 'camp machine'

lancekrogers:
## Problem

Everything after adding a machine was flags: `list`, `remove`, `diagnose`, `--reset`. The only interactive surface was the one-shot tailnet picker behind `machine add --discover`, and it saved the chosen device immediately.

So changing a host meant re-running `add` with the same id and trusting `Upsert` to replace rather than duplicate, and seeing socket state meant a separate `diagnose` run in another terminal.

## Change

A bare `camp machine` in a terminal opens the fleet manager. The subcommands are untouched and stay the interface for scripts and agents; a non-terminal `camp machine` still prints help, so nothing scripted moves.

Left pane is the fleet, right pane is the selected machine: label, host, auth, ssh user, identity, and its ControlMaster socket state with the socket path.

| key | action |
| --- | --- |
| `a` | add a machine by hand |
| `s` | scan the tailnet, pick a device to prefill the form |
| `e` | edit the selected machine |
| `d` | remove the selected machine |
| `r` | re-probe every ControlMaster socket |
| `R` | clear the selected machine's socket |

Socket probing runs as a command, never inline, so opening the TUI cannot block on it. A machine with no socket file resolves without touching ssh at all, and a present socket is checked against the local multiplex process, so the whole probe stays off the network.

## Three details the data model forced

- **`local` refuses writes.** It is the current machine, not an entry in machines.yaml, so edit, remove, and socket reset all decline it using the wording the CLI already uses.
- **The id is fixed on edit** and skipped by the field walk, in both directions. It is the key `Upsert` writes under, so an editable id would leave the original entry behind rather than rename it.
- **A duplicate id on add is rejected** rather than silently replacing that machine, and the error points at the edit that does mean to replace it.

Discovery prefills the form instead of saving. The device supplies an id, label, host, and `tailscale-ssh`; the ssh user and identity are still the operator's call. A test asserts that picking a device writes nothing.

## Two rendering fixes

**Overlay background.** lipgloss applies a container background to a line's **first segment only**, so a picker row built from name, host, and state lost it at the first nested reset and rendered as a patchwork block. Confirmed in the raw bytes before changing anything:

```
\x1b[48;5;236m> \x1b[1;38;5;205mname  \x1b[0m \x1b[38;5;246mhost.example.ts.net  \x1b[0m ...
                                  ^ reset: background gone for the rest of the row
```

The box now carries no background of its own and the dimmed backdrop does the separating, which leaves its interior uniform.

**`$HOME` in socket paths.** The detail pane and the `diagnose` table printed the ControlMaster socket as an absolute path, which spells out the operator's home directory and account name in any screenshot, recording, or pasted output. Both now abbreviate `$HOME` the way the same pane already renders `~/.obey/machines.yaml` and `~/.ssh/id_ci`. `--json` keeps the absolute path, since a consumer needs the real one.

## Recording safety

Discovery shells out to `tailscale status --json`. Recording that against a real tailnet bakes the recorder's tailnet name, device inventory, and online state into a GIF in a public repo, and makes the recording depend on whichever machines happen to be awake.

`docs/demos/fixtures/tailscale` shadows the real binary on PATH and returns an invented tailnet, so the tape is safe and reproducible. The tape header states this as a requirement.

This PR also replaces a real tailnet name that was already present in three test fixtures on `main` (`cmd/camp/machine_discover_test.go`, `cmd/camp/testdata/tailscale_status.json`, `tests/integration/machine_discover_test.go`, from #384). The values are arbitrary to those assertions.

## Verification

Recorded with `just vhs record docs/demos/machine-tui.tape` against the fixture tailnet and a fixture `machines.yaml`. The demo adds a discovered machine, sets its ssh user, edits an existing machine's host, then re-runs `camp machine list` so the file corroborates what the TUI claimed:

```
ID           HOST                            AUTH
local                                                       (this machine)
buildbox     10.0.0.99                       ssh-agent
devbox       devbox.example-net.ts.net       tailscale-ssh
workstation  workstation.example-net.ts.net  tailscale-ssh
4 machines
```

All 41 frames of the recording were extracted and read, not spot-checked. Fixtures ran through `CAMP_MACHINES_PATH` throughout, so no real `~/.obey/machines.yaml` was read or written.

## Tests

`cmd/camp/machine_tui_test.go` covers row ordering, the three `local` refusals, add/edit/remove writes against an isolated `CAMP_MACHINES_PATH`, duplicate-id and reserved-id and empty-host rejection, the fixed id on edit including the backwards field walk, discovery prefill writing nothing, auth cycling, badge shedding, and `$HOME` abbreviation in both human surfaces.

`TestMachinePanesStayWithinTheirRowBudget` renders both panes at four terminal sizes and every cursor position, asserting exact row counts and per-line widths, so the sidebar-overflow class of bug fixed in #452 cannot appear here.

`just gate-fast` passes: vet and lint across stable/dev/integration, 3512/3512 tests.

## Note for review

Enter saves from any field rather than advancing to the last one, because four of the six fields are optional. This differs from `fresh configure`, where enter advances. Happy to align them either way.


obey-agent:
## Demo

Recorded with `just vhs record docs/demos/machine-tui.tape`. The tailnet is a fixture (`docs/demos/fixtures/tailscale` shadows the real binary on PATH), so the recording captures no real network.

camp machine

In order:

1. `camp machine list` as it is today, then bare `camp machine` opening the fleet manager.
2. Walking the fleet: `local` explains itself, then each machine shows its ssh detail and ControlMaster socket state.
3. `s` scans the tailnet. Picking a device prefills the form rather than saving it, so the ssh user is set by hand before the write.
4. `e` edits an existing machine. The id is fixed and shown as such; the host is replaced and the other fields survive.
5. Quit, then `camp machine list` again: 4 machines, `workstation` added, `buildbox` now on `10.0.0.99`. The file matches what the TUI claimed.

PR #454, opened 2026-07-18
gate: mark camp audit and camp event as dev-only commands

lancekrogers:
## Summary
- `camp audit` and `camp event` are now `//go:build dev`, following the same pattern already used for `quest` and `flow` — every file in `cmd/camp/audit/` and `cmd/camp/event/` carries the tag, and registration moves out of `root.go` into new `audit_register.go` / `event_register.go` files (`init()` + `release.MarkDevOnly`), gated the same way `quest_register.go` already is.
- Added `"audit"` and `"event"` to `devOnlyCommands` in `release_profile_test.go`, the existing single-source-of-truth list the dev/stable release-profile tests assert against.
- Regenerated `docs/cli-reference` from a stable build (separate commit) — the `camp audit*` / `camp event*` pages and their index-listing lines drop out since those commands no longer exist in stable/rc builds.
- Noted the new dev-only requirement in `docs/audit-trail-rollout.md` so the rollout walkthrough doesn't silently assume a stable build can run it.

## Why
Both command groups are audit-trail/ledger internals intended for campaign-tooling development, not end-user surface — same rationale as `quest`.

## Test plan
- [x] `just gate-fast` (stable + dev build, vet, lint, dev unit tests — 3500/3500 passed)
- [x] `go build -tags '' ./cmd/camp` then `camp audit`/`camp event` → `unknown command` (absent from stable)
- [x] `go build -tags dev ./cmd/camp` then `camp audit`/`camp event` → both work
- [x] `camp --help` on the stable binary no longer lists `audit`/`event`
- [x] `just docs` regen diff reviewed — only expected additions/removals, no unrelated drift


obey-agent (commented):
## Verdict: Approve

## Overview
Clean dev-gating of `camp audit` and `camp event` using the same `//go:build dev` + `*_register.go` + `release.MarkDevOnly` + `devOnlyCommands` pattern as `quest`/`flow`. Stable builds drop the commands and the regenerated CLI reference matches.

## Key Findings
- None.
- Every file under `cmd/camp/audit/` and `cmd/camp/event/` is tagged; registration only via dev `init()`.
- `devOnlyCommands` updated so release-profile tests stay the single source of truth.
- Rollout doc notes the dev-build requirement so operators are not surprised by stable binaries.

## What's Done Well
- No parallel registration path left in `root.go`.
- Docs regen isolated in a second commit.
- Gate-fast + explicit stable/dev binary checks called out in the PR.

## Staff Standard
Yes. Merge-ready.

obey-agent (commented):
## Verdict: Approve

### Overview
Clean, minimal gating change that follows the exact `quest`/`flow` pattern already in tree. `//go:build dev` on every package file, registration via `*_register.go` + `release.MarkDevOnly`, single-source `devOnlyCommands` update, stable CLI docs regenerated, and the rollout walkthrough called out so operators do not chase missing commands on a stable build.

### Key Findings
No blocking issues. The change is small, consistent, and test-gated (`just gate-fast` claimed; pattern matches `quest_register.go` field-for-field).

### What's Done Well
- **Integration over parallel implementation**: same register/`MarkDevOnly` path as existing dev-only commands rather than inventing a new gate.
- **Docs as part of the gate**: regenerating `docs/cli-reference` and noting the requirement in `docs/audit-trail-rollout.md` prevents a class of "why is this missing?" confusion.
- **Test list as SSOT**: extending `devOnlyCommands` keeps the release-profile tests honest for both profiles.

### Staff Standard
Yes — I would be proud to submit this as my own work. Merge when ready.

PR #455, opened 2026-07-18
feat(fresh): configure every fresh.yaml key in the configure TUI, grouped by what you can change

lancekrogers:
## Problem

`camp fresh configure` could only edit `follow_up` — one of the five keys in `fresh.yaml`. The other four (`branch`, `push_upstream`, `prune`, `prune_remote`) had no interactive surface at all.

Worse, the workflow pane rendered every step as one flat numbered list, so a row reading **"Create working branch [off] · not configured"** sat next to a footer advertising **`e: edit`** — and pressing it returned `✗ select a follow-up step to edit`. The pane showed a setting asking to be configured, offered a key to configure it, and refused.

Three genuinely different classes of row looked identical: fixed steps, settings-backed steps, and follow-ups. The only cue that one differed was the literal word "Follow-up:" in its title.

## Changes

**Group the workflow pane by what each step is**, with the verbs that apply on each section header rather than a footer that lists all of them and rejects most:

```
 Sync  always runs · not configurable
   ✓ 1. Safety checks        · no merge/rebase in progress; worktree is clean
   ✓ 2. Checkout default branch
   ✓ 3. Pull                 · fast-forward only
 Settings  enter: change
   ✓ 4. Prune merged branches
   ✓ 5. Prune remote tracking refs
   ○ 6. Create working branch  · no branch configured
   ◌ 7. Push branch upstream   · needs a working branch
 Follow-ups  a: add · e: edit · d: delete · K/J: reorder
   ✓ 8. Follow-up: install   · $ npm install · stop on failure
   ✓ 9. Ready to work
```

**A settings editor on `enter`**, tri-state in a project scope (`inherit` / `on` / `off`) so clearing a key restores inheritance rather than writing `false`. `FreshProjectConfig` uses `*bool` precisely for that distinction; collapsing it to a checkbox would write override keys the user never asked for.

**`prune` / `prune_remote` stay campaign-wide**, matching what `FreshProjectConfig` actually stores. A project scope redirects to Global defaults rather than writing a key `fresh` never reads. Adding per-project prune would change `camp fresh` resolution semantics for existing users, which shouldn't ride along on a TUI fix — happy to file it separately if wanted.

**New config writers** (`SetFreshBranch`, `SetFreshPushUpstream`, `SetFreshPrune`, `SetFreshPruneRemote`) routed through the existing `withFreshConfigLock`, so they inherit the atomic write, comment preservation, and empty-scope cleanup the follow-up writers already had.

**Separate "off" from "not configured"** in glyph and color (`✓` on, `○` unset, `◌` blocked, `·` off), and route refusals through a notice level instead of the red used for real failures.

## Drive-by fix

The scopes-pane override badge counted follow-ups only. A project overriding just `branch` and `push_upstream` showed **no badge at all** — already wrong before this change, since per-project `branch` ships today.

## Verification

Driven in a pty via VHS against a fixture campaign with 10 linked projects, not just unit tests. That caught a bug tests would not have: the branch input was seeded with the *inherited* branch, so selecting "create a branch" and typing appended to it and wrote `developfeat/storefront`. Fixed, and pinned by `TestSettingEditorDoesNotSeedInheritedBranch`.

Confirmed on screen and on disk:
- branch and push_upstream written correctly under `projects.<name>`, file comments preserved
- `web-storefront · here · override 2` — badge counts both keys
- prune in a project scope shows `· prune is a campaign-wide setting · select Global defaults to change it` as a muted notice, not an error
- narrow single-pane mode (700x460) still renders both section headers within the row budget

`just gate`: 3535/3535 tests pass, lint clean.

## Note

The demo for this now lives in the `social-fitness-app` experience-lab campaign rather than in this repo, per that campaign's charter ("capture VHS recordings ... without adding demo-only fixtures to production codebases"):

- fixture builder: `vhs/fixtures/build-fresh-campaign.sh` (10-project fitness campaign, built outside the repo so no username appears in the prompt)
- tape + scenario notes: `vhs/camp/configure-fresh-workflow.{tape,md}`

That leaves `docs/demos/fresh-configure.{tape,gif}` here recording a TUI that no longer exists, under the older in-repo convention. I've deliberately **not** touched it in this PR — removing it implies `machine-tui.tape` follows and `just vhs record` loses its purpose here, which is a call about this repo's demo assets generally, not something to decide inside a TUI change.

### DIFF
--- a/internal/commands/fresh/configure_tui.go
+++ b/internal/commands/fresh/configure_tui.go
@@ -32,6 +32,38 @@ const (
 	followUpFormOverlay
 	followUpDeleteOverlay
 	followUpHelpOverlay
+	followUpSettingOverlay
+)
+
+// freshSettingAction is what saving the settings editor does to the selected
+// scope's key. Inherit clears it, which is a distinct outcome from writing
+// false: FreshProjectConfig stores pointers precisely so a project can leave a
+// key unset and follow the global value as it changes.
+type freshSettingAction int
+
+const (
+	freshSettingInherit freshSettingAction = iota
+	freshSettingOn
+	freshSettingOff
+	freshSettingNoBranch
+	freshSettingCustomBranch
+)
+
+type freshSettingOption struct {
+	label  string
+	action freshSettingAction
+}
+
+// statusLevel separates a completed change from a note about why a key did
+// nothing. A refusal rendered in the same red as a genuine failure trains
+// users to ignore the status line, and most refusals here are only the TUI
+// saying a row has nothing to configure.
+type statusLevel int
+
+const (
+	statusOK statusLevel = iota
+	statusNotice
+	statusError
 )
 
 type followUpScope struct {
@@ -69,11 +101,19 @@ type followUpTUIModel struct {
 	formError     string
 	pendingDelete string
 
-	status    string
-	statusErr bool
-	width     int
-	height    int
-	quitting  bool
+	// settingStep is the workflow row the settings editor is open on, with the
+	// options and cursor that editor is showing.
+	settingStep    freshWorkflowStep
+	settingOptions []freshSettingOption
+	settingChoice  int
+	settingInput   textinput.Model
+	settingError   string
+
+	status      string
+	statusLevel statusLevel
+	width       int
+	height      int
+	quitting    bool
 }
 
 // runConfigureTUI is the human-facing entry point for `camp fresh configure`.
@@ -150,6 +190,10 @@ func newFollowUpTUIModel(ctx context.Context, root string, projects []project.Pr
 	m.inputs[0].Placeholder = "install"
 	m.inputs[1].Placeholder = "npm install"
 	m.inputs[2].Placeholder = "optional, relative to project root"
+	m.settingInput = textinput.New()
+	m.settingInput.Prompt = "  "
+	m.settingInput.CharLimit = 256
+	m.settingInput.Placeholder = "feat/my-work"
 	m.rebuildScopes(globalFollowUpScope)
 	return m
 }
@@ -180,9 +224,15 @@ func (m *followUpTUIModel) rebuildScopes(selected string) {
 			name:        name,
 			current:     name == m.cwdProject,
 		}
-		if pc, ok := m.cfg.Projects[name]; ok && pc.FollowUp != nil {
-			scope.override = true
-			scope.overrideCount = len(pc.FollowUp)
+		// Count every key the project sets, not just its follow-ups. A project
+		// that overrides only its branch is still a project that deviates from
+		// the global defaults, and the badge is the one place the scopes pane
+		// says so.
+		if pc, ok := m.cfg.Projects[name]; ok {
+			if keys := config.ProjectOverrideKeys(pc); keys > 0 {
+				scope.override = true
+				scope.overrideCount = keys
+			}
 		}
 		scopes = append(scopes, scope)
 	}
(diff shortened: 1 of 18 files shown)


obey-agent (commented):
## Verdict: Approve

## Overview
Fixes a real UX bug (settings rows looked editable, `e` rejected them) and finishes the configure TUI for all `fresh.yaml` keys. Grouped workflow pane, enter-to-edit settings with tri-state inherit/on/off, campaign-only prune keys redirected from project scope, and config writers that clear keys instead of writing misleading overrides.

## Key Findings
- None blocking.
- `SetFreshBranch` / `SetFreshPushUpstream` / `SetFreshPrune` / `SetFreshPruneRemote` use `withFreshConfigLock` and preserve YAML shape when possible.
- Project-scope empty branch vs inherit (`*""` vs nil) is deliberate and tested.
- Prune writers are global-only; TUI blocks project-scope edits with a clear notice rather than writing dead keys.
- Unsupported verbs on fixed/settings rows use notices instead of red errors — right UX.

## What's Done Well
- Section headers carry the verbs that apply (sync / settings / follow-ups).
- Config unit tests for inherit clear, project prune isolation, empty-project prune.
- CLI reference regen isolated in a second commit.

## Staff Standard
Yes. Merge-ready (TUI changes should still get a quick PTY smoke if not already run; logic and tests look sound).

internal/commands/fresh/configure_tui.go
```diff
@@ -263,12 +313,171 @@ func (m *followUpTUIModel) selectedFollowUp() (int, config.FollowUpConfig, bool)
...
+		}
+		return freshSettingCustomBranch
+	case freshSettingPushUpstream:
+		if scope != "" {
+			if !hasProject || pc.PushUpstream == nil {
+				return freshSettingInherit
+			}
+			return boolAction(*pc.PushUpstream)
+		}
+		return boolAction(m.cfg.ResolveFreshPushUpstream(""))
```
obey-agent: **Global bools collapse “unset (default)” into “on”, and save always writes.**

`currentSettingAction` for global `prune` / `prune_remote` / `push_upstream` uses `Resolve*`, which returns the built-in default (`true`) when the key is absent. `settingOptionsFor` only offers on/off in the global scope (no clear/inherit). `saveSettingEditor` always calls the setter.

Concrete failure mode on a scaffolded `fresh.yaml` (all keys commented):

1. Open **Prune merged branches** under Global defaults (cursor lands on “on”).
2. Press enter without changing anything.
3. Disk now has `prune: true` — an explicit override the user never asked for.

Once written, there is also no TUI path back to “absent key / use default”: only on→write true, off→write false. Today that is behaviorally equivalent because the default is `true`, but it:

- dirties the carefully commented scaffold (the PR itself cares about comment preservation),
- locks the value if the built-in default ever changes,
- breaks the “nil means default” contract the writers carefully implement (`SetFreshPrune(nil)` exists but is unreachable from the global editor).

**Suggestion:** either (a) add a third global option “default (on)” that writes `nil`/clears, and select it when `cfg.Prune == nil`, or (b) no-op the save when the selected action equals the currently stored pointer state (not the resolved bool). Project scope already gets this right via inherit.

internal/commands/fresh/configure_tui_update.go
```diff
@@ -201,11 +274,159 @@ func (m *followUpTUIModel) updateOverlay(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
...
+	return m.settingOptions[m.settingChoice].action
+}
+
+func (m *followUpTUIModel) closeSettingEditor() {
+	m.overlay = followUpNoOverlay
+	m.settingError = ""
+	m.settingInput.Blur()
+}
+
+func (m *followUpTUIModel) saveSettingEditor() tea.Cmd {
```
obey-agent, line 345: **Save path has no dirty check.**

Even if the global-bool UX above is fixed, this path still rewrites the file on a pure confirm. Prefer comparing the action to the stored pointer (not the resolved value) and short-circuiting with a notice like `prune already on · nothing written` when unchanged. That keeps accidental enter from creating a git dirty state in `.campaign/settings/fresh.yaml`.

internal/commands/fresh/configure_tui.go
```diff
@@ -180,9 +224,15 @@ func (m *followUpTUIModel) rebuildScopes(selected string) {
...
 		}
-		if pc, ok := m.cfg.Projects[name]; ok && pc.FollowUp != nil {
-			scope.override = true
-			scope.overrideCount = len(pc.FollowUp)
+		// Count every key the project sets, not just its follow-ups. A project
+		// that overrides only its branch is still a project that deviates from
+		// the global defaults, and the badge is the one place the scopes pane
+		// says so.
+		if pc, ok := m.cfg.Projects[name]; ok {
+			if keys := config.ProjectOverrideKeys(pc); keys > 0 {
```
obey-agent, line 234: **Comment drift on `overrideCount`.**

`rebuildScopes` now correctly uses `ProjectOverrideKeys` (keys, not follow-up length) — good fix for the badge. The struct field above still documents:

```go
// overrideCount is the length of this project's own follow-up list.
// override distinguishes an explicit empty list from no list at all.
```

That is no longer true (`overrideCount` is a key count; empty follow-ups still count as one key via `FollowUp != nil`). Please update the field comment so the next reader does not “fix” a regression by counting `len(FollowUp)` again.

Also: `configuredProjectNames` still only returns projects with follow-up overrides and is still unit-tested as such — dead for the TUI path now. Either retire it or make it call `ProjectOverrideKeys` so a future reuse does not reintroduce the badge bug.

internal/commands/fresh/configure_tui_settings_test.go
```diff
@@ -0,0 +1,308 @@
...
+	if kinds[freshStepDone] != 1 {
+		t.Errorf("done steps = %d, want 1", kinds[freshStepDone])
+	}
+}
+
+// An unconfigured branch and a branch turned off are different answers, and
+// the pane renders them differently, so the state has to distinguish them.
+func TestBranchStateSeparatesUnsetFromOff(t *testing.T) {
+	unset := buildFreshWorkflow(&config.FreshConfig{}, "")
+	idx := 5
```
obey-agent: **Fragile indices in brand-new tests.**

This file already introduces `stepIndexFor` precisely so tests do not encode sequence positions. `TestBranchStateSeparatesUnsetFromOff` / `TestPushUpstreamBlockedWithoutBranch` / `TestConfigurableExcludesGlobalOnlyKeysInProjectScope` still hardcode `5`, `6`, `3`.

If a fixed step is inserted above Settings (or order shifts), these fail for the wrong reason and silently assert the wrong row until then. Prefer `stepIndexFor` (or a small `stepBySetting` helper that also works on a raw `[]freshWorkflowStep`).

internal/commands/fresh/configure_tui_update.go
```diff
@@ -201,11 +274,159 @@ func (m *followUpTUIModel) updateOverlay(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
...
 
+func (m *followUpTUIModel) updateSettingEditor(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
+	switch msg.String() {
+	case "ctrl+c":
+		m.quitting = true
+		return m, tea.Quit
+	case "esc":
+		m.closeSettingEditor()
+		return m, nil
+	case "up", "shift+tab":
```
obey-agent: **Settings editor drops `j`/`k`, and focusing the branch field mid-nav skips `Blink`.**

Browse mode and the footer are j/k-first. In `updateSettingEditor` only `up`/`down`/`tab` move the choice; bare `j`/`k` fall through and are discarded (unless the branch input is focused, in which case they insert into the name — also surprising after a habit of j/k navigation).

Same function: `moveSettingChoice` calls `settingInput.Focus()` when landing on “create a branch...” but the caller returns `nil` rather than `textinput.Blink`, so the caret may not appear until the next key. `openSettingEditor` / `activateSelectedStep` get this right via `return m, textinput.Blink`.

Low severity, but this is the surface the PR is selling — polish it to match the rest of the TUI.

internal/config/fresh_settings.go
```diff
@@ -0,0 +1,154 @@
...
+// the global setting and the global falls back to the built-in default (true).
+func SetFreshPushUpstream(ctx context.Context, campaignRoot, projectName string, value *bool) error {
+	return withFreshConfigLock(ctx, campaignRoot, func(mapping *yaml.Node) error {
+		return setFreshBool(mapping, projectName, "push_upstream", value)
+	})
+}
+
+// SetFreshPrune writes whether merged branches are pruned. The key is global
+// only: FreshProjectConfig has no per-project counterpart, so a project scope
+// is rejected rather than silently written somewhere fresh will not read it.
```
obey-agent: **Doc comment overclaims the API.**

“a project scope is rejected rather than silently written somewhere fresh will not read it” — the function does not accept a project name at all, so nothing is rejected. Callers that mistakenly think they are writing a project key cannot be told no; they just mutate global.

That is a reasonable design for a global-only key, but the comment should match the contract: e.g. “global only; there is no projectName parameter because FreshProjectConfig has no prune field.” The TUI’s redirect notice is the real safety net; the config API itself is trust-the-caller.

Optional hardening: if you ever add a projectName param for symmetry with `SetFreshPushUpstream`, do return a validation error on non-empty project rather than forcing `""`.

obey-agent (commented):
## Staff adversarial review — PR #455

**Verdict: APPROVE with nits (non-blocking).** Ship after addressing the global-bool materialization issue if you want the scaffold to stay clean; nothing here is a correctness break against current resolve defaults.

### What this PR gets right

- **Problem framing is real.** Flat list + footer advertising `e` while refusing non-follow-up rows is a genuine UX footgun; grouping by configurability (Sync / Settings / Follow-ups) with verbs on section headers is the correct structural fix.
- **Tri-state project settings match the data model.** `*bool` / `*string` inherit vs off vs on is preserved end-to-end (`SetFreshBranch(nil)` vs `&""`, `SetFreshPushUpstream(nil)`). The inherited-branch seed bug (`developfeat/storefront`) is well fixed and pinned by `TestSettingEditorDoesNotSeedInheritedBranch`.
- **Write path reuses the hard parts.** New setters go through `withFreshConfigLock`, share `freshScopeMapping` / `pruneEmptyProjectScope` with follow-up writers, replace scalars in place (no duplicate keys), preserve comments — tested.
- **Global-only prune is handled honestly in the TUI.** Project scope notices and redirects instead of writing a key `ResolveFreshPrune` never reads. Correct product call to not invent per-project prune here.
- **Status levels** (ok / notice / error) stop training users to ignore the status line. Glyph split (`✓` / `○` / `◌` / `·`) matches the state model.
- **Override badge** now counts all project keys via `ProjectOverrideKeys` — the drive-by was a real bug.
- **Tests are purposeful** (classifiers, blocked push, global-only redirect, empty follow-up section, round-trips, comment preservation, empty-scope prune). VHS/pty claim in the body is the right bar for a TUI change.

### Inline comments (6)

| # | Severity | Topic |
|---|----------|--------|
| 1 | **Medium** | Global bool editor collapses unset→on via `Resolve*`, save always writes → casual enter materializes `prune: true` on scaffolded files; no TUI path back to absent key |
| 2 | **Medium/Low** | `saveSettingEditor` has no dirty/no-op short-circuit |
| 3 | **Low** | Stale `overrideCount` field comment + leftover `configuredProjectNames` only-follow-ups helper |
| 4 | **Low** | New tests hardcode step indices despite `stepIndexFor` |
| 5 | **Low** | Settings overlay ignores `j`/`k`; `Focus()` without `textinput.Blink` when navigating onto custom branch |
| 6 | **Nit** | `SetFreshPrune` doc says “rejected”; API simply has no project param |

### Additional notes (no separate inline)

- **Defense in depth:** `saveSettingEditor` trusts `settingStep` for global-only keys; only `activateSelectedStep` blocks project-scope prune. Safe today; a `GlobalOnly && inProjectScope()` guard in save would make the invariant local.
- **“Ready to work”** sits under the Follow-ups header (by design in the PR mock). Slightly odd grouping, but non-configurable and cursor behavior is fine.
- **Agent/script surface** still cannot set branch/push/prune without editing YAML (documented). Acceptable for this PR; file as follow-up if agents need parity.
- Files `configure_tui{,_update,_view}.go` are all ~530–590 lines (house guideline 500). Not a merge blocker.

### Completeness checklist

| fresh.yaml key | TUI | Writer | Global-only handled |
|----------------|-----|--------|---------------------|
| branch | yes (inherit / none / custom) | `SetFreshBranch` | n/a (per-project ok) |
| push_upstream | yes | `SetFreshPushUpstream` | n/a |
| prune | yes (global) | `SetFreshPrune` | notice in project |
| prune_remote | yes (global) | `SetFreshPruneRemote` | notice in project |
| follow_up | pre-existing | pre-existing | n/a |

### Failures

None blocking. No security issues, no wrong-scope writes on the guarded paths, no missing keys relative to `FreshConfig` / `FreshProjectConfig`.

### Suggested merge bar

Optional before merge: dirty-check or global “default” option for bools so scaffolded campaigns stay comment-only until the user actually changes something. Rest is polish.

— obey-agent review (token-injected; active account remains personal)

obey-agent (commented):
## Re-review after 531b2c8 — Verdict: APPROVE

Head: `531b2c832914855fa73d752486780d67652f6e92` (`fix(fresh): polish configure TUI settings nits from review`). PR is already merged; this is a post-merge verification that the prior nits landed cleanly.

### Prior nits — all closed

| # | Nit | Status | Evidence in 531b2c8 |
|---|-----|--------|---------------------|
| 1 | Global bools collapsed unset → on / pure enter materializes explicit true | **Fixed** | `settingOptionsFor` offers `"default · on"` (via `freshSettingInherit` + `builtInBoolDefault`); `currentSettingAction` inspects stored pointers (`PushUpstream`/`Prune`/`PruneRemote == nil` → inherit), not `Resolve*` |
| 2 | Save path had no dirty check | **Fixed** | `settingUnchanged` short-circuits before writers; notice `"… nothing written"`; covered by `TestSettingUnchangedShortCircuitsSave` and `TestGlobalBoolOptionsOfferDefaultAndOpenOnAbsentKey` |
| 3 | Settings editor dropped `j`/`k`; mid-nav focus skipped `Blink` | **Fixed** | `updateSettingEditor` binds `j`/`k` with arrows; `moveSettingChoice` returns `textinput.Blink` on custom-branch focus (open path already blinked via `activateSelectedStep`) |
| 4 | `overrideCount` comment drift + `configuredProjectNames` follow-ups-only | **Fixed** | Struct docs match `ProjectOverrideKeys`; `configuredProjectNames` uses the same rule; test renamed/expanded to include branch-only `alpha` |
| 5 | Fragile hard-coded step indices in new tests | **Fixed** | `stepIndexBySetting` / `stepIndexFor` used for branch/push/prune lookups |
| 6 | `SetFreshPrune` docs overclaimed project rejection | **Fixed** | Docs state global-only (no `projectName` param), nil clears to built-in default; TUI owns project-scope redirect |

### New issues on this head

None. No remaining findings that warrant inline comments or a request for changes.

### Residual non-blocking observations (intentionally left)

- `builtInBoolDefault` duplicates the `true` defaults from `ResolveFresh{PushUpstream,Prune,PruneRemote}` — documented tradeoff so labels do not re-derive from a live override. Fine unless defaults ever diverge.
- `j`/`k` intentionally steal branch-name typing (paste/arrows still work) — called out in-code; correct product choice for this TUI.

### Summary

531b2c8 addresses every nit from the previous approve-with-nits pass with matching tests. Ship state is clean; no further action needed on this PR.

PR #457, opened 2026-07-19
feat(machine): make the fleet TUI teach rather than assume

lancekrogers:
## Problem

Opened with nothing configured, `camp machine` looked like this:

```
Machines  the fleet camp can reach for switch and list --remote
╭─────────────────────────────╮ ╭──────────────────────────────────────────────╮
│ Fleet                       │ │ Detail · local                               │
│ 1 machine                   │ │ the current machine                           │
│ > local · this machine      │ │                                              │
│                             │ │ Always reachable, and never written to        │
│                             │ │ machines.yaml.                                │
│                             │ │                                              │
│         (20 empty rows)     │ │         (20 empty rows)                       │
╰─────────────────────────────╯ ╰──────────────────────────────────────────────╯
j/k select · a add · s scan · e edit · d remove · r refresh · ? help · q quit
```

Two empty panes framing one row the user cannot act on. It names the file it edits and lists eight keys at equal weight. It never says what a machine is, why anyone would add one, or what they get afterwards. It is a config table, and it only reads as a product to someone who already knows the feature.

The populated screen had the same problem in a subtler form: it showed `Socket none · a hop opens a fresh one`, an ssh multiplexing detail, while never answering the question anyone actually opens this screen with, which is whether camp can reach the machine at all.

## The design

**Empty fleet, tailscale installed.** The step is the user's own machines, not a form asking for a hostname they would have to go look up.

```
  Machines

  Work on campaigns that live on your other computers.
  Add one here and you can reach it from this terminal:

    camp switch devbox:my-campaign    open a campaign over there
    camp list --remote                campaigns on every machine at once

  ──────────────────────────────────────────────────────────────────────

  4 machines on your Tailscale network
  Pick one to add. Nothing is saved until you confirm the details.

  > buildfarm        buildfarm.example-net.ts.net       online
    gpu-box          gpu-box.example-net.ts.net         online
    media-server     media-server.example-net.ts.net    offline
    workstation      workstation.example-net.ts.net     online

    a  Add a machine by hand
       For an ssh host that is not on your tailnet.

  ──────────────────────────────────────────────────────────────────────

  This computer is always available as "local" and needs no setup.

  enter add the highlighted device · a add by hand · s rescan · ? what is this · q quit
```

**Empty fleet, no tailscale.** Same teaching, honest about what is missing.

```
  No machines yet

    a  Add a machine
       Its hostname, your ssh user, and a key your agent already holds.

  Tailscale is not installed. With it, camp can list the machines on
  your network and fill this in for you.
```

**Populated.** The row leads with reachability; the pane leads with the answer, then the payoff, then the config.

```
╭──────────────────────────────╮ ╭────────────────────────────────────────────╮
│ Machines                     │ │ buildbox                                   │
│ 2 configured                 │ │ CI builder                                 │
│   local      this computer   │ │                                            │
│ > buildbox   ○ not tested    │ │ ○ Not tested yet                           │
│   devbox     ● reachable     │ │   t checks whether camp can reach it.      │
│                              │ │                                            │
│                              │ │ Work on it                                 │
│                              │ │   camp switch buildbox:<campaign>          │
│                              │ │   camp list --remote                       │
│                              │ │                                            │
│                              │ │ Connection                                 │
│                              │ │   Host      10.0.0.12                      │
│                              │ │   Sign-in   ssh agent key                  │
│                              │ │   SSH user  ci                             │
│                              │ │                                            │
│                              │ │ Connection reuse  idle                     │
│                              │ │   The first hop opens a session camp       │
│                              │ │   keeps warm.                              │
╰──────────────────────────────╯ ╰────────────────────────────────────────────╯
t test connection  ·  e edit · d remove  ·  a add · s scan tailnet  ·  ? help · q quit
```

**A machine that does not work** says why, and what to press:

```
│ ✗ Could not reach it                                       │
│   connect to host 10.0.0.12 port 22: Operation timed out    │
│   e edits it · t tries again                                │
```

## What changed

- **Onboarding screen** for an empty fleet, replacing the two-empty-panes layout.
- **Tailscale-aware.** `camp machine` detects the CLI and scans on open when the fleet is empty, so the first screen lists real devices. Without tailscale it says so and offers the manual path. Both paths still produce ordinary ssh machines; tailscale is a convenience, not a requirement.
- **Connection test** (`t`, or `enter`) runs the remote's own `camp --version` over ssh. One round trip establishes whether ssh authenticates, whether the host resolves, and whether camp exists over there. That last failure is invisible to socket state, which is what the screen used to show instead.
- **Saving a machine tests it immediately**, because that is when the user wants to know.
- **Plain language.** `tailscale-ssh` reads as Tailscale SSH, the ControlMaster socket is Connection reuse described by what it does, and it only names its socket file when it is stuck.
- **Help teaches concepts** before listing keys, and covers the three sign-in methods.
- **Footer is ranked**, leading with the action worth pressing on a screen of untested machines.
- **Already-added tailnet devices are marked**, so one machine does not end up in the fleet twice under two ids.
- **A test in flight animates**, since an ssh connect can take the full eight-second `ConnectTimeout` and a still screen reads as hung.
- **The id column sizes to the widest id**, so a long name cannot run into its status badge.

## Verification

Recorded with `just vhs record docs/demos/machine-tui.tape` against a stubbed tailnet and **no** `machines.yaml`, so the recording opens where a first-time user does. All 30 frames were extracted and read.

The demo ends with the added machine reported unreachable. That is correct and deliberate: the fixture tailnet has no real hosts behind it, so the test truthfully reports the DNS failure. It doubles as the failure-path demo.

Fixtures ran through `CAMP_MACHINES_PATH` throughout; no real `~/.obey/machines.yaml` was read or written, and the recording contains no real network.

## Tests

New coverage for the onboarding screen with and without tailscale, the tailnet device list, already-added marking, the local refusal for connection tests, health badge and status wording, remote version parsing, and ssh failure-message reduction. The existing suite (row ordering, the three `local` refusals, add/edit/remove writes, duplicate and reserved id rejection, fixed id on edit, discovery prefill writing nothing, pane row budgets, `$HOME` abbreviation) still passes unchanged.

`just gate-fast` passes: vet and lint across stable/dev/integration, 3541/3541 tests.


obey-agent:
## Demo

Recorded with `just vhs record docs/demos/machine-tui.tape` against a stubbed tailnet and no `machines.yaml`, so it opens where a first-time user does.

camp machine

1. First run with nothing configured: the screen says what machines are for, shows the two commands they unlock, and lists the devices on the tailnet ready to add.
2. Picking one prefills the form. The ssh user is still set by hand, and nothing is written until confirmed.
3. Saving tests the connection straight away rather than waiting to be asked.
4. Help explains the concepts before the keys.
5. Quitting and running `camp machine list` shows the file matches.

The added machine reports unreachable at the end. That is the fixture tailnet having no real hosts behind it, so the test truthfully reports the DNS failure, which doubles as the failure-path demo.

cmd/camp/machine_tui.go
```diff
@@ -220,14 +336,74 @@ func (m *machineTUIModel) probeSockets() tea.Cmd {
...
+		return rest
 	}
+	return line
+}
+
+// connectionFailureDetail reduces a failed hop to the part a person can act
+// on. The wrapper says which command ran and what it exited with, which the
+// screen already implies; what matters is ssh's own line, or camp's hint when
+// the binary is missing on the far side.
+func connectionFailureDetail(err error) string {
```
obey-agent, line 427: **Bug (High):** This function's own comment says the screen needs "camp's hint when the binary is missing on the far side," but the implementation discards it.

`remote.RunCampCommand` wraps exit-127 with `campNotFoundHint`, which carefully tells the user about login-shell PATH and `CAMP_REMOTE_CAMP_PATH`. That wrapped error string looks like:

```
remote camp not found on buildbox (...): command "ssh ..." exited with code 127: bash: camp: command not found
```

Cutting on `exited with code ` and then `: ` reduces that to just `bash: camp: command not found` — the one failure mode this health check was introduced to surface beyond socket state.

Prefer `errors.As` on `*camperrors.CommandError` (exit 127 / wrap) rather than string surgery, or short-circuit when the outer message already carries the camp-not-found guidance. Add a unit test with a `campNotFoundHint`-shaped error so this cannot regress.

cmd/camp/machine_tui_onboarding.go
```diff
@@ -0,0 +1,166 @@
...
+	body := strings.Join(ui.ClampLines(lines, max(width-4, 10)), "\n")
+	framed := []string{"", lipgloss.NewStyle().PaddingLeft(2).Render(body)}
+	if status := m.statusLine(); status != "" {
+		framed = append(framed, "", "  "+status)
+	}
+	framed = append(framed, "", "  "+machineHelpStyle.Render(m.onboardingFooter()))
+	return ui.FitFullscreenView(strings.Join(ui.CapFrame(framed, lay.width, lay.height), "\n"), lay.height)
+}
+
+func (m *machineTUIModel) onboardingFooter() string {
```
obey-agent, line 52: **UX inconsistency (Medium):** Footer and body disagree about `s`.

- Scan failed body (below): "s tries the scan again."
- Empty-device body (tailscale ready): no rescan affordance in the footer either, though `updateOnboarding` still accepts `s`.
- Footer when `len(devices) == 0`: only `a · ? · q` — no `s`.

If tailscale is ready, the footer should always surface rescan (and while scanning, either disable double-fire or show a scanning footer). Right now the teaching copy points at a key the chrome does not list.

cmd/camp/machine_tui_update.go
```diff
@@ -22,33 +23,115 @@ func (m *machineTUIModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+		if len(m.devices) > 0 {
+			m.deviceCursor = (m.deviceCursor + 1) % len(m.devices)
+		}
+	case "enter":
+		if len(m.devices) > 0 {
+			return m, m.prefillFromDevice()
+		}
+	case "a":
+		return m, m.openAddForm()
+	case "s":
```
obey-agent, line 118: **Race (Medium):** `s` while already scanning launches another `scanTailnet` with no generation token.

`applyDevices` always writes whatever `devicesMsg` arrives last-to-complete, not last-to-start. Two overlapping `tailscale status --json` calls can land out of order and briefly (or lastingly) show a stale device list / error.

Guard: if `m.scanning` already, ignore another `s` (or cancel/replace with a monotonic `scanGen` checked in `applyDevices`). Same pattern is worth applying to the discover-overlay path if it can re-enter while scanning.

cmd/camp/machine_tui.go
```diff
@@ -200,8 +298,26 @@ func (m *machineTUIModel) setError(err error) {
...
+	return hosts
+}
+
 func (m *machineTUIModel) Init() tea.Cmd {
-	return tea.Batch(textinput.Blink, m.probeSockets())
+	cmds := []tea.Cmd{textinput.Blink, m.probeSockets()}
+	// An empty fleet opens on the onboarding screen, whose whole value is
+	// showing the machines the user could add. Scanning up front is what
+	// turns that screen from an instruction into a list they can act on.
+	if m.empty() && m.tailscaleReady {
```
obey-agent, line 330: **Bug (High):** Auto-scan only happens here in `Init`. Any later transition into `empty()` does not.

Concrete path: open with one configured machine (no onboarding scan), remove it → `View` routes to onboarding with `tailscaleReady && !scanning && devices==nil && scanErr==""` → body claims **"Tailscale is running but reports no other devices."** That is false; we never scanned this session.

Mirror this Init branch wherever the fleet becomes empty (`removePending` after successful save/reload is the main one). Without that, the onboarding screen's most valuable path is only correct on cold start.

cmd/camp/machine_tui_onboarding.go
```diff
@@ -0,0 +1,166 @@
...
+
+		host := machineMuted.Render(pad(ui.Truncate(device.Host, hostWidth), hostWidth))
+
+		state := machineMuted.Render("offline")
+		if device.Online {
+			state = machineOKStyle.Render("online ")
+		}
+
+		row := prefix + name + " " + host + " " + state
+		if id, ok := configured[device.Host]; ok {
```
obey-agent, line 155: **Incomplete guarantee (Medium):** Marking "already added as X" teaches the right thing, but `enter` still opens a prefilled form for that host (`updateOnboarding` / `prefillFromDevice` do not consult `configuredHosts`).

The comment above says this prevents the same machine being added twice under a second id — it doesn't. The user can change the id on the form and save a second entry for the same host, or if `deriveMachineID` collides they only learn at save time.

Minimum: refuse enter (or confirm) when `configured[device.Host]` hits, with status like `already added as X · e it from the fleet list`. Stronger: also match case-folded / MagicDNS variants if discovery host and stored host can diverge.

internal/commands/fresh/workflow.go
```diff
@@ -60,10 +159,74 @@ func buildFreshWorkflow(cfg *config.FreshConfig, projectName string) []freshWork
...
+	if !own {
+		return freshStateOff
+	}
+	if !dependency {
+		return freshStateBlocked
+	}
+	return freshStateOn
+}
+
+func branchState(branch string) freshStepState {
```
obey-agent: **State model gap (Medium):** `branchState` only sees the *resolved* branch string, so it can never return `freshStateOff`.

For a project scope:
- inherit → resolved non-empty → `On` (fine)
- explicit `branch: ""` (the "no branch" decision `SetFreshBranch` carefully preserves) → resolved `""` → **`Unset` (○ "not configured")**
- never set, no global → resolved `""` → also `Unset`

The design comment on `freshStepState` and the test name `TestBranchStateSeparatesUnsetFromOff` claim off vs unset are distinct, but this function collapses explicit "no branch" into unset. The editor knows the difference (`currentSettingAction` / pointer semantics); the workflow pane does not.

Pass enough raw scope state into `branchState` (or compute state in `buildFreshWorkflow` from `cfg.Projects[name].Branch` / `cfg.Branch`) so explicit empty renders as off (`·`) and only a truly absent decision is ○.

cmd/camp/machine_tui_onboarding.go
```diff
@@ -0,0 +1,166 @@
...
+		return "enter add the highlighted device · a add by hand · s rescan · ? what is this · q quit"
+	}
+	return "a add a machine · ? what is this · q quit"
+}
+
+// onboardingBody is the part of the screen that changes with what camp can
+// see: the tailnet when there is one, and the reason there is not otherwise.
+func (m *machineTUIModel) onboardingBody(width int) []string {
+	switch {
+	case m.scanning:
```
obey-agent, line 72: **Polish (Low/Medium):** Connection tests got a spinner specifically because an 8s still screen reads as hung. Scanning the tailnet has the same failure mode ("Looking for machines...") with no animation and no `spin.Tick` on this path.

Either reuse `m.spin` while `m.scanning`, or drop a lightweight animated glyph into this branch. First-run onboarding is exactly when a frozen frame hurts most.

obey-agent (commented):
## Verdict: **Request changes**

Staff-level adversarial review of `feat(machine): make the fleet TUI teach rather than assume` (also carries the full `camp fresh configure` settings editor work). Design intent is excellent; a few production paths still teach the wrong lesson.

**Identity:** review posted as `obey-agent` via token injection. Active `gh` account remains personal (confirmed).

---

### Summary

This PR does the right product work: empty-fleet onboarding that leads with payoff, reachability over ControlMaster trivia, plain-language auth/socket copy, auto-test on save, ranked footer, and VHS evidence of the first-run path. The fresh half correctly separates inherit vs off, global-only keys, and notice-vs-error status levels.

I am **not** signing off yet because the new teaching surface has regressions/holes on the exact failure modes it claims to own (missing remote camp; empty fleet after remove; "already added" not enforced), and the fresh branch state model under-delivers on the off/unset distinction it advertises.

Green unit tests and a recorded tape are necessary but not sufficient for TUIs — confirmed no pty-driven interaction suite for the new key paths (`t`, empty→add, remove-last→onboarding). The tape helps first paint; it does not cover mid-session transitions.

---

### Blocking / high

1. **`connectionFailureDetail` strips the camp-not-found hint** (`cmd/camp/machine_tui.go`)
   Comment claims to keep "camp's hint when the binary is missing"; implementation digs past `campNotFoundHint` to bare `bash: camp: command not found`, dropping `CAMP_REMOTE_CAMP_PATH` guidance. That is the failure mode health checks were added to expose. Use typed errors / preserve the outer wrap; add a regression test with a 127 + hint-shaped error.

2. **Empty-fleet transition after removing the last machine**
   Auto-scan only runs in `Init`. Mid-session `empty()` onboarding with tailscale installed and no prior device list falsely says "Tailscale is running but reports no other devices." Trigger the same scan path whenever the fleet becomes empty.

---

### Medium

3. **"Already added" is display-only** — enter still prefills/saves a second id for the same host.
4. **Onboarding footer omits `s`** when the body tells the user to rescan (scan error / empty list).
5. **Overlapping `s` scans** have no generation token; last-to-finish wins, can clobber a newer result.
6. **Fresh `branchState` never returns Off** — project explicit `branch: ""` renders as unset (○), same as never configured, despite the pointer-preserving write path and the test name claiming separation.
7. **Scan path has no spinner** while connection tests do — first-run wait is the hung-looking case the PR itself argues against.

---

### Lower / process

8. **Two large product surfaces in one PR** (machine fleet TUI + fresh configure settings). Harder to review, bisect, and revert.
9. **`TestBranchStateSeparatesUnsetFromOff` hardcodes step index `5`** and never asserts `freshStateOff` — name overclaims.
10. **Password auth remains in the cycle** while hops are unsupported — acceptable if intentional.
11. **Host equality for "already added" is exact string match** — IP vs MagicDNS twin entries still slip through.

---

### What looks strong

- Separation of health vs ControlMaster is the right model; auto-test-on-save matches user intent.
- Onboarding copy hierarchy (why → commands → action → local always works) is product-grade.
- Fresh configure: inherit vs write-false, global-only prune redirect, not seeding inherited branch into the input (the developfeat bug), comment-preserving YAML writers + good unit coverage.
- Demo tape re-recorded from empty machines.yaml with fixtures; honest unreachable end-state.
- Context passed into ssh/discovery cmds; machine copy used for async test avoids pointer-into-slice races.

---

### Suggested fix order

1. Preserve camp-not-found detail + test.
2. Rescan (or clear false empty-tailnet copy) on transition into empty fleet.
3. Block or confirm enter on already-configured hosts; fix footer for `s`.
4. Fix `branchState` with raw pointer semantics; fix the test.
5. Scan generation / in-flight guard; optional scan spinner.

---

### Inline comments

**7** inline review comments filed on the PR (connection detail, empty-fleet Init gap, already-added, footer/`s`, concurrent scan, branchState, scan spinner).

### Verification performed

- Full `gh pr diff` / file list / head `d705a44` read in worktree `projects/worktrees/camp/machine-tui-ux`
- Traced `testMachine` → `RunCampCommand` → `campNotFoundHint` → `connectionFailureDetail` reduction
- Traced empty onboarding branches, remove path, device mark/prefill, fresh `branchState` vs `SetFreshBranch` empty pointer
- Did **not** re-run `just gate-fast` or a live pty session in this review pass (author claims 3541 tests + VHS; logic bugs above are independent of green tests)

### Failures / open risks

| Risk | Severity |
|------|----------|
| Missing-camp errors teach the wrong next step | High |
| Remove-last → false "no devices" onboarding | High |
| Duplicate host under second id still easy | Medium |
| Fresh explicit no-branch looks unconfigured | Medium |
| No pty coverage for new interaction graph | Process |

---

**Bottom line:** Approve the direction; do not merge until (1) and (2) are fixed and covered. The rest can ship as follow-ups if needed, but (3)–(6) are cheap and still in the "teach rather than assume" blast radius.

cmd/camp/machine_tui_update.go
```diff
@@ -244,11 +349,19 @@ func (m *machineTUIModel) updateDiscover(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
...
+// A host already present in the fleet is refused rather than prefilled: the
+// row already teaches "already added as X", and opening the form would let the
+// user save a second id for the same machine.
 func (m *machineTUIModel) prefillFromDevice() tea.Cmd {
 	if len(m.devices) == 0 {
 		return nil
 	}
 	device := m.devices[clampIndex(m.deviceCursor, len(m.devices))]
+	if existing, ok := m.configuredHosts()[device.Host]; ok {
+		m.setStatus(fmt.Sprintf("already added as %s · select it in the fleet and press e to edit", existing))
```
obey-agent, line 362: **Polish (Low / non-blocking):** Refuse is correct and closes the original “already added is display-only” gap — enter no longer opens a second form. One UX hole remains on the *discover overlay* path (browse → `s` → enter on an already-configured host):

`m.setStatus(...)` writes the guidance, but `overlayView` never renders `statusLine`. The user sees no reaction to enter while the picker is open. The row’s “already added as X” mark mitigates this, so this is not a re-open of the prior Medium finding — just residual feedback wiring.

Cheap fix options:
1. Append the status into `discoverBody` while `m.status != ""`, or
2. Close the overlay on refuse so the browse status line is visible, or
3. Inline a one-line muted note under the selected row.

Not a merge blocker.

obey-agent (commented):
## Verdict: **Approve**

Re-review of `e848ba81` (`fix(machine): address staff review on fleet TUI teaching paths`) against the prior REQUEST CHANGES blockers. Active `gh` remains `lancekrogers`; this review posted as `obey-agent` via token injection only.

---

### Prior blockers — all closed

| # | Finding | Status | Evidence |
|---|---------|--------|----------|
| 1 | `connectionFailureDetail` strips `CAMP_REMOTE_CAMP_PATH` on exit 127 | **Closed** | Early `errors.As(*CommandError)` + ExitCode 127 returns `firstLine(err.Error())` before the strip path. `TestConnectionFailureDetailPreservesCampNotFoundHint` guards the wrap. |
| 2 | Remove-last machine → false “no other devices” onboarding | **Closed** | `removePending` mirrors `Init`: `empty() && tailscaleReady` → `beginScan(false)`. Covered by `TestRemoveLastMachineStartsOnboardingScan`. |
| 3 | “Already added” display-only (enter still prefills) | **Closed** | `prefillFromDevice` consults `configuredHosts()` and refuses with status. `TestPrefillFromDeviceRefusesAlreadyConfiguredHost`. One **Low** residual on discover-overlay status visibility (inline). |
| 4 | Footer omits `s` when body teaches rescan | **Closed** | `onboardingFooter` lists `s rescan` whenever `tailscaleReady` (scanning → “scanning · …”). `TestOnboardingFooterListsRescanWhenTailscaleReady`. Empty-device body also mentions `s`. |
| 5 | Overlapping scans / no generation token | **Closed** | `scanGen` + `beginScan` / `devicesMsg.gen` / stale drop in `applyDevices`. `TestApplyDevicesDropsStaleGeneration`. |
| 6 | `branchState` never `Off` for project `branch: ""` | **Closed** | `branchState(cfg, projectName, branch)` treats project `Branch != nil && *Branch == ""` as `freshStateOff`. Extended `TestBranchStateSeparatesUnsetFromOff`. |
| 7 | Scan spinner | **Closed** | `busy()` = testing \|\| scanning; spinner ticks while scanning; onboarding + discover bodies prepend `m.spin.View()`. |

---

### Residual (non-blocking)

- Discover overlay: refuse path sets status that `overlayView` does not render. Row already marks “already added as X”. Inline only — polish, not a re-open of #3.

No new High/Medium issues found on head.

---

### Notes

- Fix commit is tightly scoped (+252/−22 across 7 files) and matches its message.
- Regression tests land with each closed finding — good staff-level hygiene.
- Fresh configure settings work + machine teaching UX remain coherent; the branch off/unset distinction now matches `SetFreshBranch`’s pointer semantics.

**Ship it.**

PR #458, opened 2026-07-19
fix: expose auto-commit session diagnostics

lancekrogers:
## Summary

- Forward configured commit-message hook diagnostics to the operator's stderr.
- Keep hook stderr captured for useful error messages while preserving stdout as the commit message.
- Make newly scaffolded `camp commit --auto-write` hooks use `ob commit --print-session-id`.

## Why

`camp commit -aw` runs the configured hook successfully but previously discarded successful hook stderr. That hid the daemon session identifier emitted by `ob`, making a long-running auto-commit request difficult to inspect or recover.

## Validation

- `go test ./...`
- `git diff --check`


pkg/commitkit/autowrite.go
```diff
@@ -111,7 +122,15 @@ func RunCommitMessageCommandWithEnv(ctx context.Context, repoPath, command strin
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
 	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
+	// Keep stderr available for error reporting while forwarding successful
+	// hook diagnostics to the operator. Commands such as `ob commit` report
+	// the daemon session ID on stderr so the session can be inspected while
```
obey-agent: **Accuracy issue:** This comment overstates what `ob` does today.

In `ob` (`printSessionIDIfNeeded` / `finalizeCommandSession`), `session_id=<id>` is emitted only *after* `RunSession` returns — not when the daemon session is created. With default create-and-stop lifecycle, the session is often already stopped by then.

So MultiWriter correctly streams mid-run *progress* (`ob: connecting...`, `ob: generating...`), but it does **not** give operators a live session id to inspect/recover a hung auto-commit. A hung `ob commit` never reaches finalize, so no `session_id=` is ever printed.

Please reword to match real behavior (live progress + post-completion id when `--print-session-id` is set), or track a follow-up in `ob` to emit the id immediately after `CreateSession` if mid-run recovery is the goal.

internal/config/campaign.go
```diff
@@ -161,7 +161,8 @@ func commentedHooksPlaceholder() []byte {
 #   commit_message:
 #     # Command run from the target repo working tree by 'camp commit --auto-write'
 #     # (-aw). Its stdout is used verbatim as the commit message. Any tool works;
-#     # 'ob commit' is one example.
-#     command: ob commit
+#     # 'ob commit --print-session-id' is one example; its session ID is
+#     # forwarded to the terminal so it can be inspected while it runs.
```
obey-agent: **User-facing scaffold overclaims mid-run inspectability.**

New campaigns will copy this example. As written, operators will expect to see a session id *while* `camp commit -aw` is running and use it for recovery.

That is not what happens with `ob commit --print-session-id` today:
1. `session_id=` is printed only at command finalization (after generation completes).
2. Default lifecycle stops the session, so the id is mostly post-hoc.
3. If the hook hangs during generation, finalize never runs → still no id.

Suggest accurate copy, e.g.:
```
# 'ob commit --print-session-id' is one example; progress and any
# session_id= line on stderr are forwarded to the terminal (session_id is
# emitted when the hook finishes, not while it is generating).
command: ob commit --print-session-id
```
If live recovery is the intended product story, that needs an `ob` change (emit id right after create) and likely `--keep-session` tradeoffs — not just camp stderr forwarding.

pkg/commitkit/autowrite_internal_test.go
```diff
@@ -0,0 +1,28 @@
+package commitkit
+
+import (
+	"bytes"
+	"context"
+	"testing"
+)
+
+func TestRunCommitMessageCommandForwardsDiagnostics(t *testing.T) {
```
obey-agent, line 10: **Coverage gap (non-blocking but worth closing):** only the success path is asserted.

Please add a table case (or second test) that:
1. writes diagnostics to stderr then `exit 1`
2. asserts diagnostics were still forwarded to `diagnosticOut`
3. asserts the returned error still contains the captured stderr text (the MultiWriter dual-path contract)

Without that, a future refactor could break error wrapping while keeping the happy-path forward green. Existing `TestRunCommitMessageCommand/fails_on_non-zero_exit` covers public API capture, but not the dual-write inject path introduced here.

obey-agent (commented):
## Verdict: REQUEST_CHANGES

### Overview
Small, focused change: tee commit-message hook stderr to the operator while still buffering it for error wrapping, and point new scaffolds at `ob commit --print-session-id`. The MultiWriter approach is the right integration pattern for dual capture+forward. Production confidence on the mechanism is high; confidence on the *stated product outcome* (mid-run session inspect/recover) is low against current `ob` behavior.

### Key Findings
1. **[Blocking] Misleading mid-run session recovery claims** — comments in `autowrite.go` and the user-facing `campaign.yaml` scaffold say the session ID is forwarded so it can be inspected *while the request runs*. In `ob`, `session_id=` is emitted only in `finalizeCommandSession` after `RunSession` returns. Default create-and-stop often stops the session first; a hung generation never finalizes, so no id appears. MultiWriter still correctly streams progress (`ob: connecting...` / `ob: generating...`) mid-run — reword claims to that reality, or open an `ob` follow-up to print the id immediately after `CreateSession`.
2. **[Non-blocking] Thin dual-path test** — new internal test covers success-only forward. Add a failure case that asserts both live diagnostic forward *and* stderr still present in the wrapped error (the MultiWriter contract).
3. **[Minor] Failure UX double-print** — on non-zero exit, operators now see stderr live *and* again inside `auto-write commit message command failed: ...`. Acceptable for diagnostics; call out if camp later wants quieter error wrapping when stderr was already streamed.

### What's Done Well
- Correct split: public API always forwards to `os.Stderr`; unexported helper injects `io.Writer` for tests (good DI without expanding the public surface unnecessarily).
- Preserves prior error-wrapping behavior by keeping a `bytes.Buffer` tee alongside the live writer.
- Context cancellation still checked up front; existing cancel / empty-stdout / non-zero-exit coverage remains valid.
- Scaffold + error-string example stay aligned (`--print-session-id`); config test pins discoverability.
- Scope is tight; no parallel implementation or magic strings introduced.

### Staff Standard
Would I approve this for production as written? **Not yet** — the mechanism is mergeable, but shipping a scaffold every new campaign will read that teaches false mid-run recovery behavior fails the staff bar for operator-facing accuracy. Fix the wording (and optionally extend the failure-path test); no redesign of MultiWriter required.

---
*Adversarial staff review as obey-agent on head `7e6e493c`.*

obey-agent (approved):
## Verdict: APPROVE

### Re-review of prior REQUEST_CHANGES (head `2c5f4a17`, fix commit `15796c0f`)

Prior blockers are closed. Mechanism was already sound; the accuracy and test gaps that blocked merge are addressed.

### Closed

1. **Mid-run session recovery overclaims (was blocking)** — Closed.
   - `pkg/commitkit/autowrite.go`: MultiWriter comment now correctly separates live progress (`ob: connecting...` / `ob: generating...`) from post-completion `session_id=` diagnostics, and states a hung generation never finalizes / never emits an id.
   - `internal/config/campaign.go` scaffold matches that reality: “session_id is emitted when the hook finishes, not while it is generating.”
   - No residual operator-facing code comment teaches false mid-run inspect/recover.

2. **Thin dual-path test (was non-blocking, requested)** — Closed.
   - `TestRunCommitMessageCommandForwardsDiagnostics` is table-driven with a failure case that asserts both live diagnostic forward *and* stderr retained in the wrapped error (`auto-write commit message command failed` + stderr text). That pins the MultiWriter contract.

3. **Failure double-print (was minor)** — Closed as documented intentional.
   - Comment explicitly notes operators may see stderr twice on failure (live stream + wrapped error) and that this is intentional for now.

### Residual (non-blocking)

1. **PR body “Why” still slightly overclaims product outcome.**
   - Current text: discarded stderr “hid the daemon session identifier… making a long-running auto-commit request difficult to inspect or recover.”
   - That still reads as mid-run recovery. Prefer aligning the PR description with the fixed code comments before merge (e.g. progress is live; `session_id=` is post-completion diagnostics when `--print-session-id` is set). Not a code blocker — not scaffolded into campaigns.

### Validation checked

- Relevant packages green: `go test ./pkg/commitkit/ ./internal/config/`
- PR file set still tight: `autowrite.go`, `autowrite_internal_test.go`, `campaign.go`, `campaign_test.go`
- Merge-from-main present; no reintroduction of the inaccurate mid-run recovery wording in the changed files

### Staff Standard

Would I approve this for production as written? **Yes.** Operator-facing scaffold and code comments match `ob` behavior; dual-path tee is tested on success and failure; scope remains correct.

---
*Adversarial re-review as obey-agent after fix commit `15796c0f` + merge main.*

PR #460, opened 2026-07-19
fix(worktrees): clean enumerates via git worktree list

### LINKED ISSUE #436: worktrees list/clean miss non-preferred-location worktrees (directory scan, not git)
## Summary

`camp worktrees list`, `camp project worktree list`, and `camp worktrees clean` enumerate worktrees by scanning the preferred directory (`projects/worktrees/<project>/`) instead of asking git. A worktree registered with `git worktree add` in any other location is invisible to all three commands, their table output, and `--json`.

This is the same root cause that PR #429 fixed for navigation but explicitly scoped out of the list/clean surfaces. It surfaced again during review of #431 (interactive worktrees TUI), which reuses the same data source, so the polished browser is also blind to non-preferred-location worktrees.

## Where

- `cmd/camp/worktrees/list.go:154` — `listWorktrees` calls `pm.ListProjectWorktrees(projectName)` (a bare `os.ReadDir` of `projects/worktrees/<project>/`).
- `cmd/camp/worktrees/list.go:213-226` — `filterRegisteredWorktreeNames` cross-checks against `git worktree list` entries but only ever *removes* names from the directory scan (it keys the allowlist on `pm.WorktreePath(projectName, name)`, the preferred path). It can never *add* a git-registered worktree that lives elsewhere.
- `cmd/camp/project/worktree/list.go:82` — the canonical command shares the same `ListProjectWorktrees` scan.
- `cmd/camp/worktrees/clean.go:222` — `clean` shares it too (so cleanup reasoning is also incomplete).

Contrast: navigation was fixed in #429 via `internal/nav/index/builder.go` `scanWorktrees` → `project.List` + `worktree.NewGitWorktree(projectPath).List(ctx)` (git as source of truth).

## Reproduce

```
git -C projects/<proj>/<some-checkout> worktree add ../../rogue-fix
camp worktrees list          # rogue-fix absent; count off by one
camp project worktree list   # absent
camp worktrees list --json   # total omits it
```

## Fix direction

Make `listWorktrees` (and the `clean` enumeration) source the worktree set from `git worktree list --porcelain` the way `scanWorktrees` already does, rather than the directory scan, then annotate preferred-vs-external location for display. Consolidating on the `internal/nav/index` / `worktree.GitWorktree` enumeration avoids a third parallel implementation.

## Notes

- Found reviewing merged PR #431; the TUI itself is fine, it just inherits this data source.
- Also tracked as campaign intent `camp-worktrees-list-and-camp-20260715-024134`.
- A regression test under `tests/integration/` asserting a non-preferred-location worktree appears in `list`/`project worktree list` would guard the fix.

lancekrogers:
## Summary

- `camp worktrees clean` previously only scanned `projects/worktrees/<project>/`, so stale worktrees registered elsewhere were invisible (same root cause as #436).
- **List** already uses `git worktree list` as source of truth; this PR aligns **clean** with that model.
- Preferred-location directory scan remains as a secondary pass for orphan dirs not present in git's list.
- Unit tests for staleness classification.

Closes #436

## Test plan

- [x] `go test ./cmd/camp/worktrees/`
- [ ] existing integration `worktrees_clean` tests in CI

### DIFF
--- a/cmd/camp/worktrees/clean.go
+++ b/cmd/camp/worktrees/clean.go
@@ -13,6 +13,7 @@ import (
 	git "github.com/Obedience-Corp/camp/internal/git"
 	navtui "github.com/Obedience-Corp/camp/internal/nav/tui"
 	"github.com/Obedience-Corp/camp/internal/paths"
+	"github.com/Obedience-Corp/camp/internal/project"
 	"github.com/Obedience-Corp/camp/internal/ui"
 	"github.com/Obedience-Corp/camp/internal/worktree"
 	"github.com/spf13/cobra"
@@ -98,8 +99,9 @@ func runWorktreesClean(cmd *cobra.Command, args []string) error {
 	resolver := paths.NewResolver(campRoot, cfg.Paths())
 	pathManager := worktree.NewPathManager(resolver)
 
-	// Find stale worktrees
-	stale, err := findStaleWorktrees(ctx, pathManager, cfg, cleanProject)
+	// Find stale worktrees (git worktree list is the primary source of truth,
+	// same as camp worktrees list; preferred-dir scan catches orphans only).
+	stale, err := findStaleWorktrees(ctx, campRoot, pathManager, cleanProject)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


cmd/camp/worktrees/clean.go
```diff
@@ -202,48 +204,105 @@ func runWorktreesClean(cmd *cobra.Command, args []string) error {
...
+			for _, entry := range gitEntries {
+				if !worktree.IsLinkedWorktree(target.path, entry) {
+					continue
+				}
+				cleanPath := filepath.Clean(entry.Path)
+				if _, ok := seen[cleanPath]; ok {
+					continue
+				}
+				seen[cleanPath] = struct{}{}
+				if cr, ok := staleIfRemovable(target.name, filepath.Base(cleanPath), cleanPath); ok {
```
obey-agent, line 242: **No regression coverage for the bug this PR claims to fix.**

`findStaleWorktrees` now correctly primary-enumerates via `git worktree list` (mirroring list), but neither unit nor integration tests exercise that path. New tests only cover `staleIfRemovable` classification.

#436 explicitly called for a regression under `tests/integration/` for a **non-preferred-location** worktree. List got that in #446 (`TestWorktreesList_ResolvesAcrossAllLocations`). Clean still only covers preferred-layout stale dirs (`worktrees_clean_test.go`).

Please add an integration case roughly:
1. `git worktree add` outside `projects/worktrees/<project>/`
2. break gitdir (or equivalent removable stale)
3. `camp worktrees clean --all --yes` removes it

Without that, this fix can regress with every unit suite green — same class of gap that left clean broken after list was fixed.

cmd/camp/worktrees/clean.go
```diff
@@ -202,48 +204,105 @@ func runWorktreesClean(cmd *cobra.Command, args []string) error {
...
+	}
+	return out, nil
+}
+
+func staleIfRemovable(projectName, worktreeName, path string) (cleanResult, bool) {
+	reason := checkWorktreeStale(path)
+	if reason == "" {
+		return cleanResult{}, false
+	}
+	if reason != "gitdir target does not exist" && !stalenessIsGitDir(reason) {
```
obey-agent, line 274: **Git-listed but vanished worktrees still never become clean candidates.**

With git as source of truth, entries whose directory is gone (or whose `.git` file is missing) now appear in enumeration, then `checkWorktreeStale` returns `"missing .git file"`, and this filter drops them:

```go
if reason != "gitdir target does not exist" && !stalenessIsGitDir(reason) {
    return cleanResult{}, false
}
```

So the case Long help describes — "the directory no longer exists" / broken git registration with no checkout left — is still not cleaned, and git's admin entry remains until something else prunes it.

This filter is pre-existing, but the PR makes the gap more visible: primary enumeration now *finds* these entries and then silently ignores them. Either:
- extend removable reasons + route through `git worktree remove`/`prune` for missing paths, or
- document explicitly that clean only auto-removes "gitdir target does not exist" (and never missing-dir), and stop claiming directory-gone cleanup in Long help.

Not necessarily blocking if out of scope, but do not close #436 as "clean half fully fixed" without acknowledging this residual class.

cmd/camp/worktrees/clean.go
```diff
@@ -202,48 +204,105 @@ func runWorktreesClean(cmd *cobra.Command, args []string) error {
...
 
 	return stale, nil
 }
 
+type cleanProjectTarget struct {
+	name string
+	path string
+}
+
+func cleanProjectTargets(ctx context.Context, campRoot, filterProject string) ([]cleanProjectTarget, error) {
```
obey-agent: **Duplicate of `listWorktreeProjectTargets`.**

This is effectively a copy of `listWorktreeProjectTargets` in `list.go` (same Resolve / List / ResolveProjectPath shape). Fine for a surgical fix, but a shared helper (or package-local `worktreeProjectTargets`) would prevent list/clean from drifting again — which is exactly how clean fell behind list after #446.

Non-blocking nit if you prefer a follow-up; worth a one-line TODO if not shared now.

obey-agent (commented):
## Verdict: REQUEST_CHANGES

### Overview
Aligns `camp worktrees clean` enumeration with the git-as-source-of-truth model already used by `camp worktrees list` (#446). Correct architectural fix for the clean half of #436: primary `git worktree list`, secondary preferred-dir orphan scan, shared project target resolution via `project.List`/`Resolve`. Implementation shape is solid; **production confidence is not** — the bug path is untested.

### Key Findings
1. **Missing non-preferred-location regression test (blocking)** — The actual #436 failure (stale worktree outside `projects/worktrees/<project>/`) has no integration or unit coverage of `findStaleWorktrees`. Only `staleIfRemovable` is unit-tested. List already has `TestWorktreesList_ResolvesAcrossAllLocations`; clean needs the analogous case.
2. **Residual: "directory no longer exists" still not removable** — Git-listed vanished paths → `"missing .git file"` → filtered out by `staleIfRemovable`. Pre-existing policy, but more visible now that git enumeration surfaces those entries. Long help still over-claims this class.
3. **Duplicated project-target helper** — `cleanProjectTargets` ≈ `listWorktreeProjectTargets`. Drift risk (how clean lagged list). Nit / follow-up.
4. **`--all` project set change** — Switches from `pm.ListAllProjects()` (dirs under worktrees root) to `project.List` (registered projects). Correct for parity with list; note orphan preferred dirs for *deregistered* project names are no longer scanned. Acceptable tradeoff if intentional.

### What's Done Well
- Primary/secondary enumeration order matches the real problem (git first, preferred-dir orphans second).
- `IsLinkedWorktree` filter reused — no parallel main-worktree / bare / `.git` path heuristics.
- Error policy mirrors list: hard-fail on `--project`, best-effort skip on `--all`.
- `staleIfRemovable` extraction improves readability and is unit-tested for the three classification cases.
- Context propagated on `project.Resolve` / `project.List` / git list.

### Staff Standard
- **Errors / context**: Good — wraps list failures, preserves ctx on project APIs.
- **Integration**: Code integrates cleanly with existing `worktree` / `project` packages; removal path unchanged and still conservative.
- **Maintainability**: Small, clear refactor; helper duplication is the only maintainability ding.
- **Production confidence**: **Insufficient without a regression test for non-preferred stale cleanup.** Green unit suite does not prove the #436 fix.

**Required before approve:** integration regression for clean of a stale non-preferred-location worktree (and ideally a note/TODO or Long-help fix for missing-dir residual).

obey-agent (commented):
## Re-review verdict: **APPROVE**

Prior: **REQUEST_CHANGES** @ `008da3f` (missing non-preferred-layout regression test; long help over-claim; duplicated project-target helper).  
This pass: head `3e14b16` — regression test + shared targets + honest help.

---

### Prior blockers — status

| # | Issue | Status |
|---|--------|--------|
| 1 | No regression test for stale worktree outside preferred layout | **Closed** — `TestWorktreesClean_StaleOutsidePreferredLayout` |
| 2 | Long help over-claim on “directory no longer exists” | **Closed** — documents conservative auto-removal + `git worktree prune` for vanished checkouts |
| 3 | Duplicated `cleanProjectTargets` / list helper | **Closed** — shared `worktreeProjectTargets` |
| 4 | `--all` project set via `project.List` | Acceptable (parity with list); unchanged residual tradeoff |

### Closed in detail

- Integration test builds a linked worktree **outside** `projects/worktrees/<project>/`, breaks `.git` pointer while keeping git admin so `git worktree list` still enumerates it, then asserts `--all --yes` removes it. Correctly targets the #436 clean failure mode.
- Help text no longer promises auto-prune of fully vanished git-listed paths; policy is explicit and conservative.
- List/clean cannot drift on project target resolution.

### Residual (nits)

1. Vanished-path git admin leftovers still require `git worktree prune` — intentional, documented.
2. Orphan preferred dirs for *deregistered* project names remain unscanned under `--all` — same as list; OK if intentional.

### Mergeability

| Check | Status |
| --- | --- |
| `mergeable` | **MERGEABLE** |
| `mergeStateStatus` | **CLEAN** |

Production confidence for the #436 clean path is now sufficient. **Approve.**

PR #464, opened 2026-07-19
feat: v1alpha8 reader schema, forward-compat rule, and JSON contract v1alpha9

lancekrogers:
## Summary

This is the reader half of a two-phase rollout of the v1alpha8 workitem marker schema (tags + projects fields). It makes every read path tolerant of the new schema before any write path emits it.

- `Metadata` gains `Tags []string` and `Projects []string` (yaml omitempty); loader allowlist widens to v1alpha4-v1alpha8.
- New forward-compat rule: an unknown future `v1alphaN` with a numeric suffix greater than the current max loads with a warning instead of failing closed, so the bump after v1alpha8 will not need this two-phase dance again. Exported as `IsForwardCompatVersion`.
- Tag validation (`^[a-z0-9][a-z0-9-]{0,63}$`, duplicate rejection) and projects validation (relative, non-escaping, duplicate rejection; existence deliberately not checked at load).
- JSON list contract bumps `workitems/v1alpha8` -> `workitems/v1alpha9`: `WorkItem` carries `tags`/`projects`, always serialized as `[]` when empty, never `null`, guaranteed at every construction path (ApplyMetadata plus the three marker-less discovery sites) and pinned by tests that invoke the real discovery functions.

**No write path changed.** `WorkitemSchemaVersion` still stamps `v1alpha7`; `create`/`adopt`/`repair`/`gather` and `links.yaml` handling are untouched, verified by diff scope (only `internal/workitem/{metadata,model,merge,json,discover_*}.go` and tests).

## Rollout contract

The writer release (003_WRITER_RELEASE) is blocked until this release is cut as an rc and reaches an installed channel. A pre-reader binary hard-fails on a v1alpha8 marker by design; that documented behavior gets an explicit contract test in the hardening phase.

Full context: `obey-campaign` festival `festivals/active/workitem-schema-tags-and-projects-WS0001` (design package `workflow/design/workitem-schema-tags-and-projects/`, docs 01-09).

## Testing

- `go build ./...`, `go test ./...` (whole module), `just lint` all green.
- New tests: version acceptance table (legacy, current, v1alpha8, forward-compat incl. leading-zero corner, malformed rejects), tag/project validation table, v1alpha7-marker-unaffected rollout test, JSON never-null tests including real-invocation coverage of all three marker-less discovery construction sites with a verified mutation check.

### DIFF
--- a/internal/workitem/discover_festivals.go
+++ b/internal/workitem/discover_festivals.go
@@ -103,6 +103,8 @@ func discoverFestivals(ctx context.Context, campaignRoot string, resolver *paths
 				SourceMetadata: map[string]any{
 					"festival_type": meta.FestivalType,
 				},
+				Tags:     []string{},
+				Projects: []string{},
 			}
 			item.SortTimestamp = DeriveSortTimestamp(item.UpdatedAt, item.CreatedAt)
 			if primaryDocAbs != "" {
--- a/internal/workitem/discover_intents.go
+++ b/internal/workitem/discover_intents.go
@@ -67,6 +67,8 @@ func discoverIntents(ctx context.Context, campaignRoot string, resolver *paths.R
 					"concept":     i.Concept,
 					"priority":    string(i.Priority),
 				},
+				Tags:     []string{},
+				Projects: []string{},
 			}
 			item.SortTimestamp = DeriveSortTimestamp(item.UpdatedAt, item.CreatedAt)
 			item.Summary = extractSummary(i.Content, 200)
--- a/internal/workitem/discover_test.go
+++ b/internal/workitem/discover_test.go
@@ -505,3 +505,68 @@ func TestTimestampDerivation(t *testing.T) {
 		}
 	})
 }
+
+func TestDiscovery_TagsProjectsNonNilAtConstruction(t *testing.T) {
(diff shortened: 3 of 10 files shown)


obey-agent (commented):
Reviewed across two independent passes during WS0001 festival execution (opus implementer, sonnet adversarial reviewer, orchestrator gate run).

Sequence 01 (metadata schema + forward-compat): CLEAN. Verified the 3-branch version rule matches design doc 02 exactly, rejection error string byte-identical to pre-change, WorkitemSchemaVersion untouched at v1alpha7, no write path in diff scope. Version edge cases traced: malformed forms rejected, strconv overflow safe, leading-zero corner pinned by test.

Sequence 02 (JSON contract v1alpha9): one P2 finding raised and fixed before merge: the never-null regression test originally used a hand-built literal; it now invokes the real discovery functions for all three marker-less construction sites, with a verified mutation check (reverting a site literal fails the test). Every WorkItem construction site in the module was traced to confirm no zero-value item reaches the JSON contract. Payload shape strictly additive for external consumers.

Gates re-run by the orchestrator at merge time: go build, full-module go test, just lint (0 issues, no new host-fs violators, no fmt.Errorf regressions). Merge-ready.

PR #468, opened 2026-07-19
feat: v1alpha8 writer release: tags, projects, filters, file workitems, lifecycle verbs

lancekrogers:
## Summary

The writer half of the two-phase v1alpha8 workitem schema rollout. The reader half shipped in #464 and reached the installed rc channel as camp v0.3.0-rc.3 (pinned by festival PR #109) before any commit here landed, per the rollout contract: every binary on the rc channel can already read what these writers produce.

Seven sequences, one commit each:

- `e1e0bca5` write-path flip: WorkitemSchemaVersion stamps v1alpha8 on create/adopt/repair/gather; all legacy versions still load.
- `62dca6e5` tags: repeatable `--tag` on create/adopt with lowercase-kebab write normalization (canonical regex exported as `wkitem.ValidTag`); repair normalizes and dedupes existing tags, recording genuine drops distinctly from reformats.
- `23832e77` projects: repeatable `--project` with path normalization (Clean + trailing-slash trim, empty-after-normalize rejected naming the input), exported `ValidateProjectPaths`, doctor warning `workitem.project.not-found` (plus `workitem.project.unvalidatable` for stat errors), consolidated doctor discovery scans.
- `32a1b710` list filters: `--tag` (AND semantics, deliberately diverging from the OR-within-dimension convention per the design) and `--project` (OR), with filter-input normalization matching the write side.
- `b69035ff` frontmatter discovery: `workflow/**` markdown files with `kind: workitem` frontmatter discovered as `item_kind: file`; collision rule (directory marker owns identity), conflict rule (`workitem.identity.conflict` hard error), one-sided ref state surfaced as `workitem.identity.ref-pending` warning, type-authority (path wins), and the forward-compat validate finding `workitem.schema.forward-compat` (warning, exit 0, no repair hint since repair would downgrade).
- `cc8ccd51` adopt/create `--file`: frontmatter stamping that never touches the body, preserves foreign keys byte-for-byte (2-space indent; non-2-space input reflows with a printed notice), refuses conflicting identity unconditionally, unions foreign tags under `--force` with a printed notice for non-conforming drops, deterministic output, explicit `agent_allowed: false` on adopt.
- `98c1b672` file lifecycle: dungeon/promote/gather parity for file workitems, including two latent impossible-path bugs fixed (gather and promote identity both loaded `<file>.md/.workitem` for file sources).

## Scope note

[...]

PR #469, opened 2026-07-19
feat: deprecate links.yaml related-project rows: CLI rejection, doctor migration, merged JSON view

lancekrogers:
## Summary

Phase three of the v1alpha8 workitem schema rollout (after reader #464 and writer #468): the `links.yaml` `role: related` + `scope.kind: project` overlap is deprecated. Semantic project membership now has exactly one writable store, the workitem's own `projects:` field, and one merged read view.

- `e10458d7` CLI rejection: `camp workitem link --role related --project <name>` is rejected at the command layer with an error naming the replacement (`create`/`adopt --project`). Registry primitives (`AddLink`) are untouched, so doctor keeps its ability to manage rows. `--allow-missing` deliberately does not bypass. All other role and scope combinations behave exactly as before.
- `58e1b5e4` doctor migration: a new auto-fixable warning `workitem.link.related-project-deprecated` per legacy row; `--fix` writes the project into the workitem's identity document (marker or frontmatter, placement matching the create/adopt key convention), idempotent set semantics, then removes the row via `RemoveLinkByID`, both inside one locked pass. A row whose workitem cannot be resolved is reported as unmigratable and preserved, never deleted as a broken-link side effect (data preservation over cleanup, recorded in the festival).
- `4290dbf4` merged JSON view: the `--json` `projects` field becomes `[{path, primary}]`, the semantic set annotated with primary from `PrimaryForScope`, always `[]` when empty, at both output call sites. A malformed `links.yaml` degrades gracefully (nil-registry fallback, stderr warning pointing at `doctor --fix`) instead of failing read-only listing.

## JSON shape decision

Doc 04 left the merged-entry encoding (plain string vs object) open. Decision: object encoding `{path, primary}` in place, within `workitems/v1alpha9`, no version bump. Rationale: doc 04's binding constraint is one array that IS the merged view (an additive second field would re-introduce the split-view problem this phase removes); no released binary has ever emitted a populated `projects` array (the field shipped empty-only in v0.3.0-rc.3 and the writer release is unreleased); festival-app does not model the field but does pin `schema_version`, so bumping is the riskier move. The originally scaffolded `project_refs`/v1alpha10 sketch is superseded; the v1alpha9 changelog entry documents the object shape.

## Dependencies and verification

Depends on #468 (the `projects:` field must exist), already merged. Every sequence passed an independent adversarial review before its commit; findings fixed pre-merge included the malformed-registry regression above and a frontmatter key-placement divergence. Gates at PR time: build clean, whole-module unit green, lint 0 issues, containerized integration 511/511.

Full context: `obey-campaign` festival `festivals/active/workitem-schema-tags-and-projects-WS0001` (design doc `workflow/design/workitem-schema-tags-and-projects/04-projects-and-links.md`).

### DIFF
--- a/internal/commands/workitem/doctor.go
+++ b/internal/commands/workitem/doctor.go
@@ -28,18 +28,19 @@ import (
 // Doctor finding codes (dotted-domain form). Stable strings; consumers
 // dispatch on them.
 const (
-	codeBrokenLink           = "workitem.link.broken"
-	codeBrokenScope          = "workitem.scope.broken"
-	codeOutOfBounds          = "workitem.scope.out-of-bounds"
-	codeScopeUnvalidatable   = "workitem.scope.unvalidatable"
-	codeDuplicatePrimary     = "workitem.link.duplicate-primary"
-	codeSchemaViolation      = "workitem.schema.violation"
-	codeCurrentMissing       = "workitem.current.missing"
-	codeMissingRefField      = "workitem.ref.missing"
-	codeWorkitemScanFailed   = "workitem.scan.failed"
-	codeRegistryParseError   = "workitem.registry.parse-error"
-	codeProjectNotFound      = "workitem.project.not-found"
-	codeProjectUnvalidatable = "workitem.project.unvalidatable"
+	codeBrokenLink               = "workitem.link.broken"
+	codeBrokenScope              = "workitem.scope.broken"
+	codeOutOfBounds              = "workitem.scope.out-of-bounds"
+	codeScopeUnvalidatable       = "workitem.scope.unvalidatable"
+	codeDuplicatePrimary         = "workitem.link.duplicate-primary"
+	codeSchemaViolation          = "workitem.schema.violation"
+	codeCurrentMissing           = "workitem.current.missing"
+	codeMissingRefField          = "workitem.ref.missing"
+	codeWorkitemScanFailed       = "workitem.scan.failed"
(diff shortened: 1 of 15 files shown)


obey-agent (commented):
Reviewed sequence-by-sequence during WS0001 festival execution: three adversarial review passes by an independent reviewer agent, findings fixed before each sequence committed, plus an orchestrator gate run at PR time.

Sequence 01 (CLI rejection): CLEAN on first review. Guard truth table verified exact (only related+project rejected), byte-for-byte no-mutation on rejection proven, no legitimate --allow-missing caller breaks (repo-wide grep), registry primitives untouched.

Sequence 02 (doctor migration): crash-window traced by hand (marker-write-then-row-removal ordering converges idempotently on re-run), the data-preservation ruling for unresolvable rows verified correctly scoped with no broken-link regression, one cosmetic frontmatter key-placement divergence found and fixed pre-commit. The implementer independently added a nil-guard the task sketch lacked.

Sequence 03 (merged view): contract surface swept for stray consumers across festival-app, camp-graph, and obey (none found), both output call sites covered, one real regression found empirically (malformed links.yaml hard-failing read-only listing) and fixed pre-commit with graceful degradation. The JSON shape decision is recorded in the PR body with rationale.

Gates at merge time, run by the orchestrator: go build clean, whole-module unit green, just lint 0 issues, containerized integration 511/511. Merge-ready.

PR #470, opened 2026-07-20
fix(register): camp register created campaigns with no festivals/ directory

lancekrogers:
## Problem

`camp register` offers to initialize a directory that isn't yet a campaign. Answer `y` and it creates a campaign **with no `festivals/` directory** — then prints a success line saying it initialized one.

```
⚠ No campaign found at /tmp/probe/bare
Would you like to initialize one? [y/N] Warning: fest init failed: ... unknown flag: --path
✓ Initialized and registered campaign at /tmp/probe/bare

festivals/: MISSING
```

## Cause

`scaffold.Init` carried its own copy of the fest invocation, passing a `--path` flag **the fest CLI has never accepted** — `fest init` takes a positional path (`Use: "init [path]"`). It failed on every run, and was non-fatal by design, so nothing surfaced.

`camp init` was unaffected *in outcome* because it calls fest a second time via `initializeFestivals`, which uses the positional form and succeeds. That second call is also why this survived six months (the `--path` line dates to `617916bd`, 2026-01-21): the only symptom on the common path was a stderr warning immediately contradicted by `Festivals: initialized` on stdout, which reads as noise.

## Fix

Give `register` the same call `init` makes, and drop scaffold's copy entirely:

```
scaffold.Init          → dirs, files, symlinks. No fest.
cmd/camp/init          → InitializeFestivals()   ← already there
cmd/camp/register (y)  → InitializeFestivals()   ← added
```

Scaffolding owns the campaign's own files; running the fest CLI is a command-level concern. Each command that initializes a campaign now makes exactly one call to one implementation.

Fixing the flag in place would also have worked, but leaves two copies of "shell out to fest init" — which is *why* one drifted unnoticed — and would leave `camp init` calling fest twice, the second call reporting *"Festival Methodology already initialized"* on a brand-new campaign.

`register` also picks up what the duplicate never had: the **30s timeout** in `fest.RunInit` (a wedged fest could previously hang the command indefinitely), plus the install/manual-init guidance `init` already showed.

After:

```
Would you like to initialize one? [y/N] Initializing Festival Methodology...
Festival Methodology initialized
✓ Initialized and registered campaign at /tmp/probe/reg
festivals/: PRESENT (9 entries)
```

`camp init`'s stderr is now empty, and the bogus `festivals/ (fest init failed - run manually)` skip entry is gone.

## Why it wasn't caught, and the harness bug that hid it

Nothing in the suite executed camp's fest invocation. A wrong argv inside `exec.Command` is invisible to unit tests and to `go vet`.

Adding integration coverage surfaced a **second problem**: the container's fest binary was built from `projectRoot/../fest`, which resolves to `projects/fest` only when camp sits at `projects/camp`. Camp is routinely developed from `projects/worktrees/camp/<branch>`, where that expression lands on `projects/worktrees/camp/fest` and misses. `festAvailable` went false and **all 9 fest-gated tests skipped** — the suite still reported `ok`, so a green integration run from a worktree carried no signal for that coverage at all. First commit walks the ancestors instead (works from either layout) and adds `CAMP_TEST_FEST_SRC`.

Full integration suite passes with the un-skipped tests now actually running (423s), so nothing was hiding behind them.

## Verification

Both new tests assert on **outcome, not argv**, so they keep holding regardless of which path owns the call. Confirmed they actually catch the bug:

```
# against the unfixed binary
--- FAIL: TestRegister_InitializesFestivals
    festivals/ should carry a fest marker (.festival, fest.yaml, .fest)
--- FAIL: TestInit_NoFestWarningOnStderr
    "Warning: fest init failed ... unknown flag: --path" should not contain "fest init failed"

# against the fixed binary
--- PASS: TestRegister_InitializesFestivals (3.36s)
--- PASS: TestInit_NoFestWarningOnStderr (3.15s)
```

`just gate`: 3614/3614 pass, lint clean. Full integration suite: pass.

## One caveat on how far this protects

`just gate` only *vets* the integration build tag — it does not run those tests (`just test` does), and GitHub CI is paused. So this catches the class **on demand, not automatically**. Making it automatic means having the gate run integration, which is a separate call with a real runtime cost (~7 min).

Also worth knowing: these tests shell out to a binary built at run time, so Go's test cache can't see source changes behind them. `-count=1` is required when using them to check a behavior change — without it I got a cached PASS against deliberately broken source.

### DIFF
--- a/cmd/camp/register.go
+++ b/cmd/camp/register.go
@@ -2,13 +2,16 @@ package main
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
 
+	initcmd "github.com/Obedience-Corp/camp/cmd/camp/init"
 	"github.com/Obedience-Corp/camp/internal/config"
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/fest"
 	"github.com/Obedience-Corp/camp/internal/scaffold"
 	"github.com/Obedience-Corp/camp/internal/ui"
 	"github.com/spf13/cobra"
@@ -84,6 +87,15 @@ func runRegister(cmd *cobra.Command, args []string) error {
 			result, err := scaffold.Init(ctx, absPath, opts)
 			if err != nil {
 				return err
+			}
+			// Scaffolding alone leaves a campaign without festivals/, which is
+			// the same campaign `camp init` would have set up completely. A
+			// missing fest CLI is reported by InitializeFestivals and is not
+			// fatal, matching how init treats it.
+			if _, festErr := initcmd.InitializeFestivals(ctx, result.CampaignRoot, initcmd.Writers{HumanOut: os.Stdout}); festErr != nil {
+				if !errors.Is(festErr, fest.ErrFestNotFound) {
+					return festErr
+				}
 			}
 			fmt.Printf("%s Initialized and registered campaign at %s\n", ui.SuccessIcon(), ui.Value(result.CampaignRoot))
 			return nil
--- a/cmd/camp/init/command.go
+++ b/cmd/camp/init/command.go
@@ -312,7 +312,7 @@ func RunFlow(ctx context.Context, p Params, w Writers, isInteractive bool) error
 	var festInitialized bool
 	if !p.DryRun {
 		var festErr error
-		festInitialized, festErr = initializeFestivals(ctx, result.CampaignRoot, w)
+		festInitialized, festErr = InitializeFestivals(ctx, result.CampaignRoot, w)
 		if festErr != nil && !errors.Is(festErr, fest.ErrFestNotFound) {
 			return festErr
 		}
--- a/cmd/camp/init/fest.go
+++ b/cmd/camp/init/fest.go
@@ -10,9 +10,15 @@ import (
 	"github.com/Obedience-Corp/camp/internal/ui"
 )
 
-// initializeFestivals runs fest init in the campaign directory.
+// InitializeFestivals runs fest init in the campaign directory.
 // Returns true if successful, false with guidance if fest is unavailable.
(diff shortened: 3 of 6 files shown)


cmd/camp/register.go
```diff
@@ -85,6 +88,15 @@ func runRegister(cmd *cobra.Command, args []string) error {
 			if err != nil {
 				return err
 			}
+			// Scaffolding alone leaves a campaign without festivals/, which is
+			// the same campaign `camp init` would have set up completely. A
+			// missing fest CLI is reported by InitializeFestivals and is not
+			// fatal, matching how init treats it.
+			if _, festErr := initcmd.InitializeFestivals(ctx, result.CampaignRoot, initcmd.Writers{HumanOut: os.Stdout}); festErr != nil {
```
obey-agent, line 95: InitializeFestivals is wired correctly here — good fix. Prefer cmd.OutOrStdout()/ErrOrStderr() (or the Writers already on the command path) over bare os.Stdout so tests and redirected invocations capture festival init output. Minor, not a correctness issue for the festivals/ presence bug.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Exemplary root-cause writeup and fix. `scaffold.Init` owned a broken `fest init --path` that never worked; `register` swallowed it and lied about success. Moving festival init to a single command-level `InitializeFestivals` (shared with `camp init`), plus integration tests that assert **outcome** (festivals/ + markers), is exactly how this class of dual-path drift should be fixed. Bonus: fest source location now works from camp worktrees.

## Key Findings
- Minor: `register` passes `os.Stdout` into Writers instead of cmd writers (inline) — polish for testability.
- Caveat in the PR body about integration not being in `just gate` is honest and important operational knowledge.

## What's Done Well
- Single ownership of fest init (no second copy to drift).
- Timeout + install guidance inherited from init’s path.
- Integration tests fail on the old binary and pass on the new (evidence in the PR).
- `locateFestSource` fixes silent skip of fest-gated tests from worktrees.

## Staff Standard
Yes — this is staff-level incident remediation. Merge when green.

PR #473, opened 2026-07-20
fix(remote): surface Tailscale SSH check URLs instead of bare timeouts

lancekrogers:
## Summary

- Preserve SSH stderr when a remote hop times out so Tailscale check-mode URLs are not discarded.
- Detect Tailscale SSH check-mode failures (`Tailscale SSH requires an additional check` + `login.tailscale.com` URL) and return actionable errors while still wrapping `context.DeadlineExceeded`.
- Surface that guidance in `camp machine` health (detail + “approve in browser, then t”) instead of a generic “unreachable / timed out” message.

Camp still correctly uses `BatchMode=yes` for agent/script hops and cannot complete browser approval itself. This fixes the diagnosis path so operators see what to do next.

## Test plan

- [x] `go test ./internal/remote/ ./cmd/camp/`
- [ ] Manually: with Tailscale SSH check pending, run `camp machine` and press `t` on a `tailscale-ssh` machine — pane should show the `login.tailscale.com` URL, not only “timed out”
- [ ] Approve the check in a browser, press `t` again — machine should report reachable
- [ ] Confirm plain timeout/network failures still report useful stderr (non-Tailscale)

### DIFF
--- a/cmd/camp/machine_tui_view.go
+++ b/cmd/camp/machine_tui_view.go
@@ -255,11 +255,22 @@ func (m *machineTUIModel) healthSection(id string, width int) []string {
 	case healthTesting:
 		return []string{style.Render(m.spin.View() + " Testing the connection...")}
 	case healthUnreachable:
-		return []string{
-			style.Render(glyph + " Could not reach it"),
-			machineMuted.Render("  " + ui.Truncate(health.Detail, max(width-4, 20))),
-			machineMuted.Render("  e edits it · t tries again"),
+		// Tailscale check URLs are long; prefer wrapping the actionable detail
+		// over hard-truncating the only thing the operator needs to open.
+		detail := health.Detail
+		if !strings.Contains(detail, "login.tailscale.com") {
+			detail = ui.Truncate(detail, max(width-4, 20))
 		}
+		lines := []string{
+			style.Render(glyph + " Could not reach it"),
+			machineMuted.Render("  " + detail),
+		}
+		if strings.Contains(health.Detail, "login.tailscale.com") {
+			lines = append(lines, machineMuted.Render("  Approve in the browser, then press t to try again."))
+		} else {
+			lines = append(lines, machineMuted.Render("  e edits it · t tries again"))
+		}
+		return lines
 	case healthUnsupported:
 		return []string{
 			style.Render(glyph + " Cannot be used yet"),
--- a/cmd/camp/machine_tui_view.go
+++ b/cmd/camp/machine_tui_view.go
@@ -255,17 +255,21 @@ func (m *machineTUIModel) healthSection(id string, width int) []string {
 	case healthTesting:
 		return []string{style.Render(m.spin.View() + " Testing the connection...")}
 	case healthUnreachable:
-		// Tailscale check URLs are long; prefer wrapping the actionable detail
-		// over hard-truncating the only thing the operator needs to open.
-		detail := health.Detail
-		if !strings.Contains(detail, "login.tailscale.com") {
-			detail = ui.Truncate(detail, max(width-4, 20))
+		tailscaleCheck := strings.Contains(health.Detail, "login.tailscale.com")
+		headline := "Could not reach it"
+		if tailscaleCheck {
+			// Check-mode is auth policy, not network failure — do not frame it
+			// as unreachable or operators chase connectivity.
+			headline = "Needs Tailscale SSH check"
 		}
-		lines := []string{
-			style.Render(glyph + " Could not reach it"),
-			machineMuted.Render("  " + detail),
+		lines := []string{style.Render(glyph + " " + headline)}
+		// Width-aware detail: wrap long check URLs so the actionable token
+		// stays visible in a narrow pane instead of clipping off-screen.
+		detailWidth := max(width-4, 20)
+		for _, line := range healthDetailLines(health.Detail, detailWidth, tailscaleCheck) {
+			lines = append(lines, machineMuted.Render("  "+line))
 		}
-		if strings.Contains(health.Detail, "login.tailscale.com") {
+		if tailscaleCheck {
 			lines = append(lines, machineMuted.Render("  Approve in the browser, then press t to try again."))
 		} else {
 			lines = append(lines, machineMuted.Render("  e edits it · t tries again"))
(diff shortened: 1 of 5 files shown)


cmd/camp/machine_tui_view.go
```diff
@@ -255,11 +255,22 @@ func (m *machineTUIModel) healthSection(id string, width int) []string {
...
 	case healthUnreachable:
-		return []string{
+		// Tailscale check URLs are long; prefer wrapping the actionable detail
+		// over hard-truncating the only thing the operator needs to open.
+		detail := health.Detail
+		if !strings.Contains(detail, "login.tailscale.com") {
+			detail = ui.Truncate(detail, max(width-4, 20))
+		}
+		lines := []string{
 			style.Render(glyph + " Could not reach it"),
```
obey-agent: Headline still says "Could not reach it" for Tailscale check-mode, which is network-unreachable framing. When Detail contains login.tailscale.com, prefer something like "Needs Tailscale SSH check" so the operator does not chase connectivity. The detail URL is good; the title fights it.

internal/remote/ssh.go
```diff
@@ -215,19 +217,140 @@ func Run(ctx context.Context, target string, opts []string, remoteCmd string) ([
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	if err := cmd.Run(); err != nil {
+		trimmed := strings.TrimSpace(stderr.String())
 		if ctx.Err() != nil {
-			return nil, camperrors.Wrapf(ctx.Err(), "ssh to %s timed out", target)
+			return nil, sshTimeoutError(target, trimmed, ctx.Err())
```
obey-agent, line 222: Good: stderr is preserved on the timeout path (this was the actual bug). Consider also routing list --remote / switch failures through TailscaleCheckDetail so those surfaces get the same one-line guidance without relying on the full wrapped Error() string alone. Not blocking if %v already prints the wrap.

obey-agent (commented):
## Verdict: **Approve** (with one UX polish)

## Overview
Correct diagnosis and a focused fix. Camp was discarding SSH stderr on timeout, which hid Tailscale check-mode URLs and made a policy/auth step look like network failure. Preserving stderr, classifying the marker+URL, and surfacing it in machine health is the right product boundary (BatchMode stays non-interactive).

## Key Findings
- **UX:** Health headline still says “Could not reach it” for check-mode failures — detail is good, title fights it (inline on `machine_tui_view.go`).
- **Coverage:** Unit tests for parse/timeout/exit/detail paths are solid; manual Tailscale check still worth a smoke before merge.

## What's Done Well
- Timeout path no longer drops stderr (root bug).
- `errors.Is(..., context.DeadlineExceeded)` preserved via wrap.
- TUI keeps long check URLs instead of hard-truncating the only actionable token.

## Staff Standard
Yes — I would ship this. Small headline polish preferred but not blocking.

cmd/camp/machine_tui_view.go
```diff
@@ -255,11 +255,22 @@ func (m *machineTUIModel) healthSection(id string, width int) []string {
 	case healthTesting:
 		return []string{style.Render(m.spin.View() + " Testing the connection...")}
 	case healthUnreachable:
-		return []string{
+		// Tailscale check URLs are long; prefer wrapping the actionable detail
+		// over hard-truncating the only thing the operator needs to open.
+		detail := health.Detail
```
obey-agent: The comment says these URLs should be wrapped, but this branch skips `ui.Truncate` and passes the full URL as one styled line. In a narrow detail pane, a long unbreakable URL can still overflow/clip, hiding the actionable token that this PR is meant to surface. Render a width-aware wrapped URL (or split the message into multiple lines while preserving the full URL) and cover a narrow-width pane in the PTY/UI check.

obey-agent (commented):
## Verdict: Request Changes

## Overview

This is a focused diagnosis of a real operator problem, and the remote tests plus build path pass. Preserving SSH stderr and retaining errors.Is behavior are the right boundaries. The actionable URL still is not reliably usable in the narrow TUI layout.

## Key Findings

- The new health rendering bypasses truncation for Tailscale URLs but does not actually guarantee wrapping; a long unbreakable URL can overflow or clip in the detail pane. This is the new inline finding.
- The existing inline finding remains open: the headline still says “Could not reach it” for a browser-auth check, which contradicts the actionable detail and can send the operator toward network debugging.
- The manual Tailscale approval flow is still unverified.

## What's Done Well

- Timeout stderr is preserved instead of discarded.
- The Tailscale marker and URL are classified in one place.
- Context deadline matching remains intact through wrapping.
- Unit coverage covers parse, timeout, exit, and generic-stderr paths.

## Staff Standard

Almost. I would ship this after the narrow-pane rendering and check-mode wording are corrected and the real approval flow is smoke-tested.

lancekrogers:
## Ready for re-review

Addressed UX polish / request-changes items:

| Feedback | Fix |
|----------|-----|
| “Could not reach it” for check-mode | Headline is now **Needs Tailscale SSH check** when detail has `login.tailscale.com` |
| Long URL overflow in narrow pane | `healthDetailLines` **width-wraps** check URLs (prefer break after `/`) instead of skipping truncate without wrap |
| Tests | `TestHealthDetailLinesWrapsTailscaleURL`, `TestHealthSectionTailscaleCheckHeadline` |

Stderr-on-timeout + TailscaleCheckDetail classification unchanged.

cmd/camp/machine_tui_view.go
```diff
@@ -274,6 +289,36 @@ func (m *machineTUIModel) healthSection(id string, width int) []string {
...
+	if detail == "" {
+		return nil
+	}
+	if maxWidth < 8 {
+		maxWidth = 8
+	}
+	if !keepFullURL {
+		return []string{ui.Truncate(detail, maxWidth)}
+	}
+	// Prefer breaking after path separators so "https://…/a/…" remains readable.
```
obey-agent, line 311: Non-blocking correctness note: maxWidth here is a terminal-column budget, but the loop uses byte len and byte slicing. The detail string emitted by this PR contains a Unicode em dash, and other SSH text may contain non-ASCII or wide runes; a boundary can split a multi-byte rune into invalid UTF-8 or exceed the visual pane width. A rune or display-width-aware wrapper plus a non-ASCII regression case would make this robust.

obey-agent (commented):
Verdict: Approve

Overview: Reviewed the full remote SSH and machine TUI diff. go test ./internal/remote ./cmd/camp passes, and the change correctly preserves stderr across timeout and exit paths so Tailscale browser-check URLs remain actionable.

Key findings: no blocking correctness findings. I left one non-blocking inline note about byte-based wrapping in healthDetailLines and Unicode display width.

What was done well: the error-shape distinction is explicit, Tailscale parsing is narrowly gated by its marker, the UI explains the required operator action, and the tests cover both timeout and exit behavior.

Would I be proud to ship this? Yes; the wrapping note is suitable for a follow-up hardening change.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Re-reviewed tip `1071f407`. Timeout path preserves stderr; Tailscale check-mode is classified and shown with a correct headline; long check URLs wrap by **display width** on rune boundaries so multi-byte text cannot corrupt or overflow the pane.

## Key Findings
No blocking issues remaining. Non-blocking prior notes (list --remote wording) are acceptable because `%v` on the wrapped error already surfaces the check URL.

## What's Done Well
Narrow, well-tested classification; operator guidance is actionable without weakening BatchMode.

## Staff Standard
Yes — ready to merge.

lancekrogers:
## Ready for re-review

Addressed non-blocking display-width note:

- `healthDetailLines` now wraps by **lipgloss display width** on **rune boundaries** (no byte slicing through multi-byte runes such as the em dash)
- Tests assert valid UTF-8 and `lipgloss.Width(line) <= maxWidth`

PR #474, opened 2026-07-20
feat(skills): project campaign skills into project worktrees

lancekrogers:
## Summary

- Project campaign skill bundles into each project worktree (`.agents/skills`, `.claude/skills`, and `.grok/skills → .agents/skills`) so Grok/Claude sessions whose git root is the worktree still discover `.campaign/skills`.
- Run projection automatically on `camp project worktree add` (and the deprecated `camp worktrees create`), best-effort so create still succeeds if projection fails.
- Add `camp skills link --worktrees` (tools + all worktrees) and `--worktrees-only` (worktrees only) for repair/sync.
- Extend `camp skills status` with a `worktree` row per git worktree under `projects/worktrees/`.
- Enumerate only directories that look like git checkouts (have `.git`), so nested `src`/`bin`/`node_modules` are not treated as worktrees.
- Gitignore local `.agents/` and `.grok/` harness dirs in the camp project checkout.

## Motivation

Grok skill discovery walks CWD → **git root**. Nested worktrees under `projects/worktrees/<project>/<name>/` are their own git roots, so campaign skills at the campaign root never load. Projection mirrors the existing per-bundle symlink model into each worktree.

## Test plan

- [x] `go test ./internal/skills/`
- [x] `go build ./cmd/camp`
- [x] Dry-run: `camp skills link --worktrees-only -n` lists real git worktrees only
- [ ] `camp project worktree add smoke-skills --start-point main` shows “Skills: projected…”
- [ ] From that worktree: `grok inspect` lists campaign skills as `project`
- [ ] `camp skills link --worktrees-only` repairs existing worktrees
- [ ] `camp skills status` shows linked worktree rows

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,5 +37,8 @@ __debug_bin*
 .envrc
 CLAUDE.md
 .claude/
+# Local agent harness skill projections (worktree projection / camp skills link)
+.agents/
+.grok/
 dist/
 projects/worktrees/
--- a/cmd/camp/project/worktree/add.go
+++ b/cmd/camp/project/worktree/add.go
@@ -177,10 +177,11 @@ func runProjectWorktreeAdd(cmd *cobra.Command, args []string) error {
 	// Project campaign skills into the worktree so Grok/Claude sessions whose
 	// git root is the worktree still discover .campaign/skills. Failure here
 	// must not undo a successful worktree create.
-	if err := intskills.ProjectIntoWorktreeBestEffort(campRoot, result.Path); err != nil {
+	projected, err := intskills.ProjectIntoWorktreeBestEffort(campRoot, result.Path)
+	if err != nil {
 		fmt.Println(ui.Warning(fmt.Sprintf("  Skills: could not project into worktree: %v", err)))
 		fmt.Println(ui.Dim("  Fix later with: camp skills link --worktrees-only"))
-	} else {
+	} else if projected {
 		fmt.Println(ui.Dim("  Skills: projected campaign skill bundles into worktree (.agents/.claude/.grok)"))
 	}
 
--- a/cmd/camp/project/worktree/add.go
+++ b/cmd/camp/project/worktree/add.go
@@ -12,6 +12,7 @@ import (
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 	"github.com/Obedience-Corp/camp/internal/paths"
 	"github.com/Obedience-Corp/camp/internal/project"
+	intskills "github.com/Obedience-Corp/camp/internal/skills"
 	"github.com/Obedience-Corp/camp/internal/ui"
 	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
 	"github.com/Obedience-Corp/camp/internal/workitem/links"
@@ -171,6 +172,16 @@ func runProjectWorktreeAdd(cmd *cobra.Command, args []string) error {
 		}
 		fmt.Printf("  Workitem: %s (%s)\n", ui.Value(link.WorkitemID), ui.Dim(link.WorkitemKey))
 		fmt.Println(ui.Dim("  camp p commit in this worktree will include WI-* in the campaign tag"))
+	}
+
+	// Project campaign skills into the worktree so Grok/Claude sessions whose
+	// git root is the worktree still discover .campaign/skills. Failure here
+	// must not undo a successful worktree create.
+	if err := intskills.ProjectIntoWorktreeBestEffort(campRoot, result.Path); err != nil {
+		fmt.Println(ui.Warning(fmt.Sprintf("  Skills: could not project into worktree: %v", err)))
+		fmt.Println(ui.Dim("  Fix later with: camp skills link --worktrees-only"))
+	} else {
+		fmt.Println(ui.Dim("  Skills: projected campaign skill bundles into worktree (.agents/.claude/.grok)"))
 	}
 
 	fmt.Println()
--- a/cmd/camp/project/worktree/add.go
+++ b/cmd/camp/project/worktree/add.go
@@ -177,7 +177,7 @@ func runProjectWorktreeAdd(cmd *cobra.Command, args []string) error {
 	// Project campaign skills into the worktree so Grok/Claude sessions whose
 	// git root is the worktree still discover .campaign/skills. Failure here
 	// must not undo a successful worktree create.
-	projected, err := intskills.ProjectIntoWorktreeBestEffort(campRoot, result.Path)
+	projected, err := intskills.ProjectIntoWorktreeBestEffort(ctx, campRoot, result.Path)
 	if err != nil {
 		fmt.Println(ui.Warning(fmt.Sprintf("  Skills: could not project into worktree: %v", err)))
 		fmt.Println(ui.Dim("  Fix later with: camp skills link --worktrees-only"))
--- a/cmd/camp/skills/worktree_project.go
+++ b/cmd/camp/skills/worktree_project.go
@@ -1,0 +1,67 @@
+package skills
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/paths"
+	intskills "github.com/Obedience-Corp/camp/internal/skills"
+)
+
+// linkAllWorktrees projects skill bundles into every projects/worktrees/*/*
+// checkout under the campaign.
+func linkAllWorktrees(out, errOut io.Writer, root, skillsDir string, force, dryRun bool) error {
+	cfg, err := config.LoadCampaignConfig(context.Background(), root)
+	if err != nil {
+		// Paths defaults still work when campaign config is missing/unloadable.
+		cfg = &config.CampaignConfig{}
+	}
+	resolver := paths.NewResolver(root, cfg.Paths())
+	wtRoot := resolver.Worktrees()
+
+	results, err := intskills.LinkAllWorktrees(root, wtRoot, skillsDir, dryRun, force, errOut)
+	if err != nil {
+		return err
+	}
+	if len(results) == 0 {
+		if _, err := fmt.Fprintf(out, "no project worktrees under %s\n", wtRoot); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	verb := "projected"
+	if dryRun {
+		verb = "would project"
+	}
+	var conflicts int
+	var conflictNames []string
+	var hardErrs int
+	for _, res := range results {
+		if res.Err != nil {
+			hardErrs++
+			if _, err := fmt.Fprintf(errOut, "worktree %s: %v\n", res.RelPath, res.Err); err != nil {
+				return err
+			}
+			continue
+		}
+		s := res.Agents
+		if _, err := fmt.Fprintf(out, "%s %d skill bundle(s) into worktree %s (created=%d replaced=%d unchanged=%d)\n",
+			verb, s.Created+s.Replaced+s.AlreadyLinked, res.RelPath, s.Created, s.Replaced, s.AlreadyLinked); err != nil {
+			return err
+		}
+		conflicts += s.Conflicts + res.Claude.Conflicts
+		conflictNames = append(conflictNames, s.ConflictNames...)
+		conflictNames = append(conflictNames, res.Claude.ConflictNames...)
+	}
+	if hardErrs > 0 {
+		return camperrors.Newf("worktree skill projection failed for %d worktree(s)", hardErrs)
+	}
+	if conflicts > 0 {
+		return camperrors.Newf("worktree projection incomplete: %d conflicting skill path(s): %v", conflicts, conflictNames)
+	}
+	return nil
+}
--- a/cmd/camp/skills/worktree_project.go
+++ b/cmd/camp/skills/worktree_project.go
@@ -26,7 +26,7 @@ func linkAllWorktrees(ctx context.Context, out, errOut io.Writer, root, skillsDi
 	resolver := paths.NewResolver(root, cfg.Paths())
 	wtRoot := resolver.Worktrees()
 
-	results, err := intskills.LinkAllWorktrees(root, wtRoot, skillsDir, dryRun, force, errOut)
+	results, err := intskills.LinkAllWorktrees(ctx, root, wtRoot, skillsDir, dryRun, force, errOut)
 	if err != nil {
 		return err
 	}
--- a/cmd/camp/skills/worktree_project.go
+++ b/cmd/camp/skills/worktree_project.go
@@ -12,9 +12,13 @@ import (
 )
 
 // linkAllWorktrees projects skill bundles into every projects/worktrees/*/*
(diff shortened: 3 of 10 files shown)


cmd/camp/project/worktree/add.go
```diff
@@ -173,6 +174,16 @@ func runProjectWorktreeAdd(cmd *cobra.Command, args []string) error {
...
 	}
 
+	// Project campaign skills into the worktree so Grok/Claude sessions whose
+	// git root is the worktree still discover .campaign/skills. Failure here
+	// must not undo a successful worktree create.
+	if err := intskills.ProjectIntoWorktreeBestEffort(campRoot, result.Path); err != nil {
+		fmt.Println(ui.Warning(fmt.Sprintf("  Skills: could not project into worktree: %v", err)))
+		fmt.Println(ui.Dim("  Fix later with: camp skills link --worktrees-only"))
+	} else {
+		fmt.Println(ui.Dim("  Skills: projected campaign skill bundles into worktree (.agents/.claude/.grok)"))
```
obey-agent, line 185: ProjectIntoWorktreeBestEffort returns nil when .campaign/skills is missing or empty, but this branch always prints "Skills: projected…". That is a false success. Either have BestEffort return (projected bool, err) or only print the success line when at least one link was created/already-linked.

cmd/camp/skills/worktree_project.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/paths"
+	intskills "github.com/Obedience-Corp/camp/internal/skills"
+)
+
+// linkAllWorktrees projects skill bundles into every projects/worktrees/*/*
+// checkout under the campaign.
+func linkAllWorktrees(out, errOut io.Writer, root, skillsDir string, force, dryRun bool) error {
+	cfg, err := config.LoadCampaignConfig(context.Background(), root)
```
obey-agent: Uses context.Background() instead of the command context. LoadCampaignConfig and bulk worktree walks should take the caller ctx so cancel/timeout from the CLI propagates (and tests can inject deadlines). Same pattern appears in status.go worktreeSkillStatusEntries.

internal/skills/worktree.go
```diff
@@ -0,0 +1,262 @@
...
+
+	var roots []string
+	for _, projectEntry := range entries {
+		if !projectEntry.IsDir() || strings.HasPrefix(projectEntry.Name(), ".") {
+			continue
+		}
+		projectDir := filepath.Join(worktreesRoot, projectEntry.Name())
+		// Some checkouts land directly under worktreesRoot/<name> (no project
+		// nesting). Accept those when they are git roots.
+		if isGitCheckoutRoot(projectDir) {
```
obey-agent, line 86: When projectDir itself is a git root you continue without scanning children. That correctly handles mis-nested checkouts (fest-gif/*), but it also means a real layout of worktreesRoot/<project-as-repo>/<nested-wt> would be invisible. Worth a one-line comment in the public docs/help that only projects/worktrees/<project>/<name> git roots (or a lone git root at the project level) are projected.

obey-agent (commented):
## Verdict: **Request changes** (small correctness/UX, then ship)

## Overview
Right architectural answer to “skills disappear in worktrees”: reuse per-bundle projection into each worktree git root, auto on `worktree add`, bulk repair via `--worktrees` / `--worktrees-only`. Git-root filtering (`.git` present) correctly avoids projecting into `src`/`node_modules`. This is the product fix we wanted.

## Key Findings
1. **False success message** on `worktree add`: BestEffort no-ops when skills are missing/empty but still prints “Skills: projected…” (inline on `add.go`).
2. **`context.Background()`** in worktree link/status config load — should use command context (inline on `worktree_project.go`; same in status).
3. **Enumeration edge case:** if `worktreesRoot/<project>` is itself a git root, nested children are skipped — document that layout assumption (inline on `ListWorktreeRoots`).
4. No `unlink --worktrees` yet — acceptable for v1 if status+link repair is enough; call out in docs.

## What's Done Well
- Extends existing `ProjectSkillEntries` instead of a parallel linker.
- Best-effort on create (does not roll back worktree).
- Status rows for operators; unit tests for list/project/idempotency/grok alias.
- `.agents/` / `.grok/` gitignore for harness noise.

## Staff Standard
Almost — fix the false “projected” success and pass real `ctx` through config load, then I’d be proud to merge.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Exemplary root-cause writeup and fix. `scaffold.Init` owned a broken `fest init --path` that never worked; `register` swallowed it and lied about success. Moving festival init to a single command-level `InitializeFestivals` (shared with `camp init`), plus integration tests that assert **outcome** (festivals/ + markers), is exactly how this class of dual-path drift should be fixed. Bonus: fest source location now works from camp worktrees.

## Key Findings
- Minor: `register` passes `os.Stdout` into Writers instead of cmd writers (inline) — polish for testability.
- Caveat in the PR body about integration not being in `just gate` is honest and important operational knowledge.

## What's Done Well
- Single ownership of fest init (no second copy to drift).
- Timeout + install guidance inherited from init’s path.
- Integration tests fail on the old binary and pass on the new (evidence in PR).
- `locateFestSource` fixes silent skip of fest-gated tests from worktrees.

## Staff Standard
Yes — this is staff-level incident remediation. Merge when green.

obey-agent (commented):
**Correction:** An earlier review body on this PR accidentally pasted the **#470** summary (“Approve / fest init”). Please ignore that review.

**Authoritative verdict for #474 remains Request changes** (false “Skills: projected…” success + `context.Background()` in worktree link/status). Inline comments above still apply.

lancekrogers:
Addressed request-changes review:

1. **False success** — `ProjectIntoWorktreeBestEffort` now returns `(projected bool, err)`. Worktree add/create only print “Skills: projected…” when `projected` is true (at least one agents skill link created/already-linked). Missing/empty skills is a silent no-op.
2. **Context** — `linkAllWorktrees` and `worktreeSkillStatusEntries` take `ctx` and use it for `LoadCampaignConfig` (plus early `ctx.Err()` check).
3. **Enumeration docs** — `ListWorktreeRoots`, `camp skills link` Long, and `camp skills` root help document that only git checkouts are projected and that a loose git root at `projects/worktrees/<name>/` does not scan nested children.

Tests: `go test ./internal/skills/` (includes BestEffort cases).

internal/skills/worktree.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+		raw, readErr := os.Readlink(linkPath)
+		if readErr != nil {
+			return camperrors.Wrap(readErr, "read grok skills alias")
+		}
+		gotAbs := resolveSymlinkTargetAbs(linkPath, raw)
+		if filepath.Clean(gotAbs) == wantAbs {
+			return nil
+		}
+		// Wrong target: replace when not dry-run.
```
obey-agent: `ProjectSkillEntries` treats `force` as the safety switch for replacing foreign symlinks, but this alias path unconditionally removes any non-matching `.grok/skills` symlink. A user can already have that path linked to their own skill directory; `camp skills link --worktrees-only` would silently delete it without `-f`. Thread `force` through `EnsureGrokSkillsAlias` and report a conflict unless the caller explicitly forces replacement, with coverage for both paths.

internal/skills/worktree.go
```diff
@@ -0,0 +1,275 @@
...
+func EnsureGrokSkillsAlias(worktreeRoot string, dryRun bool) error {
+	linkPath := filepath.Join(worktreeRoot, GrokSkillsRel)
+	// Target relative to the symlink's parent (.grok/): ../.agents/skills
+	const relTarget = "../.agents/skills"
+	wantAbs := filepath.Clean(filepath.Join(filepath.Dir(linkPath), relTarget))
+
+	info, err := os.Lstat(linkPath)
+	if err == nil {
+		if info.Mode()&os.ModeSymlink == 0 {
+			if info.IsDir() {
```
obey-agent: If `.grok/skills` is an existing real directory, this returns success without projecting the campaign bundles into it. Grok follows that directory, so an empty or unrelated directory makes the promised worktree discovery false even though `.agents/skills` is populated. Merge managed per-bundle links into the directory or surface it as an explicit conflict/status state; please add a regression test.

internal/skills/worktree_test.go
```diff
@@ -0,0 +1,298 @@
...
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		t.Fatalf("mkdir skill: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(dir, "SKILL.md"), []byte("---\nname: "+slug+"\n---\n"), 0o644); err != nil {
+		t.Fatalf("write SKILL.md: %v", err)
+	}
+}
+
+func TestListWorktreeRoots(t *testing.T) {
+	tmp := t.TempDir()
```
obey-agent: These new tests create and mutate real directory trees with `t.TempDir()`, including fake `.git` checkouts. The repository's filesystem-safety guidance requires mutation tests to run in the containerized filesystem harness rather than ad hoc host temp directories. Please port this coverage to the project test harness before merging so the tests exercise the supported isolation boundary.

obey-agent (commented):
## Verdict: Request Changes

## Overview

The worktree projection architecture is strong and the author addressed the prior false-success, context-loading, and enumeration-documentation findings. The current branch passes go test ./internal/skills/, go build ./cmd/camp, and the diff check. Two filesystem safety cases remain before this is production-ready.

## Key Findings

- A non-matching .grok/skills symlink is deleted and replaced without requiring --force, unlike the existing per-bundle safety model.
- An existing real .grok/skills directory is treated as success without receiving the campaign links, so Grok discovery can still fail while status reports the worktree as linked.
- The new filesystem-mutating tests use host t.TempDir() trees; the repository guidance requires the containerized filesystem harness for this class of test.

## What's Done Well

- Reuses ProjectSkillEntries instead of creating a parallel projection implementation.
- Best-effort projection preserves successful worktree creation.
- Git-root filtering, repair commands, and status rows are clearly documented.
- The follow-up commit correctly removed the false “projected” message and threaded command context into config loads.

## Staff Standard

Not yet. The architecture is staff-level, but the alias safety and existing-directory behavior need to be resolved before I would merge it.

lancekrogers:
## Ready for re-review

Addressed remaining request-changes items:

| Feedback | Fix |
|----------|-----|
| Foreign `.grok/skills` symlink replaced without force | `ensureWorktreeGrokSkills` reports **conflict** unless `--force` / `force=true` |
| Real `.grok/skills` directory left empty | Projects managed per-bundle links **into** the directory |
| Host FS unit tests only | Added container integration test `TestSkills_LinkWorktreesOnlyProjectsGitWorktree` |
| Prior false-success + context | Still in place from previous push |

Also still covered: git-root discovery docs, BestEffort `projected` bool, cmd `ctx` for config load.

.gitignore
```diff
@@ -37,5 +37,8 @@ __debug_bin*
 .envrc
 CLAUDE.md
 .claude/
+# Local agent harness skill projections (worktree projection / camp skills link)
+.agents/
```
obey-agent, line 41: This ignore rule only applies to the Camp checkout. ProjectIntoWorktree writes .agents/, .claude/, and .grok/ into arbitrary project worktrees such as fest, festival, and obey, whose gitignore files do not inherit this rule. Those managed symlinks will appear as untracked files and can be accidentally committed. Please make the projection install equivalent rules in the target repository local Git exclude, or otherwise ensure the target repo ignores all three managed paths, and cover git status after projection with a test.

internal/skills/worktree.go
```diff
@@ -0,0 +1,325 @@
...
+	if err != nil {
+		return false, err
+	}
+	n := proj.Agents.Created + proj.Agents.Replaced + proj.Agents.AlreadyLinked
+	return n > 0, nil
+}
+
+// InspectWorktreeProjection reports projection state for a worktree's
+// .agents/skills directory (primary harness path).
+func InspectWorktreeProjection(worktreeRoot, skillsDir string, slugs []string) (ProjectionState, error) {
```
obey-agent, line 354: Status only inspects .agents/skills, although projection manages three surfaces and can independently encounter a missing or broken .claude/skills directory or a wrong or foreign .grok/skills alias. In those cases this function can still report linked and the strict check can exit successfully while one supported harness is unusable. Please aggregate all managed destinations, or explicitly report primary versus alias health and include every surface in strict attention.

internal/skills/worktree_test.go
```diff
@@ -0,0 +1,372 @@
...
```
obey-agent: The repository instructions require filesystem-mutating workflow tests to run in container isolation, but this new unit suite uses t.TempDir plus os.MkdirAll, os.WriteFile, and os.Symlink to exercise real worktree projection on the host. Please move these cases into the containerized integration harness, or make the unit layer pure and keep real filesystem coverage there, before merging.

obey-agent (commented):
Verdict: Request Changes

Overview: Reviewed the full worktree skill-projection diff. The focused Go tests and the skills integration test pass, but three issues need attention before merge.

Key findings:
- The Camp .gitignore does not apply to the target project worktrees where .agents, .claude, and .grok are created, so projections can show up as untracked files and be committed accidentally.
- The new unit suite performs real filesystem mutation with t.TempDir, MkdirAll, WriteFile, and Symlink, which conflicts with the repository filesystem-test isolation requirement.
- Status only inspects .agents/skills, so broken Claude or Grok projections can still report linked and pass strict status.

What was done well: projection ownership is centralized, the force and conflict behavior is explicit, the integration path is covered, and best-effort worktree creation keeps projection failures from undoing a successful checkout.

Would I be proud to ship this? Not until the target-repo ignore behavior, test isolation, and complete status reporting are addressed.

lancekrogers:
## Ready for re-review

Addressed latest request-changes:

| Feedback | Fix |
|----------|-----|
| Camp `.gitignore` does not cover target project worktrees | Projection installs **local** `.git/info/exclude` patterns (`.agents/`, `.claude/`, `.grok/`) in the target worktree |
| Status only inspected `.agents/skills` | `InspectWorktreeProjection` aggregates **agents + claude + grok** (min linked; sum conflicts/broken) |
| Host `t.TempDir` FS mutation suite | Unit suite is **pure-logic only**; container integration covers real git worktrees, force/foreign grok, dir projection, status aggregation, and exclude hygiene |

`go test ./internal/skills/` and `go test -tags=integration -run 'TestSkills_.*Worktree' ./tests/integration/` pass.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Re-reviewed tip `473eec12`. Prior request-changes are addressed: local `.git/info/exclude` in the **target** worktree, multi-surface status (agents+claude+grok), force-safe foreign `.grok/skills`, directory projection into real `.grok/skills`, pure unit tests + container integration with real git worktrees.

## Key Findings
No blocking issues. Soft-fail warning when info/exclude cannot be updated is acceptable (projection still succeeds). `ProjectIntoWorktreeBestEffort` success uses agents summary counts that include grok alias create/already-linked — intentional and correct for the success message.

## What's Done Well
Extends existing projection primitives; auto-project on worktree add; integration coverage for force/dir/status/exclude hygiene.

## Staff Standard
Yes — ready to merge.

PR #475, opened 2026-07-20
fix: handle large dungeon migration commits

lancekrogers:
## Summary

- batch large path lists when resetting the real Git index after scoped commits
- prevent `camp dungeon migrate` from failing during cleanup on large campaigns
- add regression coverage for argument-size-safe batching

## Root cause

Dungeon migration expands affected directory pathspecs into every tracked file. On the reported campaign this produced roughly 4.9 MB of reset arguments, exceeding macOS's process argument limit. The migration commit could be created, but index cleanup failed and the command reported the commit as failed.

## Validation

- `just test unit` — 3,586/3,586 tests passed
- dungeon migration integration tests passed
- stable/dev builds and `go vet` passed
- `just gate-fast` reached lint; its only failure was the pre-existing unused `renderStatusBar` finding in `internal/intent/tui/explorer/render.go`

### DIFF
--- a/internal/git/commit.go
+++ b/internal/git/commit.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
@@ -286,7 +287,12 @@ func ResetIndexToHead(ctx context.Context, repoPath string, paths []string) erro
 		// thousands of files (a dungeon migration is one example). Passing all
 		// of those paths to one git process can exceed the platform's argument
 		// size limit even though each individual path is valid.
-		for _, batch := range splitPathspecsForExec(paths) {
+		//
+		// Batches are applied sequentially. A mid-loop non-lock failure can
+		// leave earlier batches already reset; re-running is safe because
+		// `git reset HEAD -- <paths>` is idempotent for this cleanup.
+		limit := resetPathspecPayloadLimit(repoPath)
+		for _, batch := range splitPathspecsForExec(paths, limit) {
 			args := []string{"-C", repoPath, "reset", "-q", "HEAD", "--"}
 			args = append(args, batch...)
 			cmd := exec.CommandContext(ctx, "git", args...)
@@ -303,19 +309,58 @@ func ResetIndexToHead(ctx context.Context, repoPath string, paths []string) erro
 	})
 }
 
-// Keep reset invocations comfortably below common Unix ARG_MAX values. The
-// paths are already expanded before this point, so a single migration can
-// otherwise create a multi-megabyte argv.
-const resetPathspecArgLimit = 32 * 1024
+// platformCommandLineBudget is the max bytes we will put on a single git
+// process command line (executable + args). Unix ARG_MAX is typically
+// 256KiB–1MiB; Windows CreateProcess is 32,767 characters for the whole line.
+// We keep Unix large so huge migrations need few batches (performance), and
+// Windows conservative so fixed argv overhead cannot reintroduce the failure.
+func platformCommandLineBudget() int {
+	if runtime.GOOS == "windows" {
+		return 32767
+	}
+	// Large enough that multi-megabyte path lists become a handful of batches,
+	// not thousands of git invocations. Still far below typical ARG_MAX.
+	return 128 * 1024
+}
 
-func splitPathspecsForExec(paths []string) [][]string {
+// resetPathspecPayloadLimit is the max bytes of path arguments (+ NULs) per
+// git reset process after reserving space for fixed args
+// (`git -C <repo> reset -q HEAD --`) and a small safety margin.
+func resetPathspecPayloadLimit(repoPath string) int {
+	// Fixed argv: git, -C, repoPath, reset, -q, HEAD, --  (each +1 for separator/NUL).
+	fixed := len("git") + 1 +
+		len("-C") + 1 +
+		len(repoPath) + 1 +
+		len("reset") + 1 +
+		len("-q") + 1 +
+		len("HEAD") + 1 +
+		len("--") + 1
+	margin := 1024
+	if runtime.GOOS == "windows" {
+		// Extra headroom for quoting / CreateProcess encoding.
+		margin = 2048
+	}
+	payload := platformCommandLineBudget() - fixed - margin
+	if payload < minResetPathspecPayload {
+		return minResetPathspecPayload
+	}
+	return payload
+}
+
+// splitPathspecsForExec partitions paths so each batch's path payload
+// (bytes + one NUL per arg) stays within limit. A single path larger than
+// limit still forms its own batch (cannot split further).
+func splitPathspecsForExec(paths []string, limit int) [][]string {
+	if limit <= 0 {
+		limit = minResetPathspecPayload
+	}
 	var batches [][]string
 	var batch []string
 	batchBytes := 0
 
 	for _, path := range paths {
 		pathBytes := len(path) + 1 // include the argument's NUL terminator
-		if len(batch) > 0 && batchBytes+pathBytes > resetPathspecArgLimit {
+		if len(batch) > 0 && batchBytes+pathBytes > limit {
 			batches = append(batches, batch)
 			batch = nil
 			batchBytes = 0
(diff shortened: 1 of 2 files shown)


internal/git/commit.go
```diff
@@ -282,21 +282,53 @@ func ResetIndexToHead(ctx context.Context, repoPath string, paths []string) erro
...
 			}
-			return camperrors.NewGit("reset", "", errType.String(), strings.TrimSpace(string(out)), err)
 		}
 		return nil
 	})
 }
 
+// Keep reset invocations comfortably below common Unix ARG_MAX values. The
+// paths are already expanded before this point, so a single migration can
+// otherwise create a multi-megabyte argv.
```
obey-agent: 32 KiB of *path* bytes alone sits at the Windows CreateProcess ceiling (32,767 chars) before counting `git -C <repo> reset -q HEAD --` and the executable path. On macOS you are fine (ARG_MAX is much larger; the reported failure was multi-MB). Prefer a smaller budget (e.g. 16 KiB) so fixed argv overhead cannot reintroduce the failure on constrained platforms, and note the constant is path-payload only.

internal/git/commit.go
```diff
@@ -282,21 +282,53 @@ func ResetIndexToHead(ctx context.Context, repoPath string, paths []string) erro
...
-		if err != nil {
-			errType := ClassifyGitError(string(out), cmd.ProcessState.ExitCode())
-			if errType == GitErrorLock {
-				return &LockError{Path: "index.lock", Err: err}
+		// Scoped commits may expand a directory pathspec into tens of
+		// thousands of files (a dungeon migration is one example). Passing all
+		// of those paths to one git process can exceed the platform's argument
+		// size limit even though each individual path is valid.
+		for _, batch := range splitPathspecsForExec(paths) {
+			args := []string{"-C", repoPath, "reset", "-q", "HEAD", "--"}
```
obey-agent, line 296: Good that empty `paths` still short-circuits above. One behavioral change vs the single-shot reset: a non-lock failure mid-batch leaves earlier batches already applied. For `reset HEAD -- paths` that is effectively idempotent on retry, which is fine — worth a one-line comment so future readers do not try to wrap this in a multi-batch transaction.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Focused, correct fix for a real production failure mode: dungeon migration (and any scoped commit) expands directory pathspecs into huge file lists, then `ResetIndexToHead` invoked `git reset` with a multi-megabyte argv and blew past platform limits after the commit had already succeeded. Batching pathspecs below a conservative byte budget is the right place to fix this — not in dungeon-specific code.

## Key Findings
- **Arg budget headroom (non-blocking):** `resetPathspecArgLimit = 32 * 1024` counts only path payload (+ NULs). That is essentially the Windows CreateProcess string ceiling *before* fixed args (`git -C … reset -q HEAD --`). Recommend ~16 KiB (or similar) so constrained platforms keep headroom (inline on the constant).
- **Partial batches:** mid-loop failure leaves earlier batches applied; for this operation that is retry-safe/idempotent — a short comment would help (inline on the loop).
- Empty `paths` still returns early; single oversize path still forms its own batch — good edge handling.
- Unit test proves multi-batch split, per-batch size, and order preservation — appropriate for a pure helper.

## What's Done Well
- Fix lives in the shared git primitive used by all scoped commits, not a one-off dungeon hack.
- Clear root-cause narrative in the PR body (commit OK, cleanup failed, user saw failure).
- Lock-retry still wraps the full multi-batch sequence; lock errors stay classified.
- No API surface churn.

## Staff Standard
Yes — I would ship this as-is for the reported macOS failure. Lowering the limit for Windows/headroom is a small follow-up if you want belt-and-suspenders, not a merge blocker for the incident.

lancekrogers:
## Follow-up: platform-aware argv budget

Addressed the review note without reducing Unix batch performance:

| Platform | Command-line budget | Path payload |
|----------|---------------------|--------------|
| **Unix/macOS** | **128 KiB** (larger batches than before → fewer `git` spawns) | budget − fixed argv − 1 KiB margin |
| **Windows** | **32,767** (CreateProcess ceiling) | budget − fixed argv − 2 KiB margin |

Also:
- Fixed argv for `git -C <repo> reset -q HEAD --` is reserved from the budget
- Documented that mid-batch failure is retry-safe (`reset HEAD -- paths` is idempotent)
- Tests for empty input, repo-path-sensitive payload, and Unix budget size

Unix path is **more** performant than the previous flat 32 KiB path-only limit.

internal/git/commit.go
```diff
@@ -282,21 +283,101 @@ func ResetIndexToHead(ctx context.Context, repoPath string, paths []string) erro
...
+		len("-C") + 1 +
+		len(repoPath) + 1 +
+		len("reset") + 1 +
+		len("-q") + 1 +
+		len("HEAD") + 1 +
+		len("--") + 1
+	margin := 1024
+	if runtime.GOOS == "windows" {
+		// Extra headroom for quoting / CreateProcess encoding.
+		margin = 2048
```
obey-agent: The Windows budget is still set to the exact CreateProcess ceiling, while resetPathspecPayloadLimit estimates only raw path bytes plus a 2 KiB margin. Go os/exec adds quoting for paths containing spaces and escaping for special characters, and the executable path is longer than the literal git token used in the fixed-argument estimate. A batch with many ordinary space-containing Git paths can still exceed 32767 and recreate this failure. Please use a materially lower Windows ceiling such as 16 KiB, or calculate the actual platform command-line representation, and add a Windows-specific regression case.

obey-agent (commented):
Verdict: Request Changes

Overview: Re-reviewed the updated two-commit diff. The batching design is correct for the reported macOS failure, and the validation is strong: internal Git tests, vet, dungeon migration integration tests, and the full unit suite pass. One cross-platform correctness gap remains in the Windows budget calculation.

Key Findings:
- The Windows path reserves only 2 KiB while budgeting against the exact 32,767-character CreateProcess ceiling, but Go os/exec adds quoting and escaping for path arguments and the executable path is not represented by the literal git token used in the estimate. A batch containing many ordinary space-containing paths can still exceed the limit. Please lower the Windows ceiling materially or calculate the actual platform command-line representation, with a Windows-specific regression case. Inline comment is on internal/git/commit.go:341.

What was Done Well: The fix remains in the shared ResetIndexToHead primitive, preserves lock retry behavior, documents retry-safe partial progress, reserves fixed arguments, handles empty and oversized inputs, and adds useful pure batching tests.

Staff Standard: I would be proud to submit the Unix/macOS portion as my own work, but I would not ship the claimed Windows support until its command-line sizing is conservative against os/exec quoting.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Re-reviewed tip `37aacc73`. Batching `git reset` pathspecs fixes the real dungeon-migration ARG_MAX failure. Windows now budgets 16 KiB with long-exe stand-in and 4 KiB quoting margin (not the raw 32767 ceiling); Unix remains 128 KiB so large migrations stay few-batch and fast.

## Key Findings
No blocking issues. Mid-batch failure remains retry-safe/idempotent (documented). Unit coverage for split, empty input, and platform budget is adequate for this pure helper.

## What's Done Well
Fix lives in the shared scoped-commit cleanup path; clear root-cause narrative; platform split preserves non-Windows performance.

## Staff Standard
Yes — ready to merge.

lancekrogers:
## Ready for re-review

Addressed Windows CreateProcess budgeting:

| Platform | Command-line budget | Notes |
|----------|---------------------|--------|
| **Windows** | **16 KiB** (was exact 32767) | Long `git.exe` path stand-in + 4 KiB quoting margin |
| **Unix/macOS** | **128 KiB** (unchanged) | Large batches preserved for performance |

Does not reduce non-Windows batch performance.

obey-agent (commented):
## Verdict: **Approve**

## Overview
Re-reviewed tip `37aacc73`. Batching `git reset` pathspecs fixes the dungeon-migration ARG_MAX failure. Windows budgets **16 KiB** with a long `git.exe` path stand-in and **4 KiB** quoting margin (not the raw 32767 ceiling); Unix remains **128 KiB** so large migrations stay few-batch and fast.

## Key Findings
No blocking issues. Mid-batch failure remains retry-safe/idempotent (documented). Unit coverage for split, empty input, and platform budget is adequate.

## Staff Standard
Yes — ready to merge.

obey-agent (commented):
Verdict: Approve

Overview: The requested Windows argv-safety fix is now addressed. Windows uses a conservative 16 KiB budget below the 32,767-character CreateProcess ceiling, reserves a realistic git executable path, and keeps a wider quoting margin; Unix/macOS batching remains unchanged.

Key Findings: The prior Windows command-line sizing concern is resolved. Focused Git tests and vet pass on the final commit; the full unit suite passed with 3,585/3,585 tests, and the dungeon migration integration test passed.

What Was Done Well: The fix stays in the shared Git primitive, preserves lock retry and idempotent partial progress, and now documents the platform-specific safety rationale in both code and tests.

Staff Standard: Yes, I would be proud to ship this updated change.

PR #478, opened 2026-07-21
fix(worktrees): make list project-aware

lancekrogers:
## Summary

- Make `camp worktrees list` automatically scope to the current registered project.
- Detect project context from both the main checkout and linked Git worktrees.
- Preserve `--project` as an explicit filter when invoked outside a project.
- Add a container-backed regression test and a deterministic VHS demo.

## Root cause

The list command only used `--project` when supplied; otherwise it scanned every registered project. Generic project resolution also misidentified a submodule worktree as `projects/worktrees`, so linked worktree context needed to be matched against Git's worktree entries directly.

## Validation

- `just gate-fast`
- `just test unit`
- `go test -tags=integration ./tests/integration -run 'TestWorktreesList_' -count=1 -timeout 15m`
- VHS recording from a disposable two-project campaign

## Demo

See [`docs/demos/worktree-list.gif`](https://github.com/Obedience-Corp/camp/blob/worktree-list-project-aware/docs/demos/worktree-list.gif) and the reproducible [`worktree-list.tape`](https://github.com/Obedience-Corp/camp/blob/worktree-list-project-aware/docs/demos/worktree-list.tape).


cmd/camp/worktrees/list.go
```diff
@@ -182,6 +189,94 @@ func listWorktrees(ctx context.Context, campRoot string, filterProject string, s
...
+		}
+	}
+	if match != nil {
+		return []worktreeProjectTarget{*match}, nil
+	}
+
+	// A linked git worktree lives beside the registered project path rather
+	// than underneath it. Match the current directory against git's worktree
+	// entries so projects/worktrees/<project>/<name> remains project-aware too.
+	for _, proj := range projects {
```
obey-agent, line 286: **Maintainability / production cost:** when the user is *not* under a project path (including the common case of campaign root), this loop still runs `git worktree list` for every registered project, only to conclude there is no match and fall through to campaign-wide enumeration.

`listWorktrees` then lists those worktrees again. Net: ~2× `git worktree list` vs main for the default "show everything" path.

Prefer either:
1. Short-circuit when `cwd` is the campaign root (or not under `projects/`), or
2. Reuse the `projects` slice already loaded and skip the second full scan by returning targets built from it without re-calling `worktreeProjectTargets` (which re-lists projects).

Detection is right; this is pure wasted I/O on the happy path that did not need detection.

cmd/camp/worktrees/list.go
```diff
@@ -182,6 +189,94 @@ func listWorktrees(ctx context.Context, campRoot string, filterProject string, s
...
+		}
+	}
+	if match != nil {
+		return []worktreeProjectTarget{*match}, nil
+	}
+
+	return worktreeProjectTargets(ctx, campRoot, "")
+}
+
+func pathWithin(child, parent string) bool {
```
obey-agent, line 360: **Correctness edge case:** `pathWithin` does not `filepath.Clean` either argument, and only `cwd` is `EvalSymlinks`'d above. `target.path` / `logicalPath` are not.

On machines where the campaign root is reached via a symlink (common on macOS), `cwd` can be the resolved path while `ResolveProjectPath` still returns the unresolved form — then the registered-path loop misses and you pay the worktree scan (or mis-scope if git paths also differ).

Clean both sides here (and ideally symlink-resolve parents the same way as `cwd`) so containment is stable:

```go
child, parent = filepath.Clean(child), filepath.Clean(parent)
```

cmd/camp/worktrees/list.go
```diff
@@ -182,6 +189,94 @@ func listWorktrees(ctx context.Context, campRoot string, filterProject string, s
...
+	if child == "" || parent == "" {
+		return false
+	}
+	if child == parent {
+		return true
+	}
+	rel, err := filepath.Rel(parent, child)
```
obey-agent, line 369: **Test gap:** `pathWithin` is pure and security/UX-adjacent (wrong containment = wrong project scope). There is solid container coverage for the end-to-end cases, but no table-driven unit tests for equal path, nested child, sibling, empty inputs, or `parent` prefix tricks.

Please add cases in `list_test.go` (same package already has pure tests for disambiguation). Cheap insurance for the next path-layout change.

cmd/camp/worktrees/list.go
```diff
@@ -182,6 +189,94 @@ func listWorktrees(ctx context.Context, campRoot string, filterProject string, s
 	}, nil
 }
 
+// listWorktreeProjectTargets resolves the scan scope for the list command.
+// An explicit --project always wins. Without it, a project checkout or one of
+// its linked worktrees scopes the list to that project; otherwise the command
+// retains its campaign-wide behavior.
+func listWorktreeProjectTargets(ctx context.Context, campRoot, filterProject string) ([]worktreeProjectTarget, error) {
```
obey-agent, line 196: **SRP / size:** this function is ~70 lines and owns three policies (explicit filter, registered-path match, linked-worktree match). The two match loops also rebuild the same `worktreeProjectTarget` shape.

Extracting `matchRegisteredProject(cwd, campRoot, projects)` and `matchLinkedWorktree(ctx, cwd, projects)` would keep each under ~50 lines and make unit-testing the match rules without git (or with a small interface) straightforward. Not a merge blocker if you take the short-circuit + `pathWithin` cleanups above, but it will pay off the next time list/clean scope drifts.

obey-agent (commented):
## Verdict: Comment (approve with non-blocking polish)

## Overview

Solid, user-facing fix: `camp worktrees list` now scopes to the current project (main checkout or linked worktree) instead of always dumping the whole campaign, while `--project` still overrides and campaign-root stays global. The root-cause story (avoid generic project resolution treating `projects/worktrees` as a project) is handled correctly by matching registered paths first, then git worktree entries. Integration coverage and a reproducible VHS demo match the behavior change.

## Key Findings

1. **Campaign-root cost regression** — auto-detect always runs `git worktree list` for every project before falling back to campaign-wide list, then listing runs those again. Short-circuit campaign root / non-project cwd, or reuse the already-loaded project set. (inline on the linked-worktree loop)
2. **`pathWithin` should Clean (and ideally symlink-align) both args** — only `cwd` is `EvalSymlinks`’d; unresolved `ResolveProjectPath` can miss main-checkout detection on symlink-heavy layouts. (inline on `pathWithin`)
3. **Missing pure unit tests for `pathWithin`** — end-to-end integration is good; table-driven containment cases belong in `list_test.go`. (inline)
4. **`listWorktreeProjectTargets` is ~70 lines / three policies** — extract registered-path vs linked-worktree matchers for SRP and easier tests. Non-blocking if the short-circuit + Clean land. (inline)

None of these look like ship-stoppers for the intended UX, but (1) and (2) are worth a small follow-up commit before or right after merge.

## What's Done Well

- **Right layering**: list-specific auto-scope, not a global project.Resolve hack — explicitly avoids the `projects/worktrees` mis-identity.
- **Policy clarity**: `--project` always wins; docs/root help updated consistently.
- **Longest-path match** for nested project layouts is the correct disambiguation rule.
- **Regression test** covers campaign-wide, main checkout, linked worktree, and `--project` filter.
- **Demo** (`worktree-list.tape` + gif) exercises cgo navigation and interactive list — good operator proof.
- **Errors**: still uses `camperrors.Wrap` on list failures; context flows through I/O.

## Staff Standard

I would be comfortable shipping this as my own work for the product behavior and test evidence. I would **not** be proud to leave the double worktree enumeration on the default campaign-root path without a short-circuit, or `pathWithin` without Clean + unit tests — those are small, high-leverage cleanups. Address findings 1–2 (and ideally 3) in a polish commit; then this is merge-ready without reservation.

cmd/camp/worktrees/list_test.go
```diff
@@ -129,3 +131,109 @@ func TestOutputListTable_StaleReasonInStatus(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			if got := pathWithin(tc.child, tc.parent); got != tc.want {
+				t.Errorf("pathWithin(%q, %q) = %v, want %v", tc.child, tc.parent, got, tc.want)
+			}
+		})
+	}
+}
+
+func TestPathWithin_SymlinkParent(t *testing.T) {
+	root := t.TempDir()
```
obey-agent: This test uses t.TempDir, MkdirAll, and Symlink to mutate the host filesystem. The repository's test policy requires filesystem-mutating workflow tests to run in the containerized harness rather than host temp directories. Please move this symlink case into the container-backed integration test (or make the path-normalization logic injectable and keep the unit test purely in-memory) so the new test follows that policy.

obey-agent (commented):
## Verdict: Request Changes

## Overview

The project-aware scoping logic is well structured and the integration test covers campaign-root listing, project checkout detection, linked-worktree detection, and explicit filtering. The latest hardening commit also removes the earlier redundant campaign-root worktree scans and adds useful containment tests. The targeted unit and integration tests pass.

## Key Findings

- The new `TestPathWithin_SymlinkParent` mutates the host filesystem with `t.TempDir`, `MkdirAll`, and `Symlink`. This conflicts with the repository test policy requiring filesystem-mutating tests to use the containerized harness; please address the inline comment before merge.

## What is Done Well

- Explicit `--project` precedence is clear.
- Longest-path matching handles nested project layouts.
- Git worktree entries remain the source of truth, including worktrees outside the conventional directory.
- The full-TUI VHS demo exercises `cgo` navigation and color-capable PTY setup.

## Staff Standard

I would be comfortable shipping the implementation after the symlink test is moved to the container-backed integration harness or made purely in-memory. No additional production-logic blockers were found.

cmd/camp/worktrees/list.go
```diff
@@ -182,6 +189,177 @@ func listWorktrees(ctx context.Context, campRoot string, filterProject string, s
...
+
+// matchLinkedWorktree returns the project that owns the linked git worktree
+// containing cwd. Longest entry path wins when multiple entries nest.
+func matchLinkedWorktree(ctx context.Context, cwd, campRoot string, projects []project.Project) *worktreeProjectTarget {
+	var (
+		match     *worktreeProjectTarget
+		matchPath string
+	)
+	for _, proj := range projects {
+		if err := ctx.Err(); err != nil {
```
obey-agent, line 287: **Non-blocking (context hygiene):** when `ctx` is cancelled mid-loop this returns `nil`, so `listWorktreeProjectTargets` falls through to campaign-wide targets with `nil` error instead of surfacing cancellation.

Prefer:

```go
if err := ctx.Err(); err != nil {
    return nil // and have the caller treat cancel as error
}
```

or change the signature to `(*worktreeProjectTarget, error)` and return `ctx.Err()` so cancel never becomes a silent "list everything". Not a ship-stopper for interactive list, but it is the staff rule for context-first I/O paths.

cmd/camp/worktrees/list.go
```diff
@@ -182,6 +189,177 @@ func listWorktrees(ctx context.Context, campRoot string, filterProject string, s
...
+	all := targetsFromProjects(campRoot, projects)
+
+	cwd := resolveListCwd()
+	if cwd == "" {
+		return all, nil
+	}
+
+	campReal := normalizePath(campRoot)
+	// Campaign root is the common "list everything" case — skip detection work.
+	if sameNormalizedPath(cwd, campReal) {
```
obey-agent, line 220: **Prior finding closed:** campaign-root and non-`projects/worktrees` short-circuits plus `targetsFromProjects` reuse correctly eliminate the earlier 2× `git worktree list` cost on the default unscoped path. Clear, documented, and easy to follow.

tests/integration/worktrees_list_test.go
```diff
@@ -124,3 +124,98 @@ git -C %[1]s worktree add %[2]s/projects/worktrees/dup -b dup-loose
...
+	require.NoError(t, err, "filtered list: %s", out)
+	assert.Contains(t, out, "other-context")
+	assert.NotContains(t, out, "proj-context")
+}
+
+// TestWorktreesList_ScopesViaSymlinkedProjectPath proves path normalization
+// keeps project detection stable when the project checkout is reached through
+// a symlink (common on macOS). Filesystem mutation stays in the container
+// harness per repo test policy — not host TempDir/Symlink unit tests.
+func TestWorktreesList_ScopesViaSymlinkedProjectPath(t *testing.T) {
```
obey-agent, line 188: **Prior finding closed:** symlink coverage now lives in the container harness (`TestWorktreesList_ScopesViaSymlinkedProjectPath`). Pure `pathWithin` tables remain host-safe. Matches repo filesystem-mutation policy.

obey-agent (commented):
## Verdict: Approve

## Overview

Re-review of HEAD `71853a08` after the hardening and test-policy follow-ups. Project-aware `camp worktrees list` is production-ready: correct scoping policy, cheap unscoped path, symlink-stable containment, and solid unit + container coverage. Prior Request Changes (host TempDir/Symlink unit test) is fully addressed.

## Key Findings

Prior blockers are **closed**:
- Campaign-root / non-worktree short-circuit and `targetsFromProjects` reuse (no double project.List / wasted detection scans)
- `normalizePath` + `pathWithin` Clean/EvalSymlinks alignment
- Helpers extracted (`matchRegisteredProject`, `matchLinkedWorktree`)
- Symlink case moved to container integration; pure pathWithin tables stay host-safe

**Non-blocking only:**
- `matchLinkedWorktree` returns `nil` on `ctx.Err()` so cancel can fall through to campaign-wide targets instead of an error (inline). Optional polish; not a merge gate for this interactive command.

## What's Done Well

- **Policy order is right:** `--project` → campaign root short-circuit → registered path → conventional worktree home → git worktree match → all.
- **Longest-path match** for nested monorepos is correct and unit-tested.
- **Integration matrix** covers campaign-wide, main checkout, linked worktree, `--project` filter, and symlinked project cwd.
- **VHS demo** documents the UX with `cgo` navigation.
- **camperrors** wrapping and context checks on the hot path are in good shape.

## Staff Standard

Yes — I would be proud to ship this as my own work. Approve for merge; the only remaining note is optional cancel-propagation hygiene in `matchLinkedWorktree`.

PR #479, opened 2026-07-21
fix: prevent stale workitem context in generic commits

lancekrogers:
## Summary

- prevent generic `camp commit`, project commits, and worktree commits from silently inheriting stale `.campaign/workitems/current.yaml` context
- preserve explicit, ancestor, linked-project, and festival context resolution
- keep `camp workitem commit` as the intentional current-workitem workflow
- add resolver and commit-path regression tests

## Root cause

The generic commit wrappers used the resolver's lowest-priority `current.yaml` fallback. Because that file is per-machine and session-wide, an old selected festival could be attached to unrelated commits with no local context.

## Validation

- `just gate-push` passed: stable/dev builds, vet, lint, and dev short tests
- targeted resolver, workitem, commit, project, and worktree tests passed
- unrelated dungeon-migration worktree changes were left unstaged

### DIFF
--- a/cmd/camp/commit_workitem.go
+++ b/cmd/camp/commit_workitem.go
@@ -13,6 +13,9 @@ import (
 // and returns the captured quest id and ref for inclusion in the commit
 // tag. Resolution failures are non-fatal: empty strings are returned so the
 // caller can still produce a quest- and workitem-free tag.
+// Generic camp commits intentionally do not inherit current.yaml; callers
+// that want session-wide workitem scoping should use camp workitem commit or
+// pass --workitem explicitly.
 //
 // explicit, when non-empty, is the user-supplied --workitem selector.
 //
@@ -21,7 +24,8 @@ import (
 // disk so future commits inherit it. A stderr warning notifies the user.
 func resolveCommitContext(ctx context.Context, campaignRoot, explicit string) (questID, festivalRef, workitemRef string) {
 	res, err := resolver.Resolve(ctx, campaignRoot, resolver.Options{
-		Explicit: explicit,
+		Explicit:       explicit,
+		DisableCurrent: true,
 	})
 	if err != nil || res == nil || res.Workitem == nil {
 		return "", "", ""
@@ -39,7 +43,8 @@ func resolveCommitContext(ctx context.Context, campaignRoot, explicit string) (q
 // workitem context resolves so the hook sees no leaked vars.
 func workitemEnvForCommit(ctx context.Context, campaignRoot, explicit string) []string {
 	res, err := resolver.Resolve(ctx, campaignRoot, resolver.Options{
-		Explicit: explicit,
+		Explicit:       explicit,
+		DisableCurrent: true,
 	})
 	if err != nil || res == nil || res.Workitem == nil {
 		return nil
--- a/cmd/camp/commit_workitem_test.go
+++ b/cmd/camp/commit_workitem_test.go
@@ -5,22 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-
-	"github.com/Obedience-Corp/camp/internal/workitem/links"
 )
-
-func writeCommitContextCampaign(t *testing.T) string {
-	t.Helper()
-	root := t.TempDir()
-	if err := os.MkdirAll(filepath.Join(root, ".campaign"), 0o755); err != nil {
-		t.Fatal(err)
-	}
-	if err := os.WriteFile(filepath.Join(root, ".campaign", "campaign.yaml"),
-		[]byte("version: campaign/v1\nid: test-campaign\nname: Test\ntype: product\n"), 0o644); err != nil {
-		t.Fatal(err)
-	}
-	return root
-}
 
 func TestResolveCommitContext_GracefulOutsideCampaign(t *testing.T) {
 	tmp := t.TempDir()
@@ -38,28 +23,5 @@ func TestResolveCommitContext_GracefulOutsideCampaign(t *testing.T) {
 	questID, festivalRef, ref := resolveCommitContext(context.Background(), filepath.Join(tmp, "no-campaign"), "")
 	if questID != "" || festivalRef != "" || ref != "" {
 		t.Fatalf("expected empty strings for non-campaign root, got quest=%q festival=%q ref=%q", questID, festivalRef, ref)
-	}
-}
-
-func TestResolveCommitContext_DoesNotInheritCurrentWorkitem(t *testing.T) {
-	root := writeCommitContextCampaign(t)
-	wiDir := filepath.Join(root, "workflow", "design", "example")
-	if err := os.MkdirAll(wiDir, 0o755); err != nil {
-		t.Fatal(err)
-	}
-	if err := os.WriteFile(filepath.Join(wiDir, ".workitem"), []byte(
-		"version: v1alpha5\nkind: workitem\nid: design-example-2026-05-24\ntype: design\ntitle: Example\n"), 0o644); err != nil {
-		t.Fatal(err)
-	}
-	if err := links.SaveCurrent(context.Background(), root, &links.Current{
-		Version:    links.CurrentSchemaVersion,
-		WorkitemID: "design-example-2026-05-24",
-	}); err != nil {
-		t.Fatal(err)
-	}
-
-	gotQuest, gotFestival, gotWorkitem := resolveCommitContext(context.Background(), root, "")
-	if gotQuest != "" || gotFestival != "" || gotWorkitem != "" {
-		t.Fatalf("commit context inherited current workitem: quest=%q festival=%q workitem=%q", gotQuest, gotFestival, gotWorkitem)
 	}
 }
--- a/cmd/camp/commit_workitem_test.go
+++ b/cmd/camp/commit_workitem_test.go
@@ -5,7 +5,22 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/Obedience-Corp/camp/internal/workitem/links"
 )
+
+func writeCommitContextCampaign(t *testing.T) string {
+	t.Helper()
+	root := t.TempDir()
+	if err := os.MkdirAll(filepath.Join(root, ".campaign"), 0o755); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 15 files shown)


cmd/camp/commit_workitem.go
```diff
@@ -21,7 +24,8 @@ import (
 // disk so future commits inherit it. A stderr warning notifies the user.
 func resolveCommitContext(ctx context.Context, campaignRoot, explicit string) (questID, festivalRef, workitemRef string) {
 	res, err := resolver.Resolve(ctx, campaignRoot, resolver.Options{
-		Explicit: explicit,
+		Explicit:       explicit,
+		DisableCurrent: true,
```
obey-agent, line 28: [P1] Stale current context still leaks through auto-commit paths

This opt-out is wired into the three CLI commit wrappers, but the shared `internal/commands/workitem.ResolveCommitContext` / `AmbientCommitOptions` path still calls `resolver.Resolve` with the default options. That helper is used by the intent/note auto-commit commands, so an unrelated `camp intent ...` or note operation can still stamp the stale `.campaign/workitems/current.yaml` workitem—the exact misattribution this PR is intended to prevent. Either pass `DisableCurrent` there as well (while retaining explicit/ancestor/link tiers), or explicitly document and regression-test those commands as intentional current-scoped exceptions; as written, the root-cause fix is incomplete.

cmd/camp/project/commit_workitem.go
```diff
@@ -24,8 +24,9 @@ import (
 // short-circuits cwd-based resolution.
 func resolveProjectCommitContext(ctx context.Context, campaignRoot, cwd, explicit string) (questID, festivalRef, workitemRef string) {
 	res, err := resolver.Resolve(ctx, campaignRoot, resolver.Options{
-		Explicit: explicit,
-		Cwd:      cwd,
+		Explicit:       explicit,
+		Cwd:            cwd,
+		DisableCurrent: true,
```
obey-agent, line 29: [P2] Cover the public commit entry points, not only the helper

The regression test exercises `resolveCommitContext` directly, but the PR changes three user-facing commands (`camp commit`, `camp project commit`, and `camp worktrees commit`) plus their auto-write environment paths. Add command-level tests that create a current.yaml selection and assert the actual git commit subject/environment for each wrapper, while also covering explicit, ancestor, and link contexts. Otherwise a future wiring omission can leave one command inheriting stale context while this helper-level test remains green.

cmd/camp/commit_workitem_test.go
```diff
@@ -25,3 +40,26 @@ func TestResolveCommitContext_GracefulOutsideCampaign(t *testing.T) {
 		t.Fatalf("expected empty strings for non-campaign root, got quest=%q festival=%q ref=%q", questID, festivalRef, ref)
 	}
 }
+
+func TestResolveCommitContext_DoesNotInheritCurrentWorkitem(t *testing.T) {
+	root := writeCommitContextCampaign(t)
```
obey-agent: [P2] Keep the new filesystem-mutating test in the container harness

This new test creates a campaign, writes workitem state, and updates current.yaml directly under t.TempDir. The repository policy requires tests that exercise real filesystem mutation to use the containerized harness; the current test file is already legacy-allowlisted, but extending it with more host-side mutation preserves the violation. Move this regression to the container-backed integration suite or refactor the test around an injectable in-memory resolver.

cmd/camp/worktrees/commit_workitem.go
```diff
@@ -11,8 +11,9 @@ import (
 
 func resolveWorktreeCommitContext(ctx context.Context, campaignRoot, cwd, explicit string) (questID, festivalRef, workitemRef string) {
 	res, err := resolver.Resolve(ctx, campaignRoot, resolver.Options{
-		Explicit: explicit,
-		Cwd:      cwd,
+		Explicit:       explicit,
+		Cwd:            cwd,
+		DisableCurrent: true,
```
obey-agent, line 16: [P2] Document the behavior change for project/worktree commits

`camp commit --help` now warns that current.yaml is ignored, but `camp project commit --help` and `camp worktrees commit --help` still describe automatic context without explaining this change. Since this PR silently changes tagging behavior for those commands too, update their help/reference docs (preferably from one shared wording) and add a regression assertion for the documented contract.

obey-agent (commented):
## Verdict: Request Changes

## Overview

The core `DisableCurrent` mechanism is small and the affected commit wrappers are wired consistently, and `just gate-push` passes. However, the PR does not fully close the stale-context failure mode it describes: shared intent/note auto-commit paths still use the resolver default and can inherit current.yaml, while the new tests do not exercise the public commands or auto-write paths. The user-visible behavior change is also only documented for `camp commit`.

## Key Findings

- **P1:** The shared `internal/commands/workitem.ResolveCommitContext` / `AmbientCommitOptions` path still resolves `current.yaml`; intent and note auto-commits can therefore continue stamping stale workitem refs. Either disable that tier there too or explicitly define and test those paths as intentional exceptions.
- **P2:** Add command-level coverage for `camp commit`, `camp project commit`, and `camp worktrees commit`, including their auto-write environment paths and explicit/ancestor/link contexts.
- **P2:** Document the current.yaml behavior change in project/worktree commit help and reference docs, not only `camp commit` help.
- **P2:** Move the new filesystem-mutating host-temp tests into the container-backed harness or make the test purely in-memory.

## What is Done Well

- The resolver option is backward-compatible for existing callers and preserves explicit, ancestor, link, and festival tiers.
- The changed wrappers consistently apply the opt-out to both tag resolution and auto-write environment resolution.
- Error handling remains best-effort at the wrapper boundary, matching existing commit behavior.
- The full project quality gate passes.

## Staff Standard

I would not approve this as-is. The P1 auto-commit leakage means the stated stale-context fix is incomplete in production, and the missing command-level coverage makes that easy to regress. After the behavior boundary is made explicit and covered, this should be a straightforward change to ship.

internal/workitem/resolver/resolver.go
```diff
@@ -81,6 +85,12 @@ func Resolve(ctx context.Context, root string, opts Options) (*Resolution, error
 	}
 
 	result := &Resolution{Source: SourceNone, Reason: "no tier matched"}
+	currentTier := func() (*workitem.WorkItem, TraceStep, error) {
+		if opts.DisableCurrent {
```
obey-agent, line 89: **Test gap (non-blocking):** `DisableCurrent` is the whole product fix, but `internal/workitem/resolver` has no unit coverage for it. Integration tests prove commit paths; they do not lock the resolver contract itself.

Please add a small test in `resolver_test.go` (same package already uses host fixtures for other tiers):

1. Seed only `current.yaml` so the current tier would match
2. `Resolve(..., Options{Cwd: root, DisableCurrent: true})`
3. Assert `Source == SourceNone`, `Workitem == nil`, and a trace step with `Tier=current`, `Result=skip`, detail containing `disabled`

That keeps the flag from regressing if someone rewires the tier table.

cmd/camp/commit_workitem.go
obey-agent, line 28: **Maintainability:** `DisableCurrent: true` is now repeated at 7 call sites (`camp`/`project`/`worktrees` commit env+context + `ResolveCommitContext`). The next generic commit path that forgets the flag reintroduces the stale-tag bug.

Prefer a single constructor in the resolver package, e.g.:

```go
// GenericCommitOptions builds Options for path/explicit context only
// (never current.yaml). Use for camp/p/worktrees commit and auto-commit.
func GenericCommitOptions(explicit, cwd, festivalID string) Options {
    return Options{Explicit: explicit, Cwd: cwd, FestivalID: festivalID, DisableCurrent: true}
}
```

Call sites become one-liners and the policy lives next to the flag docs. Non-blocking if you would rather land the behavior first.

tests/integration/commit_tags_test.go
```diff
@@ -115,6 +115,76 @@ func TestIntegration_CommitTags_NoContext(t *testing.T) {
 		"subject should still carry the campaign tag: %s", subject)
 }
 
+func TestIntegration_CommitTags_CampCommitIgnoresCurrentSelection(t *testing.T) {
```
obey-agent, line 118: **Coverage balance (non-blocking):** excellent negative matrix for "generic commits ignore current.yaml". Consider one positive regression that `camp workitem commit` *still* attaches the current selection when cwd has no other context — that is the intentional remaining user of the tier, and staging still calls `Resolve` without `DisableCurrent`.

Without that, a future accidental `DisableCurrent: true` on the workitem-commit path could ship as "all green" against these new tests alone. Optional if existing workitem-commit suites already pin SourceCurrent end-to-end.

obey-agent (commented):
## Verdict: Approve

## Overview

Correct, well-scoped fix for a real footgun: generic `camp commit` / `p commit` / `worktrees commit` and intent auto-commits no longer silently inherit a stale per-machine `current.yaml`. The resolver gains an explicit `DisableCurrent` opt-out with a clear `skip` trace, while `camp workitem commit` keeps the full tier ladder (no DisableCurrent). Docs and help text match the new contract.

## Key Findings

Non-blocking polish only:

1. **Unit-test the flag** — add `resolver` coverage that `DisableCurrent` yields `SourceNone` + `current`/`skip` when only current.yaml would match (inline on `resolver.go`).
2. **Centralize the policy** — seven `DisableCurrent: true` call sites; a `GenericCommitOptions` helper would prevent the next commit wrapper from forgetting the flag (inline on `commit_workitem.go`).
3. **Positive workitem-commit regression** — optional container test that `camp workitem commit` still uses current when that is the only context (inline on integration tests).

None of these block merge; product behavior and integration negatives are already solid.

## What's Done Well

- **Right layer:** policy switch on the shared resolver, not ad-hoc special-casing in each command.
- **Trace honesty:** disabled tier records `skip` with detail instead of disappearing.
- **Preserved higher tiers:** explicit / ancestor / link / festival still apply for generic commits.
- **Intentional escape hatch:** `camp workitem commit` remains the session-wide workflow; docs say so consistently.
- **Integration matrix:** camp, `p`, worktrees, auto-write env, and intent auto-commit all assert no stale `WI-` / env leak.
- **camperrors / context:** existing resolver cancel and error bubbling unchanged; no new bare `fmt.Errorf`.

## Staff Standard

Yes — I would be proud to ship this as my own work. Approve for merge; the unit helper / positive-test items are follow-ups, not gates.

PR #480, opened 2026-07-21
fix(leverage): correct personal --author leverage (no multi-count actual PM)

lancekrogers:
## Summary

`--author` was **overcounting actual effort** (summing each repo’s first→last commit span), which made personal leverage look artificially low (e.g. ~37x vs ~334x campaign) while COCOMO/LOC totals stayed identical.

### Root cause
- Personal actual PM = Σ(author span per project), including monorepo subprojects that share one `GitDir`
- Campaign aggregate skipped the deduped `CampaignActualPersonMonths` path whenever `--author` was set
- Estimated effort was not scaled by blame ownership

### Fix
- **Actual PM**: union of the author’s commit span across **unique git dirs** (not summed per project)
- **Estimated PM**: scale each project’s COCOMO by blame ownership fraction
- **`--author` expansion**: match `authors.json` identity groups (all emails)
- **`--by-author`**: report `est PM / union actual PM` per author (drops the old `weightedPM/campaignActual × campaignLeverage` formula)
- Docs + unit tests for multi-repo and monorepo dedupe

### Verified on obey-campaign (obey-chat-2 excluded)
| Mode | Actual PM | Full leverage |
|------|-----------|---------------|
| Default | ~14.4 | ~334x |
| `--author lance…` (before) | ~128 | ~37x |
| `--author lance…` (after) | ~14.7 | ~318x |

Personal leverage now tracks campaign leverage when the author owns nearly all remaining code.

Related explore: `workflow/explore/camp-leverage-scoring-audit-2026-07-21` (WI-a61e82)

## Test plan
- [x] `go test ./internal/leverage/ ./cmd/camp/leverage/`
- [x] Smoke: `camp leverage --json` vs `camp leverage --author <you> --json` on a multi-project campaign
- [x] Confirm personal `actual_person_months` ≈ campaign actual (not ~N×)
- [x] Confirm personal full leverage within ~band of campaign when sole majority author
- [ ] CI green on this PR

### DIFF
--- a/cmd/camp/leverage/main_command.go
+++ b/cmd/camp/leverage/main_command.go
@@ -53,6 +53,10 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	authorFilter, _ := cmd.Flags().GetString("author")
 	byAuthor, _ := cmd.Flags().GetBool("by-author")
 
+	if authorFilter != "" && byAuthor {
+		return camperrors.New("--author and --by-author are mutually exclusive: use --by-author for the full per-author breakdown, or --author for personal leverage")
+	}
+
 	setup, err := initLeverageSetup(ctx)
 	if err != nil {
 		return err
@@ -148,8 +152,11 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	if authorFilter != "" {
 		// Personal headline: union author calendar across unique git dirs
 		// (never sum per-project spans) with already ownership-scaled estimates.
-		authorPM, pmErr := intleverage.AuthorActualPersonMonths(ctx, scoredProjects, authorMatch)
-		if pmErr == nil && authorPM > 0 {
+		authorPM, pmErr := intleverage.AuthorActualPersonMonths(ctx, scoredProjects, authorMatch, setup.Resolver)
+		if pmErr != nil {
+			return camperrors.Wrap(pmErr, "computing personal actual person-months")
+		}
+		if authorPM > 0 {
 			estPM := agg.EstimatedPeople * agg.EstimatedMonths
 			agg.ActualPersonMonths = authorPM
 			agg.ActualPeople = 1
--- a/cmd/camp/leverage/main_command.go
+++ b/cmd/camp/leverage/main_command.go
@@ -87,8 +87,12 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	now := time.Now()
 	elapsed := intleverage.ElapsedMonths(cfg.ProjectStart, now)
 
+	authorMatch := intleverage.ExpandAuthorFilter(setup.Resolver, authorFilter)
+
 	var scores []*intleverage.LeverageScore
 	var snapshotInputs []intleverage.CurrentSnapshotInput
+	// Keep projects aligned with scores for personal aggregate rewrite.
+	var scoredProjects []intleverage.ResolvedProject
 	for _, proj := range resolved {
 		if ctx.Err() != nil {
 			return ctx.Err()
@@ -106,10 +110,13 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 
 		score := intleverage.ComputeProjectScore(ctx, proj, result, intleverage.ProjectScoreParams{
 			AuthorFilter:    authorFilter,
+			AuthorMatch:     authorMatch,
+			Resolver:        setup.Resolver,
 			PeopleOverride:  peopleOverride,
 			FallbackElapsed: elapsed,
 		})
 		scores = append(scores, score)
+		scoredProjects = append(scoredProjects, proj)
 		snapshotInputs = append(snapshotInputs, intleverage.CurrentSnapshotInput{
 			Project: proj,
 			Result:  result,
@@ -122,7 +129,10 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	}
 
 	effectivePeople := cfg.ActualPeople
-	if effectivePeople == 0 {
+	if authorFilter != "" {
+		// Personal mode is always one effective person for simple leverage.
+		effectivePeople = 1
+	} else if effectivePeople == 0 {
 		for _, s := range scores {
 			if s.AuthorCount > effectivePeople {
 				effectivePeople = s.AuthorCount
@@ -135,7 +145,20 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 
 	agg := intleverage.AggregateScores(scores, effectivePeople, elapsed)
 
-	if authorFilter == "" && peopleOverride == 0 {
+	if authorFilter != "" {
+		// Personal headline: union author calendar across unique git dirs
+		// (never sum per-project spans) with already ownership-scaled estimates.
+		authorPM, pmErr := intleverage.AuthorActualPersonMonths(ctx, scoredProjects, authorMatch)
+		if pmErr == nil && authorPM > 0 {
+			estPM := agg.EstimatedPeople * agg.EstimatedMonths
+			agg.ActualPersonMonths = authorPM
+			agg.ActualPeople = 1
+			agg.FullLeverage = estPM / authorPM
+			if effectivePeople > 0 {
+				agg.SimpleLeverage = agg.EstimatedPeople / float64(effectivePeople)
+			}
+		}
+	} else if peopleOverride == 0 {
 		campaignPM, pmErr := intleverage.CampaignActualPersonMonths(ctx, resolved, setup.Resolver)
 		if pmErr == nil && campaignPM > 0 {
 			estPM := agg.EstimatedPeople * agg.EstimatedMonths
@@ -170,7 +193,7 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	}
 
 	if byAuthor {
-		return leverageOutputByAuthor(cmd, agg, resolved, setup.Resolver, opts)
+		return leverageOutputByAuthor(cmd, ctx, agg, scores, scoredProjects, setup.Resolver, opts)
 	}
 
 	return leverageOutputTable(cmd, agg, scores, cfg, setup.AutoDetected, recent, opts)
--- a/cmd/camp/leverage/main_command.go
+++ b/cmd/camp/leverage/main_command.go
@@ -53,6 +53,9 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	authorFilter, _ := cmd.Flags().GetString("author")
 	byAuthor, _ := cmd.Flags().GetBool("by-author")
 
+	// Explicit CLI conflict only. Config AuthorEmail is applied after setup and
+	// is suppressed when --by-author is requested so the breakdown still works
+	// in campaigns that set a personal default.
 	if authorFilter != "" && byAuthor {
 		return camperrors.New("--author and --by-author are mutually exclusive: use --by-author for the full per-author breakdown, or --author for personal leverage")
 	}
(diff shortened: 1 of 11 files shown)


cmd/camp/leverage/output.go
```diff
@@ -160,56 +161,100 @@ func leverageOutputTable(cmd *cobra.Command, agg *intleverage.LeverageScore, sco
...
+	for i, proj := range resolved {
+		var score *intleverage.LeverageScore
+		if i < len(scores) {
+			score = scores[i]
+		}
+		projEstPM := 0.0
+		if score != nil {
+			// When --by-author is used without --author, scores are full-tree.
+			// Recompute from unscaled people×months.
+			projEstPM = score.EstimatedPeople * score.EstimatedMonths
```
obey-agent, line 189: [P1] Handle the supported combination of --author and --by-author. With both flags, ComputeProjectScore has already scaled each score to the selected authors ownership, but this code then distributes that reduced estimate across every author in the project. The table consequently reports understated EST PM and leverage for all rows while presenting them as a full author breakdown. Either reject the combination explicitly or compute this table from unscaled project scores and apply the intended author filter only to the requested row.

cmd/camp/leverage/output.go
```diff
@@ -160,56 +161,100 @@ func leverageOutputTable(cmd *cobra.Command, agg *intleverage.LeverageScore, sco
...
+		}
+		match := intleverage.AuthorMatch{
+			Filter:    authorID,
+			AuthorIDs: map[string]bool{authorID: true},
+			Emails:    authorEmailsForID(resolver, authorID),
+		}
+		if len(match.Emails) == 0 {
+			match.Emails = []string{authorID}
+		}
+		actualPM, err := intleverage.AuthorActualPersonMonths(ctx, resolved, match)
```
obey-agent: [P2] Avoid running a full git-log matrix once per author. AuthorActualPersonMonths loops every unique GitDir and every matched email, so --by-author is O(authors × repos × emails) subprocesses on top of the existing scans. A campaign with many contributors can make this reporting command very slow. Reuse one precomputed date-span map per GitDir, such as the existing gitDirAuthors/allAuthorDates path, and look up each canonical author from it.

internal/leverage/author_filter_test.go
```diff
@@ -0,0 +1,280 @@
...
+	empty := ResolvedProject{}
+	if AuthorOwnershipFraction(empty, match, resolver) != 1.0 {
+		t.Error("empty authors should default ownership to 1.0")
+	}
+}
+
+func TestAuthorActualPersonMonths_DeduplicatesAcrossRepos(t *testing.T) {
+	ctx := context.Background()
+
+	repo1 := initGitRepo(t)
```
obey-agent: [P2] These new tests create repositories and commits through initGitRepo and commitFileWithDate, which use host t.TempDir filesystem mutation. The repository policy requires filesystem-mutating workflow tests to run through the container integration harness under tests/integration. Please move these git-backed cases to GetSharedContainer or keep unit tests pure; the current static lint does not detect this indirect helper usage.

internal/leverage/author_filter.go
```diff
@@ -0,0 +1,275 @@
...
+
+	queries := match.Emails
+	if len(queries) == 0 && match.Filter != "" {
+		queries = []string{match.Filter}
+	}
+
+	var found bool
+	for _, q := range queries {
+		f, l, qErr := AuthorDateRange(ctx, gitDir, q)
+		if qErr != nil {
```
obey-agent, line 241: [P1] Preserve cancellation and operational git errors instead of treating every query failure as no commits. AuthorDateRange wraps failures from git log, including context cancellation, but this loop discards all qErr values. If the command is cancelled or a repo becomes inaccessible, ComputeProjectScore falls back to the minimum 0.1 PM and can emit fabricated leverage while ignoring the failure. Suppress only the expected no-matching-commits case and return context or execution errors.

internal/leverage/author_filter.go
```diff
@@ -0,0 +1,275 @@
...
+			m.AuthorIDs[id] = true
+			for _, email := range identity.Emails {
+				addEmail(email)
+			}
+		}
+	}
+
+	// Always keep the raw filter for git substring match (covers unlisted emails
+	// and partial strings like "lancekrogers").
+	addEmail(filter)
```
obey-agent, line 97: [P1] Do not retain the raw filter after a configured identity match. For example, with configured IDs alice and malice, --author alice adds the alice group emails and then adds alice as a git substring; git log and MatchesAuthor therefore include malice commits and blame lines as well. That defeats the stated identity-group expansion and makes both actual PM and ownership-scaled estimates incorrect. Only use the raw substring fallback when no configured identity matched, or make MatchesAuthor honor the matched AuthorIDs exclusively.

obey-agent (commented):
Verdict: Request Changes

Overview: This PR addresses the multi-count actual person-month problem and adds useful identity-group and ownership-aware scoring. The targeted internal/leverage and cmd/camp/leverage tests pass, but the implementation still has correctness, cancellation, performance, and test-isolation issues that prevent production approval.

Key Findings:
- P1: Configured identity matches still append the raw filter, so substring collisions can include unrelated authors in git spans and blame ownership.
- P1: Author-date query errors, including cancellation and repository failures, are discarded and can produce a fabricated minimum-effort score.
- P1: Combining --author with --by-author reapportions an already author-scaled estimate across every author in the table.
- P2: The new git-backed tests mutate host temp directories instead of using the required container integration harness.
- P2: Per-author actual PM recomputes git logs for every author, repository, and email, creating an avoidable subprocess explosion.

What Is Done Well:
- The PR identifies the main aggregation defect and deduplicates actual calendar spans by GitDir.
- Ownership scaling and alternate-email expansion are covered by focused tests, and the documentation explains the new personal formula.

Staff Standard: No. I would not submit this as my own work until the five findings above are resolved and the command-level behavior is rechecked.

internal/leverage/author_filter.go
```diff
@@ -0,0 +1,448 @@
...
+func AuthorHasCommitsMatch(ctx context.Context, gitDir string, match AuthorMatch) (bool, error) {
+	if err := ctx.Err(); err != nil {
+		return false, err
+	}
+	queries := match.Emails
+	if len(queries) == 0 && match.Filter != "" {
+		queries = []string{match.Filter}
+	}
+	for _, q := range queries {
+		ok, err := AuthorHasCommits(ctx, gitDir, q)
```
obey-agent, line 196: [P1] The preflight still swallows cancellation and operational git failures. AuthorHasCommits returns false with a nil error for any exec failure, including CommandContext cancellation, so this loop treats a cancelled or broken repository as no matching commits and the campaign path can continue with projects silently removed. The date-range path now propagates errors, but this earlier author-presence path needs the same treatment.

internal/leverage/author_filter.go
```diff
@@ -0,0 +1,448 @@
...
+
+// identityMatchesFilter reports whether a configured identity matches filter
+// (already lowercased). Matching is intentionally stricter than raw git
+// --author substring so short filters do not collide across identities.
+func identityMatchesFilter(id string, identity AuthorIdentity, filterLower string) bool {
+	if filterLower == "" {
+		return false
+	}
+	if strings.ToLower(id) == filterLower {
+		return true
```
obey-agent, line 109: [P2] Full display names are not matched even though the documentation presents display-name matching as supported. containsWord splits Alice Smith into two tokens and compares each token to the entire filter, so --author "Alice Smith" falls through to the ad-hoc path and can hit the partial-filter bug above. Support an exact normalized display-name match before token matching, or document that only one display-name word is accepted.

cmd/camp/leverage/main_command.go
```diff
@@ -53,6 +53,10 @@ func runLeverage(cmd *cobra.Command, args []string) error {
 	authorFilter, _ := cmd.Flags().GetString("author")
 	byAuthor, _ := cmd.Flags().GetBool("by-author")
 
+	if authorFilter != "" && byAuthor {
```
obey-agent, line 59: [P1] Validate the effective author mode after applying the configured default. This guard runs before cfg.AuthorEmail is loaded; in a campaign with AuthorEmail configured, camp leverage --by-author silently sets authorFilter from config later and computes personal scores while displaying the full by-author table. That contradicts the new mutual-exclusion contract. Either make explicit --by-author override the configured personal default or perform the conflict check after resolving defaults; apply the same fix in directory mode.

internal/leverage/author_filter.go
```diff
@@ -0,0 +1,448 @@
...
+		return m
+	}
+
+	// No configured identity: raw filter is the sole git author query.
+	addEmail(filter)
+
+	// Seed AuthorIDs from resolver so ownership lookups work for exact emails
+	// that are not grouped. Skip excluded identities.
+	if resolver != nil && !resolver.IsExcluded(filter) {
+		id := resolver.Resolve(filter)
```
obey-agent: [P1] Only seed AuthorIDs for an exact known identity, not every ad-hoc filter. With no matching authors.json entry, --author alice@co now creates AuthorIDs[alice@co], which forces AuthorActualPersonMonths onto CollectMergedAuthorSpans. That map is keyed by full canonical commit emails, so partial filters such as the documented alice@co form find no author and return the 0.1 minimum, even though each project score still finds commits through the raw git query. Leave AuthorIDs empty for ad-hoc filters or prove the filter is an exact configured email before using the fast path.

internal/leverage/author_filter_test.go
```diff
@@ -0,0 +1,237 @@
...
+	cfg := &AuthorConfig{
+		Authors: map[string]AuthorIdentity{
+			"alice":  {DisplayName: "Alice", Emails: []string{"alice@example.com"}},
+			"malice": {DisplayName: "Malice", Emails: []string{"malice@example.com"}},
+		},
+	}
+	resolver := NewAuthorResolver(cfg)
+	match := ExpandAuthorFilter(resolver, "alice")
+
+	if !match.MatchesAuthor(resolver, "alice@example.com") {
```
obey-agent, line 127: [P2] These tests still create repositories and commits through initGitRepo and commitFileWithDate in host t.TempDir directories. The project policy requires filesystem-mutating workflow tests to run through the container integration harness under tests/integration. The new integration cases are good, but these host-side cases remain part of this PR and should be moved or reduced to pure unit tests.

obey-agent (commented):
Verdict: Request Changes

Overview: The follow-up commit resolves the original raw-collision, flag-combination, date-range error, and by-author performance findings. Targeted unit, vet, and integration tests pass, but several edge cases still make the reported personal leverage incorrect or make cancellation unsafe.

Key Findings:
- P1: Ad-hoc partial author filters are incorrectly promoted to canonical IDs, causing campaign actual PM to fall back to 0.1 even when project queries find matching commits.
- P1: The mutual-exclusion check runs before the configured AuthorEmail default is applied, so --by-author can still run in personal mode in configured campaigns.
- P1: Author presence checks still convert CommandContext cancellation and operational git failures into no-commit exclusions.
- P2: Host-mutating git tests remain in the new unit test file despite the containerized filesystem-test policy.
- P2: Full display-name filters such as Alice Smith do not match the documented identity expansion.

What Is Done Well:
- The main collision and raw-filter issues from the first review are addressed.
- The new merged-span implementation removes the per-author repository subprocess explosion and adds useful container integration coverage.
- Error handling in the date-range aggregation path is substantially improved.

Staff Standard: Not yet. I would not approve until the remaining personal-filter and cancellation paths are corrected and the new filesystem-mutating tests are moved to the integration harness.

PR #482, opened 2026-07-23
fix: restore Camp integration suite consistency

lancekrogers:
## Summary

- align machine-discovery integration expectations with the intentional `ssh-agent` default while retaining explicit `tailscale-ssh` coverage
- preserve the exact worktree `camp fresh` detaches while reclaiming the default branch
- continue pruning unrelated merged detached worktrees

## Root cause

The dual-auth change updated the CLI contract and unit coverage but left two older integration assertions expecting `tailscale-ssh`. Separately, the Fresh reclaim flow detached the worktree holding the default branch, after which the same run immediately classified that new detached worktree as merged and removed it.

## Validation

- original three failing integration tests pass
- adjacent machine-discovery and Fresh worktree-pruning regression cases pass
- `just gate-fast` passes, including stable/dev builds, vet, lint, and 3,669 unit tests
- full container integration suite passes: 537/537 tests in 449.86s
- `git diff --check` passes

### DIFF
--- a/internal/commands/fresh/fresh.go
+++ b/internal/commands/fresh/fresh.go
@@ -293,6 +293,13 @@ func executeFresh(ctx context.Context, name, path string, opts freshOptions) err
 			BaseRef:       syncState.baseRef,
 			RefreshRemote: true,
 		}
+		// Reclaiming the default branch detaches its former worktree. Preserve
+		// that exact worktree during this prune pass: fresh created the detached
+		// state as a safe branch handoff, so it must not immediately classify
+		// the worktree as merged and remove it.
+		if syncState.reclaimed && syncState.worktreePath != "" {
+			pruneOpts.PreserveDetachedWorktrees = []string{syncState.worktreePath}
+		}
 		pr := prune.Execute(ctx, name, path, pruneOpts)
 		if pr.Error != "" {
 			return camperrors.Wrapf(errors.New(pr.Error), "prune merged branches")
--- a/internal/prune/prune.go
+++ b/internal/prune/prune.go
@@ -56,6 +56,12 @@ type Options struct {
 
 	// Remote name used when RefreshRemote is true. Empty means "origin".
 	RemoteName string
+
+	// PreserveDetachedWorktrees lists detached worktree paths that this prune
+	// pass must retain even when their commits are merged into BaseRef. This is
+	// used when a caller detached a worktree as part of the surrounding
+	// operation and must not immediately undo that work by deleting it.
+	PreserveDetachedWorktrees []string
 }
 
 // Result holds the outcome for a single branch.
@@ -279,6 +285,11 @@ func deleteDetachedWorktrees(ctx context.Context, path, baseRef string, opts Opt
 	}
 	if gitDir, err := git.Output(ctx, path, "rev-parse", "--absolute-git-dir"); err == nil {
 		skipPaths[filepath.Clean(gitDir)] = struct{}{}
+	}
+	for _, preserved := range opts.PreserveDetachedWorktrees {
+		if preserved = strings.TrimSpace(preserved); preserved != "" {
+			skipPaths[filepath.Clean(preserved)] = struct{}{}
+		}
 	}
 
 	removedAny := false
--- a/tests/integration/fresh_test.go
+++ b/tests/integration/fresh_test.go
@@ -530,8 +530,10 @@ git -C %[1]s worktree add %[2]s main
 	branchOnStuck := tc.GitOutput(t, stuckWT, "rev-parse", "--abbrev-ref", "HEAD")
 	require.Equal(t, "main", branchOnStuck, "precondition: stuck worktree should hold main")
 
-	// --no-prune so the reclaimed worktree is not removed after detach
-	// (prune deletes clean detached worktrees that match merged history).
+	// --no-prune keeps this test focused on the reclaim handoff itself. The
+	// reclaim-then-prune preservation contract (the detached worktree surviving
+	// the same run's prune pass) is locked by
+	// TestFresh_ReclaimedWorktreePreservedDuringPrune.
 	output, err := tc.RunCampInDir(projectDir, "fresh", "--no-branch", "--no-push", "--no-follow-up", "--no-prune")
 	require.NoError(t, err, "camp fresh should reclaim main:\n%s", output)
 	assert.Contains(t, output, "Free", "fresh should report freeing the default branch")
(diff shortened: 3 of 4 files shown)


internal/commands/fresh/fresh.go
```diff
@@ -293,6 +293,13 @@ func executeFresh(ctx context.Context, name, path string, opts freshOptions) err
 			BaseRef:       syncState.baseRef,
 			RefreshRemote: true,
 		}
+		// Reclaiming the default branch detaches its former worktree. Preserve
+		// that exact worktree during this prune pass: fresh created the detached
+		// state as a safe branch handoff, so it must not immediately classify
+		// the worktree as merged and remove it.
+		if syncState.reclaimed && syncState.worktreePath != "" {
```
obey-agent, line 300: **Request changes — missing regression lock for this exact path.**

This preserve wiring is the right fix for the reclaim-then-prune race described in the PR body. The production change looks correct.

However, `tests/integration/fresh_test.go::TestFresh_ReclaimsDefaultBranchFromOtherWorktree` still runs with `--no-prune` and still comments:

```
// --no-prune so the reclaimed worktree is not removed after detach
// (prune deletes clean detached worktrees that match merged history).
```

That comment documents the old bug. After this PR, the default prune path is supposed to keep the reclaimed worktree — but nothing asserts that. If someone later drops `PreserveDetachedWorktrees` or only passes it on dry-run, unit gates and the existing reclaim test can still pass.

Please either:
1. Drop `--no-prune` from that test and assert the reclaimed worktree still exists (detached) after default prune, or
2. Add a sibling integration test that reclaims with prune enabled and proves both: reclaimed path kept, unrelated merged detached worktrees still removed.

Without that, this fix is production-correct but not permanently locked.

internal/prune/prune.go
```diff
@@ -56,6 +56,12 @@ type Options struct {
 
 	// Remote name used when RefreshRemote is true. Empty means "origin".
 	RemoteName string
+
+	// PreserveDetachedWorktrees lists detached worktree paths that this prune
+	// pass must retain even when their commits are merged into BaseRef. This is
+	// used when a caller detached a worktree as part of the surrounding
+	// operation and must not immediately undo that work by deleting it.
+	PreserveDetachedWorktrees []string
```
obey-agent, line 64: **Maintainability:** the new option is well-documented and the skip-path integration is clean.

Consider a thin host-side unit test (even table-driven over a fake List seam if available, or a small container case) that `PreserveDetachedWorktrees` keeps a merged detached path while still removing a non-preserved sibling. Right now `internal/prune` only has a stub `TestOptions_DiscardDirty`, so this new contract is only enforced transitively via fresh integration — which currently still disables prune on the reclaim case (see comment on fresh.go).

Not a blocker by itself; the integration-test update is the higher-value lock.

tests/integration/machine_discover_test.go
```diff
@@ -64,28 +64,28 @@ func TestMachineAddDiscoverYes(t *testing.T) {
 	row, ok := findMachine(payload.Machines, "devbox")
 	require.True(t, ok, "devbox missing from list --json: %+v", payload.Machines)
 	require.Equal(t, "devbox.example-net.ts.net", row.Host)
-	require.Equal(t, "tailscale-ssh", row.AuthMethod)
-	require.Empty(t, row.SSHUser, "tailscale-ssh needs no ssh_user")
-	require.Empty(t, row.IdentityFile, "tailscale-ssh needs no identity_file")
+	require.Equal(t, "ssh-agent", row.AuthMethod)
```
obey-agent, line 67: **Looks correct.** Aligning `--discover --yes` with the dual-auth OpenSSH default (`ssh-agent`) matches unit coverage in `cmd/camp/machine_discover_test.go` and the published CLI contract.

Keeping `TestMachineAddDiscoverByID` on explicit `--auth tailscale-ssh` is the right split: default path + opt-in Tailscale path both stay covered, and upsert idempotency is still exercised. No change requested here.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Small, well-scoped regression fix: machine-discovery integration assertions now match the dual-auth `ssh-agent` default, and `camp fresh` no longer immediately prunes the worktree it just detached while reclaiming the default branch. Production code and comments are clear and minimal. The reclaim/prune fix is correct, but the suite still does not lock the failure mode this PR claims to fix.

## Key Findings
- **Missing regression lock (blocking):** `TestFresh_ReclaimsDefaultBranchFromOtherWorktree` still uses `--no-prune` with a comment that documents the old delete-after-detach bug. After this change, reclaim + default prune should preserve that worktree — but no test asserts it. If `PreserveDetachedWorktrees` is dropped later, the suite can stay green while the original failure returns. See inline on `internal/commands/fresh/fresh.go`.
- **Thin unit coverage for the new prune contract:** `PreserveDetachedWorktrees` is only exercised transitively (and currently not on the reclaim path). Prefer updating the reclaim integration test; optional host-side coverage in `internal/prune` is a nice secondary lock. See inline on `internal/prune/prune.go`.
- **Machine discovery tests:** Correct and complete for this PR — default OpenSSH path + explicit `tailscale-ssh` override. No change requested. See inline on `tests/integration/machine_discover_test.go`.

## What's Done Well
- Root-cause analysis matches the code: dual-auth contract drift in older integration expectations, plus reclaim → prune undoing its own handoff.
- Preserve API is caller-driven and surgical — only the exact reclaimed path is retained; unrelated merged detached worktrees still prune.
- Comments on both the option and the fresh call site explain *why*, not just *what*.
- Machine-discovery update preserves Tailscale coverage instead of deleting it.

## Suggested follow-up (before merge)
1. Drop `--no-prune` from `TestFresh_ReclaimsDefaultBranchFromOtherWorktree` (or add a sibling), and assert the reclaimed worktree remains after prune.
2. Remove/rewrite the outdated comment that says prune deletes the reclaimed worktree.
3. Optionally assert an unrelated merged detached sibling is still removed in the same run.

## Staff Standard
Almost — the production fix is something I'd ship. I would not be proud of merging a bugfix without a test that fails without the fix and passes with it. Land the reclaim+prune regression lock, then this is merge-ready.

lancekrogers:
Thanks for the review. Addressed in e204d9f6.

**Blocking finding (regression lock).** You were right that nothing pinned this exact path. Added `TestFresh_ReclaimedWorktreePreservedDuringPrune` (your option 2): it reclaims the default branch with prune **enabled** (no `--no-prune`) and, alongside the reclaimed worktree, plants an unrelated clean/merged/detached sibling. It asserts both halves of the contract:

- the reclaimed (now detached) worktree **survives** the same run's prune pass, and
- the unrelated merged detached sibling is **still removed** (proving `PreserveDetachedWorktrees` is scoped to the reclaimed path, not a blanket prune disable).

I verified it fails without the fix and passes with it: neutralizing the preserve wiring in `fresh.go` makes it fail with `reclaimed worktree must be preserved during prune, not deleted`, and restoring it makes it pass. I also rewrote the misleading `--no-prune` comment on `TestFresh_ReclaimsDefaultBranchFromOtherWorktree` that documented the old delete-after-detach bug; that test stays focused on the reclaim handoff, and the preservation contract now lives in the new sibling.

**Secondary finding (host-side `internal/prune` unit test).** I deliberately did not add one. `deleteDetachedWorktrees` operates on real `git worktree` state with no List seam to fake, so a host-side test would have to create and mutate real worktrees, which this repo's own `lint-no-host-fs-tests` gate forbids (FS-mutating tests must be containerized). The containerized integration lock above is the correct home for this contract. If you'd still like belt-and-suspenders, I can extract a pure `shouldPreserve(path, list)` helper and table-test it, but that would only cover a trivial `filepath.Clean` map lookup, so I left it out unless you feel strongly.

**Verification.** All 22 `TestFresh*` integration tests green. `lint-no-host-fs-tests` clean (no new violators), and `fresh_test.go` is clean under golangci-lint. Machine-discovery assertions unchanged per your note.

tests/integration/fresh_test.go
```diff
@@ -545,3 +547,61 @@ git -C %[1]s worktree add %[2]s main
...
 	assert.Equal(t, "HEAD", stuck, "occupying worktree should be detached, got %s", stuck)
 }
+
+// TestFresh_ReclaimedWorktreePreservedDuringPrune locks the reclaim-then-prune
+// contract. Reclaiming the default branch detaches its former worktree at merged
+// history, so the same run's prune pass would immediately delete it unless fresh
+// preserves that exact path (PreserveDetachedWorktrees in fresh.go). With prune
+// ENABLED, the reclaimed worktree must survive while an unrelated merged detached
+// worktree is still removed. This fails without the preserve wiring.
+func TestFresh_ReclaimedWorktreePreservedDuringPrune(t *testing.T) {
```
obey-agent, line 557: **Resolved / LGTM.** This is exactly the regression lock the previous review asked for.

Strengths:
- Prune is **enabled** (the real failure path).
- Dual assertion: reclaimed path kept **and** unrelated merged detached sibling removed — so a total prune disable would still fail.
- Commit message notes it was verified to fail without the `fresh.go` preserve wiring and pass with it.

No further change requested here.

obey-agent (commented):
## Verdict: Approve

## Overview
Follow-up commit addresses the prior review cleanly. Production fix was already correct; the suite now locks the reclaim-then-prune preservation contract with a dual-assertion integration test that fails without the preserve wiring and passes with it. Machine-discovery alignment remains correct.

## Prior findings — disposition
- **Missing reclaim+prune regression lock** → **Resolved** via `TestFresh_ReclaimedWorktreePreservedDuringPrune` (prune enabled; reclaimed path kept; unrelated merged detached still removed). Misleading `--no-prune` comment on the original reclaim test rewritten to point at the new lock.
- **Optional host-side prune unit test** → **Accepted as deferred.** Integration coverage is the high-value lock for this path; not blocking.
- **Machine discovery defaults** → Unchanged and still correct.

## Key Findings
- No blocking issues on HEAD (`e204d9f`).
- New test design is solid: dual assertion prevents false greens from either total prune disable or preserve-everything mistakes.

## What's Done Well
- Minimal, surgical production change (`PreserveDetachedWorktrees` only for the reclaimed path).
- Regression test written to prove both sides of the contract in one run.
- Honest commit message documenting the review gap and the fail/pass verification.

## Staff Standard
Yes — I would be proud to submit this as my own work and would merge it.

Approved.

PR #484, opened 2026-07-23
feat(workitem): add camp workitem rename verb

lancekrogers:
## What this adds

`camp workitem rename <selector> <new-name>` renames a workitem's directory or
file basename while keeping its identity coherent. Only the path basename
changes: the stable id, ref, title, workflow type, and lifecycle status are
untouched. Every path-derived reference is repaired in the same commit as the
move.

Flags: `--dry-run`, `--json`, `--no-commit`. The command carries
`agent_allowed: "true"` annotations (matching `promote.go`) because it is fully
specified by positional args and flags with no interactive prompt.

Motivation: workitems could not be renamed without a manual `mv` plus by-hand
fixups of every reference, because the workitem `Key` is path-derived and
several stores key off that path.

## Design principle

Rename is a move within the same parent, so its fixup surface must equal the
fixup surface of a physical move. Rather than reimplement, the command reuses
the exact primitives `promote` and `camp dungeon move` already use:
`statusmove.Move` for the physical relocation and `mdlinks.RewriteForMove` for
markdown link rewriting, then re-keys the path-derived stores the way `gather`
does when it relocates workitems.

## Enumerated fixup surface (each verified against source)

1. **Physical move** via `statusmove.Move` with `BoundaryRoot` set
   (`internal/statusmove/move.go`). No dated bucket. `noReplaceMove` gives
   collision safety at the OS level; the command also pre-checks the
   destination and returns a clean "already exists" error. Handles both files
   and directories.
2. **Markdown links** via `mdlinks.RewriteForMove`
   (`internal/mdlinks/rewrite.go`): rewrites both the moved subtree's own
   relative links and other campaign files that pointed at the moved item, and
   returns the campaign-relative modified files to stage. A pure sibling rename
   leaves same-depth relative links unchanged (correct) and rewrites external
   references into the renamed item.
3. **Priority + attention store** `.campaign/settings/workitems.json`, keyed by
   `WorkItem.Key` (`internal/workitem/priority/store.go`): the old key's manual
   priority and attention (stage + group) entries are re-keyed 1:1 to the new
   key (`rekeyRenamePriorities`, `rename_migrate.go`). This file is gitignored
   per-machine state (`internal/scaffold/init.go` writes `settings/workitems.json`
   into `.campaign/.gitignore`), so it is migrated on disk and never staged,
   matching how `gather` treats the same file.
4. **Link registry** `.campaign/workitems/links.yaml`
   (`internal/workitem/links/`): `rehomeRenameLinks` migrates the subject
   links' `workitem_key` (old key -> new key; `workitem_id`/stable id is left
   unchanged) and rewrites any `scope.path` that targets the renamed directory
   exactly or as a subtree. links.yaml is tracked, so a change is staged into
   the commit.
5. **Current-workitem pointer** `.campaign/workitems/current.yaml`
   (`internal/commands/workitem/current.go`): `current.yaml` stores the stable
   id when the item has one, otherwise the path `Key`. The rename preserves the
   stable id, so only the key-fallback case (unadopted item) needs migration;
   `migrateRenameCurrent` updates it. current.yaml is also gitignored per-machine
   and is not staged.
6. **Workitem audit ledger** `.campaign/workitems/.workitems.jsonl`: an
   `EventMove` record is appended (`stageAuditEvent`), the same event type
   `camp dungeon move` uses for relocations
   (`internal/workitem/audit/audit.go` doc for `EventMove`). It is staged only
   when the file exists on disk, because `git add` hard-fails on a
   never-created pathspec (same guard as `cmd/camp/dungeon/workitem_ledger.go`).
7. **Navigation index**: `navindex.Delete(root)` invalidates the nav cache, as
   promote does.
8. **Auto-commit** via `dungeoncmd.StageAndCommitDungeonMove` (the same helper
   promote uses): pre-stages the source deletion so git records a rename, and
   commits the destination, rewritten markdown, links.yaml, and the audit log
   together. `--no-commit` skips it.

`WorkItem.Key` is always `"<prefix>:" + RelativePath` across every discover path
(`discover_workflow_docs.go`, `discover_frontmatter.go`, `discover_intents.go`,
`discover_festivals.go`), so the new key is derived by swapping the trailing
path segment (`renameKey`); the derivation errors out rather than guessing if
that invariant ever fails to hold.

## Design decisions and rationale

- **Festivals and intents are rejected** (`ensureRenamable`). A festival's
  identity is owned by the fest CLI and an intent has its own title-based
  rename (`internal/intent/rename.go`); relocating their directories through a
  generic mover would desync that state. Every other workflow type
  (design, explore, custom) and file workitems are supported.
- **File workitems are supported** because the shared machinery already handles
  files cheaply (`statusmove` Lstat's any source, `mdlinks` handles a single
  `.md`, the `file:` key re-keys like any other). The original extension is
  preserved: a new name with no extension gets the original appended, a
  matching extension is kept, and a conflicting extension is rejected
  (`renameBasename`). Note: because the selector's slug matcher compares against
  the full basename, a file workitem is selected by its full filename
  (e.g. `note.md`), its stable id, ref, key, or path, not by an
  extension-stripped stem. This is pre-existing selector behavior, unchanged
  here.
- **Dungeon items are allowed.** Renaming an item already inside a dungeon falls
  out of the machinery naturally (statusmove + mdlinks + key migration all work
  on the dungeon path), so it is permitted rather than special-cased.
- **Quest links are intentionally NOT handled.** I verified that promote, the
  dungeon move path, gather, and mdlinks never touch quest link paths
  (`internal/quest/types.go` `Link.Path`); a workitem relocation already leaves
  quest links stale today. This is a pre-existing gap across all move flows, not
  something rename introduces, so rename matches the existing behavior and does
  not add divergent quest handling.
- **The `.workitem` marker / frontmatter is not restamped.** No stamped field is
  re-derived from the slug (`internal/workitem/metadata.go`): the id is minted at
  creation and frozen, the title is independent. Preserving stable_id therefore
  means simply not touching the marker.
- **No ledgerkit event is emitted.** A rename is a relocation, not a lifecycle
  transition, and there is no `KindRenamed` in `pkg/ledgerkit`; misusing
  `KindTransitioned` or adding a new kind (which ripples into camp-graph
  consumers) is unwarranted. The workitem audit log's purpose-built `EventMove`
  is the correct surface, matching `camp dungeon move`.
- **JSON contract** uses the `jsoncontract` envelope with schema version
  `workitem-rename/v1alpha1`, matching the newer subcommands (create, current,
  priority, stage) rather than promote's bare object, so agent callers get a
  stable, versioned error envelope.

## Tradeoffs and deferred

- Git branches and worktrees previously derived from the old workitem name are
  intentionally not renamed; existing branches keep their names.
- The priority store and current.yaml are migrated on disk but not committed,
  because they are gitignored per-machine state. In a campaign that force-tracks
  them (against the generated `.gitignore`), a rename would leave those files
  dirty. This matches gather's existing treatment and is the correct behavior
  for the standard gitignored layout.
- Quest link staleness on move is left as the documented pre-existing gap above.

## Test coverage

Unit tests (`internal/commands/workitem/rename_test.go`), table-driven, error
cases first: `renameBasename` (extension handling), `renameKey` (key
derivation invariant), `ensureRenamable` (festival/intent rejection),
`rewriteScopePath`, `rekeyRenamePriorities` (the priority migration proof,
including attention stage + group), and `rehomeRenameLinks` (key migration with
stable-id preserved, scope-path subtree rewrite).

Integration tests, containerized with the `integration` build tag
(`tests/integration/workitem_rename_test.go`): end-to-end reference repair
(directory move + external markdown link + priority store re-key + clean git
tree + rename commit), links.yaml repair (key + scope path, stable id
preserved), `--json` shape, `--dry-run` (no change, no commit), `--no-commit`
(moved on disk, HEAD unchanged), sibling collision rejection, and file-workitem
rename with extension preservation.

Gates run from the worktree, all passing: `just build` (stable + dev),
`go vet ./...` (stable, dev, integration), `just lint` (stable + dev; golangci
`--new-from-merge-base`, the fmt.Errorf ratchet, and the host-fs-tests ratchet
all clean), `just test unit` (3643/3643 stable, 3675/3675 dev), the 7 rename
integration tests, and `just docs` (regenerated; the only doc changes are the
new `camp_workitem_rename.md` and the rename entries in `camp_workitem.md` and
`camp-reference.md`, no unrelated drift).

## Real CLI output

Dry run (human):

```
dry-run: would rename workitem timeline (directory) to workflow/design/release-timeline
```

Dry run (`--json`):

```json
{
  "schema_version": "workitem-rename/v1alpha1",
  "generated_at": "2026-07-23T22:50:39.084564Z",
  "workitem": {
    "id": "design-timeline-fixed",
    "key": "design:workflow/design/release-timeline",
    "type": "design",
    "item_kind": "directory",
    "from": "workflow/design/timeline",
    "to": "workflow/design/release-timeline",
    "committed": false,
    "links_updated": false,
    "priority_migrated": false,
    "current_updated": false
  }
}
```

Real rename with a link present (`--json`):

```json
{
  "schema_version": "workitem-rename/v1alpha1",
  "generated_at": "2026-07-23T22:51:31.678282Z",
  "workitem": {
    "id": "design-alpha-fixed",
    "key": "design:workflow/design/alpha-renamed",
    "type": "design",
    "item_kind": "directory",
    "from": "workflow/design/alpha",
    "to": "workflow/design/alpha-renamed",
    "committed": true,
    "commit_message": "Committed changes to git",
    "links_updated": true,
    "priority_migrated": false,
    "current_updated": false
  }
}
```

links.yaml before and after that rename (subject key and scope path migrated,
stable id preserved):

```yaml
# before
  - id: lnk_20260723_1ef657
    workitem_id: design-alpha-fixed
    workitem_key: design:workflow/design/alpha
    scope:
      kind: campaign_path
      path: workflow/design/alpha/sub
    role: primary
# after
  - id: lnk_20260723_1ef657
    workitem_id: design-alpha-fixed
    workitem_key: design:workflow/design/alpha-renamed
    scope:
      kind: campaign_path
      path: workflow/design/alpha-renamed/sub
    role: primary
```

Rename commit (git records the directory move as a rename and stages the
repaired references together):

```
 .campaign/workitems/.workitems.jsonl                     | 1 +
 docs/note.md                                             | 2 +-
 workflow/design/{timeline => release-timeline}/.workitem | 0
 workflow/design/{timeline => release-timeline}/README.md | 0
 workflow/design/{timeline => release-timeline}/sub/x.md  | 0
```

### DIFF
--- a/cmd/camp/zz_manifest_annotations.go
+++ b/cmd/camp/zz_manifest_annotations.go
@@ -94,6 +94,7 @@ var manifestAgentAllowedReasons = map[string]string{
 	"workitem group":             "Non-interactive group update with explicit selector",
 	"workitem priority":          "Non-interactive priority update with explicit selector",
 	"workitem promote":           "Non-interactive promote fully specified by --target and flags",
+	"workitem rename":            "Non-interactive rename fully specified by positional args and flags",
 	"workitem repair":            "Idempotent, non-destructive marker repair with --json and --dry-run",
 	"workitem resolve":           "Read-only workitem context resolution",
 	"workitem stage":             "Non-interactive attention-stage update with explicit selector",
--- a/docs/cli-reference/camp-reference.md
+++ b/docs/cli-reference/camp-reference.md
@@ -7567,6 +7567,47 @@ camp workitem promote [id] --target <target> [flags]
 ```
 ---
 
+## camp workitem rename
+
+Rename a workitem and repair references
+
+### Synopsis
+
+Rename the workitem matched by <selector> so its directory (or file)
+basename becomes <new-name>. Identity is preserved: the stable id, ref, title,
+type, and lifecycle status do not change; only the path basename moves.
+
+References are repaired in the same commit as the move:
+  - relative markdown links pointing at the workitem are rewritten
+  - the workitem link registry (links.yaml) key and any scope paths under the
+    renamed directory are updated
+  - manual priority and attention-stage entries are re-keyed on disk
+  - the current-workitem pointer is updated when it referenced the old path key
+
+Festivals and intents are managed by their own tooling and cannot be renamed
+here. For file workitems, pass the full new filename; the original extension is
+kept when omitted.
+
+```
+camp workitem rename <selector> <new-name> [flags]
+```
+
+### Options
+
+```
+      --dry-run     Print the planned rename, change nothing
+  -h, --help        help for rename
+      --json        Output result as a single JSON object
+      --no-commit   Skip the auto-commit
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+---
+
 ## camp workitem repair
 
 Repair a workflow directory into a workitem
--- a/docs/cli-reference/camp_workitem.md
+++ b/docs/cli-reference/camp_workitem.md
@@ -59,6 +59,7 @@ camp workitem [flags]
 * [camp workitem list](camp_workitem_list.md)	 - List or browse filtered workitems
 * [camp workitem priority](camp_workitem_priority.md)	 - Set or clear the manual priority
 * [camp workitem promote](camp_workitem_promote.md)	 - Promote a workitem to a festival, doc, or dungeon
+* [camp workitem rename](camp_workitem_rename.md)	 - Rename a workitem and repair references
 * [camp workitem repair](camp_workitem_repair.md)	 - Repair a workflow directory into a workitem
 * [camp workitem resolve](camp_workitem_resolve.md)	 - Print the workitem for the current context
 * [camp workitem stage](camp_workitem_stage.md)	 - Set or clear the attention stage
--- a/docs/cli-reference/camp_workitem_rename.md
+++ b/docs/cli-reference/camp_workitem_rename.md
@@ -1,0 +1,43 @@
+## camp workitem rename
+
+Rename a workitem and repair references
+
+### Synopsis
+
+Rename the workitem matched by <selector> so its directory (or file)
+basename becomes <new-name>. Identity is preserved: the stable id, ref, title,
+type, and lifecycle status do not change; only the path basename moves.
+
+References are repaired in the same commit as the move:
+  - relative markdown links pointing at the workitem are rewritten
+  - the workitem link registry (links.yaml) key and any scope paths under the
+    renamed directory are updated
+  - manual priority and attention-stage entries are re-keyed on disk
+  - the current-workitem pointer is updated when it referenced the old path key
+
+Festivals and intents are managed by their own tooling and cannot be renamed
+here. For file workitems, pass the full new filename; the original extension is
+kept when omitted.
+
+```
+camp workitem rename <selector> <new-name> [flags]
+```
+
+### Options
+
+```
+      --dry-run     Print the planned rename, change nothing
+  -h, --help        help for rename
+      --json        Output result as a single JSON object
+      --no-commit   Skip the auto-commit
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+
+### SEE ALSO
+
+* [camp workitem](camp_workitem.md)	 - View active campaign work items
--- a/internal/commands/workitem/json_contract.go
+++ b/internal/commands/workitem/json_contract.go
@@ -25,4 +25,6 @@ const (
 	WorkitemRepairJSONVersion = "workitem-repair/v1alpha1"
 	// WorkitemGatherJSONVersion is the schema version of camp gather <type> --json.
 	WorkitemGatherJSONVersion = "workitem-gather/v1alpha1"
+	// WorkitemRenameJSONVersion is the schema version of camp workitem rename --json.
+	WorkitemRenameJSONVersion = "workitem-rename/v1alpha1"
 )
--- a/internal/commands/workitem/rename.go
+++ b/internal/commands/workitem/rename.go
@@ -1,0 +1,259 @@
+package workitem
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/jsoncontract"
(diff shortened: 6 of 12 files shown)


obey-agent:
## Terminal demo (real binary + VHS)

Recorded from the PR branch binary in a disposable fixture. Real rename + markdown fixup on disk.

camp PR #484 real TTY demo

**Journey shown**
1. `camp workitem --list` — item at `workflow/design/old-name`
2. `ls workflow/design/`
3. `camp workitem rename old-name better-name --dry-run`
4. `camp workitem rename old-name better-name --no-commit --json` — identity preserved, path updated
5. Post-rename listing + `OBEY.md` link rewrite verification

**Gist:** https://gist.github.com/lancekrogers/e3fabf969ad3f80e4f820a7e61e5de99 (secret/unlisted)

---

## Staff review

**Verdict:** Approve (comment)

**Overview:** Rename is implemented as a true “move + same fixup surface as promote/dungeon move,” not a bespoke partial rewrite. That is the correct design for a path-keyed workitem system.

**What's done well**
- Reuses `statusmove.Move` + `mdlinks.RewriteForMove` + shared commit staging
- Explicit reject of festivals/intents (identity owned elsewhere)
- Priority/links/current migrations called out, including gitignore vs tracked differences
- Honest callout that quest links stay stale (pre-existing gap) rather than inventing divergent behavior
- JSON contract versioned (`workitem-rename/v1alpha1`)

**Notes / non-blockers**
- Quest-link staleness is the biggest product footgun; tracking it as a follow-up across *all* move paths (not just rename) would be valuable
- Stable id preserved correctly in the demo (`feature-old-name-2026-07-24` after rename) — good operator story

**Staff standard:** Yes — production-shaped relocation command.

obey-agent (commented):
## Staff review

**Verdict:** Approve (comment)

**Overview:** Rename is implemented as a true “move + same fixup surface as promote/dungeon move,” not a bespoke partial rewrite. That is the correct design for a path-keyed workitem system.

**What's done well**
- Reuses `statusmove.Move` + `mdlinks.RewriteForMove` + shared commit staging
- Explicit reject of festivals/intents (identity owned elsewhere)
- Priority/links/current migrations called out, including gitignore vs tracked differences
- Honest callout that quest links stay stale (pre-existing gap) rather than inventing divergent behavior
- JSON contract versioned (`workitem-rename/v1alpha1`)

**Notes / non-blockers**
- Quest-link staleness is the biggest product footgun; tracking it as a follow-up across *all* move paths (not just rename) would be valuable
- Stable id preserved correctly in the demo (`feature-old-name-2026-07-24` after rename) — good operator story

**Staff standard:** Yes — production-shaped relocation command.

PR #485, opened 2026-07-23
feat(workitem): add camp workitem id verb

lancekrogers:
## Summary

Adds `camp workitem id [selector-or-path]`, a read-only command that prints a workitem's durable identifier. It is the identifier sibling of the existing `camp workitem --print` (which prints a path): the bare id goes to stdout so scripts and agents can capture it.

## Output contract

- **Default (bare id):** the durable stable id on stdout, nothing else, exit 0.
- **`--key`:** the path-derived `<type>:<path>` key form instead.
- **`--json`:** a small structured object aligned with the existing WorkItem / resolve vocabulary.

Real captured output (from a throwaway fixture campaign):

```
$ camp workitem id design-example-2026-05-24
design-example-2026-05-24

$ camp workitem id --key design-example-2026-05-24
design:workflow/design/example

$ camp workitem id --json design-example-2026-05-24
{
  "schema_version": "workitem-id/v1alpha1",
  "generated_at": "2026-07-23T23:11:02.326311Z",
  "id": "design-example-2026-05-24",
  "id_kind": "stable",
  "key": "design:workflow/design/example",
  "stable_id": "design-example-2026-05-24",
  "source": "explicit",
  "relative_path": "workflow/design/example"
}
```

## Which identifier is the default (stable id vs key)

There are two identifier forms in the codebase: the durable stable id and the path-derived key (`WorkItem.Key`, always `<type>:<path>`). The stable id is not one uniform thing across all workitem types, so "stable id" here means the single-segment identifier the selector can resolve back to the item, delegated to `internal/workitem/link_target.go:LinkWorkitemID`:

1. an adopted workitem's stable `.workitem` id (`WorkItem.StableID`), when present;
2. a festival's `fest.yaml` metadata id (`WorkItem.SourceID`, e.g. `SC0001`) for festivals, which have no `StableID`;
3. otherwise the path-derived key as a documented fallback.

`id.go:durableID` takes its value from `LinkWorkitemID` so the id printed here cannot drift from what a link stores, and derives an `id_kind` (`stable` / `festival` / `key`) so `--json` consumers can see when an item without a real stable id fell back to its key. `stable_id` is `omitempty`, so its absence is the machine-readable fallback signal too.

Festival and key-fallback `--json` (real output):

```
$ camp workitem id --json SC0001
{ "id": "SC0001", "id_kind": "festival", "key": "festival:festivals/active/social-campaign-SC0001", "source": "explicit", "relative_path": "festivals/active/social-campaign-SC0001", ... }

$ camp workitem id --json workflow/design/plain
{ "id": "design:workflow/design/plain", "id_kind": "key", "key": "design:workflow/design/plain", "source": "explicit", "relative_path": "workflow/design/plain", ... }
```

## How resolution is shared with `resolve`

`id` is a thin veneer over the same resolver machinery `camp workitem resolve` uses; it writes no second resolution path and differs only in output shape, positional-argument handling, and completion registration.

- No argument: `internal/commands/workitem/id.go:runID` calls `internal/workitem/resolver.Resolve` with empty options, so context detection runs the identical tiered pipeline as `resolve` (explicit, cwd ancestor, link, festival, current-workitem pointer). See `internal/workitem/resolver/resolver.go:63`.
- With an argument: `runID` sets `resolver.Options.Explicit`, so the argument flows through the resolver's explicit tier into `internal/workitem/selector.Resolve` (`resolver.go:164`), the same selector family every other workitem command uses (`ref` / `stable id` / `key` / `path` / `slug` / `festival id`).
- Output shape differs deliberately: `resolve` prints `no workitem context` and exits 0; `id` treats "no context and no selector" as an error (nonzero exit) because it is a scripting primitive.

`TestDurableID_MatchesLinkTarget` asserts `durableID`'s value equals `LinkWorkitemID` for stable / festival / key-only items, guarding against drift between the two.

## Filesystem path argument

`id.go:selectorFromArg` translates a positional argument that names an existing on-disk path (absolute or relative to the cwd) into the campaign-relative form the selector's path matcher expects; anything else passes through unchanged as an id/key/slug/festival selector. Both sides are canonicalized with `EvalSymlinks` the same way `resolver.Resolve` canonicalizes its root, so a symlinked campaign resolves consistently. A path that does not exist on disk is treated as a selector and fails loudly rather than silently succeeding.

## Completion wiring

Dynamic shell completion for the positional argument is registered as the command's cobra `ValidArgsFunction` (`completeWorkitemSelector`), reachable through camp's existing `__complete` machinery (verified: `camp __complete workitem id design` returns candidates). It gathers candidates from `discoverWorkitems` (the same gather that serves `camp workitem --json`) and, via the pure, unit-tested `workitemSelectorCandidates`, offers one durable id per item (which covers festival ids), prefix-filtered, deduplicated, sorted, and annotated with the item title as a cobra description. Discovery is bounded by a 200ms timeout so a large campaign cannot stall a keystroke. This adds no shell-template changes, so the `internal/shell` wrapper-transparency tests are unaffected.

## Error behavior

- No context and no argument: clear message on stderr, nonzero exit, never a hang or interactive prompt.

  ```
  $ camp workitem id
  Error: validation error on workitem: no workitem in the current context; pass a selector or run inside a workitem directory
  exit=1
  ```
- Unknown selector / nonexistent path: the resolver's explicit tier fails loudly (selector not found), nonzero exit.
- Ambiguous selector (e.g. a slug matching two items): surfaced as an ambiguity error from the shared selector.
- With `--json`, all of the above render through the existing `jsoncontract` error envelope with `schema_version: workitem-id/v1alpha1`.

## Cobra annotations

`agent_allowed: "true"` with a reason on the command, plus the matching entry in `cmd/camp/zz_manifest_annotations.go` (the hand-maintained allowlist the manifest guard test enforces).

## Test coverage

New tests in `internal/commands/workitem/id_test.go` (host-side fixture trees via the existing `linkTestCampaign` + `chdir` helpers, matching the `resolve`/`selector` test convention; this command is read-only), error cases first:

- no context + no argument is an error; nonexistent selector is an error; nonexistent path is an error; ambiguous slug is an error;
- stable id is the default; `--key` prints the key; cwd inside a workitem resolves via the ancestor tier; an explicit selector beats cwd;
- campaign-relative path resolves; absolute filesystem path resolves; festival durable id is the `fest.yaml` id;
- `--json` shape (all fields) and the key-fallback `--json` case;
- `durableID` matches `LinkWorkitemID` (drift guard);
- `workitemSelectorCandidates` (dedup, prefix filter, sort, title annotation) and the wired `completeWorkitemSelector` (directive + candidates, and no completion after the first positional arg).

Updated the two guard tests a new subcommand is expected to touch: `internal/commands/workitem/workitem_test.go` (registered-subcommand list/count) and the manifest allowlist above.

## Gates

Run from the worktree, all green:

- `just build-camp` (stable) and dev-profile build via `just gate`
- `just gate`: stable + dev build, `go vet` stable/dev/integration, `just lint` stable + dev (0 issues each), unit tests dev (3687 passed) and stable (3655 passed)
- `just docs` regenerated the CLI reference (committed separately as generated output)

No pre-existing failures encountered.

## Tradeoffs and deferred items

- **No interactive picker/TUI.** Completion is the interactive surface. A reusable fuzzy picker is being built concurrently elsewhere; unifying `id` with it can happen later.
- **Read-only.** The command mutates no campaign state; the only writes it could trigger are whatever `discoverWorkitems` already does on read for `camp workitem --json` (it does not write).
- **Interactive shell-wrapper `--path-output`.** The `camp` shell function intercepts `camp workitem ...` on a TTY and appends `--path-output` unless a passthrough flag (`--help`/`--json`/`--print`) is present. This is a pre-existing interaction that affects every workitem subcommand equally (verified: `camp workitem resolve --path-output` already exits 2 today), not something this PR introduces. `id`'s primary consumers are scripts/agents that call the binary directly or pass `--json` (which passes through cleanly), so it is left as-is; making the wrapper pass through recognized subcommands is a separate, broader change.

### DIFF
--- a/cmd/camp/zz_manifest_annotations.go
+++ b/cmd/camp/zz_manifest_annotations.go
@@ -88,6 +88,7 @@ var manifestAgentAllowedReasons = map[string]string{
 	"workitem create":            "Non-interactive workitem creation with explicit slug",
 	"workitem current":           "Non-interactive current-workitem selection and query",
 	"workitem doctor":            "Read path (--json) is safe; never pass --fix from an agent",
+	"workitem id":                "Read-only identifier lookup with a bare-id stdout contract and --json for automation",
 	"workitem link":              "Non-interactive workitem link operation",
 	"workitem links":             "Read-only workitem link listing",
 	"workitem list":              "Read-only filtered listing with non-interactive compact and JSON output",
--- a/docs/cli-reference/camp-reference.md
+++ b/docs/cli-reference/camp-reference.md
@@ -7353,6 +7353,52 @@ camp workitem group <selector> <group|clear> [flags]
 ```
 ---
 
+## camp workitem id
+
+Print the identifier of a workitem
+
+### Synopsis
+
+Print the stable identifier of a workitem.
+
+With no argument, the workitem is detected from the current context using the
+same tiered resolution as `camp workitem resolve` (explicit selector, cwd
+ancestor, linked scope, festival, current-workitem pointer). With an argument,
+the workitem is resolved through the shared selector family: workitem ref,
+stable id, key, campaign-relative path, directory slug, or festival id. A
+filesystem path (absolute or relative to the current directory) is accepted and
+translated to the campaign-relative form the selector expects.
+
+The bare stable id is written to stdout for shell scripting; it is the id sibling
+of `camp workitem --print`, which prints a path. Use --key for the
+path-derived key instead, or --json for a structured object.
+
+Examples:
+  camp workitem id                       # id of the workitem for the cwd
+  camp workitem id design-x-2026-05-24   # echoes back the stable id
+  camp workitem id ./workflow/design/x   # resolves a filesystem path
+  camp workitem id --key                 # print the <type>:<path> key form
+  camp workitem id --json SC0001         # structured object for a festival
+
+```
+camp workitem id [selector-or-path] [flags]
+```
+
+### Options
+
+```
+  -h, --help   help for id
+      --json   emit a structured JSON result
+      --key    print the path-derived key instead of the stable id
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+---
+
 ## camp workitem link
 
 Create a workitem link
--- a/docs/cli-reference/camp_workitem.md
+++ b/docs/cli-reference/camp_workitem.md
@@ -54,6 +54,7 @@ camp workitem [flags]
 * [camp workitem current](camp_workitem_current.md)	 - Get, set, or clear the current workitem
 * [camp workitem doctor](camp_workitem_doctor.md)	 - Report link-registry health issues
 * [camp workitem group](camp_workitem_group.md)	 - Set or clear the group
+* [camp workitem id](camp_workitem_id.md)	 - Print the identifier of a workitem
 * [camp workitem link](camp_workitem_link.md)	 - Create a workitem link
 * [camp workitem links](camp_workitem_links.md)	 - List workitem links
 * [camp workitem list](camp_workitem_list.md)	 - List or browse filtered workitems
--- a/docs/cli-reference/camp_workitem_id.md
+++ b/docs/cli-reference/camp_workitem_id.md
@@ -1,0 +1,48 @@
+## camp workitem id
+
+Print the identifier of a workitem
+
+### Synopsis
+
+Print the stable identifier of a workitem.
+
+With no argument, the workitem is detected from the current context using the
+same tiered resolution as `camp workitem resolve` (explicit selector, cwd
+ancestor, linked scope, festival, current-workitem pointer). With an argument,
+the workitem is resolved through the shared selector family: workitem ref,
+stable id, key, campaign-relative path, directory slug, or festival id. A
+filesystem path (absolute or relative to the current directory) is accepted and
+translated to the campaign-relative form the selector expects.
+
+The bare stable id is written to stdout for shell scripting; it is the id sibling
+of `camp workitem --print`, which prints a path. Use --key for the
+path-derived key instead, or --json for a structured object.
+
+Examples:
+  camp workitem id                       # id of the workitem for the cwd
+  camp workitem id design-x-2026-05-24   # echoes back the stable id
+  camp workitem id ./workflow/design/x   # resolves a filesystem path
+  camp workitem id --key                 # print the <type>:<path> key form
+  camp workitem id --json SC0001         # structured object for a festival
+
+```
+camp workitem id [selector-or-path] [flags]
+```
+
+### Options
+
+```
+  -h, --help   help for id
+      --json   emit a structured JSON result
+      --key    print the path-derived key instead of the stable id
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+
+### SEE ALSO
+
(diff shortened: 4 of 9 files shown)


obey-agent:
## Terminal demo (real binary + VHS)

Recorded from the PR branch binary in a disposable fixture campaign (`HOME` isolated). Real CLI output, not mocked.

camp PR #485 real TTY demo

**Journey shown**
1. `camp workitem --list` — fixture workitems
2. `camp workitem id billing-redesign` — bare stable id on stdout
3. `camp workitem id --key billing-redesign` — path-derived key
4. `camp workitem id --json billing-redesign` — structured envelope (`workitem-id/v1alpha1`, `id_kind: stable`)
5. Unadopted path / key-fallback (`plain-notes`)
6. Context resolution from inside a workitem directory
7. No-arg outside context → validation error + nonzero exit

**Gist:** https://gist.github.com/lancekrogers/a9e46873dcc943d2d4dff90b80ca05e3 (secret/unlisted)

---

## Staff review

**Verdict:** Approve (comment)

**Overview:** Clean scripting primitive. Thin veneer over shared `resolver` / `LinkWorkitemID` is the right shape; stdout contract (bare id vs path from `--print`) is agent-friendly and well documented.

**What's done well**
- Reuses resolver + selector instead of a parallel lookup path
- `durableID` locked to `LinkWorkitemID` (drift test is the right guard)
- Completion bounded at 200ms; no shell-template churn
- `--json` uses the versioned envelope; no-context is an error (correct for a scripting verb)

**Notes / non-blockers**
- Key-fallback `id_kind: key` is documented; worth a one-liner in the CLI long help that scripts depending on “always stable” should check `id_kind` in JSON
- Path argument canonicalization via `EvalSymlinks` is good; keep an eye on non-existent paths treated as selectors (already fails loud — good)

**Staff standard:** Yes — would ship this as a staff-level scripting surface.

obey-agent (commented):
## Staff review

**Verdict:** Approve (comment)

**Overview:** Clean scripting primitive. Thin veneer over shared `resolver` / `LinkWorkitemID` is the right shape; stdout contract (bare id vs path from `--print`) is agent-friendly and well documented.

**What's done well**
- Reuses resolver + selector instead of a parallel lookup path
- `durableID` locked to `LinkWorkitemID` (drift test is the right guard)
- Completion bounded at 200ms; no shell-template churn
- `--json` uses the versioned envelope; no-context is an error (correct for a scripting verb)

**Notes / non-blockers**
- Key-fallback `id_kind: key` is documented; worth a one-liner in the CLI long help that scripts depending on “always stable” should check `id_kind` in JSON
- Path argument canonicalization via `EvalSymlinks` is good; keep an eye on non-existent paths treated as selectors (already fails loud — good)

**Staff standard:** Yes — would ship this as a staff-level scripting surface.

obey-agent:
## Full TTY TUI demo (VHS)

Follow-up to the CLI-only recording. Interactive **workitem dashboard** on a real PTY, then the new `id` verb against the selected item.

camp PR #485 full TTY TUI demo

**Journey**
1. `camp workitem` → full dashboard (list + detail pane shows **stable id**)
2. `j`/`k` navigate
3. `q` quit
4. `camp workitem id billing-redesign` + `--json` — scripting sibling of the TUI detail field

**Gist:** https://gist.github.com/lancekrogers/1e98c6cd52ea2a86fd218631d1174905 (secret)

PR #490, opened 2026-07-24
feat(commit): signal amend mode to message writers

lancekrogers:
## Summary

- add the explicit `CAMP_COMMIT_AMEND=1` contract for auto-write commit-message hooks
- send the amend signal from root, project, and worktree commit paths
- document the hook environment and cover helper plus subprocess propagation

## Why

Commit-message writers cannot safely infer amend mode from an empty staged index: a message-only amend has no newly staged delta, while an ordinary commit with an empty index should still fail. Camp now tells configured writers explicitly when it is preparing an amend.

## Impact

Writers such as `ob commit` can generate a message for the complete amended commit while ordinary auto-write behavior remains unchanged.

## Validation

- `just test unit` — 3,678/3,678 tests passed
- `just lint` — passed

### DIFF
--- a/pkg/commitkit/autowrite_internal_test.go
+++ b/pkg/commitkit/autowrite_internal_test.go
@@ -11,12 +11,13 @@ func TestRunCommitMessageCommandForwardsDiagnostics(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name              string
-		command           string
-		wantMessage       string
-		wantErr           bool
-		wantDiagContains  []string
-		wantErrContains   []string
+		name             string
+		command          string
+		extraEnv         []string
+		wantMessage      string
+		wantErr          bool
+		wantDiagContains []string
+		wantErrContains  []string
 	}{
 		{
 			name:        "success forwards stderr and returns stdout",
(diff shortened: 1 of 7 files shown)


pkg/commitkit/workitem_env.go
```diff
@@ -6,6 +6,18 @@ import (
 	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
 )
 
+const commitAmendEnv = "CAMP_COMMIT_AMEND=1"
+
+// WithCommitAmendEnv adds the Camp-to-writer amend signal when amend is true.
+// Writers use this explicit contract instead of inferring amend mode from an
+// empty staged index.
+func WithCommitAmendEnv(env []string, amend bool) []string {
```
obey-agent, line 14: **[praise]** Explicit amend contract beats empty-index heuristics. Keep the value as `1` only — writers should not treat other truthy strings as amend.

pkg/commitkit/autowrite_internal_test.go
```diff
@@ -26,6 +27,12 @@ func TestRunCommitMessageCommandForwardsDiagnostics(t *testing.T) {
 				"session_id=session-123\n",
 			},
 		},
+		{
+			name:        "passes explicit amend contract to writer",
+			command:     `test "$CAMP_COMMIT_AMEND" = "1" && printf 'fix: amended\n'`,
```
obey-agent, line 32: **[praise]** Subprocess-level env check is the right regression for a cross-process contract.

obey-agent (commented):
## Verdict: Approve

## Overview

Small, correct contract: writers must not infer amend from an empty index. `CAMP_COMMIT_AMEND=1` is explicit, wired through root/project/worktree commit paths, documented in the campaign hooks comment, and covered by helper + subprocess env tests.

## Key Findings

- No blocking issues.
- Pairs cleanly with Obedience-Corp/obey#215 (`resolveCommitAmend` only accepts `"1"`).

## What's Done Well

- `WithCommitAmendEnv` is pure and easy to compose with workitem env.
- Subprocess test actually checks the env contract reaches the writer command.
- Docs touch in campaign.go matches the code path.

## Staff Standard

Yes — merge.

— review by obey-agent

PR #486, opened 2026-07-23
feat(list,switch): surface remote machines in picker and list browser

lancekrogers:
## Summary

Implements design **WI-56d496** (remote list/switch discoverability) on top of dual-auth #481.

Operators can already **connect** (`csw machine:campaign`) but could not **see or pick** remotes from ordinary list/switch surfaces. This PR:

1. **Switch picker (R3/R5)** — Opens immediately with local campaigns; when `machines.yaml` has entries, remotes append via go-fuzzyfinder hot reload (never blocks open on SSH). Labels `machine · name`. Selecting a remote uses the existing `runRemoteSwitch` / `--shell-connect` hop path. Remotes-only fleets (empty local registry) are allowed.
2. **List TUI (R1)** — Key `r` toggles remote load asynchronously via `tea.Cmd` (status line can paint). Remote rows are read-only for status/org mutate. `g`/enter on a remote row writes `ssh-hop:<machine>:<campaign>` for shell-init hop.
3. **Shell list arms** — zsh/bash/fish list branches detect the `ssh-hop:` marker and `eval` `camp switch … --shell-connect` (local absolute paths still `cd`).
4. **Table hint (R2)** — Human table without `--remote` emits one dim stderr line when machines are configured **and stderr is a TTY** (pipes stay silent).
5. **Help (R6)** — list/switch long help mention shell-init, `csw`, and remotes.

Hop contract (R4) is unchanged: bare `command camp switch machine:…` still refuses silent hop.

## Test plan

- [x] `go test ./cmd/camp/ -count=1`
- [x] `go test ./internal/shell/ -count=1`
- [x] `go test ./... -count=1`
- [x] `just build-camp`
- [ ] Operator gate (live fleet):
  ```bash
  eval "$(camp shell-init zsh)"   # from built binary
  camp list                       # local TUI; r loads remotes
  camp list --remote              # still works
  # table hint: camp list --format table 2>&1 (TTY stderr)
  csw                             # picker: locals first, remotes append
  csw archdtop:lance-arch         # hop still works
  command camp switch archdtop:x  # still no silent hop
  ```

## Design refs

- Campaign design: `workflow/design/machine-remote-list-switch-surfaces/` (WI-56d496)
- Fathom amendments: non-blocking picker open; TTY-gated table hint; async list TUI load

PR #492, opened 2026-07-24
feat(list,switch,machine): remote surfaces v2 from operator gate feedback

lancekrogers:
Follow-up to #486 driven by the live operator gate on WI-56d496. Intent: remote-surfaces-v2-gate-feedback.

## Changes

1. **List TUI auto-loads remotes on open** when machines.yaml is non-empty (async via tea.Cmd, fail-open; locals render immediately, remote rows append). `r` now hides or reloads. This revises design decision R1 after operator experience: the toggle-only default hid the fleet.
2. **`camp list --remote` opens the same TUI in a TTY.** Pipes and `--json` keep the scriptable table path unchanged.
3. **Switch picker load-failure visibility**: loader errors and unreachable machines now print stderr warnings after the picker closes (fuzzyfinder headers are fixed at open, so in-TUI reporting is not possible). Header wording no longer claims perpetual loading.
4. **Remote switch suggestions**: a failed `machine:name` resolve appends near matches from the per-machine completion cache (cache-only, no extra SSH): `did you mean archdtop:lance-arch?`
5. **`camp machine diagnose` probes the remote camp version** (`version --json` over the existing RunCampCommand path) and warns on skew, including dev builds from different commits. The gate's misleading `csw archdtop:lance-arch` errors were an old fleet binary; skew is now visible instead of mysterious.
6. **Test hermeticity**: list TUI tests now isolate CAMP_MACHINES_PATH so a developer's real fleet cannot leak into model construction.

## Not changed on purpose

- The reported "picker fuzzy matching broken for remote rows" could NOT be reproduced: a standalone PTY repro with identical wiring (WithHotReloadLock, pointer slice, same label format, both type-first and append-first orderings) matches appended remote rows correctly. The observed failures trace to the old remote binary instead.
- Hop contract (R4), shell templates, BatchMode: untouched.

## Verification

- `just gate` green: 3683 tests.
- PTY-driven (pyte) against a fixture registry + dead machine: TUI opens with locals immediately, auto-load fails open with `loaded remotes (1 unreachable: deadbox)` status, `--remote` opens the TUI in a TTY, piped `--remote` still prints the MACHINE table.

Workitem: WI-56d496

### DIFF
--- a/cmd/camp/list_tui.go
+++ b/cmd/camp/list_tui.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"strings"
 
 	"github.com/charmbracelet/bubbles/textinput"
@@ -78,10 +79,12 @@ func listTUIRequested(cmd *cobra.Command, isTTY bool) bool {
 	if interactive, _ := cmd.Flags().GetBool("interactive"); interactive {
 		return true
 	}
-	for _, f := range []string{"sort", "org", "tag", "status", "all", "group", "no-group", "remote"} {
-		if cmd.Flags().Changed(f) {
-			return false
-		}
+	// --remote is deliberately NOT a shaping flag: in a TTY it opens the same
+	// browser (remotes auto-load when machines are configured); piped and
+	// --json runs still take the table path via the isTTY/format checks above.
+	shaping := []string{"sort", "org", "tag", "status", "all", "group", "no-group"}
+	if slices.ContainsFunc(shaping, cmd.Flags().Changed) {
+		return false
 	}
 	return isTTY
 }
@@ -144,9 +147,22 @@ func newListTUIModel(ctx context.Context, reg *config.Registry, orgFilter string
 	m := listTUIModel{ctx: ctx, input: ti, orgFilter: orgFilter}
 	if mf, err := machines.Load(); err == nil && len(mf.Machines) > 0 {
 		m.machinesConfigured = true
+		// Remotes load automatically on open (fail-open, async); `r` hides or
+		// reloads them. remoteLoading here lets the first frame show progress.
+		m.remoteLoading = true
 	}
 	m.loadFromRegistry(reg)
 	return m
+}
+
+// remoteListFilter mirrors the browser's current view (org + active-only) for
+// the remote fan-out so local and remote rows filter consistently.
+func (m listTUIModel) remoteListFilter() listFilter {
+	filter := listFilter{org: m.orgFilter}
+	if !m.activeOnly {
+		filter.all = true
+	}
+	return filter
 }
 
 func (m *listTUIModel) loadFromRegistry(reg *config.Registry) {
@@ -204,7 +220,12 @@ func (m *listTUIModel) reload() error {
 	return nil
 }
 
-func (m listTUIModel) Init() tea.Cmd { return textinput.Blink }
+func (m listTUIModel) Init() tea.Cmd {
+	if m.remoteLoading {
+		return tea.Batch(textinput.Blink, loadRemoteCampaignsCmd(m.ctx, m.remoteListFilter()))
+	}
+	return textinput.Blink
+}
 
 func (m *listTUIModel) cycleStatus() error {
 	e := m.visible[m.cursor]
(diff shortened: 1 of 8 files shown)


cmd/camp/list_tui.go
```diff
@@ -78,10 +79,12 @@ func listTUIRequested(cmd *cobra.Command, isTTY bool) bool {
...
 		return true
 	}
-	for _, f := range []string{"sort", "org", "tag", "status", "all", "group", "no-group", "remote"} {
-		if cmd.Flags().Changed(f) {
-			return false
-		}
+	// --remote is deliberately NOT a shaping flag: in a TTY it opens the same
+	// browser (remotes auto-load when machines are configured); piped and
+	// --json runs still take the table path via the isTTY/format checks above.
+	shaping := []string{"sort", "org", "tag", "status", "all", "group", "no-group"}
```
obey-agent, line 85: **[praise]** Treating `--remote` as non-shaping so TTY opens the browser (with auto-load) while pipes stay table-shaped is the right dual-audience split.

cmd/camp/machine.go
```diff
@@ -320,6 +322,38 @@ type machineDiagnoseRow struct {
...
+	if err := json.Unmarshal(out, &info); err != nil {
+		return "", ""
+	}
+	return info.Version, info.Commit
+}
+
+// campVersionSkew reports whether a probed remote version differs from this
+// binary. Same version string but different commits (the dev-build case) also
+// counts — two "dev" builds from different commits are not the same camp.
+func campVersionSkew(local version.Info, remoteVersion, remoteCommit string) bool {
```
obey-agent, line 349: **[praise]** Dev builds with the same version string but different commits correctly count as skew — that is exactly how fleets drift in practice.

cmd/camp/switch.go
```diff
@@ -466,6 +466,31 @@ func runRemoteSwitch(ctx context.Context, cmd *cobra.Command, msel cmdutil.Parse
 		"; run via the csw shell wrapper to hop there")
 }
 
+// withRemoteSuggestions augments a remote resolve failure with near matches
+// from the per-machine completion cache. Cache-only on purpose: the failed
+// resolve already paid one SSH round-trip, and suggestions must not add
+// another. A cold cache simply yields no suggestions.
+func withRemoteSuggestions(err error, msel cmdutil.ParsedMachineSelector) error {
```
obey-agent, line 473: **[praise]** Cache-only suggestions avoid a second SSH after a failed resolve. Cold cache returning the original error unchanged is correct.

cmd/camp/list_tui_test.go
```diff
@@ -29,6 +29,10 @@ func setListRegistry(t *testing.T) {
 	t.Helper()
 	path := filepath.Join(t.TempDir(), "registry.json")
 	t.Setenv("CAMP_REGISTRY_PATH", path)
+	// Isolate from the runner's real fleet: without this, a developer machine
+	// with ~/.obey/machines.yaml constructs every test model in remote-loading
+	// state (remotes auto-load on open when machines are configured).
+	t.Setenv("CAMP_MACHINES_PATH", filepath.Join(t.TempDir(), "no-machines.yaml"))
```
obey-agent, line 35: **[praise]** Isolating `CAMP_MACHINES_PATH` is mandatory once remotes auto-load; without it every developer machine becomes a flaky test environment.

obey-agent (commented):
## Verdict: Approve

## Overview

Operator-driven remote surfaces polish that matches real fleet pain: remotes auto-load when machines exist, `--remote` stops forcing a different TTY path, switch failures become visible, remote resolve gets cache-only suggestions, and `machine diagnose` surfaces camp version skew (including dev commit mismatches). Test hermeticity via `CAMP_MACHINES_PATH` isolation is required given auto-load.

## Key Findings

1. **Extra SSH on diagnose** — `probeRemoteCampVersion` adds a `version --json` hop per machine. Best-effort and fail-open is correct; if diagnose feels slow on large fleets, consider parallelizing probes later (not blocking).
2. **Picker failures only on stderr after close** — documented limitation of fuzzyfinder fixed headers; wording change is honest.
3. **Fuzzy remote match unrepro** — body attributes to old remote binary; skew warning should prevent that class of mystery next time.

## What's Done Well

- Auto-load + fail-open preserves local rows immediately.
- Scriptable path preserved for pipes/`--json`.
- `withRemoteSuggestions` is cache-only (no second SSH on failure).
- `campVersionSkew` treats empty remote version as "unknown", not skew — avoids false alarms on ancient camps.
- Test isolation for real developer fleets.

## Staff Standard

Yes — merge as the post-gate v2 follow-up.

— review by obey-agent

PR #495, opened 2026-07-24
feat(fresh,wi): tier-1 sweep in camp fresh + camp wi banner (WF0001 P1/S3)

lancekrogers:
## Summary

Wires the tier-1 workitem sweep into `camp fresh` behind a `completed_runs`
setting and adds a read-only "N workitems have completed runs" banner to
`camp wi`. Third and final sequence of phase 001 of the
`workitem-festival-lifecycle` festival (WF0001), spec `WI-75d77f` docs 03
(evidence-driven completion, "Where the sweep runs") and 05 (phase 1 items 4-5).

## Stack

Base: **`fest/wf0001-p1s2-sweep-command`** (PR #494), not `main`. This uses the
`camp workitem sweep` internals from that PR. Full chain: #493 (planner) <-
#494 (sweep verb) <- this. Rebase onto `main` if the parents merge first. This
is phase 001's tip; phase 002 stacks on it.

## What changed

- `internal/config/settings.go`: `FreshConfig.CompletedRuns` +
  `ResolveFreshCompletedRuns()` (default `"sweep"`; any unrecognized value
  defaults to `"sweep"`, never fails closed to `"off"`). Campaign-root scoped
  like `Prune`.
- `internal/commands/fresh/workitem_sweep.go`: `runCampaignWorkitemSweep`, called
  from **exactly two** sites, the single-project `RunE` (after `executeFresh`
  succeeds) and once after the `runFreshBatch` loop, **never inside
  `executeFresh`**, which stays project-scoped. `off` does no discovery at all;
  `report` prints the banner; `sweep` promotes. A sweep failure is reported but
  never fails the fresh run. `camp fresh all` / `--list` sweep exactly once.
- `internal/commands/workitem/sweep.go`: refactored into
  `gatherSweepCandidates`/`executeSweepCandidates`/`newSweepResult`/`fillSweepPlan`,
  and a new exported `RunFreshSweep` so fresh reuses the sweep path rather than
  duplicating it. New import edge `fresh -> commands/workitem` (no cycle).
- `internal/workitem/sweep.go`: `SweepBannerText` (wording lives once in the pure
  package, shared by `camp wi` and fresh's `report`).
- `camp wi` banner: `internal/commands/workitem/list_output.go` (compact list)
  and the TUI footer (`Model.sweepEligible` + `renderFooter`), computed from
  `PlanSweep` with zero mutation. `camp wi --json` is byte-identical.

## Once-per-invocation (the highest-risk decision)

[...]

PR #494, opened 2026-07-24
feat(workitem): camp workitem sweep verb for tier-1 completion (WF0001 P1/S2)

lancekrogers:
## Summary

Adds `camp workitem sweep`, the tier-1 mutation path of evidence-driven
completion: it promotes every workitem whose active workflow run has completed
to that item's local `dungeon/completed`. Second sequence of the
`workitem-festival-lifecycle` festival (WF0001), spec of record `WI-75d77f`
docs 03 (evidence-driven completion) and 05 (implementation plan, phase 1
item 3).

## Stack

Base: **`fest/wf0001-p1s1-sweep-planner`** (PR #493), not `main`. This sequence
uses `PlanSweep`/`SweepCandidate`/`resolveSweepLocation` from that PR and will
not compile without them. If #493 merges to `main` first, rebase this onto
`main`. Sequence 03 (`camp fresh` wiring + `camp wi` banner) stacks on this
branch next.

## What changed

- `internal/commands/workitem/sweep.go`: `newSweepCommand` + `runWorkitemSweep`.
  Gathers workitems, plans via `PlanSweep`, and for each candidate runs the same
  per-item sequence `runWorkitemPromote` uses for a single item -
  `MoveToDungeon(…, "completed")`, audit event, ledger `KindTransitioned` emit,
  `StageAndCommitDungeonMove`, nav-cache invalidation, inside a loop with
  per-item error isolation. `--dry-run` returns the plan before any mutation;
  `--json` emits a `workitem-sweep/v1alpha1` envelope shaped like
  `workitemGatherResult`. Agent-allowed; `cobra.NoArgs`, no interactive selector.
- `internal/workitem/audit/audit.go`: additive `Evidence string
  json:"evidence,omitempty"` on `Event`. Set to `workflow_run_completed` for
  sweep promotes; left empty for the untouched manual-promote path.
- `internal/commands/workitem/json_contract.go`: `WorkitemSweepJSONVersion`.
- `internal/commands/workitem/workitem.go` + `cmd/camp/zz_manifest_annotations.go`:
  register the command and its agent-allowed allowlist entry.
- `tests/integration/workitem_sweep_test.go`: six containerized scenarios.
  Host unit tests for the JSON envelope shape and version constant.

## Evidence-tier invariant

Only loop-completion evidence promotes, and only ever to `completed` (never
`archived`/`someday`). No merged-branch (inference-tier) evidence is acted on
here; that is phase 2, prompt/report only.

## Exit-code / --json decision (called out)

[...]

PR #493, opened 2026-07-24
feat(workitem): sweep eligibility planner + local-dungeon destination rule (WF0001 P1/S1)

lancekrogers:
## Summary

Adds the pure sweep-eligibility planner and the local-dungeon destination
resolver that the tier-1 evidence-driven completion feature is built on. This is
the first sequence of the `workitem-festival-lifecycle` festival (WF0001), spec
of record `WI-75d77f` doc 03 (evidence-driven completion) and doc 05
(implementation plan, phase 1).

No user-facing behavior changes yet: this sequence lands the planner and the
resolver seam only. The `camp workitem sweep` verb (sequence 02) and `camp fresh`
wiring (sequence 03) execute against these in the following stacked PRs.

## What changed

- `internal/workitem/sweep.go` (new): `PlanSweep(items []WorkItem)
  []SweepCandidate`, a pure filter (no I/O, no context, no error return) over
  already-gathered `Discover()` output. Eligibility, cheapest checks first:
  workflow type is not `festival`/`intent` (explicit named exclusion, not an
  accident of which discovery paths populate `WorkflowMeta`), `WorkflowMeta`
  non-nil, `RunStatus == "completed"`, non-empty `ActiveRunID`, and a defensive
  whole-segment dungeon-path guard. `SweepCandidate` carries the item, a
  `Reason` (constant `EvidenceWorkflowRunCompleted`) future-proofed for a second
  evidence tier, and the `ActiveRunID` for the promote event payload.
- `internal/commands/workitem/sweep.go` (new): `resolveSweepLocation`, a thin
  named wrapper over `locate.DetectFromCwd`. Every candidate `PlanSweep` can
  produce today has a `workflow/<type>/<slug>` path, so this is currently a
  pass-through; phase 3 extends `DetectFromCwd` itself for `festivals/` homes and
  this wrapper needs no change. It is the single seam a future contributor greps
  for.
- Table-driven tests for both, exclusion cases before happy paths per campaign
  convention. Includes a multi-run fixture (`TestPlanSweep_MultiRunCaveat`)
  driving `LoadLocalRunFS` against an `fstest.MapFS` where `active_run_id` points
  at an in-progress `run-002` while `run-001` holds a `workflow_run_completed`
  event, proving the stale `completed` never leaks. The `resolveSweepLocation`
  negative test asserts the exact current error for a `festivals/` path, a locked
  baseline phase 3 must change deliberately.

## Evidence-tier invariant

[...]

PR #496, opened 2026-07-24
feat(fresh): tier-2 merged-branch backstop mapping + guard (WF0001 P2/S1)

lancekrogers:
## Summary

Adds the tier-2 merged-branch backstop mapping and open-work guard, wired into
`camp fresh` in report mode. When fresh prunes a merged branch, it maps that
branch back to a still-active workitem and (in `report` mode) prints the exact
`camp workitem promote` command, inference-tier evidence that **never
auto-promotes** (FESTIVAL_RULES rule 2). First sequence of phase 002 of the
`workitem-festival-lifecycle` festival (WF0001), spec `WI-75d77f` doc 03 tier 2.

## Stack

Base: **`fest/wf0001-p1s3-fresh-integration`** (PR #495), the phase-001 tip.
Full chain: #493 <- #494 <- #495 <- this. Rebase onto `main` if the parents merge
first. Sequence 02 (interactive prompt + `merged_branch` evidence promote + pty)
stacks on this.

## What changed

- `internal/commands/fresh/merged_backstop.go` (new): `MapMergedBranchesToWorkitems`
  maps a project's just-pruned branches to still-active workitems via two signals
 , worktree-scope links (branch name == worktree dir basename) and `WI-` commit
  tags on commits newly reachable from the default branch since a pre-pull
  `beforeSHA` (the pruned ref is gone by the time prune returns). `HasOpenWork`
  suppresses a match while the workitem has other open linked scopes (a second
  project/repo, or a still-existing other worktree), the WI-ca06e1 multi-PR
  shape. `reportMergedBackstop` prints the promote command for surviving matches.
- `internal/config/settings.go`: `merged_workitems` (`prompt|report|off`) +
  `ResolveFreshMergedWorkitems`. **Default is `off` in this sequence** (dormant
  until the interactive prompt exists); sequence 02 changes the default to the
  spec value `prompt` and adds the TTY prompt + `merged_branch` evidence.
- `internal/commands/fresh/fresh.go` + `batch.go`: capture `beforeSHA` before the
  pull, thread `campRoot`/`mergedWorkitems`, call the backstop once per project
  after prune (per-project, unlike phase 1's campaign-wide sweep, because pruned
  branches are only known inside a project's own prune pass).
- Exported `WorkitemAliases` (was `workitemAliases`) in `commits.go` for reuse.

## Design notes (called out)

- **Ledger fast-path** (doc-06 open question): uses a direct `beforeSHA..HEAD`
  git-log scan + `commitkit.ParseTag`, not the `camp workitem commits` ledger
  path, because the tag lives in the commit message regardless of ledger
  completeness (a raw `git commit` never reaches the ledger). The phase-3 spike
  may later prove the ledger path reusable as an optimization; follow-up, not a
  blocker.
- **Sequence boundary**: the `merged_workitems` setting + report wiring live here
  (not deferred to sequence 02 as the task doc split them) so the mapping is
  CLI-reachable and containerized-integration-testable in this PR; a standalone
  sequence-01 PR could otherwise neither compile the `beforeSHA` capture nor
  test the mapping. Detailed in the sequence results doc.
- **Guard scope-path**: `MergedBackstopMatch.ScopePath` carries the matched
  scope so the guard doesn't count the just-merged worktree itself as open work.

## Test output

Unit (pure mapping/guard):

```
--- PASS: TestWorkitemRefsFromSubjects
--- PASS: TestMatchWorktreeLinkBranch
--- PASS: TestRefMatchesActiveItem
--- PASS: TestActiveBackstopItems_ExcludesFestivalsAndIntents
--- PASS: TestHasOpenWork  (single-scope, WI-ca06e1 two-project suppressed, stale worktree, existing worktree)
```

Integration (end-to-end via `camp fresh`):

```
--- PASS: TestIntegration_FreshMergedBackstop_WorktreeLinkReported
--- PASS: TestIntegration_FreshMergedBackstop_OffIsSilent
--- PASS: TestIntegration_FreshMergedBackstop_NoMatchIsSilent
```

The worktree-link test asserts the report prints `camp workitem promote
design-myfeature --target completed` and the workitem is not moved. `just lint`
clean (0 issues, both ratchets); `just test unit` 3702/3702.

Not merged by an agent; left for review.

### DIFF
--- a/internal/commands/fresh/merged_backstop.go
+++ b/internal/commands/fresh/merged_backstop.go
@@ -23,11 +23,11 @@ import (
 // backstopRoot resolves the campaign root for the backstop: the threaded
 // campRoot when set, else a cwd detect (fresh runs at or under the campaign
 // root). Empty on failure disables the backstop for this project.
-func backstopRoot(opts freshOptions) string {
+func backstopRoot(ctx context.Context, opts freshOptions) string {
 	if opts.campRoot != "" {
 		return opts.campRoot
 	}
-	root, err := campaign.DetectCached(context.Background())
+	root, err := campaign.DetectCached(ctx)
 	if err != nil {
 		return ""
 	}
@@ -59,7 +59,7 @@ func reportMergedBackstop(ctx context.Context, out io.Writer, root, projectPath 
 		return
 	}
 	for _, m := range matches {
-		if HasOpenWork(root, registry, m.Workitem, m.ScopePath) {
+		if HasOpenWork(root, registry, m.Workitem, m.ScopePath, projectRelPath(root, projectPath)) {
 			continue
 		}
 		_, _ = fmt.Fprintf(out, "%s workitem %s had a merged branch and is still active; promote when done:\n    %s\n",
@@ -293,21 +293,31 @@ func refFromItem(wi wkitem.WorkItem) string {
 // HasOpenWork reports whether wi still has open work beyond the branch that just
 // merged, in which case a tier-2 match must be suppressed (doc 03 conservatism
 // guard: one merged PR out of several open ones is progress, not completion). A
-// workitem linked to a second project/repo, or with a still-existing worktree
+// workitem linked to a DIFFERENT project/repo, or with a still-existing worktree
 // other than the just-merged one, is "open." A stale worktree link whose
-// directory is already gone does not count as open.
-func HasOpenWork(root string, registry *links.Links, wi wkitem.WorkItem, justMergedScopePath string) bool {
+// directory is already gone does not count as open. mergedScopePath is the
+// scope that just merged (the matched worktree link's path, or the project path
+// for a commit-tag match); projectPath is the just-pruned project's
+// campaign-relative path, so the just-merged project's own project/repo link is
+// not mistaken for other open work.
+func HasOpenWork(root string, registry *links.Links, wi wkitem.WorkItem, mergedScopePath, projectPath string) bool {
 	for _, link := range registry.Links {
 		if !linkMatchesBackstopItem(link, wi) {
 			continue
 		}
 		switch link.Scope.Kind {
 		case links.ScopeProject, links.ScopeRepo:
-			if link.Scope.Path != justMergedScopePath {
+			// A project/repo link is open work only when it points at a
+			// DIFFERENT project than the one whose branch just merged. The
+			// just-merged project's own link is not "other" work (compare
+			// against projectPath, not the matched worktree's scope path).
+			if link.Scope.Path != projectPath {
 				return true
 			}
 		case links.ScopeWorktree:
-			if link.Scope.Path == justMergedScopePath {
+			// Skip the worktree that just merged; any OTHER worktree that still
+			// exists on disk is open work.
+			if link.Scope.Path == mergedScopePath {
 				continue
 			}
 			if worktreeDirExists(root, link.Scope.Path) {
--- a/internal/commands/fresh/merged_backstop.go
+++ b/internal/commands/fresh/merged_backstop.go
@@ -1,0 +1,329 @@
+package fresh
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/Obedience-Corp/camp/internal/campaign"
+	wkcmd "github.com/Obedience-Corp/camp/internal/commands/workitem"
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/git"
+	"github.com/Obedience-Corp/camp/internal/paths"
+	"github.com/Obedience-Corp/camp/internal/ui"
+	wkitem "github.com/Obedience-Corp/camp/internal/workitem"
+	"github.com/Obedience-Corp/camp/internal/workitem/links"
+	"github.com/Obedience-Corp/camp/pkg/commitkit"
+)
+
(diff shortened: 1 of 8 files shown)


internal/commands/fresh/merged_backstop.go
```diff
@@ -0,0 +1,335 @@
...
+// other than the just-merged one, is "open." A stale worktree link whose
+// directory is already gone does not count as open.
+func HasOpenWork(root string, registry *links.Links, wi wkitem.WorkItem, justMergedScopePath string) bool {
+	for _, link := range registry.Links {
+		if !linkMatchesBackstopItem(link, wi) {
+			continue
+		}
+		switch link.Scope.Kind {
+		case links.ScopeProject, links.ScopeRepo:
+			if link.Scope.Path != justMergedScopePath {
```
obey-agent: **[bug]** For a worktree-link match, `justMergedScopePath` is the worktree path (e.g. `projects/worktrees/camp/foo`). Any co-existing **project/repo** link on the same workitem (almost always present after `camp workitem worktree` / normal link workflows) has a *different* path (`projects/camp`), so this branch returns true and **suppresses the promote report for the common case**.

The multi-project guard intent (WI-ca06e1) is right, but project/repo open-work must be relative to the *project that just pruned*, not compared to the worktree path.

Suggested shape:
- Thread the project’s campaign-relative path into `HasOpenWork` (or the match)
- Count `ScopeProject`/`ScopeRepo` as open only when `link.Scope.Path != projectPath`
- Keep worktree existence checks as they are for *other* worktrees

Add a unit test: project link + matching worktree link + just-merged worktree path → **not** open.

internal/commands/fresh/merged_backstop.go
```diff
@@ -0,0 +1,335 @@
...
+)
+
+// backstopRoot resolves the campaign root for the backstop: the threaded
+// campRoot when set, else a cwd detect (fresh runs at or under the campaign
+// root). Empty on failure disables the backstop for this project.
+func backstopRoot(opts freshOptions) string {
+	if opts.campRoot != "" {
+		return opts.campRoot
+	}
+	root, err := campaign.DetectCached(context.Background())
```
obey-agent: **[nit]** `backstopRoot` uses `context.Background()` when `opts.campRoot` is empty. Prefer accepting/propagating the caller `ctx` so cancellation and any future deadline on `camp fresh` still apply to the detect.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview
Tier-2 mapping is thoughtfully designed: worktree-link then WI-tag signals, default `merged_workitems: off` until the interactive prompt lands, report-only (never auto-promote), and solid pure unit tests for tag parsing / open-work shapes. The `beforeSHA` capture + post-prune wiring is the right I/O shape.

## Key Findings
1. **Blocking product bug**: `HasOpenWork` compares project/repo link paths against the *worktree* scope path. A normal workitem with both a project link and a worktree link will always look “open” when the worktree branch merges, so the promote report never fires for the most common camp workflow. See inline + suggested test.
2. **Nit**: `backstopRoot` uses `context.Background()` instead of the fresh cycle ctx.
3. **Note**: `beforeSHA` from `rev-parse` ignores errors (empty → commit-tag signal skipped). Acceptable best-effort if intentional; worth a debug log.

## What's Done Well
- Explicit inference-tier invariant (never auto-promote) + integration test asserting the workitem stays put
- Export of `WorkitemAliases` for reuse rather than a parallel alias helper
- `off` / no-match silence tests
- Clear sequence boundary documentation for why setting + report land together

## Staff Standard
Not yet — the open-work guard false-positive will make the feature look broken in real campaigns. Fix that comparison and land a regression test, then this is a strong phase-2 start.

lancekrogers:
Addressed all three review points (merged updated p1s3 with the #495 fixes, then pushed 89f4bf84):

1. **BLOCKING guard bug** — `HasOpenWork` now takes the just-pruned project's campaign-relative path separately and compares `ScopeProject`/`ScopeRepo` links against *that* (not the matched worktree scope path). So a workitem with both a project link and its own just-merged worktree link no longer reads as perpetually open, and the report fires. Worktree-existence checks are unchanged. Added the regression test you asked for: project link + matching worktree link + just-merged worktree path => NOT open.
2. **ctx** — `backstopRoot` propagates the caller's `ctx` instead of `context.Background()`.
3. **beforeSHA** — the `rev-parse HEAD` error/empty case is now logged (dim, best-effort) rather than silently ignored; the commit-tag scan is skipped for that project while worktree-link matching still works.

Stack: brought p2s1 up to date by **merging** the updated p1s3 (kept your emdash commit b1b2e1a7, no rebase). Full gate green (0 lint issues, both ratchets; 3748/3748 gate-fast, 3703/3703 stable; 3 backstop integration tests pass).

obey-agent (approved):
## Verdict: Approve (re-review)

## Overview
Follow-up `89f4bf84` fixes the blocking open-work guard bug and the smaller review nits. Stack tip for phase-2 sequence 1 is ready once #495 lands (or merge stack order is preserved).

## Prior findings — disposition
1. **Blocking HasOpenWork false-positive** — fixed: project/repo links compared to `projectPath` (just-pruned project), worktree links still skip only `mergedScopePath`. Regression test `project link + just-merged worktree link is not open` locks the common camp worktree workflow.
2. **backstopRoot ctx** — fixed: propagates caller context.
3. **beforeSHA silence** — fixed: dim stderr log when rev-parse fails/empty; worktree-link path still works.

## What's Done Well
- Fix matches the suggested shape from the prior review exactly
- Named regression test documents the failure mode for the next reader

## Staff Standard
Yes — approve. Keep stacked on #495.

PR #498, opened 2026-07-24
fix(machine): stop diagnose probes from healing the ControlMaster socket they report

lancekrogers:
## Summary

`TestMachineDiagnoseControlMasterLifecycle` has been failing since #492. It is not a flaky test: #492 added a per-machine camp version probe to `camp machine diagnose`, and that probe hops with `ControlMaster=auto`. Against a **stale** ControlPath, OpenSSH unlinks the socket and opens a fresh master, so diagnose healed the exact stale state it was reporting.

The operator recovery path the feature exists for was broken:

| Operator step | Reported | Socket after the call |
| --- | --- | --- |
| `camp machine diagnose self` | `stale` | probe replaced it, now **live** |
| `camp machine diagnose self --reset` | `live`, `reset:false` | nothing to clear |

See a stale socket, run `--reset`, clear nothing. And in a single `--reset` run the order is check(stale) -> probe(opens fresh master) -> reset, so `--reset` tore down a healthy connection it had just created.

History confirms the direction: the test landed green in 2f7f8a85; `probeRemoteCampVersion` arrived later in aae497f6 (#492).

## Root cause, verified

Reproduced in an Alpine 3.21 container against a real loopback sshd. After SIGKILLing the master (socket left behind, `-O check` refused), one normal `ControlMaster=auto` hop:

```
before: stale
hop:    PROBE_OK
after:  socket present / check: Master running (pid=46)   <-- healed
```

## Fix

`OptsReuseOnly` / `RunCampCommandReuseOnly` in `internal/remote/ssh.go`, built on `ControlMaster=no`: reuse a live master when one exists, but never open one and never unlink a stale socket. Verified against all three socket states in the same container:

- **none** -> probe works, no socket created
- **live** -> reuses the existing master (same pid, no extra handshake, one-auth property kept)
- **stale** -> probe works, socket stays stale

`switch`/`ResolveRoot` keep `ControlMaster=auto`, so the one-auth resolve-plus-hop property is untouched and `remote_switch_ssh_test.go`'s single-socket assertion still holds.

`ControlPath` comes from `ControlSocketPath` rather than `controlPath`, so a read-only diagnostic no longer creates `~/.obey/ssh-ctl` as a side effect.

## Second bug found while tracing

The machine TUI's `testMachine` had the same defect, and it is worse there: `probeSockets` runs only at `Init` and no `healthMsg` refreshes it, so testing a machine silently replaced a stale socket while the screen kept displaying `stale` for the rest of the session. Same one-line call change.

## Verification

- `TestMachineDiagnoseControlMasterLifecycle` — **PASS** (was FAIL)
- Full integration suite — **552 passed, 0 failed, 8 skipped**, exit 0 (was 551 passed / 1 failed)
- `just gate-fast` — vet + lint clean across stable/dev/integration profiles, 3733 unit tests pass
- New `TestOptsReuseOnlyDoesNotDisturbSocket` pins the no-create / no-unlink / no-side-effect-dir properties

## Tradeoffs

A probe against a machine with **no** live master now pays its own handshake instead of leaving a warm master behind for a subsequent hop. That is the intended cost: a surface that reports socket state must not mutate it. Operators who want a warm master get one from `camp switch`, which is unchanged.

### DIFF
--- a/cmd/camp/machine.go
+++ b/cmd/camp/machine.go
@@ -331,8 +331,11 @@ type machineDiagnoseRow struct {
 // best-effort: any failure (unreachable, camp missing, ancient binary without
 // `version --json`) reports as an empty version, never a diagnose error —
 // diagnose must keep working against exactly the broken machines it exists for.
+// The hop is reuse-only: under ControlMaster=auto this probe would unlink a
+// stale socket and open a fresh master, so diagnose would heal the exact stale
+// state it is reporting and the follow-up --reset would find nothing to clear.
 func probeRemoteCampVersion(ctx context.Context, m *machines.Machine) (versionStr, commit string) {
-	out, err := remote.RunCampCommand(ctx, m, "version --json")
+	out, err := remote.RunCampCommandReuseOnly(ctx, m, "version --json")
 	if err != nil {
 		return "", ""
 	}
--- a/cmd/camp/machine_tui.go
+++ b/cmd/camp/machine_tui.go
@@ -369,6 +369,11 @@ func (m *machineTUIModel) scanTailnet(intoOverlay bool, gen uint64) tea.Cmd {
 // round trip answers everything the screen needs: whether ssh authenticates,
 // whether the host resolves, and whether camp is installed over there, which
 // is the failure a socket state can never show.
+//
+// The hop is reuse-only so testing a machine cannot rewrite the socket column
+// next to it: probeSockets runs once at Init and no healthMsg refreshes it, so
+// a ControlMaster=auto probe would silently replace a stale socket while the
+// screen kept displaying "stale" for the rest of the session.
 func (m *machineTUIModel) testMachine(target machines.Machine) tea.Cmd {
 	ctx := m.ctx
 	return func() tea.Msg {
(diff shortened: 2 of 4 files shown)


obey-agent (approved):
## Summary

Re-review of head `42f90e67`: the fix still looks correct. Version/health probes that co-report ControlMaster state now hop via `OptsReuseOnly` / `RunCampCommandReuseOnly` (`ControlMaster=no` + side-effect-free `ControlSocketPath`), so they reuse a live master but neither open one nor unlink a stale socket — the exact defect that broke `TestMachineDiagnoseControlMasterLifecycle` after #492. Both diagnose (`probeRemoteCampVersion`) and the machine TUI (`testMachine`) call sites are wired; `ResolveRoot` / list remote keep `ControlMaster=auto`. Residual risk is low: unit coverage pins the flag/path/no-mkdir properties, and the existing integration lifecycle test exercises the stale → diagnose → `--reset` path that previously failed.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

---

Re-review by obey-agent: no remaining issues; approving.

PR #499, opened 2026-07-24
fix(sweep): key tier-1 eligibility on the latest completed run (sweep never fired)

lancekrogers:
## Summary

Fixes a correctness bug in the tier-1 workitem sweep (shipped in #493/#494): as
merged, `camp workitem sweep` never fires on a genuinely completed workitem. This
was found by spike B of the WF0001 verification-spikes sequence
(`003_FESTIVAL_RAIL/01_verification_spikes/results/spike-b-localrun-eligibility.md`).

## Root cause (verified anchors)

- fest clears `active_run_id` from `.workflow/workflow.yaml` the instant a run
  reaches a terminal event: `projects/fest/internal/workflow/localstore/store.go:321-322`
  (`if m.ActiveRunID == evt.RunID { m.ActiveRunID = "" }`, inside the
  `EventWorkflowRunCompleted`/`Abandoned` branch at `store.go:290`). The completed
  run survives only in the manifest's `runs[]` index with `status: completed`.
- camp's `LoadLocalRunFS` only reads the run referenced by `active_run_id`
  (`internal/workitem/localrun.go:105`, `if mf.ActiveRunID != ""`). Once it is
  cleared, the loader returns `RunStatus:""`, `ActiveRunID:""`, an empty
  workflow.
- The shipped `PlanSweep` keyed eligibility on `RunStatus == "completed" &&
  ActiveRunID != ""`. Real completed workitems have `RunStatus == "" &&
  ActiveRunID == ""`, so they were never eligible. The phase-1 fixtures hand-set
  `active_run_id` at the completed run, a state fest never emits, so the tests
  were green but the sweep never fired in production.

## Fix

- `internal/workitem/localrun.go`: when `active_run_id` is empty, surface the
  LATEST run from the manifest's `runs[]` index (newest by `ended_at`, tie-broken
  by list order) and verify its terminal status by REPLAYING that run's
  `progress_events.jsonl` (events stay authoritative; the manifest is only the
  index of which run is latest). New `LatestRunID` / `LatestRunStatus` on
  `LocalRunProgress`, and additive `latest_run_id` / `latest_run_status` (omitempty)
  on `WorkItemWorkflow`.
- `internal/workitem/sweep.go`: re-key `PlanSweep` to "the latest run reached
  completed AND no newer run is active (`active_run_id` empty)". Dropped the
  `ActiveRunID != ""` requirement. `SweepCandidate.RunID` (json `run_id`, renamed
  from `active_run_id` since it is no longer an active run) carries the completed
  run's id for the promote evidence. The multi-run caveat still holds: a new run
  repoints `active_run_id`, making the item ineligible again.
- Fixtures rewritten to the REAL post-completion shape (`active_run_id` cleared,
  `runs[]` terminal). Banner/report/promote layers needed no change (verified).

## Acceptance-bar test (drives real fest)

`TestIntegration_WorkitemSweep_EndToEndFestDrivenCompletion` drives the real
`fest` binary in the container: `fest create workflow` + `fest workflow advance`
to completion, asserts `active_run_id` is actually cleared in the on-disk
manifest, then runs `camp workitem sweep` and asserts the workitem is promoted
(source dir gone, `run_id` recorded). This is the exact end-to-end path that
would have failed before the fix.

```
--- PASS: TestIntegration_WorkitemSweep_EndToEndFestDrivenCompletion (4.29s)
```

All six `TestIntegration_WorkitemSweep_*` scenarios pass against the corrected
fixtures; unit tests include `TestLoadLocalRunFS_LatestRunWhenActiveCleared`.
`just gate` green: 0 lint issues, both ratchets clean.

## Notes

- Base `main`. Independent of the WF0001 stack (#495/#496/#497); this fixes the
  already-merged #493/#494 detection logic.
- `camp wi --json` gains additive `latest_run_id`/`latest_run_status`
  (omitempty). The `workitem-sweep/v1alpha1` per-item `active_run_id` field is
  renamed to `run_id` (the feature is fresh and the value is the completed, not
  active, run); called out here as a deliberate contract correction.

Not merged by an agent; left for review.

### DIFF
--- a/internal/workitem/localrun.go
+++ b/internal/workitem/localrun.go
@@ -38,21 +38,40 @@ const (
 // Camp and fest are separate Go modules; this is a thin duplicate of the
 // fields camp needs from fest's localstore. Fest is the source of truth.
 type LocalRunProgress struct {
-	WorkflowID     string
-	ActiveRunID    string
-	RunStatus      string
-	CurrentStep    int
-	TotalSteps     int
-	CompletedSteps int
-	Blocked        bool
-	DocHashChanged bool
+	WorkflowID  string
+	ActiveRunID string
+	RunStatus   string
+	// LatestRunID / LatestRunStatus surface the most recent run's terminal
+	// state when there is no active run. fest clears active_run_id from
+	// workflow.yaml the instant a run completes (see fest localstore), so a
+	// genuinely completed standalone workflow has an empty ActiveRunID and its
+	// completed run survives only in the manifest's runs[] index. The status is
+	// verified by replaying that run's events (events are authoritative), with
+	// the manifest used only as the index of which run is latest.
+	LatestRunID     string
+	LatestRunStatus string
+	CurrentStep     int
+	TotalSteps      int
+	CompletedSteps  int
+	Blocked         bool
+	DocHashChanged  bool
 }
 
 type localWorkflowManifest struct {
-	WorkflowID  string `yaml:"workflow_id"`
-	ActiveRunID string `yaml:"active_run_id"`
-	DocHash     string `yaml:"doc_hash"`
-	DocPath     string `yaml:"doc_path"`
+	WorkflowID  string          `yaml:"workflow_id"`
+	ActiveRunID string          `yaml:"active_run_id"`
+	DocHash     string          `yaml:"doc_hash"`
+	DocPath     string          `yaml:"doc_path"`
+	Runs        []localRunIndex `yaml:"runs"`
+}
+
+// localRunIndex is one entry in workflow.yaml's runs[] list: the index fest
+// keeps of every run, newest appended last. Used only to find WHICH run is
+// latest; its cached status is not trusted (the run's events are replayed).
+type localRunIndex struct {
+	RunID   string `yaml:"run_id"`
+	Status  string `yaml:"status"`
+	EndedAt string `yaml:"ended_at"`
 }
 
 type localRunManifest struct {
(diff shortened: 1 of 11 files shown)


internal/workitem/localrun.go
```diff
@@ -134,6 +161,48 @@ func LoadLocalRunFS(ctx context.Context, fsys fs.FS, base string) (*LocalRunProg
...
+	// Replay does not carry TotalSteps; keep it from the run.yaml cache.
+	replayed.Summary.TotalSteps = rm.Summary.TotalSteps
+	return replayed, true, nil
+}
+
+// latestRunID picks the most recent run from the manifest's runs[] index: the
+// run with the newest ended_at, tie-broken by list order (fest appends newest
+// last). Returns "" when there are no runs. The chosen run's status is verified
+// by replay elsewhere; this only decides WHICH run is latest.
+func latestRunID(runs []localRunIndex) string {
```
obey-agent, line 193: **[suggestion]** `latestRunID` chooses the max `ended_at` string with list-order tie-break, but there is no unit test with multiple `runs[]` entries (out-of-order `ended_at`, empty `ended_at`, missing `run_id`). The only covered path is a single completed run in `TestLoadLocalRunFS_LatestRunWhenActiveCleared`.

**Suggestion:** Add a small table test for `latestRunID` covering: newest-by-`ended_at`, tie → last list entry, skip empty `run_id`, and all-empty `ended_at` → last entry (fest append-last). Optionally document that string compare assumes fest’s consistent UTC/Zulu timestamps.

internal/workitem/localrun.go
```diff
@@ -103,24 +122,32 @@ func LoadLocalRunFS(ctx context.Context, fsys fs.FS, base string) (*LocalRunProg
...
+			return nil, rErr
+		}
+		if ok {
+			out.RunStatus = rm.Status
+			out.CurrentStep = rm.Summary.CurrentStep
+			out.CompletedSteps = rm.Summary.CompletedSteps
+			out.Blocked = rm.Summary.Blocked
+			out.TotalSteps = rm.Summary.TotalSteps
+		}
+	} else if latest := latestRunID(mf.Runs); latest != "" {
```
obey-agent, line 136: **[suggestion]** When there is no active run, progress is written only to `LatestRunID` / `LatestRunStatus`; `RunStatus` stays empty. Discovery copies that through (`discover_workflow_docs.go:128-130`). The workitem TUI still only renders `RunStatus` (`tui/view.go:562-563`), so a genuinely completed workitem will show progress steps but not a “completed” status line—even though the new fields exist. Not a sweep correctness regression (production already had empty `RunStatus` after fest cleared `active_run_id`), but the new fields are not surfaced for humans/JSON consumers that only read `run_status`.

**Suggestion:** Follow-up: in TUI (and any list formatting), fall back to `LatestRunStatus` when `ActiveRunID` is empty; consider documenting `latest_run_id` / `latest_run_status` on the workitems JSON contract changelog.

obey-agent (approved):
## Summary

PR #499 correctly fixes the tier-1 sweep never-fires bug: `LoadLocalRunFS` now falls back to the manifest `runs[]` index when `active_run_id` is cleared, replays that run’s events for an authoritative terminal status, and `PlanSweep` keys eligibility on `LatestRunStatus == "completed"` with an empty `ActiveRunID`. Field renames (`ActiveRunID` → `RunID` / `active_run_id` → `run_id` on the sweep envelope) match the new meaning and are covered by unit, plan-envelope, and integration tests, including a real fest-driven e2e. Residual risk is low: multi-run ordering relies on lexicographic `ended_at` (safe for consistent RFC3339/Zulu from fest) and a few stale docs/comments remain. **READY TO MERGE.**

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 2

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[nit]** `internal/workitem/sweep.go:17` — The `SweepCandidate` type comment still describes the pre-fix model (“its active workflow run reached completed” / “ActiveRunID is the run whose completion made the item eligible”), but the struct field is now `RunID` and eligibility is based on the latest completed run with no active run.
  - **Suggestion:** Rewrite the godoc to match the new `RunID` field comment immediately below (latest completed run; fest clears `active_run_id` on completion).
- **[nit]** `internal/commands/workitem/sweep.go:78` — Command `Long` help still says “Promote every workitem whose **active** workflow run has completed…”, which is exactly the incorrect mental model this PR removes.
  - **Suggestion:** Rephrase to something like “whose latest workflow run has completed (and no run is currently active)”.

---

Review by obey-agent: no correctness blockers; approving (nits/suggestions optional).

obey-agent:
Both suggestions addressed in 6990a735.

**1. `latestRunID` coverage** — added `TestLatestRunID`, a table test over the four ordering rules: newest-by-`ended_at` regardless of list order, tie -> last list entry, entries with empty `run_id` skipped, and all-empty `ended_at` -> last entry. Plus empty-input, all-missing-`run_id`, and a dated run beating an undated one written after it. All 8 cases pass against the existing implementation, so this pins behavior rather than fixing a bug. The doc comment now records why the string compare is safe (fest writes RFC 3339 Zulu, which sorts lexicographically in time order).

**2. TUI fallback** — confirmed and fixed. `renderPreview` now falls back to `LatestRunStatus` when `RunStatus` is empty (the loader populates exactly one of the two), and `LatestRunStatus` was added to the workflow-block visibility gate so an item carrying only that field still renders the block. Covered by `TestRenderPreviewWorkflowStatus`, which drives `renderPreview` itself; verified it fails with the fix reverted.

While tracing the other `RunStatus` consumer: `gatherBlockedByRun` (`gather_plan.go:105-114`) needs no fallback. It returns early when `ActiveRunID == ""`, so it only reads `RunStatus` while a run is live.

**Not done, deliberately:** documenting `latest_run_id` / `latest_run_status` in `docs/json-contracts.md`. That file is a rules-and-versioning document (path semantics, field rules, error envelope) and documents none of the sibling `workflow_meta` fields — not `run_status`, `active_run_id`, `workflow_id`, or `current_step`. Adding two of eight fields there would misrepresent it as a field catalog. The additive/`omitempty` contract is recorded on the struct in `model.go`. Happy to add a real `workflow_meta` schema section if you want that surface documented as a whole.

Gate: `just gate-fast` clean (0 lint issues, 3743 unit tests).

PR #502, opened 2026-07-26
fix(workitem): validate link writes against the new row, not the whole registry

lancekrogers:
## The bug

One stale row in `.campaign/workitems/links.yaml` breaks every subsequent link write in the campaign.

Both writers add the new link and then validate the **whole registry**, failing on `errs[0]`:

- `internal/commands/workitem/link.go` (`camp workitem link`)
- `internal/workitem/links/attach.go` (`camp project worktree add --workitem`)

So a deleted directory fails an unrelated write, with an error naming a path the user never touched and no link id to identify it. Reproduced against a real campaign carrying 7 rot rows:

```
$ camp workitem link WI-30dd80 --worktree fest/live
Error: validation error on workitem_id: no workitem with id bug-tui-chat-bugs-2026-07-19 found on disk
```

The only escape today is `--allow-missing`, which is backwards: it suppresses the new link's own existence checks along with the stale ones.

## Fix, part 1: judge a write on what it introduces

`links.ValidateOne` validates the subject link, seeding the uniqueness maps from the other rows so **duplicate id** and **duplicate primary per scope** still span the registry. `links.AsError` reports **all** failures, so fixing a link is one round trip instead of one per problem.

## Fix, part 2: clear the rot camp can prove

Not refusing is only half an answer; dead rows should not pile up waiting for a maintenance command. But **`links.yaml` is tracked**, so an automatic removal propagates to every machine the campaign syncs to. The discriminator is whether a missing target means *gone* or only *not here*:

| Scope | Missing means | Auto-removable |
|---|---|---|
| worktree | not on this machine — camp's own scaffold writes `# Git worktrees (machine-local)` above the ignore rule it adds | no |
| project declared in `.gitmodules` | uncloned, the normal state before `camp sync` | no |
| `workflow/`, `festivals/`, campaign paths | tracked, so absent everywhere | **yes** |

`links.MachineLocal` draws that line. `PruneDead` removes only the authoritative class and reports every row with its reason and a one-command undo:

```
$ camp workitem link WI-30dd80 --worktree fest/live
removed dead link lnk_20260719_131a3c (campaign_path:workflow/design/deleted): campaign_path workflow/design/deleted no longer exists
  undo: git checkout -- .campaign/workitems/links.yaml
linked design-real-target-2026-07-26 -> worktree:projects/worktrees/fest/live (role primary, id lnk_20260726_3e813d)
```

**A missing workitem is deliberately not an auto-prune signal**, though it looks like the stronger one. Answering it needs a full tree walk (`wkitem.Discover`, with its own handling for stable ids, keys, and festival ids). Pruning on a scan means any gap in it — or a branch that does not carry the workitem — silently deletes live links. That case stays with `camp workitem doctor`, which reports before it acts and is an explicit user action.

## Two existing bugs fall out of the same distinction

1. **`camp workitem doctor --fix` deleted machine-local worktree links.** `codeBrokenScope` called `RemoveLinkByID` unconditionally, so running it on a laptop destroyed links that were correct on the desktop that owns the worktree — and the deletion travels via git. Now a warning naming `camp workitem unlink --id <id>`.

2. **The validator required every scope target to exist**, so a healthy campaign synced across two machines reported `schema.violation` errors and `camp workitem doctor` exited non-zero. Existence is now relaxed for machine-local scopes; escaping the campaign root still is not.

## Rot stays visible

`camp status` grows a notice, ignoring machine-local scopes so it does not become permanent noise on a multi-machine campaign:

```
⚠ notice: 2 workitem links point at a path that no longer exists
  run: camp workitem doctor --fix
```

Stat-level per the `internal/notice` contract: one registry read plus one stat per row.

## Verification

Against a **faithful mirror** of the real 7-rot-row campaign (`links.yaml`, `.gitmodules`, workitem tree, and project dirs all mirrored): the link succeeds, **nothing is pruned**, no notice fires. Every one of those rows is machine-local or has a live target. The old code refused the write; a naive auto-clean would have deleted 7 real rows.

Regression tests confirmed failing against pre-fix code, e.g.:

```
--- FAIL: TestAttachPrimary_SucceedsAlongsideStaleRow
    stale row blocked a valid attach: validation error on scope.path: link target does not exist

--- FAIL: TestDoctorFix_DoesNotDeleteMachineLocalWorktreeLink
    expected workitem.scope.not-on-this-machine finding, got "... [error] workitem.scope.broken
    — scope path projects/worktrees/fest/on-another-machine does not exist
      hint: remove the link or restore the directory; auto-fix removes the link"
```

New tests: `TestValidateOne_*`, `TestAsError_ReportsAllFailures`, `TestValidate_ExemptsMachineLocalScopesFromExistence`, `TestMachineLocal`, `TestPruneDead_NeverRemovesMachineLocalRows`, `TestPruneDead_IgnoresMissingWorkitem`, `TestReportPruned_NamesEachRowAndTheUndo`, `TestAttachPrimary_PrunesDeadRowsAndReports`, `TestAttachPrimary_PreservesMachineLocalRows`, `TestAttachPrimary_StillRejectsMissingScopePath`, `TestDoctorFix_DoesNotDeleteMachineLocalWorktreeLink`, `TestDoctorFix_StillRemovesAuthoritativelyDeadScope`, `TestStaleLinks_*`.

`just gate-fast`: 111 packages, 3774/3774 passed, 0 lint issues, both ratchets clean.

### DIFF
--- a/internal/workitem/links/attach.go
+++ b/internal/workitem/links/attach.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
+	"io"
 	"time"
 
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
@@ -25,6 +26,10 @@ type AttachOptions struct {
 	Replace bool
 	// AllowMissing skips ValidateLinkPath existence checks (migrations).
 	AllowMissing bool
+	// Report receives a line for each dead row the write pruned. Never nil in
+	// command code: pruning a row the user did not ask about is only acceptable
+	// because it is reported.
+	Report io.Writer
 }
 
 // AttachPrimary records a primary workitem→scope link in links.yaml.
@@ -49,7 +54,12 @@ func AttachPrimary(ctx context.Context, campaignRoot string, opts AttachOptions)
 	}
 
 	var out Link
+	var pruned []Pruned
 	err := WithLock(ctx, campaignRoot, func(registry *Links) error {
+		// Drop rows whose scope target is provably gone while the registry is
+		// open anyway. See Dead for why this is the only signal trusted here.
+		pruned = PruneDead(campaignRoot, registry)
+
 		id, idErr := NewLinkID(registry)
 		if idErr != nil {
 			return idErr
@@ -75,7 +85,23 @@ func AttachPrimary(ctx context.Context, campaignRoot string, opts AttachOptions)
 	if err != nil {
 		return Link{}, err
 	}
+	ReportPruned(opts.Report, pruned)
 	return out, nil
+}
+
+// ReportPruned writes one line per dropped row plus the undo, so a user who did
+// not ask for the cleanup can see exactly what went, why, and how to put it
+// back. Naming the reason matters: a worktree link dropped because its workitem
+// is gone otherwise reads as camp deleting worktree links.
+func ReportPruned(w io.Writer, pruned []Pruned) {
+	if w == nil || len(pruned) == 0 {
+		return
+	}
+	for _, p := range pruned {
+		_, _ = fmt.Fprintf(w, "removed dead link %s (%s:%s): %s\n",
+			p.Link.ID, p.Link.Scope.Kind, p.Link.Scope.Path, p.Reason)
+	}
+	_, _ = fmt.Fprintf(w, "  undo: git checkout -- .campaign/workitems/links.yaml\n")
 }
 
 // WorktreeScopePath returns the campaign-relative path for a project worktree
--- a/internal/workitem/links/attach.go
+++ b/internal/workitem/links/attach.go
@@ -66,14 +66,11 @@ func AttachPrimary(ctx context.Context, campaignRoot string, opts AttachOptions)
 		if err := registry.AddLink(out, opts.Replace); err != nil {
 			return err
 		}
-		if errs := Validate(ctx, registry, ValidateOptions{
+		return AsError(ValidateOne(ctx, registry, out.ID, ValidateOptions{
 			CampaignRoot: campaignRoot,
 			AllowMissing: opts.AllowMissing,
 			Now:          out.CreatedAt,
-		}); len(errs) > 0 {
-			return camperrors.NewValidation(errs[0].Field, errs[0].Message, nil)
-		}
-		return nil
+		}))
 	})
 	if err != nil {
 		return Link{}, err
--- a/internal/workitem/links/attach.go
+++ b/internal/workitem/links/attach.go
@@ -58,7 +58,15 @@ func AttachPrimary(ctx context.Context, campaignRoot string, opts AttachOptions)
 	err := WithLock(ctx, campaignRoot, func(registry *Links) error {
 		// Drop rows whose scope target is provably gone while the registry is
 		// open anyway. See Dead for why this is the only signal trusted here.
-		pruned = PruneDead(campaignRoot, registry)
+		//
+		// A caller with nowhere to report to does not prune. Reporting is the
+		// price of removing a row the user did not ask about, so a missing
+		// Report writer makes the removal silent -- and silence is the one
+		// outcome this must not have. Skipping leaves the rows for the next
+		// reporting caller or for doctor.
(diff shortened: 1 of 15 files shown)


internal/workitem/links/attach.go
```diff
@@ -66,21 +76,34 @@ func AttachPrimary(ctx context.Context, campaignRoot string, opts AttachOptions)
...
+func ReportPruned(w io.Writer, pruned []Pruned) {
+	if w == nil || len(pruned) == 0 {
+		return
+	}
+	for _, p := range pruned {
+		_, _ = fmt.Fprintf(w, "removed dead link %s (%s:%s): %s\n",
+			p.Link.ID, p.Link.Scope.Kind, p.Link.Scope.Path, p.Reason)
+	}
+	_, _ = fmt.Fprintf(w, "  undo: git checkout -- .campaign/workitems/links.yaml\n")
+}
```
obey-agent, line 118: **Non-blocking / product clarity**

The undo line is correct for recovery, but because prune and the new row are saved in the same \`links.yaml\` write, \`git checkout -- .campaign/workitems/links.yaml\` also discards the link the user just successfully created.

Worth a half-sentence in the report (e.g. "undo restores the previous registry, including removing the link just added") so the recovery path is not surprising. Not a correctness issue — just makes the automatic cleanup honest about its undo scope.

internal/workitem/links/attach.go
```diff
@@ -66,21 +76,34 @@ func AttachPrimary(ctx context.Context, campaignRoot string, opts AttachOptions)
...
-		}); len(errs) > 0 {
-			return camperrors.NewValidation(errs[0].Field, errs[0].Message, nil)
-		}
-		return nil
+		}))
 	})
 	if err != nil {
 		return Link{}, err
 	}
+	ReportPruned(opts.Report, pruned)
```
obey-agent, line 96: **Maintainability (non-blocking)**

The comment correctly requires command code to pass \`Report\`, but the library still silently prunes when \`Report == nil\` (tests and any future caller can do it).

Given the product rule that automatic removal of tracked state must never be silent, consider either:
1. requiring a non-nil \`Report\` when \`PruneDead\` removed anything (fail closed), or
2. always writing prune lines to a default (e.g. stderr) when Report is nil.

Optional — the current call sites all pass a writer — but the contract is currently social, not enforced.

obey-agent (commented):
## Verdict: Approve

Staff-level review of the stale-link write fix. This is the right diagnosis and the right shape of fix.

## Overview

One stale row in `links.yaml` was failing every subsequent link write because both writers validated the **whole** registry after insert. This PR judges writes on the subject row (`ValidateOne`), auto-cleans only **authoritatively** dead scope targets (`PruneDead` + `MachineLocal`), stops `doctor --fix` from destroying machine-local worktree links, and surfaces residual rot via a stat-level `camp status` notice. The multi-machine / git-tracked registry constraint is treated as load-bearing rather than cosmetic — that is the quality bar for this area.

## Key Findings

Non-blocking product clarity only (see inlines):

- **Undo scope** — `git checkout -- .campaign/workitems/links.yaml` after a prune+write also drops the newly created link. Worth a half-sentence in the report so recovery is not surprising.
- **`Report` is socially required, not enforced** — silent prune is still possible if a future caller omits `Report`. Call sites today are correct.

No correctness issues found. The decision not to auto-prune on missing workitem (full tree walk / branch gaps) is the right conservative line; doctor remains the explicit path.

## What's Done Well

- **`MachineLocal` as a first-class concept** with comments that explain *why* (tracked registry + multi-machine), not just what.
- **Symmetric fixes** across writers (`link`, `AttachPrimary`, worktree commands), validator, doctor, and status notice — not a one-command patch.
- **Regression tests that fail against pre-fix behavior** and the faithful 7-rot-row mirror verification story in the PR body.
- **`AsError` reporting all failures** — small, correct DX improvement.
- Existence check relaxed only for machine-local scopes; campaign-root escape still enforced via `ErrInvalidInput`.

## Staff Standard

Yes — I would be proud to ship this as my own work. Merge when CI is green. The undo wording is optional polish, not a gate.

PR #503, opened 2026-07-26
fix(workitem): promote to dungeon releases the workitem's links

lancekrogers:
Reported by Fable.

## The bug

A link attaches an **active** workitem to a working location — almost always a worktree, which is temporary by construction. `camp workitem promote --target completed|archived|someday` ends that, but promote left the links in place and nothing downstream handled the result.

`wkitem.Discover` skips dungeon directories (`internal/workitem/discover_workflow_docs.go:35`), so the selector could no longer see the workitem. The link stayed in `links.yaml` pointing at something unresolvable:

```
$ camp workitem resolve --explain     # from inside the worktree, after promote
  [link] error — primary link lnk_20260726_b588d9 points to missing workitem design-demo-2026-07-26
  [none] match — no workitem context
```

Before the promote that same command reported `[link] match`. So **`camp p commit` in that worktree silently stops stamping the ref.**

Separately, `doctor` reported it as corruption:

```
[error] workitem.link.broken — workitem_id design-demo-2026-07-26 is not present on disk
  hint: auto-fix removes the link
```

which left the campaign looking broken until someone ran `--fix`.

## Fix

Removal was always the right outcome — a workitem that is no longer active should not hold a link to a working directory. It just happened late, by a different command, for a reason that read like corruption.

So promote does it at the moment the workitem stops being active, and reports each row with its scope and the undo:

```
$ camp workitem promote WI-953817 --target completed
✓ Promoted workitem demo to workflow/design/.dungeon/completed/2026-07-26/demo
  released link lnk_20260726_da0c31 (worktree:projects/worktrees/demo-proj/wt); demo is no longer active
  undo: git checkout -- .campaign/workitems/links.yaml
```

Afterwards the state is coherent rather than broken-and-pending-cleanup:

| | before | after |
|---|---|---|
| `camp workitem doctor` | `[error] workitem.link.broken … not present on disk` | `0 findings` |
| resolver from the worktree | `[link] error — points to missing workitem` | `[link] miss` |
| `links.yaml` | stale row | released, reported, undoable |

`--json` gains `released_links[]` with id, scope kind, scope path, and role — enough to re-link by hand.

## doctor still covers the backlog

Workitems dungeoned before this landed still hold links. `doctor` keeps removing them (correct), but now distinguishes them: `dungeonedWorkitems` walks the workflow tree for `.workitem` markers under a dungeon segment, so the finding names where the workitem went and why the link is going instead of claiming it is "not present on disk". Both dungeon spellings are matched, since campaigns predating the hidden layout still use the visible one and `camp dungeon migrate` converts lazily. The walk runs only in doctor, never on a hot path.

## Note on an earlier revision of this branch

The first version preserved the links and taught doctor to treat a shelved workitem as present. That was wrong: it kept a row that nothing could resolve, so `camp p commit` still lost the ref while doctor reported everything fine. Lance pointed out the link's lifecycle is bounded by the workitem's active life. Force-pushed to the corrected approach.

## Tests

- `TestUnlinkShelvedWorkitem` — releases matching rows, returns scope detail, leaves unrelated rows alone
- `TestPrintReleasedLinks` — names the row and the undo; silent when nothing was released
- `TestDoctor_ShelvedWorkitemLinkIsNamedAndRemoved` — accurate message, and `--fix` still removes
- `TestDungeonedWorkitems`, `TestUnderDungeon` (incl. a `dungeonesque` false-positive guard)

`just gate-fast`: 3761/3761 passed, 0 lint issues, both ratchets clean.

Independent of #502 (stale rows blocking link *writes*); no overlapping files.

### DIFF
--- a/internal/commands/workitem/doctor.go
+++ b/internal/commands/workitem/doctor.go
@@ -257,7 +257,14 @@ func collectWorkitemFindings(ctx context.Context, root string, registry *links.L
 			// is corrupt" into "this is leftover housekeeping". Promote now
 			// releases these itself, so this path covers workitems dungeoned
 			// before that landed.
-			if dungeonPath, ok := shelved[link.WorkitemID]; ok {
+			//
+			// A festival target outranks that framing and is handled below: the
+			// source sits in a dungeon because doFestivalPromote put it there,
+			// and the row has somewhere real to go rather than being cleanup.
+			// Keeping it at error severity leaves doctor's exit code pointed at
+			// the one case that still needs a decision.
+			_, hasFestivalTarget := promotedTargets[link.WorkitemID]
+			if dungeonPath, ok := shelved[link.WorkitemID]; ok && !hasFestivalTarget {
 				finding.Code = codeWorkitemShelved
 				finding.Severity = docSeverityWarning
 				finding.Message = "workitem " + link.WorkitemID + " is shelved at " + dungeonPath +
--- a/internal/commands/workitem/doctor.go
+++ b/internal/commands/workitem/doctor.go
@@ -33,6 +33,7 @@ const (
 	codeOutOfBounds              = "workitem.scope.out-of-bounds"
 	codeScopeUnvalidatable       = "workitem.scope.unvalidatable"
 	codeScopeNotLocal            = "workitem.scope.not-on-this-machine"
+	codeWorkitemShelved          = "workitem.link.shelved"
 	codeDuplicatePrimary         = "workitem.link.duplicate-primary"
 	codeSchemaViolation          = "workitem.schema.violation"
 	codeCurrentMissing           = "workitem.current.missing"
@@ -231,6 +232,7 @@ func collectWorkitemFindings(ctx context.Context, root string, registry *links.L
 	}
 
 	promotedTargets := promotedFestivalTargets(ctx, root)
+	shelved := dungeonedWorkitems(ctx, root)
 	primarySeen := make(map[string]string)
 	for _, link := range registry.Links {
 		// A deprecated related-project link (doc 04 D005) is handled solely by
@@ -248,6 +250,19 @@ func collectWorkitemFindings(ctx context.Context, root string, registry *links.L
 				Message:     "workitem_id " + link.WorkitemID + " is not present on disk",
 				FixHint:     "auto-fix removes the link",
 				AutoFixable: true,
+			}
+			// A link is an active workitem's attachment to a working location,
+			// so a shelved workitem should not hold one. Removal is the right
+			// action either way; saying which case this is turns "your registry
+			// is corrupt" into "this is leftover housekeeping". Promote now
+			// releases these itself, so this path covers workitems dungeoned
+			// before that landed.
+			if dungeonPath, ok := shelved[link.WorkitemID]; ok {
+				finding.Code = codeWorkitemShelved
+				finding.Severity = docSeverityWarning
+				finding.Message = "workitem " + link.WorkitemID + " is shelved at " + dungeonPath +
+					"; a workitem that is no longer active should not hold links"
+				finding.FixHint = "auto-fix removes the link"
 			}
 			if target, ok := promotedTargets[link.WorkitemID]; ok {
 				finding.Message = "workitem_id " + link.WorkitemID +
@@ -423,7 +438,7 @@ func autoFixWorkitemFindings(ctx context.Context, root string, registry *links.L
 			continue
 		}
 		switch f.Code {
-		case codeBrokenLink:
+		case codeBrokenLink, codeWorkitemShelved:
 			id := strings.TrimPrefix(f.Target, "link:")
 			if f.migrateToID != "" {
 				if repointLinkByID(registry, id, f.migrateToID, f.migrateToKey) {
--- a/internal/commands/workitem/promote.go
+++ b/internal/commands/workitem/promote.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -51,6 +52,18 @@ type workitemPromoteResult struct {
 	Committed     bool     `json:"committed"`
 	CommitMessage string   `json:"commit_message,omitempty"`
 	Warnings      []string `json:"warnings,omitempty"`
+	// ReleasedLinks are the links dropped because the workitem is no longer
+	// active. Reported so an automatic removal is never silent.
+	ReleasedLinks []releasedLink `json:"released_links,omitempty"`
+}
+
+// releasedLink records a link that promote removed, in enough detail to put it
+// back by hand.
+type releasedLink struct {
+	ID        string `json:"id"`
+	ScopeKind string `json:"scope_kind"`
+	ScopePath string `json:"scope_path"`
+	Role      string `json:"role"`
 }
 
 type commitInputs struct {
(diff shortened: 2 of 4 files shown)


internal/commands/workitem/promote.go
```diff
@@ -246,19 +280,75 @@ func emitPromoteJSON(cmd *cobra.Command, result workitemPromoteResult) error {
...
-	}, nil
+	}
+
+	// A link attaches an active workitem to a working location -- usually a
+	// worktree, which is temporary by construction. Shelving the workitem ends
+	// that: nothing should still be checked out "for" completed work, and a
+	// link left behind only resolves to a workitem the selector cannot see
+	// (`camp p commit` silently stops stamping the ref). So the links go with
+	// the workitem, reported rather than dropped quietly.
+	dropped, err := unlinkShelvedWorkitem(ctx, campaignRoot, oldID, oldKey)
```
obey-agent: **Correctness gap — same orphan-link bug for `--target doc`**

This release path is right for completed/archived/someday. Festival promote migrates links instead. But `doDocPromote` still only calls `appendShelve`, which moves the source into the dungeon and returns **without** `unlinkShelvedWorkitem`.

So:

```
camp workitem promote WI-… --target doc
```

still leaves rows in `links.yaml` pointing at a workitem `wkitem.Discover` cannot see — the exact post-condition this PR is fixing for dungeon targets: resolver `[link] error`, silent loss of WI-* on `camp p commit`, doctor noise until `--fix`.

`unlink` cannot live inside `appendShelve` unconditionally — festival shelves first and needs those rows present for `migratePromotedLinks`. The fix belongs on the doc path (or a shared "shelve and release unless keep" helper that festival deliberately skips in favor of migrate):

1. capture `oldID, oldKey := promotedWorkitemIdentity(...)` **before** `appendShelve`
2. if `!opts.Keep` and rows dropped, set `result.ReleasedLinks` and append `links.LinksPath` to `ci.destPaths`
3. cover with a test that promotes to doc and asserts the registry no longer holds the workitem's links

Without this, the lifecycle rule "a workitem that is no longer active should not hold a link" is only half-implemented.

internal/commands/workitem/promote_dungeon_links_test.go
```diff
@@ -0,0 +1,207 @@
...
+		"workflow/design/dungeonesque/x/.workitem":                  false,
+	}
+	for path, want := range cases {
+		if got := underDungeon(path); got != want {
+			t.Fatalf("underDungeon(%q) = %v, want %v", path, got, want)
+		}
+	}
+}
+
+func TestUnlinkShelvedWorkitem(t *testing.T) {
```
obey-agent, line 134: **Test gap (secondary)**

These unit tests cover `unlinkShelvedWorkitem`, print, and doctor well, but nothing drives `doDungeonPromote` / `runWorkitemPromote` end-to-end.

A single integration case — seed link, promote `--target completed`, assert registry empty + stdout names the released id + undo — would lock the path that actually ships (identity capture before move, `destPaths` includes `links.yaml`, report wiring). The pure unit tests would not catch a future refactor that calls unlink with post-move identity or forgets to attach `ReleasedLinks` to the result.

internal/commands/workitem/doctor.go
```diff
@@ -248,6 +250,19 @@ func collectWorkitemFindings(ctx context.Context, root string, registry *links.L
 				FixHint:     "auto-fix removes the link",
 				AutoFixable: true,
 			}
+			// A link is an active workitem's attachment to a working location,
+			// so a shelved workitem should not hold one. Removal is the right
+			// action either way; saying which case this is turns "your registry
+			// is corrupt" into "this is leftover housekeeping". Promote now
+			// releases these itself, so this path covers workitems dungeoned
+			// before that landed.
+			if dungeonPath, ok := shelved[link.WorkitemID]; ok {
```
obey-agent: **Behavior note (non-blocking if auto-fix still migrates — which it does)**

When a shelved workitem also has `promoted_to` a festival, this block sets `Code = codeWorkitemShelved` / warning, then the next block overwrites Message/FixHint and sets `migrateToID` without restoring error severity or `codeBrokenLink`.

`autoFix` still repoints because it checks `migrateToID` for both codes — good. But doctor now exits 0 on a pure "should re-link to festival" backlog (warnings only), whereas previously it was an error and forced attention.

If you want the festival-migration backlog to stay loud, apply the festival branch first (or restore error severity when `migrateToID` is set). Otherwise the current "housekeeping" framing is consistent with the PR's intent.

obey-agent (commented):
## Verdict: Request Changes

Staff-level review of promote-releases-links. The dungeon path is right and well explained; the lifecycle rule is incomplete for doc promote.

## Overview

Promoting a workitem to completed/archived/someday correctly ends the workitem's active life, but left links in place. Discover skips dungeons, so the resolver errored, `camp p commit` silently stopped stamping WI-*, and doctor called it corruption. Releasing links at promote time (reported, with undo) is the correct lifecycle boundary — and the earlier "preserve links + teach doctor they're fine" revision was correctly abandoned.

## Key Findings

1. **`--target doc` still orphans links** (blocking) — `doDungeonPromote` unlinks; festival migrates; `doDocPromote` only `appendShelve`s. After `camp workitem promote --target doc` (without `--keep`), the source is shelved and links still point at an undiscoverable workitem — the same bug this PR is closing for dungeon targets. See inline on `unlinkShelvedWorkitem` call site. Cannot put unlink inside `appendShelve` unconditionally (festival needs rows for `migratePromotedLinks`); fix on the doc path after capturing identity pre-shelve.
2. **No end-to-end promote test** (secondary) — unit coverage of `unlinkShelvedWorkitem` / print / doctor is good; one integration case through `doDungeonPromote` would lock identity capture, `destPaths`, and report wiring.
3. **Shelved + festival-migrate severity** (non-blocking) — when both apply, code stays `workitem.link.shelved` (warning) while `migrateToID` is set. Auto-fix still migrates; doctor exit code no longer forces attention on that backlog. Optional to keep error severity when `migrateToID` is set.

## What's Done Well

- Correct product call after the first revision: link lifetime is bounded by active workitem life, not "present somewhere on disk."
- Capture identity **before** the move (key encodes path).
- Doctor backlog path distinguishes shelved vs truly missing, both dungeon spellings, walk only in doctor.
- Human + JSON reporting of released links with scope detail and undo — matches campaign "never silent removal" rule.
- `ErrSkipSave` when nothing dropped; `links.yaml` added to commit inputs when something was.

## Staff Standard

Almost — dungeon path yes; full promote surface not yet. Close the doc gap (and ideally one e2e test), then this is merge-ready. Independent of #502 as claimed; still fine to land after or beside it.

obey-agent (commented):
## Verdict: Approve

Re-review of `e8704a79` against the prior Request Changes. All three findings are addressed correctly.

## Overview

The follow-up finishes the lifecycle rule instead of patching one path. Dungeon and doc promotes that shelve the source both release links through a shared helper; festival still migrates (rows have a destination); `--keep` leaves links alone. Doctor keeps festival-migrate backlog at error severity so exit code still points at work that needs a decision. End-to-end promote tests lock the command path, not only the unlink primitive.

## Prior findings status

| Prior finding | Status |
|---|---|
| `--target doc` orphans links | **Fixed** — identity captured before `appendShelve`; `releaseLinksForShelvedSource` when `!Keep` |
| No end-to-end promote test | **Fixed** — dungeon releases + reports; doc releases; `--keep` preserves |
| Shelved + festival severity softens exit code | **Fixed** — shelved warning only when `!hasFestivalTarget` |

## What's Done Well

- Shared `releaseLinksForShelvedSource` (not a second copy of unlink wiring) and a clear comment on why it cannot live in `appendShelve` (festival needs the rows for migrate).
- Doc path captures identity before the source moves — same timing discipline as dungeon.
- `ReleasedLinks` uses append so the helper stays composable.
- Tests fail against pre-fix doc behavior (stated in commit) and cover the keep branch.
- Local package tests for the new cases pass (`go test ./internal/commands/workitem/ -run 'Test(UnlinkShelved|PrintReleased|Doctor_Shelved|Dungeoned|UnderDungeon|PromoteTo)'`).

## Optional polish (non-blocking)

- PR body test list still omits the three new e2e cases — update when convenient.
- No dedicated unit test for the doctor `hasFestivalTarget` branch; the code is small and clear, so not a gate.

## Staff Standard

Yes. Merge when CI is green. This is the complete form of the original fix.

PR #504, opened 2026-07-26
feat(switch): carry hop origin, add 'camp switch -', resolve self-references locally

lancekrogers:
Ships the P0 half of the mesh-navigation loop: a hop now carries where it came from, `camp switch -` returns, and a selector naming the current machine resolves locally instead of ssh-ing to itself.

Festival MN0001, phase 004 sequence 01. Design is in `festivals/active/mesh-navigation-mutual-machine-visibility-MN0001/003_DESIGN/01_origin_hopback/outputs/origin-hopback-spec.md` (every symbol in it is anchor-verified against camp main).

## Why

A hop knows where it landed but not where it came from, so the far session cannot offer to hop back or to register the origin. sshd's default `AcceptEnv` blocks `SendEnv` forwarding, so the identity rides the remote command instead of the ssh environment: no sshd config on any fleet member.

## What changed

**`CAMP_HOP_ORIGIN` payload** (`cmd/camp/hop_origin.go`). Single line, `v1;host=…;user=…;campaign=…;id=…`, every byte outside `[A-Za-z0-9._~-]` percent-encoded. That encoding is what makes a campaign named `x;host=evil.example.com` inert: it cannot inject a second `host` pair, and it cannot break out of either shell-quoting layer. Over-cap optional fields are dropped rather than truncated, because a truncated hostname is a wrong hostname and wrong routing is worse than absent routing.

**Embedding** (`cmd/camp/switch.go`). `emitShellConnect` gains an `origin` parameter and prepends `export CAMP_HOP_ORIGIN=<quoted> && ` to the remote command. `exec` preserves it into the login shell. The local `cd --` path is untouched and pinned by a regression test.

**`camp switch -`** (`cmd/camp/hop_origin.go`). Reads the payload, builds a transient in-memory machine, resolves the origin campaign on the origin's own camp, emits the reverse hop line. Never reads or writes `machines.yaml` for the gesture itself (D008: registration-independent), which a test asserts. The reverse line carries its own origin, so the gesture is a toggle rather than a one-way trip.

**Self-reference** (`cmd/camp/hop_origin.go`). `archdtop:notes` typed on archdtop resolves locally. One rule: the id derived from this machine's reachable name, using the same derivation that fills the payload's `id`, so the two answers cannot drift.

**File split.** `switch.go` was over the project's 500-line ceiling (545 before this work, 572 with it). This performs the shared-tail extraction the design spec already planned: `switch_remote.go` holds `guardRemoteOutputFlags` and `emitHopOrRefuse`, shared by both hop entry points so their refusal text cannot drift; `switch_selector.go` holds the selector/scope helpers. `switch.go` is now 469 lines. Pure moves.

## Live verification (mac-studio to archdtop)

Emitted line:

```
exec ssh -t '-o' 'StrictHostKeyChecking=accept-new' ... 'lance@archdtop.tail37114b.ts.net' 'export CAMP_HOP_ORIGIN='\''v1;host=mac-studio.tail37114b.ts.net;user=lancerogers;campaign=obey-campaign;id=mac-studio'\'' && cd '\''/home/lance/campaigns/lance-arch'\'' && exec $SHELL -l'
```

Executing its remote half:

```
host=archdtop
pwd=/home/lance/campaigns/lance-arch
CAMP_HOP_ORIGIN=v1;host=mac-studio.tail37114b.ts.net;user=lancerogers;campaign=obey-campaign;id=mac-studio
```

Gesture rows against the real fleet:

```
$ camp switch -
Error: camp switch -: this session did not start from a camp hop (CAMP_HOP_ORIGIN is not set)

$ CAMP_HOP_ORIGIN=<archdtop payload> camp switch -
Error: resolved archdtop:lance-arch -> /home/lance/campaigns/lance-arch; run via the csw shell wrapper to hop there

$ CAMP_HOP_ORIGIN=<archdtop payload> camp switch - --print
Error: remote target archdtop: --print is local-only; use the csw shell wrapper to hop there

$ CAMP_HOP_ORIGIN='v1;host=a;host=b;user=c' camp switch -
Error: camp switch -: CAMP_HOP_ORIGIN is malformed (duplicate key "host"); hop back with 'camp switch <machine>:<campaign>'
```

The bare row resolved through a live ssh to the real archdtop, which also exercises the registered-entry path (it used id `archdtop`, not the transient `archdtop-origin`).

Self-reference on the mac:

```
$ camp switch mac-studio:obey-campaign --print      -> /Users/lancerogers/Dev/AI/obey-campaign
$ camp switch mac-studio:obey-campaign@p --print    -> /Users/lancerogers/Dev/AI/obey-campaign/projects
$ camp switch mac-studio:obc/obey-campaign --print  -> /Users/lancerogers/Dev/AI/obey-campaign
$ camp switch archdtop:lance-arch --print           -> Error: remote target archdtop: --print is local-only …
```

## Tests

`just gate` green: 3762 tests. The hop-back decision table runs with no live ssh via an injected resolver seam, mirroring the existing `tailscaleStatusFunc` pattern. Notable coverage: 11 encode rows, 10 parse-rejection rows, 7 host-detection rows including five fallbacks, the full 10-row hop-back decision table each asserting stdout stays empty, plus pins for "empty origin emits byte-identical output to before", "hop-back never writes machines.yaml", and "the reverse line carries its own origin".

## Deliberate choices a reviewer may want to push on

1. **The transient machine leaves `auth_method` empty.** The hop argv is identical across auth methods (`authArgs` branches only on `IdentityFile`), `EnsureKeyAuth` rejects only `ssh-password`, and failure classification reads ssh's own stderr. Guessing would only put a wrong label in front of a broken hop.
2. **A registered entry wins over self-detection.** A `machines.yaml` row pointing at this machine is a misconfiguration that sequence 02's adopt verb is specified to refuse creating. Honoring an explicit operator file beats second-guessing it on every hop.
3. **Origin failures are warnings, never errors.** No reachable name, no tailscale, or no campaign context each yield no export and a stderr note; the hop itself is unaffected.

## Found while testing, not fixed here

`camp switch machine:campaign` exits **127 with entirely empty stdout and stderr** when the remote camp is not on the far machine's login PATH. Reproduces identically with the released binary from `main`, so it predates this branch. The actionable hint exists in `campNotFoundHint` (`internal/remote/ssh.go:602`) but never reaches the user. Worked around during testing with `CAMP_REMOTE_CAMP_PATH`. Deserves its own issue rather than scope creep here.

### DIFF
--- a/cmd/camp/hop_origin.go
+++ b/cmd/camp/hop_origin.go
@@ -10,7 +10,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/spf13/cobra"
+
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/machines"
+	"github.com/Obedience-Corp/camp/internal/remote"
 )
 
 // HopOriginEnvVar carries the origin machine identity into a hopped session.
@@ -308,4 +312,98 @@ func suggestedMachineID(host string) string {
 		return ""
 	}
 	return id
+}
+
+// hopBackSelector is the reserved argument for the hop-back gesture. It is
+// checked before selector parsing because `-` is otherwise a legal campaign
+// query that reaches the fuzzy tier and matches an arbitrary hyphenated
+// campaign name, differently on every invocation.
+const hopBackSelector = "-"
+
+// transientOriginSuffix namespaces the in-memory machine built from a payload.
+// The id becomes a ControlMaster socket name, so colliding with a REGISTERED
+// machine id would make ssh reuse that machine's live master and land the hop
+// on the wrong host. A registered id ending in "-origin" would have to be
+// created deliberately.
+const transientOriginSuffix = "-origin"
+
+// runHopBack implements `camp switch -`: return to the machine and campaign this
+// session was hopped from. Registration-independent by design — the origin need
+// never have been adopted here — so it reads the payload and builds a transient
+// machine rather than consulting ~/.obey/machines.yaml. Nothing is written.
+func runHopBack(ctx context.Context, cmd *cobra.Command, printOnly, shellConnect, jsonOut bool) error {
+	raw := os.Getenv(HopOriginEnvVar)
+	if strings.TrimSpace(raw) == "" {
+		return camperrors.New("camp switch -: this session did not start from a camp hop (" +
+			HopOriginEnvVar + " is not set)")
+	}
+	origin, err := ParseHopOrigin(raw)
+	if err != nil {
+		return camperrors.New("camp switch -: " + HopOriginEnvVar + " is malformed (" + err.Error() +
+			"); hop back with 'camp switch <machine>:<campaign>'")
+	}
+	if origin.Campaign == "" {
+		return camperrors.New("camp switch -: origin campaign unknown (the outbound hop did not start " +
+			"inside a campaign); hop back with 'camp switch <machine>:<campaign>'")
+	}
+
+	m, registered := originTarget(origin)
+	if err := guardRemoteOutputFlags(m.ID, printOnly, jsonOut); err != nil {
+		return err
+	}
+	root, err := hopBackResolveRoot(ctx, m, origin.Campaign)
+	if err != nil {
+		return hopBackFailure(err, m, registered)
+	}
+	return emitHopOrRefuse(ctx, cmd, m, origin.Campaign, root, shellConnect)
+}
+
+// originTarget resolves the payload to a machine to hop to. A registered entry
+// matching the origin host wins: it carries the operator's own identity_file,
+// ssh_user, and auth_method, all better information than the payload has. This
+// is the expected steady state once a fleet is mutually adopted.
+//
+// Otherwise the machine is transient and in-memory. AuthMethod is left empty on
+// purpose: the hop argv is identical for tailscale-ssh and ssh-agent (authArgs
+// branches only on IdentityFile), EnsureKeyAuth rejects only ssh-password, and
+// failure classification reads ssh's own stderr. Guessing an auth mode would put
+// a wrong label on the one message an operator reads when a hop breaks.
+func originTarget(origin HopOrigin) (*machines.Machine, bool) {
+	want := strings.ToLower(normalizeDNSName(origin.Host))
+	if mf, err := machines.Load(); err == nil {
+		for i := range mf.Machines {
+			if strings.ToLower(normalizeDNSName(mf.Machines[i].Host)) == want {
+				return &mf.Machines[i], true
+			}
+		}
+	}
+	return &machines.Machine{
+		ID:      transientOriginID(origin.Host),
+		Label:   origin.Host,
+		Host:    origin.Host,
+		SSHUser: origin.User,
+	}, false
+}
+
+// transientOriginID derives the in-memory machine's id. It is only two things:
+// a ControlMaster path component and a name in error text. Both tolerate a
+// generic value; neither tolerates an empty one, which would collapse the socket
+// path to ~/.obey/ssh-ctl/.sock and share one socket across every degenerate
+// origin — the exact cross-host reuse the suffix exists to prevent.
+func transientOriginID(host string) string {
+	if id := suggestedMachineID(host); id != "" {
+		return id + transientOriginSuffix
+	}
+	return strings.TrimPrefix(transientOriginSuffix, "-")
+}
+
+// hopBackFailure explains an unreachable origin. The diagnostic pointer differs
+// by registration because `camp machine diagnose` only knows registered
+// machines: pointing an operator at it for a transient origin sends them to a
+// command that answers "unknown machine".
+func hopBackFailure(err error, m *machines.Machine, registered bool) error {
+	if registered {
+		return camperrors.Wrapf(err, "run 'camp machine diagnose %s' to check reachability", m.ID)
+	}
+	return camperrors.Wrapf(err, "the origin is not registered here, probe it with: %s", remote.ProbeCommand(m))
 }
--- a/cmd/camp/hop_origin.go
+++ b/cmd/camp/hop_origin.go
@@ -407,3 +407,31 @@ func hopBackFailure(err error, m *machines.Machine, registered bool) error {
 	}
 	return camperrors.Wrapf(err, "the origin is not registered here, probe it with: %s", remote.ProbeCommand(m))
 }
+
+// isSelfMachine reports whether a selector's machine segment names THIS
+// machine, by comparing against the id derived from this machine's reachable
+// name — character for character the derivation that fills the payload's id
+// field, so the two answers cannot disagree.
+//
+// Ordering matters and is deliberate. This is checked only AFTER
+// machines.LocalMachineID and only for a selector that would otherwise hop, so
+// the detection cost (a tailscale probe) never lands on a successful local
+// switch. A registered machine still wins: a machines.yaml entry pointing at
+// this machine is a misconfiguration that `camp machine adopt` refuses to
+// create, and honoring the operator's explicit file here is better than
+// second-guessing it on every hop.
+func isSelfMachine(ctx context.Context, id string) bool {
+	if id == "" {
+		return false
+	}
+	if mf, err := machines.Load(); err == nil {
+		if _, _, found := mf.Lookup(id); found {
+			return false
+		}
+	}
+	host, err := detectReachableName(ctx, runTailscaleStatusForSelf)
+	if err != nil || host == "" {
+		return false
+	}
+	return suggestedMachineID(host) == id
+}
--- a/cmd/camp/hop_origin.go
+++ b/cmd/camp/hop_origin.go
@@ -351,7 +351,7 @@ func runHopBack(ctx context.Context, cmd *cobra.Command, printOnly, shellConnect
 	if err := guardRemoteOutputFlags(m.ID, printOnly, jsonOut); err != nil {
 		return err
 	}
-	root, err := hopBackResolveRoot(ctx, m, origin.Campaign)
+	root, err := resolveRemoteRoot(ctx, m, origin.Campaign)
 	if err != nil {
 		return hopBackFailure(err, m, registered)
 	}
--- a/cmd/camp/hop_origin.go
+++ b/cmd/camp/hop_origin.go
@@ -149,14 +149,29 @@ func ParseHopOrigin(payload string) (HopOrigin, error) {
 		if hopOriginHasC0(val) {
 			return HopOrigin{}, camperrors.New("field \"" + key + "\" contains a control character")
 		}
+		// Re-apply encode-side field caps on the wire form (raw, pre-decode
+		// length matches encodeHopOrigin's len(encoded) checks). Required
+		// fields fail closed; optional fields drop, same as encode.
 		switch key {
 		case "host":
+			if len(raw) > hopOriginMaxHost {
+				return HopOrigin{}, camperrors.New("field \"host\" is too long")
+			}
 			o.Host = val
 		case "user":
+			if len(raw) > hopOriginMaxUser {
+				return HopOrigin{}, camperrors.New("field \"user\" is too long")
+			}
 			o.User = val
 		case "campaign":
+			if len(raw) > hopOriginMaxCampaign {
+				continue
+			}
 			o.Campaign = val
 		case "id":
+			if len(raw) > hopOriginMaxID {
+				continue
+			}
 			o.ID = val
 		}
 		// Unknown keys are ignored so additive v1 fields do not break older
--- a/cmd/camp/hop_origin.go
+++ b/cmd/camp/hop_origin.go
@@ -22,6 +22,21 @@ import (
 // the far machine can name where the session came from without any reverse
 // connection or sshd configuration. Mirrors the CAMP_QUEST pattern
 // (internal/quest/shellenv.go) with more fields.
+//
+// Trust model, because this value steers an ssh: it is session-local state, not
+// an authenticated claim. Anything that can set the environment of a camp
+// process can choose where `camp switch -` dials, and hops use
+// StrictHostKeyChecking=accept-new, so a first connection to a host named by a
+// forged payload accepts its key. Encoding bounds what a payload can SAY -- the
+// unreserved set keeps ';' and '=' out of values, parse rejects C0 controls and
+// re-applies the field caps, and remote.ShellQuote keeps the emitted line from
+// breaking out of its quoting -- but none of that binds the payload to a
+// cryptographic identity.
+//
+// What limits the blast radius is that the same trust already exists: a process
+// that can set this variable can equally invoke ssh directly. The value ranks
+// below the operator's own fleet file, which is why originTarget prefers a
+// registered row whose host matches, and hop-back never writes anything.
 const HopOriginEnvVar = "CAMP_HOP_ORIGIN"
 
 // hopOriginVersion prefixes every payload. A consumer that does not recognize
@@ -380,6 +395,14 @@ func runHopBack(ctx context.Context, cmd *cobra.Command, printOnly, shellConnect
 			"inside a campaign); hop back with 'camp switch <machine>:<campaign>'")
 	}
 
+	// A payload naming this machine is stale or forged; either way the hop it
+	// would emit is an ssh to ourselves, which yields a nested session rather
+	// than a switch. Say where the user already is instead.
+	if isSelfHost(ctx, origin.Host) {
+		return camperrors.New("camp switch -: the origin is this machine; " +
+			"you are already here — use 'camp switch " + origin.Campaign + "' to switch locally")
+	}
+
 	m, registered := originTarget(origin)
 	if err := guardRemoteOutputFlags(m.ID, printOnly, jsonOut); err != nil {
 		return err
@@ -452,16 +475,17 @@ func hopBackFailure(err error, m *machines.Machine, registered bool) error {
 // switch. A registered id wins over self-detection: Lookup(id) only — any
 // machines.yaml row whose id equals the selector suppresses local resolve,
 // regardless of that row's host. That honors the operator's explicit id
-// mapping; adopt refuses to create a true self-row, but a shared fleet file
-// that lists every node under its real id (including this one) will hop via
-// ssh rather than short-circuit locally.
+// mapping — except when that entry points back at this machine, which is the
+// one case where honoring it means sshing to ourselves. A fleet file shared
+// verbatim across every node lists this one under its real id, and that is an
+// ordinary way to run a mesh rather than a misconfiguration.
 func isSelfMachine(ctx context.Context, id string) bool {
 	if id == "" {
 		return false
 	}
 	if mf, err := machines.Load(); err == nil {
-		if _, _, found := mf.Lookup(id); found {
-			return false
+		if m, _, found := mf.Lookup(id); found {
+			return m != nil && isSelfHost(ctx, m.Host)
 		}
 	}
 	host, err := detectReachableName(ctx, runTailscaleStatusForSelf)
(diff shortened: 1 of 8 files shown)


cmd/camp/switch_remote.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/machines"
+	navfuzzy "github.com/Obedience-Corp/camp/internal/nav/fuzzy"
+	"github.com/Obedience-Corp/camp/internal/remote"
+)
+
+// hopBackResolveRoot is the seam that lets the hop-back decision table be tested
+// without a live ssh. Production is remote.ResolveRoot; tests substitute a stub.
+var hopBackResolveRoot = remote.ResolveRoot
```
obey-agent: **Naming/maintainability:** \`hopBackResolveRoot\` is now the resolve seam for *all* remote switches (\`runRemoteSwitch\` line 40), not only hop-back. Fine for test injection, but the name will mislead the next editor into thinking normal remote switch still calls \`remote.ResolveRoot\` directly.

Rename to something like \`resolveRemoteRoot\` (defaulting to \`remote.ResolveRoot\`) so both hop-back and machine:campaign share an honestly named seam.

obey-agent (commented):
## Verdict: Approve

## Overview
P0 half of mesh navigation done to a high bar: origin payload design (versioned, percent-encoded, size-capped, fail-closed parse), embedding without sshd `AcceptEnv` changes, `camp switch -` as registration-independent hop-back, and self-reference resolution that cannot drift from the payload's `id` derivation. File split of `switch.go` is pure extraction and keeps the shared refusal path from diverging. Live verification and a large, decision-table-heavy test suite make this production-credible.

## Key Findings
- **Maintainability (non-blocking):** `hopBackResolveRoot` is now the resolve seam for *all* remote switches, not only hop-back. Rename to `resolveRemoteRoot` (or similar) so the next editor is not misled. See inline.
- Deliberate choices (empty auth on transient machines, registered entry wins over self-detection, origin failures as warnings) are sound and documented. Agree with leaving the silent-exit bug to its own PR (#510).

## What's Done Well
- Encoding treats separators as structural, not conventional — hostile campaign names cannot inject fields.
- Over-cap optional fields are dropped, never truncated (wrong routing is worse than absent routing).
- Transient machine id suffix prevents ControlMaster socket collision with registered hosts.
- Injected resolve seam mirrors existing camp patterns; hop-back decision table asserts empty stdout on failures.
- Reverse hop carries its own origin (toggle, not one-way trip).

## Staff Standard
Yes. This is staff-level work — design-faithful, security-conscious, and tested at the decision boundaries that matter. Merge when stack base is ready.

cmd/camp/hop_origin.go
```diff
@@ -0,0 +1,437 @@
...
+// this machine is a misconfiguration that `camp machine adopt` refuses to
+// create, and honoring the operator's explicit file here is better than
+// second-guessing it on every hop.
+func isSelfMachine(ctx context.Context, id string) bool {
+	if id == "" {
+		return false
+	}
+	if mf, err := machines.Load(); err == nil {
+		if _, _, found := mf.Lookup(id); found {
+			return false
```
obey-agent: **[bug]** `isSelfMachine` returns false for any id present in `machines.yaml` without checking whether that entry’s `Host` is this machine. A common full-mesh fleet file (same yaml on every node listing every node, including “self”) therefore disables local self-resolve for the natural selector (`mac-studio:campaign` on mac-studio) and falls through to `runRemoteSwitch` → SSH to self — reintroducing the self-hop loop this PR claims to fix. The comment calls a self-row a misconfiguration that adopt will refuse, but nothing stops a synced/shared fleet file, and docs only say the implicit id is `local` (not that listing your own host under its real id is invalid). The unit test only pins registered-wins when `host: somewhere.else`, so the same-host case is untested.

**Suggestion:** After `Lookup(id)`, compare `normalizeDNSName(m.Host)` to `detectReachableName(...)` (case-insensitive). If equal, treat as self and resolve locally; if different, keep remote. Add a test: registered id with host == reachable name → `isSelfMachine` true; registered id with foreign host → false.

cmd/camp/hop_origin.go
```diff
@@ -0,0 +1,437 @@
...
+// session was hopped from. Registration-independent by design — the origin need
+// never have been adopted here — so it reads the payload and builds a transient
+// machine rather than consulting ~/.obey/machines.yaml. Nothing is written.
+func runHopBack(ctx context.Context, cmd *cobra.Command, printOnly, shellConnect, jsonOut bool) error {
+	raw := os.Getenv(HopOriginEnvVar)
+	if strings.TrimSpace(raw) == "" {
+		return camperrors.New("camp switch -: this session did not start from a camp hop (" +
+			HopOriginEnvVar + " is not set)")
+	}
+	origin, err := ParseHopOrigin(raw)
```
obey-agent, line 388: **[suggestion]** `CAMP_HOP_ORIGIN` is a session-local trust boundary that hop-back acts on immediately under `--shell-connect` (`eval` → `exec ssh`). Any process that can set the env (or a compromised origin that rewrites the hop line) can point `csw -` at an arbitrary `user@host`/`campaign`. Transient targets use payload host/user with empty `AuthMethod`, and hops use `StrictHostKeyChecking=accept-new`, so a first connection to a spoofed host auto-accepts its key. Encoding correctly blocks field injection and shell breakout; it does not bind the payload to a cryptographic identity of the origin machine. Agents use `switch` (`agent_allowed`) and the csw wrapper, so blind hop-back is a realistic path.

**Suggestion:** Document the threat model next to `HopOriginEnvVar`. Consider refusing transient (unregistered) hop-back unless the operator/agent passes an explicit flag, or print the resolved `user@host:campaign` on stderr and require registered-host match for non-interactive hops. At minimum, surface the destination clearly before eval in wrapper UX notes.

obey-agent (commented):
## Summary

Solid mesh-navigation P0: percent-encoded `CAMP_HOP_ORIGIN` (v1, strict parse, size caps), dual-layer shell safety (`hopOriginEncode` + `ShellQuote`), `camp switch -` registration-independent hop-back with a clean resolver seam, self-reference local resolve, and a strong decision-table test suite (encode/parse, embedding, hop-back rows, no machines.yaml write, reverse-origin toggle, shell `eval "$line"` pin). Origin failures correctly degrade to “no export + stderr warning” rather than failing the hop. Shared remote tail (`guardRemoteOutputFlags` / `emitHopOrRefuse`) keeps `-` and `machine:campaign` refusal text aligned.

Main residual risks are policy/trust-boundary choices rather than classic injection: env-sourced origin is fully trusted for routing under `--shell-connect`; `isSelfMachine` treats any registered id as non-self without comparing hosts (breaks self-local-resolve for full-mesh `machines.yaml`); parse is laxer than encode on controls/field caps. Shell injection of the export payload itself is well covered.

## Issue counts by severity

- bugs: 1
- suggestions: 4
- nits: 3

## Issues outside the diff

- **[suggestion]** cmd/camp/hop_origin_test.go:952 -- `TestRunHopBackReverseLineCarriesItsOwnOrigin` only asserts that `export CAMP_HOP_ORIGIN=` appears. It does not check that the reverse payload parses, names this machine (not the inbound origin), or survives `ParseHopOrigin`. A regression that re-exports the inbound env (or an empty/malformed value) could still pass. `TestRunHopBackDecisionTable`’s shell-connect success row only looks for `exec ssh -t `.
  - **Suggestion:** Parse the export from stdout, `ParseHopOrigin` it, and assert host/id differ from the inbound test payload (and/or match `buildHopOrigin` with an injected detect func). Tighten the decision-table success row similarly.
- **[nit]** cmd/camp/switch.go:24 -- `Long` / `Example` still document `csw archdtop:lance-arch` but not `csw -` / hop origin. Operators and agents discover the gesture only from errors or festival notes.
  - **Suggestion:** Add a short Long/Example blurb: hopped sessions carry `CAMP_HOP_ORIGIN`; `csw -` returns to the origin machine+campaign via the shell wrapper.
- **[nit]** cmd/camp/hop_origin_test.go:1079 -- Subtest comment says “A machines.yaml row pointing at this machine” but the fixture uses `host: somewhere.else`. Comment and fixture disagree; future readers may think same-host registration is covered.
  - **Suggestion:** Fix the comment to “registered id collides with derived self id but points at a foreign host”.

cmd/camp/hop_origin.go
```diff
@@ -0,0 +1,437 @@
...
+// ParseHopOrigin reads a payload. A malformed payload is an error rather than a
+// best-effort parse: the value routes a hop, and a half-understood route is the
+// failure mode the size rules already refuse.
+func ParseHopOrigin(payload string) (HopOrigin, error) {
+	payload = strings.TrimSpace(payload)
+	if payload == "" {
+		return HopOrigin{}, camperrors.New("empty payload")
+	}
+	if len(payload) > hopOriginMaxTotal {
+		return HopOrigin{}, camperrors.New("payload exceeds 1024 bytes")
```
obey-agent, line 139: **[suggestion]** `ParseHopOrigin` does not re-apply encode-side field caps or reject C0 controls after percent-decode. A hand-crafted env can supply over-long hosts, embedded `@`/whitespace, or `%0A` newlines in `host`/`user`. Production emitters cannot produce these (`encodeHopOrigin` + unreserved set). Consumers still ShellQuote Target/inner and use argv for `ResolveRoot`, so this is not a straightforward RCE, but: (1) newlines break the documented one-line `--shell-connect` stdout contract (fish command substitution splits on newlines); (2) `hopBackFailure` → `remote.ProbeCommand` builds an unquoted copy-paste string (`user@host` joined with spaces), so a hostile host with `;`/`$()` becomes a paste footgun.

**Suggestion:** After decode, enforce `hopOriginMaxHost`/`MaxUser`/… and reject values containing `\n`, `\r`, NUL, or other C0 controls for required fields. Quote or `ShellQuote` components inside `ProbeCommand` (or a hop-back-specific probe display) so error text is safe to paste.

cmd/camp/hop_origin.go
```diff
@@ -0,0 +1,437 @@
...
+// created deliberately.
+const transientOriginSuffix = "-origin"
+
+// runHopBack implements `camp switch -`: return to the machine and campaign this
+// session was hopped from. Registration-independent by design — the origin need
+// never have been adopted here — so it reads the payload and builds a transient
+// machine rather than consulting ~/.obey/machines.yaml. Nothing is written.
+func runHopBack(ctx context.Context, cmd *cobra.Command, printOnly, shellConnect, jsonOut bool) error {
+	raw := os.Getenv(HopOriginEnvVar)
+	if strings.TrimSpace(raw) == "" {
```
obey-agent, line 384: **[suggestion]** `runHopBack` never short-circuits when the origin host is this machine (spoofed or stale env). That path still pays `resolveRemoteRoot` + emits `exec ssh` to self, yielding a nested self-session rather than a local switch. No test covers origin-host-is-self.

**Suggestion:** If `originTarget` resolves to a host matching `detectReachableName`, resolve the campaign locally (or error with a clear “origin is this machine; use camp switch <campaign>”) instead of SSHing. Add a decision-table row.

cmd/camp/switch_remote.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/Obedience-Corp/camp/internal/machines"
+	navfuzzy "github.com/Obedience-Corp/camp/internal/nav/fuzzy"
+	"github.com/Obedience-Corp/camp/internal/remote"
+)
+
+// resolveRemoteRoot is the seam every remote switch resolves through, hop-back
+// and machine:campaign alike, so both can be tested without a live ssh.
+// Production is remote.ResolveRoot; tests substitute a stub.
+var resolveRemoteRoot = remote.ResolveRoot
+
```
obey-agent, line 26: **[nit]** Package-level `var resolveRemoteRoot` is swapped in tests without a mutex. Current tests avoid `t.Parallel()`, so green today; a later parallel test will race.

**Suggestion:** Document “not parallel-safe” on the var, or use a `t.Cleanup`-friendly test helper that serializes / restores under a mutex.

cmd/camp/hop_origin.go
```diff
@@ -0,0 +1,437 @@
...
+			o.User = val
+		case "campaign":
+			o.Campaign = val
+		case "id":
+			o.ID = val
+		}
+		// Unknown keys are ignored so additive v1 fields do not break older
+		// consumers; a change in meaning bumps the version instead.
+	}
+	if o.Host == "" {
```
obey-agent, line 195: **Defense-in-depth:** encode rejects raw `\n`/`\r` in the assembled payload, but parse does not reject control bytes after percent-decode.

A handcrafted env value such as `v1;host=box%0Aevil;user=u;campaign=c` parses successfully; `originTarget` then builds `Host` with an embedded newline, which flows into `remote.Target` / `ShellQuote` / ControlPath id derivation.

ShellQuote keeps this from becoming a shell-injection breakout, so this is not a remote RCE path — more “weird ssh argv / confusing failure.” Mirroring the encode guard here (reject decoded host/user/campaign/id if they contain control bytes or newlines) would keep parse symmetric with “wrong routing is worse than absent routing” and close the env-injection shape without cost.

cmd/camp/hop_origin.go
```diff
@@ -0,0 +1,437 @@
...
+// switch. A registered machine still wins: a machines.yaml entry pointing at
+// this machine is a misconfiguration that `camp machine adopt` refuses to
+// create, and honoring the operator's explicit file here is better than
+// second-guessing it on every hop.
+func isSelfMachine(ctx context.Context, id string) bool {
+	if id == "" {
+		return false
+	}
+	if mf, err := machines.Load(); err == nil {
+		if _, _, found := mf.Lookup(id); found {
```
obey-agent: **Comment/code mismatch (worth a one-line fix):** the comment says a registered entry *pointing at this machine* wins, but the check is `Lookup(id)` only — any machines.yaml row whose **id** equals the selector, regardless of that row’s `host`.

That is probably the right rule (honor the operator’s explicit id mapping), but it is not “host points at self.” Under the stated adopt-refuses-self invariant this is fine in steady state; if a row is ever present with `id == suggestedMachineID(self)` and a different host, self-ref is suppressed and the hop goes to that host.

Please either:
1. tighten the comment to “a registered **id** wins over self-detection,” or
2. if the design intent really is host-based, compare the registered host to `detectReachableName` before suppressing.

A unit test already pins “registered entry wins”; renaming the case/comment to match Lookup-by-id would prevent a later reader from “fixing” this into a host compare that changes behavior.

cmd/camp/switch.go
```diff
@@ -289,6 +289,17 @@ func runSwitch(cmd *cobra.Command, args []string) error {
...
 	}
 
+	// The hop-back gesture is handled before the registry is even loaded, for
+	// two reasons. It is checked before selector parsing because "-" is
+	// otherwise a legal campaign query that reaches the fuzzy tier and matches
+	// an arbitrary hyphenated campaign, differently on every invocation. And it
+	// is checked before the "no campaigns registered" guard because hop-back
+	// resolves against the ORIGIN machine's registry, so a local registry is not
+	// a precondition and pointing the operator at `camp init` would be wrong.
+	if len(args) == 1 && args[0] == hopBackSelector {
```
obey-agent, line 307: **Discoverability gap:** hop-back is correctly reserved here (before fuzzy parse / empty-registry guard), but `switchCmd.Long` / `Example` and `docs/cli-reference/camp_switch.md` still never mention `camp switch -` / `csw -`.

A reserved P0 gesture with no `camp switch --help` path is easy to miss. Suggest documenting:
- `csw -` returns to the machine/campaign this session hopped from (`CAMP_HOP_ORIGIN`)
- registration-independent; refuses `--print`/`--json` like other remote targets
- `-` is reserved and is no longer a fuzzy campaign query

(Help text is outside this diff hunk — fine as a follow-up commit on the branch.)

cmd/camp/hop_origin.go
```diff
@@ -0,0 +1,437 @@
...
+	if o.Host == "" {
+		return HopOrigin{}, camperrors.New("missing host")
+	}
+	if o.User == "" {
+		return HopOrigin{}, camperrors.New("missing user")
+	}
+	return o, nil
+}
+
+// hopOriginEncode percent-encodes everything outside the unreserved set, which
```
obey-agent, line 216: **Nice:** encoding the unreserved set so `;`/`=`/`%` never appear raw is the right injection boundary. Combined with `remote.ShellQuote` on the export and the wrapper’s `eval "$line"` pin in `internal/shell/shell_connect_test.go`, this is layered correctly rather than relying on a single quoting style.

obey-agent (commented):
## Verdict

**Approve with minor comments** — ship-quality P0 for mesh hop origin / hop-back / self-ref. No blocking correctness bugs found. The implementation matches the stated design tradeoffs, is well-tested (including negative/safety pins), and was live-verified on the fleet.

---

## Overview

PR #504 threads origin identity through remote hops via `CAMP_HOP_ORIGIN`, adds `camp switch -` as a registration-independent hop-back toggle, and resolves selectors that name *this* machine locally so agents/humans do not SSH to themselves after a hop. Shared remote tail (`emitHopOrRefuse` / `guardRemoteOutputFlags`) and a `resolveRemoteRoot` seam keep the two hop entry points consistent and unit-testable without live SSH. `switch.go` is brought back under the 500-line ceiling via pure extractions.

---

## Key Findings

### Medium

1. **Hop-back is undiscoverable from CLI help / reference docs** (`cmd/camp/switch.go` Long/Example; `docs/cli-reference/camp_switch.md`)
   - Reserved selector `-`, env contract, and gesture semantics are invisible to `camp switch --help`.
   - Recommend documenting `csw -`, that it needs `CAMP_HOP_ORIGIN`, and that `-` is no longer a fuzzy campaign query.
   - Inline: hop-back intercept in `runSwitch`.

### Low

2. **`ParseHopOrigin` does not reject control bytes after percent-decode** (`cmd/camp/hop_origin.go`)
   - Encode forbids raw newlines in the assembled payload; parse accepts `host=x%0Ay`.
   - ShellQuote prevents shell breakout; still produces a hostile/odd SSH target from a handcrafted env.
   - Mirror encode’s control/newline rejection on decoded fields for parse/encode symmetry.
   - Inline on `ParseHopOrigin`.

3. **`isSelfMachine` comment vs behavior** (`cmd/camp/hop_origin.go`)
   - Comment: registered entry “pointing at this machine” wins.
   - Code: any `Lookup(id)` hit suppresses self-detection (id-based, not host-based).
   - Steady state is fine under adopt-refuses-self; please align comment (or host-compare) so the next change does not “fix” the wrong invariant.
   - Inline on the Lookup short-circuit.

### Nits / non-blocking

4. Reverse-line test asserts `export CAMP_HOP_ORIGIN=` is present, not that the reverse payload describes *this* machine (toggle property only partially pinned).
5. Origin carries bare `cfg.Name` only (no org/tab). Hop-back to campaign root of that name is a reasonable scope for this sequence — worth one sentence in help when docs land.

---

## What's Done Well

- **Injection boundary is real:** percent-encoding of the unreserved set + `ShellQuote` on export + wrapper `eval "$line"` regression pin. Separator injection via hostile campaign names is tested.
- **Agent safety preserved:** bare `camp switch machine:…` / `camp switch -` still resolve-and-refuse; `--print`/`--json` remote refusals shared so text cannot drift.
- **Best-effort origin:** detection failures warn on stderr and never fail the hop; empty origin remains byte-compatible with pre-feature hop lines (regression test).
- **Registration independence (D008):** hop-back builds a transient machine, never writes `machines.yaml`, and is tested with an absent fleet file.
- **ControlMaster safety:** transient ids get `-origin` suffix; empty-id collapse to a shared socket is explicitly prevented.
- **Deliberate empty `auth_method` on transient machines:** honest vs guessing; matches `authArgs` / `EnsureKeyAuth` reality.
- **Test craft:** decision table with empty-stdout-on-failure, context cancellation on detect, over-cap drop-not-truncate, registered-host case/trailing-dot match, empty local registry still reaches hop-back.
- **File split:** shared remote tail + selector extract is the right move for the line budget without behavior drift.

---

## Staff Standard

| Criterion | Assessment |
| --- | --- |
| Read real path / no parallel system | Extends `switch` / shell-connect; reuses machines, remote, encoding patterns |
| Explicit formats over heuristics | Versioned `v1` payload; reject unknown version whole-cloth |
| Context / errors / operability | ctx on detect; stderr warnings; hop-back failures point at diagnose vs probe correctly |
| Safety / agent contract | No bare hop; stdout stays one eval line |
| Tests | Behavior + negative paths + pins for invariants that matter |
| Docs | **Gap** on the new user-facing gesture |
| Scope discipline | P0 half only; pre-existing 127 PATH issue correctly left out |

**Bottom line:** merge-ready on correctness and design. Please add help/reference coverage for `camp switch -` (and optionally the parse control-byte + comment nits) before or right after merge so the gesture is operable without reading the festival design doc.

PR #505, opened 2026-07-26
feat(machine): consented 'camp machine adopt' with decline memory and hints

lancekrogers:
> **Stacked on #504.** Base is `mesh-origin-threading`, not `main`. Review #504 first; this PR's diff assumes the payload and parser it adds.

Adds the consent half of the mesh loop: the verb that turns a hop origin into a fleet entry, and the surfaces that tell you it exists.

Festival MN0001, phase 004 sequence 02. Design: `festivals/active/mesh-navigation-mutual-machine-visibility-MN0001/003_DESIGN/02_registration/outputs/registration-spec.md`.

## Why

PR-A makes a hop carry its origin. Nothing may act on that automatically: writing to another machine's fleet file because you connected to it is exactly the behavior D001 exists to prevent. So the payload sits inert until the operator runs one command and says yes.

## What changed

**`camp machine adopt`** (`cmd/camp/machine_adopt.go`). Reads `CAMP_HOP_ORIGIN`, shows the exact YAML that will be appended plus the reachability caveat, and writes only on an explicit yes.

Consent is structural rather than advisory:
- TTY-only. Non-TTY is a hard error that writes nothing.
- Default no, via the existing `confirmForm`.
- `--json` refused: there is no machine-readable representation of a consent decision.
- **No `--yes` flag.** The design originally specified one; review removed it. The `machineAddYes` precedent is a *selection* shortcut (pick the first discovered device), and adopt has exactly one candidate, so `--yes` could only mean "skip the consent". `--force` exists but only skips the reminder that you declined before.

**Decline memory** (`internal/machines/declined.go`). `declined_origins.yaml` beside `machines.yaml`. A separate file rather than a field, because `machines.yaml` is a cross-tool contract the festival app mirrors field-for-field, a declined origin is by definition not a machine, and an older camp that cannot read the new file simply prompts again. Matching normalizes case and the trailing MagicDNS dot so one machine cannot present as two identities. Adopting clears the record. A corrupt file warns and is treated as empty: a broken suppression list must never block an adopt.

**Two guards.**
- A host already registered under a *different* id reports and exits without writing. Two rows for one host would mean two ControlMaster sockets to the same machine and an ambiguous `camp machine list`.
- Adopting **this** machine is refused. Such a row would shadow the self-reference resolution from #504. The guard lives at write time because that is the only place the row can be created.

**Id collisions prompt for `camp machine add`** rather than auto-suffixing. Two ids differing by a digit are invisible in `camp machine list` and make the next hop a coin flip.

**Hint surfaces**, all sharing one predicate so suppression cannot drift: the `camp machine list` footer, the machine TUI status line at open, and the hop-back failure when the origin is unregistered here. Silent when there is no payload, the payload is malformed, the origin is already in the fleet under any id, it was declined, or the hint already fired this process. Never in `--json`.

## Tests

`just gate` green: 3779 tests.

Four of the design's five transcripts are asserted. The fifth (confirm-and-write) cannot run headless *by construction*, and `TestAdoptNonTTYWritesNothing` asserts exactly that, which is the stronger statement. Plus: entry mapping (including that `identity_file` is never guessed), collision refusal naming the escape hatch, duplicate-host detection, self-host refusal, `--json` refusal, the exact preview bytes, decline round-trip across three spellings of one host, a 5-row hint suppression matrix, and the once-per-process bound.

## Two real defects the tests caught

**The self-host guard compared against only one of this machine's two names.** `detectReachableName` returns the tailnet name when tailscale is up and the hostname otherwise. A payload built while tailscale was down carries the hostname, so a single comparison would not recognize it as this machine and the guard would let a self-entry through, re-opening the bug #504 closes. Now both are compared.

**A test fixture used the developer's own tailnet name**, so three green tests were silently exercising the self-guard branch instead of what they claimed. The fixture now names a deliberately foreign machine. Green is not correct when a fixture accidentally matches the host environment.

## PTY verification

The TUI change was driven in a real pty with pyte, not trusted to model assertions:

```
╰─────────────────────────────╯ ╰─────────────────────────────────────────────────────────────╯
✓ origin origin-box.tail37114b.ts.net is not in your fleet; run 'camp machine adopt' to add it
t test connection  ·  e edit · d remove  ·  a add · s scan tailnet  ·  ? help · q quit
```

## One thing for you to decide

The pty capture shows the hint rendering as **`✓ origin … is not in your fleet`**. `statusLine()` prefixes any non-error status with `✓`, so an advisory line reads as a success report. It is not wrong (the glyph is the status marker, not a claim), and fixing it means adding a third status kind to a mechanism shared by every message on that screen. That is a TUI convention call for you rather than something I should change unilaterally, so I left it and am flagging it. Happy to add a neutral status kind if you want one.

### DIFF
--- a/cmd/camp/machine.go
+++ b/cmd/camp/machine.go
@@ -209,9 +209,14 @@ func runMachineList(cmd *cobra.Command, _ []string) error {
 	if err := renderMachineListTable(cmd.OutOrStdout(), mf.Machines); err != nil {
 		return err
 	}
-	if hint := OriginHint(); hint != "" {
-		_, err := fmt.Fprintln(cmd.OutOrStdout(), ui.Dim(hint))
-		return err
+	// OriginHint's contract is non-TTY silence: piped/redirected list stays pure
+	// data, and the once-per-process token is not burned on non-interactive paths.
+	// --json already returns above; this covers bare table output without a TTY.
+	if ui.IsTerminal() {
+		if hint := OriginHint(); hint != "" {
+			_, err := fmt.Fprintln(cmd.OutOrStdout(), ui.Dim(hint))
+			return err
+		}
 	}
 	return nil
 }
--- a/cmd/camp/machine.go
+++ b/cmd/camp/machine.go
@@ -201,9 +201,19 @@ func runMachineList(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 	if machineListJSON {
+		// JSON consumers get data, never advice: the hint is suppressed here
+		// rather than inside OriginHint, because whether advice belongs in the
+		// output is the caller's decision.
 		return writeMachineListJSON(cmd.OutOrStdout(), mf)
 	}
-	return renderMachineListTable(cmd.OutOrStdout(), mf.Machines)
+	if err := renderMachineListTable(cmd.OutOrStdout(), mf.Machines); err != nil {
+		return err
+	}
+	if hint := OriginHint(); hint != "" {
+		_, err := fmt.Fprintln(cmd.OutOrStdout(), ui.Dim(hint))
+		return err
+	}
+	return nil
 }
 
 func writeMachineListJSON(w io.Writer, mf *machines.File) error {
--- a/cmd/camp/machine.go
+++ b/cmd/camp/machine.go
@@ -176,6 +176,9 @@ func init() {
 	machineCmd.AddCommand(machineAddCmd)
 	machineCmd.AddCommand(machineRemoveCmd)
 	machineCmd.AddCommand(machineDiagnoseCmd)
+	machineCmd.AddCommand(machineAdoptCmd)
+	machineAdoptCmd.Flags().BoolVar(&machineAdoptForce, "force", false,
+		"Skip the reminder that you declined this origin before (never skips the confirmation)")
 
 	machineListCmd.Flags().BoolVar(&machineListJSON, "json", false, "Output as JSON")
 
--- a/cmd/camp/machine_adopt.go
+++ b/cmd/camp/machine_adopt.go
@@ -39,7 +39,7 @@ var machineAdoptCmd = &cobra.Command{
 what will be written.
 
 A hop carries its origin in CAMP_HOP_ORIGIN. This reads that value and offers to
-add the origin to ~/.obey/machines.yaml so hops and completion work in the other
+add the origin to your machines file so hops and completion work in the other
 direction too. The hop itself never registers anything: adopting is an explicit,
 interactive act, and there is no flag that skips the confirmation.
 
@@ -91,6 +91,15 @@ func runMachineAdopt(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
+	// The TTY refusal comes before any human-oriented output. A non-interactive
+	// run that has already printed a decline reminder and then errors leaves
+	// half a consent surface on stdout, which is exactly what the rest of this
+	// command avoids.
+	if !adoptIsTerminal() {
+		return camperrors.New("camp machine adopt requires an interactive terminal: registering a " +
+			"machine is an explicit consent step and cannot be automated")
+	}
+
 	declined, declineErr := machines.LoadDeclined()
 	if declineErr != nil {
 		_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "camp: warning: %v (treating as no declines)\n", declineErr)
@@ -98,11 +107,6 @@ func runMachineAdopt(cmd *cobra.Command, _ []string) error {
 	if d, was := declined.IsDeclined(origin.Host); was && !machineAdoptForce {
 		_, _ = fmt.Fprintf(cmd.OutOrStdout(), "You declined this origin on %s; adopting now will register it.\n\n",
 			d.DeclinedAt.Format("2006-01-02"))
-	}
-
-	if !adoptIsTerminal() {
-		return camperrors.New("camp machine adopt requires an interactive terminal: registering a " +
-			"machine is an explicit consent step and cannot be automated")
 	}
 
 	if _, err := fmt.Fprint(cmd.OutOrStdout(), adoptPreview(origin, entry)); err != nil {
@@ -200,7 +204,7 @@ func adoptPreview(origin HopOrigin, entry machines.Machine) string {
 	if origin.Campaign != "" {
 		fmt.Fprintf(&b, "  campaign  %s\n", origin.Campaign)
 	}
-	b.WriteString("\nThis entry will be appended to ~/.obey/machines.yaml:\n\n")
+	b.WriteString("\nThis entry will be appended to " + machines.MachinesPath() + ":\n\n")
 	// Marshal the row the same way Save does so quoting/escaping match disk.
 	yml, err := yaml.Marshal([]machines.Machine{entry})
 	if err != nil {
--- a/cmd/camp/machine_adopt.go
+++ b/cmd/camp/machine_adopt.go
@@ -22,6 +22,16 @@ var machineAdoptForce bool
 var adoptIsTerminal = ui.IsTerminal
 var adoptConfirm = confirmForm
 
+// adoptDetectTailnet / adoptDetectBare are the dual reachable-name probes used
+// by isSelfOrigin. Production uses tailscale Self + os.Hostname; tests inject
+// pure functions so the dual-name matrix is covered without a live tailnet.
+var adoptDetectTailnet = func(ctx context.Context) (string, error) {
+	return detectReachableName(ctx, runTailscaleStatusForSelf)
+}
+var adoptDetectBare = func(ctx context.Context) (string, error) {
+	return detectReachableName(ctx, nil)
+}
+
 var machineAdoptCmd = &cobra.Command{
 	Use:   "adopt",
 	Short: "Register the machine this session was hopped from",
@@ -236,14 +246,14 @@ func isSelfOrigin(ctx context.Context, origin HopOrigin) (bool, error) {
 	if want == "" {
 		return false, nil
 	}
-	tailnet, err := detectReachableName(ctx, runTailscaleStatusForSelf)
+	tailnet, err := adoptDetectTailnet(ctx)
 	if err != nil {
 		return false, err
 	}
 	if machines.NormalizeHost(tailnet) == want {
 		return true, nil
 	}
-	bare, err := detectReachableName(ctx, nil)
+	bare, err := adoptDetectBare(ctx)
 	if err != nil {
 		return false, err
 	}
--- a/cmd/camp/machine_adopt.go
+++ b/cmd/camp/machine_adopt.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
 
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 	"github.com/Obedience-Corp/camp/internal/machines"
@@ -15,6 +16,11 @@ import (
 )
 
 var machineAdoptForce bool
+
+// adoptIsTerminal and adoptConfirm are production seams for the consent path.
+// Tests swap them so cancel vs explicit-No can be asserted without a real TTY.
+var adoptIsTerminal = ui.IsTerminal
+var adoptConfirm = confirmForm
 
 var machineAdoptCmd = &cobra.Command{
 	Use:   "adopt",
@@ -30,8 +36,9 @@ interactive act, and there is no flag that skips the confirmation.
 Adopting records how to REACH a machine. It does not make that machine reachable:
 that depends on its own sshd or tailnet policy. Verify with 'camp machine diagnose'.
 
-Declining is remembered, so hints stay quiet until you ask again. Re-running this
-command always works; --force only skips the reminder that you declined before.`,
+Answering No is remembered, so hints stay quiet until you ask again. Esc/cancel
+aborts without writing decline memory. Re-running this command always works;
+--force only skips the reminder that you declined before.`,
 	Args: cobra.NoArgs,
 	RunE: runMachineAdopt,
 }
@@ -48,7 +55,13 @@ func runMachineAdopt(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-	if self, err := isSelfOrigin(ctx, origin); err == nil && self {
+	// Fail closed on detection errors: a skipped self-guard could write a
+	// machines.yaml row pointing at this host and break self-reference resolve.
+	self, err := isSelfOrigin(ctx, origin)
+	if err != nil {
+		return camperrors.Wrap(err, "camp machine adopt: cannot determine whether this origin is this machine")
+	}
+	if self {
 		return camperrors.New("camp machine adopt: this origin is this machine (" + origin.Host +
 			"); nothing to adopt")
 	}
@@ -77,7 +90,7 @@ func runMachineAdopt(cmd *cobra.Command, _ []string) error {
 			d.DeclinedAt.Format("2006-01-02"))
 	}
 
-	if !ui.IsTerminal() {
+	if !adoptIsTerminal() {
 		return camperrors.New("camp machine adopt requires an interactive terminal: registering a " +
 			"machine is an explicit consent step and cannot be automated")
 	}
@@ -85,8 +98,13 @@ func runMachineAdopt(cmd *cobra.Command, _ []string) error {
 	if _, err := fmt.Fprint(cmd.OutOrStdout(), adoptPreview(origin, entry)); err != nil {
 		return err
 	}
-	ok, err := confirmForm(ctx, fmt.Sprintf("Add machine %q?", entry.ID))
+	ok, canceled, err := adoptConfirm(ctx, fmt.Sprintf("Add machine %q?", entry.ID))
 	if err != nil {
+		return err
+	}
+	if canceled {
+		// Esc/abort is not a deliberate decline — leave decline memory alone.
+		_, err := fmt.Fprintln(cmd.OutOrStdout(), "Not adopted.")
 		return err
 	}
 	if !ok {
@@ -161,8 +179,9 @@ func adoptEntry(mf *machines.File, origin HopOrigin) (machines.Machine, error) {
 	}, nil
 }
 
-// adoptPreview renders exactly the YAML that will be appended, plus the
-// reachability caveat. The operator approves the bytes, not a summary of them.
+// adoptPreview renders the YAML that will be appended (via the same
+// yaml.Marshal path File.Save uses), plus the reachability caveat. The
+// operator approves the bytes, not a hand-formatted summary of them.
 func adoptPreview(origin HopOrigin, entry machines.Machine) string {
 	var b strings.Builder
 	b.WriteString("\nOrigin from this session:\n\n")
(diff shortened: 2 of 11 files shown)


cmd/camp/machine_tui.go
```diff
@@ -203,6 +203,14 @@ func newMachineTUIModel(ctx context.Context, mf *machines.File) *machineTUIModel
 		spin:           newMachineSpinner(),
 	}
 	m.rebuildRows("")
+	// The fleet screen is where an operator looks at machines, so it is where
+	// an unregistered origin is worth mentioning. It names the command rather
+	// than offering a key: the adopt confirmation is a huh form that takes over
+	// the terminal, and this model already owns it. One preview, one consent
+	// surface, nothing to keep in sync.
+	if hint := OriginHint(); hint != "" {
+		m.setStatus(hint)
```
obey-agent: Agreeing with the PR note: \`setStatus\` + \`statusLine\` always prefixes non-error status with ✓, so an *advisory* ("origin … is not in your fleet") reads as success. Not a functional bug, but it is a real UX footgun on the first screen an operator opens after a hop.

If you take the product call to add a neutral status kind, this is the call site that wants it — e.g. \`setStatus(msg)\` / \`setAdvice(msg)\` / \`setError(err)\` with no glyph or a distinct marker for advice. Leaving as-is is fine if intentional; just do not lose the decision.

obey-agent (commented):
## Verdict: Approve

## Overview
Consented adopt is implemented the way a fleet registration verb should be: interactive confirmation with an exact YAML preview, no automation flag that skips consent, decline memory keyed by normalized host, and hints that share one predicate so suppression cannot drift across list / TUI / hop-back. Self-host guard comparing *both* reachable names is a real defect catch; the foreign-host fixture fix is the kind of test hygiene that prevents green-but-wrong suites.

## Key Findings
- **Product call (flagged by author, reinforced in review):** advisory origin hint renders with `✓` via the shared statusLine convention. Not a functional bug; decide whether to add a neutral status kind. See inline.
- Non-TTY and `--json` hard refusals correctly make adopt non-automatable.
- Decline file beside machines.yaml (not inside it) is the right version-skew story.

## What's Done Well
- No auto-suffix on id collision; refuses and names the explicit escape hatch.
- `identity_file` never guessed; auth defaults to ssh-agent and the preview says so.
- Host-normalized decline matching (case + trailing dot).
- PTY verification of the TUI path, not model-only assertions.
- Hint once-per-process bound is honest given "no session daemon."

## Staff Standard
Yes. Consent, diagnostics, and "camp diagnoses, never manages" are held consistently. Ship.

cmd/camp/machine_adopt.go
```diff
@@ -0,0 +1,227 @@
...
+		Label:      "adopted from hop " + time.Now().Format("2006-01-02"),
+		Host:       origin.Host,
+		AuthMethod: auth,
+		SSHUser:    origin.User,
+	}, nil
+}
+
+// adoptPreview renders exactly the YAML that will be appended, plus the
+// reachability caveat. The operator approves the bytes, not a summary of them.
+func adoptPreview(origin HopOrigin, entry machines.Machine) string {
```
obey-agent, line 199: **[bug]** `adoptPreview` claims the operator “approves the bytes” and hand-formats the machines.yaml fragment with bare `%s` interpolation for `host`, `ssh_user`, `label`, and `id`. Persistence goes through `yaml.Marshal` in `machines.File.Save`, which will quote/escape values. `#504`’s `hopOriginDecode` accepts percent-decoded bytes with no post-decode rejection of whitespace or control characters, so a spoofed `CAMP_HOP_ORIGIN` (e.g. `host=good.example.com%0A%20%20%20%20auth_method:%20ssh-password` or `host=*alias`) can make the consent surface show extra lines or YAML-looking structure that is not what will be written. The on-disk file remains a single marshaled row, but the structural consent guarantee of this PR is violated: the operator is not shown the exact bytes that will be appended.

**Suggestion:** Build the preview from `yaml.Marshal` of the prospective `machines.Machine` (or a one-element `File`), and/or reject `host`/`user` that contain whitespace, controls, or other non-hostname/user characters at `adoptOrigin` / `ParseHopOrigin` time before any confirm UI. Prefer both: safe parse + marshaled preview so the “exact YAML” claim stays true under hostile env values.

obey-agent (commented):
## Summary

Re-review of `Obedience-Corp/camp#505` at `6cdab3d9586fcaa978aa97728e3c64603b84bc09` (stacked on #504). Prior head `29fc8e7d` flagged the advisory-as-success TUI glyph; this head adds `statusKind` / `statusAdvice` and renders advice with a muted `·` rather than `✓`, with tests covering that path. Consent structure is otherwise strong: TTY-only adopt, no `--yes`, `--json` refused, default-no confirm, decline memory in a sibling file, host-normalized suppressions shared by list/TUI/hop-back, and self-host / duplicate-host / id-collision guards. Residual risks center on (1) the hand-rolled “exact YAML” preview diverging from marshaled writes under spoofed `CAMP_HOP_ORIGIN` control characters, (2) list still emitting the hint on non-TTY despite the documented caller contract, (3) Esc/cancel permanently recording a decline, and (4) the self-origin guard failing open on detection errors and when MagicDNS self cannot be resolved while tailscale is down.

## Issue counts by severity

- bugs: 1
- suggestions: 3
- nits: 2

cmd/camp/machine.go
```diff
@@ -198,9 +201,19 @@ func runMachineList(cmd *cobra.Command, _ []string) error {
...
 		return writeMachineListJSON(cmd.OutOrStdout(), mf)
 	}
-	return renderMachineListTable(cmd.OutOrStdout(), mf.Machines)
+	if err := renderMachineListTable(cmd.OutOrStdout(), mf.Machines); err != nil {
+		return err
+	}
+	if hint := OriginHint(); hint != "" {
+		_, err := fmt.Fprintln(cmd.OutOrStdout(), ui.Dim(hint))
+		return err
+	}
```
obey-agent, line 220: **[suggestion]** `OriginHint`’s contract (and the PR description) says callers suppress the hint for non-TTY output. `runMachineList` suppresses only `--json`; a redirected/piped `camp machine list` still appends the dim advice line after the table. That can break line-oriented consumers and is inconsistent with the documented non-TTY silence rule. The once-per-process token is also burned on that non-interactive path.

**Suggestion:** Gate the list footer with the same terminal check used elsewhere (`ui.IsTerminal()`, or at least stdout-is-TTY) so non-interactive list stays pure data.

cmd/camp/machine_adopt.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	if _, err := fmt.Fprint(cmd.OutOrStdout(), adoptPreview(origin, entry)); err != nil {
+		return err
+	}
+	ok, err := confirmForm(ctx, fmt.Sprintf("Add machine %q?", entry.ID))
+	if err != nil {
+		return err
+	}
+	if !ok {
```
obey-agent, line 124: **[suggestion]** Any non-confirm path records a durable decline: `confirmForm` maps huh cancel/Esc to `(false, nil)`, and `!ok` always calls `declined.Decline` + `Save`. An operator who opens adopt, reads the preview, and dismisses with Esc gets the same persistent hint suppression as an explicit “No”. Re-running adopt still works, but the design text frames decline memory as remembering a deliberate refusal; accidental cancel is a weaker signal and surprises when list/TUI/hop-back stay quiet afterward.

**Suggestion:** Persist decline only on an explicit “No” from the confirm widget; treat cancel as abort without writing `declined_origins.yaml` (and keep the “Not adopted” message). If cancel-as-decline is intentional, say so in the command Long text and the post-decline line.

cmd/camp/machine_adopt.go
```diff
@@ -0,0 +1,227 @@
...
+	if jsonOut {
+		return camperrors.New("camp machine adopt does not support --json: registering a machine is " +
+			"an explicit consent step, not a query")
+	}
+
+	origin, err := adoptOrigin()
+	if err != nil {
+		return err
+	}
+	if self, err := isSelfOrigin(ctx, origin); err == nil && self {
```
obey-agent: **[suggestion]** The self-origin guard is fail-open: `if self, err := isSelfOrigin(...); err == nil && self`. Any detection error skips the guard and proceeds to consent/write. Separately, `isSelfOrigin` only compares the best-effort `detectReachableName` result (tailnet-or-hostname) and the bare hostname; when tailscale is down, a payload carrying this machine’s MagicDNS name does not match `os.Hostname()`, so a self row can still be created. That re-opens the self-reference shadowing problem this guard exists to prevent (the PR already fixed the dual-name case for the tailscale-up direction).

**Suggestion:** Fail closed on `isSelfOrigin` error (or at least refuse adopt with a clear “could not verify this is not self” message). Consider also comparing against the last-known tailnet name if available, and/or refusing hosts that equal any configured local identity aliases you already trust. At minimum, do not ignore `err`.

cmd/camp/machine_adopt.go
```diff
@@ -0,0 +1,227 @@
...
+	declined, declineErr := machines.LoadDeclined()
+	if declineErr != nil {
+		_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "camp: warning: %v (treating as no declines)\n", declineErr)
+	}
+	if d, was := declined.IsDeclined(origin.Host); was && !machineAdoptForce {
+		_, _ = fmt.Fprintf(cmd.OutOrStdout(), "You declined this origin on %s; adopting now will register it.\n\n",
+			d.DeclinedAt.Format("2006-01-02"))
+	}
+
+	if !ui.IsTerminal() {
```
obey-agent: **[nit]** The “You declined this origin on …” reminder is printed before the interactive-terminal hard check. On a non-TTY session with a prior decline, adopt writes that reminder to stdout and then errors that consent cannot be automated—partial human-oriented output on the non-interactive path the rest of the command carefully avoids.

**Suggestion:** Move the TTY requirement above any decline reminder (and above preview), so non-TTY either succeeds for no-op cases (already registered) or fails with a single error and empty stdout.

cmd/camp/machine_adopt.go
```diff
@@ -0,0 +1,227 @@
...
+func adoptPreview(origin HopOrigin, entry machines.Machine) string {
+	var b strings.Builder
+	b.WriteString("\nOrigin from this session:\n\n")
+	fmt.Fprintf(&b, "  host      %s\n", origin.Host)
+	fmt.Fprintf(&b, "  user      %s\n", origin.User)
+	if origin.Campaign != "" {
+		fmt.Fprintf(&b, "  campaign  %s\n", origin.Campaign)
+	}
+	b.WriteString("\nThis entry will be appended to ~/.obey/machines.yaml:\n\n")
+	fmt.Fprintf(&b, "    - id: %s\n", entry.ID)
```
obey-agent: **[nit]** Preview and help text hardcode `~/.obey/machines.yaml` even though `machines.DeclinedPath` / `MachinesPath` honor `CAMP_MACHINES_PATH` and `XDG_CONFIG_HOME`. Tests isolate via `CAMP_MACHINES_PATH`; an operator using the override still sees the default path in the consent surface.

**Suggestion:** Render `machines.MachinesPath()` (possibly with `~` shortening) in the preview and Long text, or say “your machines file” without a fixed path.

cmd/camp/machine_adopt.go
obey-agent: **[medium]** Self-origin guard is fail-open.

```go
if self, err := isSelfOrigin(ctx, origin); err == nil && self {
```

Any detection error skips the guard and proceeds to preview/consent/write. The guard exists specifically to stop a machines.yaml row that would shadow self-reference resolution from #504. A safety check that cannot decide should refuse the write (or at least hard-error), not continue.

Secondary gap: when Tailscale is currently down, both `detectReachableName` calls resolve to the bare hostname, so a self payload that still carries the MagicDNS name is not recognized either. The dual-name fix only fully works while Tailscale is up on *this* machine at adopt time.

Prefer: on `isSelfOrigin` error → return that error; and inject the reachable-name funcs so the dual-name matrix can be unit-tested without a live tailscale binary.

cmd/camp/machine_adopt.go
obey-agent, line 124: **[suggestion]** Any non-confirm path records a durable decline.

`confirmForm` maps huh cancel/Esc to `(false, nil)`, and `!ok` always calls `declined.Decline` + `Save`. An operator who opens adopt, reads the preview, and hits Esc did not affirmatively decline — but the next `camp machine list` / TUI open goes quiet, and the next adopt shows "You declined this origin on …".

Options that keep consent structural without over-remembering:
- Distinguish cancel from explicit No (theme.IsCancelled already exists in confirmForm’s error path; surface it), or
- Only persist decline on an explicit No answer.

Either way, the success-path message ("Run adopt again if you change your mind") understates that decline memory will suppress hints until then.

cmd/camp/machine_adopt.go
```diff
@@ -0,0 +1,227 @@
...
+// isSelfOrigin guards the one entry that would break self-reference resolution:
+// a machines.yaml row pointing at the adopting machine. The guard belongs here,
+// at write time, because this is the only place such a row can be created.
+//
+// It compares against BOTH names this machine can legitimately be called: the
+// tailnet name and the plain hostname. detectReachableName returns whichever is
+// available, so a payload built when tailscale was down would carry the hostname
+// while this machine now reports its tailnet name, and a single comparison would
+// miss it. Both are this machine, so both must count.
+func isSelfOrigin(ctx context.Context, origin HopOrigin) (bool, error) {
```
obey-agent, line 248: **[suggestion / test gap]** The dual reachable-name self-guard cannot be regression-tested as written.

`isSelfOrigin` hardcodes `runTailscaleStatusForSelf` and `nil` rather than accepting injected `reachableNameFunc`s the way `buildHopOrigin` / `detectReachableName` already do. `TestAdoptRefusesSelfHost` only exercises the live bare-hostname path:

```go
host, err := detectReachableName(context.Background(), nil)
```

The PR’s own post-mortem says a single-name comparison was a real defect (hostname payload vs tailnet detect). That exact matrix (ts-name vs bare-name cross product) has no table test. Inject the two detectors (or a small selfNames helper) and assert:

| origin host | ts detect | bare detect | want |
|---|---|---|---|
| bare | ts | bare | self |
| ts | ts | bare | self |
| foreign | ts | bare | not self |
| ts | err | bare | error / refuse |

cmd/camp/machine.go
```diff
@@ -198,9 +201,19 @@ func runMachineList(cmd *cobra.Command, _ []string) error {
...
+		// JSON consumers get data, never advice: the hint is suppressed here
+		// rather than inside OriginHint, because whether advice belongs in the
+		// output is the caller's decision.
 		return writeMachineListJSON(cmd.OutOrStdout(), mf)
 	}
-	return renderMachineListTable(cmd.OutOrStdout(), mf.Machines)
+	if err := renderMachineListTable(cmd.OutOrStdout(), mf.Machines); err != nil {
+		return err
+	}
+	if hint := OriginHint(); hint != "" {
```
obey-agent: **[suggestion]** `OriginHint`’s contract (and this PR’s description) says callers suppress the hint for non-TTY output. `runMachineList` only suppresses `--json`:

```go
if hint := OriginHint(); hint != "" {
    _, err := fmt.Fprintln(cmd.OutOrStdout(), ui.Dim(hint))
```

A redirected/piped `camp machine list` still appends the adopt line. That is fine if intentional, but then the doc comment on `OriginHint` overclaims. Either gate with `ui.IsTerminal()` here (matching TUI / adopt surfaces), or tighten the comment to "callers suppress for --json; TTY is optional."

cmd/camp/machine_adopt.go
obey-agent, line 199: **[nit]** `adoptPreview` claims the operator “approves the bytes” and hand-formats the machines.yaml fragment with bare `%s` for id/label/host/user/auth. Persistence goes through `yaml.Marshal` in `machines.encode`, which will quote/escape differently for any value that is not a plain unquoted scalar.

For camp-built hop payloads this is low risk (DNS names + percent-encoding). For a crafted `CAMP_HOP_ORIGIN` (or a user with special characters), the preview can diverge from the bytes on disk while the UX still says “exactly the YAML that will be appended.” Prefer marshaling a one-element `[]machines.Machine` (or a tiny helper) for the fragment so preview ≡ write, and keep the surrounding prose hand-written.

obey-agent (commented):
## Verdict: Approve

## Overview
Stacked on #504. This is the consent half of mesh mutual visibility: `camp machine adopt` turns an inert `CAMP_HOP_ORIGIN` into a fleet row only after an explicit interactive yes, with decline memory and three hint surfaces that share one predicate. The design holds the hard line from D001 — hop never auto-registers — and makes that structural (TTY-only, default-no confirm, no `--yes`, `--json` refused) rather than advisory.

Head `6cdab3d9` also lands the TUI fix for the product call from earlier review: `statusKind` / `statusAdvice` renders the origin hint with muted `·` instead of success `✓`, with model + glyph assertions.

## Key Findings

1. **[medium] Self-origin guard is fail-open** (`machine_adopt.go:51`) — detection error skips the guard and continues to write. A check that exists to protect self-reference resolution should refuse (or hard-error) when it cannot decide. Secondary: dual-name comparison only fully works while Tailscale is up at adopt time.

2. **[suggestion / test gap] Dual-name self-guard is untestable as written** (`isSelfOrigin` hardcodes production detectors). `TestAdoptRefusesSelfHost` only covers live bare-hostname. The exact matrix the PR’s post-mortem calls a real defect has no table test. Inject `reachableNameFunc`s.

3. **[suggestion] Esc/cancel records a durable decline** — `confirmForm` maps cancel to `(false, nil)` and `!ok` always `Decline`+`Save`. Accidental Esc suppresses hints and surfaces a “you declined” reminder on re-run. Prefer distinguishing cancel from explicit No.

4. **[suggestion] OriginHint non-TTY contract drift** — docs say callers suppress for non-TTY; `runMachineList` only suppresses `--json`. Align gate or comment.

5. **[nit] “Exact YAML” preview is hand-formatted** — can diverge from `yaml.Marshal` for non-plain scalars / crafted payloads. Prefer marshaling the entry fragment for preview ≡ write.

None of these are merge-blockers for the consent architecture; (1) and (2) are the ones worth landing before or immediately after merge if self-shadowing is treated as high-cost.

## What's Done Well
- Consent is structural: TTY hard-error writes nothing (asserted), no automation flag, preview before prompt, default-no.
- Decline file beside `machines.yaml` (not a field) — correct cross-tool / version-skew story; corrupt list treated as empty so suppression never blocks adopt.
- Host-normalized matching (case + trailing MagicDNS dot) shared by decline, `machineByHost`, and `OriginHint`.
- Id collision refuses and names `camp machine add` escape hatch (no auto-suffix / U8).
- `identity_file` never guessed; `auth_method` defaults to ssh-agent and the preview says so.
- Shared `OriginHint` predicate for list footer / TUI open / hop-back failure — suppression cannot drift.
- Tests assert fleet file absence on failure paths (not just error strings); foreign-host fixture avoids green-but-wrong self-guard branch; PTY-verified TUI path; advisory glyph covered.
- Dual reachable-name self-guard (when TS is up) is the right shape for the bug the PR caught in development.

## Staff Standard
**Yes.** This is staff-level product engineering: clear threat model (D001), structural consent, honest bounds (once-per-process hint, no session daemon), tests that catch the defects the author claims, and a deliberate TUI status kind rather than overloading success. Residual items are fail-open on a safety guard and testability of that guard — tighten those and this is exemplary.

Stacked note: do not merge ahead of #504; diff assumes its payload/parser.

PR #506, opened 2026-07-26
feat(switch): warn about camp version skew at hop time, without a probe

lancekrogers:
> **Stacked on #505**, which is stacked on #504. Base is `mesh-adopt-origin`. Review those first.

Warns about camp version skew at the moment it matters (hop time) without making hops slower.

Festival MN0001, phase 004 sequence 03.

## Why

Version skew is a permanent condition in a personal fleet, and the moment it bites is the moment you hop: a feature that exists here may not exist there. `camp machine diagnose` already probes remote versions and already knows how to judge skew (`campVersionSkew`, `probeRemoteCampVersion`, both from #492). What was missing is that the knowledge died with the command.

## What changed

Diagnose now caches each probe result per machine (`cmd/camp/version_cache.go`), beside the completion cache, and the hop path reads it.

**The hop never probes.** Adding an ssh round-trip to deliver an advisory would make every hop slower in order to occasionally say something optional. `hopSkewWarning` reads one file; the only writer is the diagnose loop, which was already paying for the probe.

**Freshness is 12 hours**, matching the shape of the underlying fact: skew changes on install cadence, not on the second. Absent, corrupt, and stale all read as *no opinion*, so the failure mode is silence rather than a wrong claim. A failed probe is never cached, so an empty version can never later be mistaken for a real answer.

**The warning goes to stderr, dimmed:**

```
camp: camp on archdtop is v0.0.1-ancient, this machine is dev; features added since may not work there (run 'camp machine diagnose archdtop' to re-check)
```

## Tests

`just gate` green: 3784.

Six-row cache matrix (miss, match, version skew, same-version-different-commit, stale, empty-probe), each seeded with an explicit age so freshness is tested rather than raced against the clock. Plus: failed probes are not written, corrupt JSON reads as a miss, and the cache lives beside the completion cache so one `rm -rf ~/.obey/cache` clears every derived artifact about a machine.

The load-bearing test is `TestSkewWarningGoesToStderrAndKeepsStdoutToOneLine`. stdout is eval'd by the shell wrapper, so a warning printed there would be **executed as a shell command**. It asserts stdout is exactly one line, does not contain the warning, and that stderr does.

## Deliberate tradeoff

The warning only fires if `camp machine diagnose` has run for that machine in the last 12 hours. A user who never runs diagnose never sees it. That is the right side of the tradeoff: diagnose is the surface that already pays for truth, so it is the surface that earns the right to warm the cache. The alternative makes every hop slower.

### DIFF
--- a/cmd/camp/machine.go
+++ b/cmd/camp/machine.go
@@ -410,6 +410,9 @@ func runMachineDiagnose(cmd *cobra.Command, args []string) error {
 		m := &targets[i]
 		d := remote.CheckControlMaster(ctx, m)
 		remoteVersion, remoteCommit := probeRemoteCampVersion(ctx, m)
+		// Diagnose is the only surface that pays for a live probe, so it is the
+		// only one that can warm the cache the hop path reads.
+		writeMachineVersionCache(m.ID, remoteVersion, remoteCommit)
 		row := machineDiagnoseRow{
 			ID:           m.ID,
 			Host:         m.Host,
--- a/cmd/camp/version_cache.go
+++ b/cmd/camp/version_cache.go
@@ -1,0 +1,85 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/Obedience-Corp/camp/internal/fsutil"
+	"github.com/Obedience-Corp/camp/internal/version"
+)
+
+// versionCacheTTL bounds how long a probed remote version is trusted for the
+// hop-time warning. Skew changes on install cadence, not on the second, so an
+// hours-scale bound is right: short enough that an upgrade stops the warning the
+// same day, long enough that the warning survives between diagnose runs. A stale
+// entry is simply ignored, so the failure mode is silence, never a wrong claim.
+const versionCacheTTL = 12 * time.Hour
+
+// versionCacheEntry is the on-disk per-machine version probe result. Derived,
+// gitignored, and disposable: deleting it costs one diagnose.
+type versionCacheEntry struct {
+	Version  string `json:"version"`
+	Commit   string `json:"commit"`
+	ProbedAt int64  `json:"probed_at"` // unix nanoseconds
+}
+
+func (e versionCacheEntry) fresh(now time.Time) bool {
+	return now.Sub(time.Unix(0, e.ProbedAt)) <= versionCacheTTL
+}
+
+// versionCachePath keeps the probe beside the completion cache, under the same
+// per-machine directory, so one `rm -rf ~/.obey/cache` clears every derived
+// artifact camp keeps about a machine.
+func versionCachePath(id string) string {
+	return filepath.Join(machineCacheDir(), id+".version.json")
+}
+
+// writeMachineVersionCache records a probe result. Best-effort: a failure just
+// means the next hop does not warn, which is the same as today's behavior.
+func writeMachineVersionCache(id, versionStr, commit string) {
+	if id == "" || versionStr == "" {
+		return
+	}
+	dir := machineCacheDir()
+	if os.MkdirAll(dir, 0o700) != nil {
+		return
+	}
+	data, err := json.Marshal(versionCacheEntry{Version: versionStr, Commit: commit, ProbedAt: time.Now().UnixNano()})
+	if err != nil {
+		return
+	}
+	_ = fsutil.WriteFileAtomically(versionCachePath(id), data, 0o600)
+}
+
+// readMachineVersionCache returns a fresh cached probe, or false. Absent,
+// corrupt, and stale are all the same answer: no opinion.
+func readMachineVersionCache(id string) (versionCacheEntry, bool) {
+	data, err := os.ReadFile(versionCachePath(id))
+	if err != nil {
+		return versionCacheEntry{}, false
+	}
+	var e versionCacheEntry
+	if json.Unmarshal(data, &e) != nil || !e.fresh(time.Now()) {
+		return versionCacheEntry{}, false
+	}
+	return e, true
+}
+
+// hopSkewWarning returns the one-line warning for a hop to id, or "" when there
+// is nothing to say. It NEVER probes: the hop path must not grow an ssh
+// round-trip to deliver an advisory, so it reports only what a previous
+// `camp machine diagnose` already learned.
+func hopSkewWarning(id string) string {
+	entry, ok := readMachineVersionCache(id)
+	if !ok {
+		return ""
+	}
+	local := version.Get()
+	if !campVersionSkew(local, entry.Version, entry.Commit) {
+		return ""
+	}
+	return "camp on " + id + " is " + entry.Version + ", this machine is " + local.Version +
+		"; features added since may not work there (run 'camp machine diagnose " + id + "' to re-check)"
+}
--- a/cmd/camp/version_cache.go
+++ b/cmd/camp/version_cache.go
@@ -13,8 +13,13 @@ import (
 // versionCacheTTL bounds how long a probed remote version is trusted for the
 // hop-time warning. Skew changes on install cadence, not on the second, so an
 // hours-scale bound is right: short enough that an upgrade stops the warning the
-// same day, long enough that the warning survives between diagnose runs. A stale
-// entry is simply ignored, so the failure mode is silence, never a wrong claim.
+// same day, long enough that the warning survives between diagnose runs.
+//
+// A TTL-expired, absent, or corrupt entry is ignored, so those failure modes are
+// silence rather than a wrong claim. Within the window the claim can still go
+// stale the other way: upgrade the remote to match and the warning persists
+// until the next diagnose or expiry. That is the accepted cost of never probing
+// on the hop path, and the message points at diagnose for exactly this reason.
 const versionCacheTTL = 12 * time.Hour
 
 // versionCacheEntry is the on-disk per-machine version probe result. Derived,
--- a/cmd/camp/version_cache.go
+++ b/cmd/camp/version_cache.go
@@ -38,6 +38,8 @@ func versionCachePath(id string) string {
 
 // writeMachineVersionCache records a probe result. Best-effort: a failure just
 // means the next hop does not warn, which is the same as today's behavior.
+// A failed probe (empty version) is a no-op on purpose — it leaves any prior
+// still-fresh entry in place rather than wiping a known-good answer with silence.
 func writeMachineVersionCache(id, versionStr, commit string) {
 	if id == "" || versionStr == "" {
 		return
@@ -54,14 +56,17 @@ func writeMachineVersionCache(id, versionStr, commit string) {
 }
 
 // readMachineVersionCache returns a fresh cached probe, or false. Absent,
-// corrupt, and stale are all the same answer: no opinion.
+// corrupt, empty version, and stale are all the same answer: no opinion.
+// Empty version is defense in depth: writeMachineVersionCache already refuses
+// to record failed probes, but a hand-written or half-written cache file must
+// not read as a real answer.
 func readMachineVersionCache(id string) (versionCacheEntry, bool) {
 	data, err := os.ReadFile(versionCachePath(id))
 	if err != nil {
 		return versionCacheEntry{}, false
 	}
 	var e versionCacheEntry
-	if json.Unmarshal(data, &e) != nil || !e.fresh(time.Now()) {
+	if json.Unmarshal(data, &e) != nil || e.Version == "" || !e.fresh(time.Now()) {
 		return versionCacheEntry{}, false
 	}
 	return e, true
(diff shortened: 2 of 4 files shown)


cmd/camp/version_cache.go
```diff
@@ -0,0 +1,85 @@
...
+func hopSkewWarning(id string) string {
+	entry, ok := readMachineVersionCache(id)
+	if !ok {
+		return ""
+	}
+	local := version.Get()
+	if !campVersionSkew(local, entry.Version, entry.Commit) {
+		return ""
+	}
+	return "camp on " + id + " is " + entry.Version + ", this machine is " + local.Version +
```
obey-agent: **Wording nit (optional):** "features added since may not work there" reads as if *this* machine is always ahead. When the remote is newer (local lagging), the warning still fires but the sentence is directionally wrong.

Something like "versions differ; features may not match (run 'camp machine diagnose …' to re-check)" stays honest for both directions without more logic.

obey-agent (commented):
## Verdict: Approve

## Overview
Version skew at hop time without adding a probe is the right tradeoff: diagnose is the only surface that pays for a live version check, the hop path reads a disposable 12h cache, and failures degrade to silence rather than wrong claims. Stderr-only warning preserves the shell-wrapper invariant (stdout is exactly one eval line).

## Key Findings
- **Optional wording:** "features added since may not work there" assumes local is ahead. Neutral "versions differ" wording would cover both directions. See inline.
- Cache miss / stale / corrupt / empty version all correctly produce no warning.

## What's Done Well
- Failure mode is silence, never a false claim.
- Version cache co-located with machine completion cache (one `rm -rf ~/.obey/cache` clears derived state).
- Integration test that warning stays on stderr and stdout stays one line.
- Matrix covers same version/different commit (dev builds).

## Staff Standard
Yes. Small, correct, and aligned with the "no new hop-path round-trip" constraint.

cmd/camp/version_cache.go
```diff
@@ -0,0 +1,88 @@
...
+		return ""
+	}
+	local := version.Get()
+	if !campVersionSkew(local, entry.Version, entry.Commit) {
+		return ""
+	}
+	// Says only that they differ. The skew check does not order versions, and
+	// the remote is as likely to be ahead as behind, so naming a direction
+	// would be wrong half the time.
+	return "camp on " + id + " is " + entry.Version + ", this machine is " + local.Version +
```
obey-agent: **[suggestion]** When skew is commit-only (same `Version`, different `Commit` — the case the matrix row `same version, different commit warns` exists for), the message only prints version strings: `camp on archdtop is dev, this machine is dev; features may not match…`. That looks like a no-op comparison even though the warning is correct. Diagnose already treats this as the reason to append commits (`campVersionDisplay` / `campLocalVersionDisplay` comment: dev builds all report version `dev`).

**Suggestion:** Format both sides the same way diagnose does when a commit is present, e.g. `dev (0000000000)` vs `dev (abc1234)`, or reuse `campLocalVersionDisplay` and a small remote helper. Keep the neutral trailing clause.

cmd/camp/version_cache.go
```diff
@@ -0,0 +1,88 @@
...
+// versionCachePath keeps the probe beside the completion cache, under the same
+// per-machine directory, so one `rm -rf ~/.obey/cache` clears every derived
+// artifact camp keeps about a machine.
+func versionCachePath(id string) string {
+	return filepath.Join(machineCacheDir(), id+".version.json")
+}
+
+// writeMachineVersionCache records a probe result. Best-effort: a failure just
+// means the next hop does not warn, which is the same as today's behavior.
+func writeMachineVersionCache(id, versionStr, commit string) {
```
obey-agent, line 48: **[suggestion]** `writeMachineVersionCache` no-ops on empty `versionStr` (failed probe) and never deletes an existing `id.version.json`. After a successful diagnose warms the cache, a later diagnose that cannot probe leaves the previous entry in place until TTL. Hop can then warn with a version that the operator's most recent diagnose just reported as unavailable / no skew. The function comment also overclaims: "a failure just means the next hop does not warn" is only true when no prior entry exists.

**Suggestion:** On empty probe from diagnose, remove the cache file (or write an explicit negative/empty marker that `readMachineVersionCache` treats as miss) so the hop stays consistent with the latest diagnose. Alternatively, if keeping last-known-good through flaky probes is intentional, tighten the comment to say failure leaves any prior entry untouched until TTL.

cmd/camp/version_cache.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/Obedience-Corp/camp/internal/fsutil"
+	"github.com/Obedience-Corp/camp/internal/version"
+)
+
+// versionCacheTTL bounds how long a probed remote version is trusted for the
+// hop-time warning. Skew changes on install cadence, not on the second, so an
+// hours-scale bound is right: short enough that an upgrade stops the warning the
+// same day, long enough that the warning survives between diagnose runs. A stale
+// entry is simply ignored, so the failure mode is silence, never a wrong claim.
```
obey-agent: **[nit]** The TTL comment says a stale entry yields "silence, never a wrong claim." TTL-stale is correctly silent (tested). Within TTL, a cache that became obsolete because the remote (or local) was upgraded to match can still produce a false-positive skew warning for up to 12h — accepted by the design, but "never a wrong claim" is slightly broader than the code guarantees.

**Suggestion:** Narrow the comment to "TTL-expired / absent / corrupt never claim skew" so it does not read as promising zero false positives while the entry is still fresh.

obey-agent (commented):
## Summary

Re-review of head `40ca5f7e` (stacked on #505). The only delta from the prior-reviewed commit `5eabe8ac` is the wording fix that prior review asked for: the hop warning no longer implies the local machine is ahead (`features added since may not work there` → neutral `features may not match`), with a comment that skew is unordered. That addresses the open inline note cleanly.

The design still holds: diagnose is the sole live probe writer (`writeMachineVersionCache` after `probeRemoteCampVersion`); hop only reads a disposable 12h per-machine cache and never SSHs for version; miss / corrupt / stale / empty-version all degrade to silence; stderr-only emission keeps the eval'd stdout line safe (covered by `TestSkewWarningGoesToStderrAndKeepsStdoutToOneLine`). Version identity is string equality via existing `campVersionSkew` (no semver parse), including same-version/different-commit for `dev` builds.

No bugs found on the focus axes (TTL-stale false warnings, missing-cache silence, parse correctness). Two residual suggestions remain around warning fidelity and cache invalidation when a later diagnose fails.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

cmd/camp/version_cache.go
obey-agent: **Nit (UX / consistency with diagnose):** when skew is same version string + different commit (the common personal-fleet `dev` case), this message prints identical version strings on both sides:

```text
camp on archdtop is dev, this machine is dev; features may not match
```

Diagnose already solved this with `campVersionDisplay` / `campLocalVersionDisplay` (commit appended when present, specifically called out for disambiguating `"dev"` builds). Reusing those (or the same formatting rule) would make the hop warning actionable without another `diagnose` just to learn *why* they differ. The matrix case `"same version, different commit warns"` currently only asserts presence of the version string, so it would not catch the opaque dual-`dev` message.

cmd/camp/version_cache.go
```diff
@@ -0,0 +1,88 @@
...
+
+// readMachineVersionCache returns a fresh cached probe, or false. Absent,
+// corrupt, and stale are all the same answer: no opinion.
+func readMachineVersionCache(id string) (versionCacheEntry, bool) {
+	data, err := os.ReadFile(versionCachePath(id))
+	if err != nil {
+		return versionCacheEntry{}, false
+	}
+	var e versionCacheEntry
+	if json.Unmarshal(data, &e) != nil || !e.fresh(time.Now()) {
```
obey-agent: **Nit (defense-in-depth):** `writeMachineVersionCache` refuses empty `versionStr`, but `readMachineVersionCache` will still treat a hand-edited / partially-written `{"version":"","probed_at":...}` as a fresh hit and defer to `campVersionSkew` (which also no-ops on empty). That is safe today, but the cache contract is cleaner if read matches write: empty version == miss, same as absent/corrupt/stale. One extra `e.Version == ""` check keeps "no opinion" localized to the cache layer.

cmd/camp/machine.go
```diff
@@ -405,6 +405,9 @@ func runMachineDiagnose(cmd *cobra.Command, args []string) error {
 		m := &targets[i]
 		d := remote.CheckControlMaster(ctx, m)
 		remoteVersion, remoteCommit := probeRemoteCampVersion(ctx, m)
+		// Diagnose is the only surface that pays for a live probe, so it is the
+		// only one that can warm the cache the hop path reads.
+		writeMachineVersionCache(m.ID, remoteVersion, remoteCommit)
```
obey-agent, line 415: **Observation (not a block):** a failed probe intentionally does not write, so a prior fresh entry stays for up to TTL. That means a remote that was upgraded to match (or lost `version --json`) can still produce a hop warning until TTL expires or a successful diagnose re-warms. That is the right default for flaky SSH — silence on every probe blip would hide real skew — and the message already points at re-check. Worth a one-line comment on the write call site if future readers wonder why failure does not invalidate.

obey-agent (commented):
## Verdict

**Approve with nits.** Design is sound, tradeoffs are explicit and correctly weighted, and the load-bearing safety property (stdout stays one eval line) is pinned by a real integration-style test through `emitHopOrRefuse`. No blocking defects. Stacked on #505 / #504 — merge order still matters for the base.

## Overview

This PR closes the gap where `camp machine diagnose` already paid for a live remote version probe and judged skew (`campVersionSkew` / `probeRemoteCampVersion` from earlier mesh work), but hop time — the moment skew actually bites — threw that knowledge away.

The shape is right for a personal fleet:

| Surface | Role |
|---|---|
| `camp machine diagnose` | Only writer: already probes, now warms `~/.obey/cache/machines/<id>.version.json` |
| `emitHopOrRefuse` / hop path | Only reader: file read + skew check, **never** ssh |
| Failure modes | Absent / corrupt / stale / empty probe → silence, not a wrong claim |

Freshness at 12h matches install cadence, not keystroke cadence (contrast the 60s completion cache next to it). Best-effort write, `0o600` files, atomic write via `fsutil`, same directory as completion cache so one `rm -rf ~/.obey/cache` clears derived machine state — all consistent with existing patterns.

The follow-up commit that dropped directional language ("features added since may not work there") is correct: `campVersionSkew` does not order versions, so claiming local is ahead would be wrong roughly half the time.

## Key Findings

1. **Nit — hop warning omits commit disambiguation diagnose already has** (`cmd/camp/version_cache.go` ~L86)
   - Same-version / different-commit skew (the common `dev` fleet case) renders as `camp on X is dev, this machine is dev; features may not match`, which is true but not actionable.
   - Diagnose already formats with `campVersionDisplay` / `campLocalVersionDisplay` for exactly this reason. Reuse (or mirror) that formatting so the hop warning names the commits when they are what differ.
   - Matrix case `"same version, different commit warns"` only asserts the version string is present, so it would not catch the opaque dual-`dev` text.

2. **Nit — read/write contract asymmetry for empty version** (`readMachineVersionCache`)
   - Write refuses empty `versionStr`; read will accept a hand-rolled empty version as a fresh hit and rely on `campVersionSkew` to no-op.
   - Safe today; cleaner if empty version is a miss at the cache boundary (same family as absent/corrupt/stale).

3. **Observation — failed probe leaves prior fresh entry** (`writeMachineVersionCache` call site in diagnose)
   - Intentional and correct for flaky SSH: invalidating on every probe blip would hide real skew.
   - Dual of the "silence over wrong claim" story: mid-TTL after a remote upgrade (or a probe that starts failing) you can still warn until TTL or a successful re-diagnose. Message already points at re-check. A one-line comment on the write site would freeze the intent for the next reader.

4. **Non-issue / accepted product tradeoff**
   - Warning only appears if diagnose has run in the last 12h. Users who never diagnose never see it. Prefer that over adding an ssh RTT to every hop for an advisory. Documented in the PR body and worth keeping as the load-bearing product decision.

No concerns on:
- stderr vs stdout placement (`cmd.ErrOrStderr` + dim styling)
- cache path co-location with completion cache
- not caching failed probes
- reusing `campVersionSkew` rather than forking judgment logic
- not probing from the hop path

## What's Done Well

- **Correct cost placement.** Diagnose already pays for truth; hop only reads a file. The explicit "hop NEVER probes" comment and function contract are the kind of constraint that keeps the fleet UX from rotting into "helpful but slow."
- **Stdout safety is tested where it matters.** `TestSkewWarningGoesToStderrAndKeepsStdoutToOneLine` drives `emitHopOrRefuse` with real buffers and asserts exactly one stdout newline and no warning leakage. That is the contract the shell wrapper depends on; a unit-only check of `hopSkewWarning` would have missed a wiring mistake.
- **Freshness tested by seeding age, not racing the clock.** Six-row matrix covers miss / match / version skew / commit skew / stale / empty, plus corrupt JSON and "failed probe not written."
- **Failure mode discipline.** Silence over wrong claim for absent/corrupt/stale; never cache empty probe results so emptiness cannot be mistaken for an answer.
- **Directional language fixed before merge.** The follow-up commit shows the review loop working on the same PR rather than shipping a confidently wrong half of the warning.
- **Small, focused surface.** New file + thin hooks into diagnose write and hop read. No parallel skew machinery, no new dependencies, no global state beyond the existing cache dir helper.

## Staff Standard

Meets staff bar for a small reliability/UX feature:

- [x] Reads existing seams (`machineCacheDir`, `campVersionSkew`, `emitHopOrRefuse`, `fsutil.WriteFileAtomically`) instead of inventing parallel ones
- [x] Product tradeoff (stale silence vs hop latency) is conscious and documented in code comments, not just the PR body
- [x] Tests protect the dangerous contract (eval'd stdout), not only the pure helper
- [x] Error/failure paths prefer degradation (no warning) over false confidence
- [ ] Minor: hop user-facing string should share diagnose's version+commit display so the fleet's real skew case (`dev` vs `dev`) is legible

**Ship when stacked base is ready; address commit display when convenient (same PR or fast follow).** Not a reason to hold the hop-time warning itself.

PR #507, opened 2026-07-26
feat(machine): push name-only snapshots so remotes can see the host

lancekrogers:
> **Stacked on #506** (which is stacked on #505, on #504). Base is `mesh-skew-warning`.

Makes campaign-name completion work in both directions, without the remote ever connecting back.

Festival MN0001, phase 005 sequence 01. Design: `003_DESIGN/03_transfer_reach/outputs/snapshot-verb-spec.md`.

## Why

Completion is one-way today. `camp list --remote` on the host enumerates a remote and caches its campaign names, so `csw archdtop:<TAB>` works from the mac. Nothing warms the *remote's* cache about the host, because that would require archdtop to connect back to the mac, and D003 is explicit that camp diagnoses reverse reachability rather than managing it.

So the information exists, just on the wrong side of a one-way link. The fix is that the side which *can* reach pushes what it knows.

## What changed

**Hidden `camp machine cache-put <id> --campaigns a,b,c`** (`cmd/camp/machine_cache_put.go`). Machine-to-machine plumbing, so it is hidden from help. The **receiver** validates everything it is handed, which is the reason this is a verb rather than an scp of a JSON file into the remote's cache directory: a file drop writes to another machine's state without that machine's code checking it.

The machine id is validated **first**, and not as a naming nicety: it becomes the cache *file name*, so `validateMachineID` is what keeps a hostile value inside the cache directory. Campaign names reject `/`, `:`, newline, and NUL, because on the receiver they are display and completion data, never paths.

**Push triggers**: after a successful hop emit, and after a successful remote enumerate. Both ride the ControlMaster socket the triggering operation already opened, so neither adds a connection.

**Per-source TTL** (`cmd/camp/completion_remote.go`). Pushed entries live 24h; pulled ones keep their 60s. Their refresh events differ: a pulled entry is replaced the next time you run `list --remote`, a pushed one by the next hop from that host, which may be tomorrow. A 60s bound would make the feature stop working a minute after it started. One file, one `source` field, so the keystroke path stays a single `os.ReadFile`.

**Failures are silent**: unreachable, timed out, rejected payload, or an old remote whose camp has no such verb (exit 127, the same POSIX signal `campNotFoundHint` already keys on). The operator asked to hop or to list, not to sync caches.

## Tests

`just gate` green: 3792.

Notable: an 8-row name-validation table; hostile machine ids (`../escape`, `/abs`, `local`, caps, leading digit); a TTL test proving a 30-minute-old *pushed* entry is fresh while an identically-aged *pulled* one is stale; decoder tolerance verified by unmarshalling a `source`-bearing document into a pre-change struct; and a 6-row push-silence matrix including the exit-127 old-remote case.

`TestPushCarriesNamesOnly` asserts the invariant as an **absence** (no identity_file, no home paths, no campaign ids, no host, no org, no path), so a future field addition fails the test rather than quietly shipping.

## Spec correction included

The design spec argued the push had to happen *before* the hop line is emitted, reasoning that `exec` replaces camp's process so "after means never". That is wrong: camp only *prints* the line, and the wrapper evals it after camp exits. Emit-then-push is strictly better, because the line the operator is waiting on is written before any bookkeeping starts. The spec section is corrected in place with the original reasoning and why it fails, rather than the code silently diverging.

### DIFF
--- a/cmd/camp/list_remote.go
+++ b/cmd/camp/list_remote.go
@@ -102,6 +102,11 @@ func enumerateRemoteFor(f listFilter) enumerateFunc {
 		// Warm the completion cache so `csw <id>:<tab>` has campaigns without the
 		// keystroke path ever doing a live ssh.
 		writeMachineCacheCampaigns(m.ID, names)
+		// A successful enumerate proves this machine can reach m, so it is also
+		// the moment to tell m about us. Silent on every failure.
+		if selfID, selfNames := selfSnapshot(ctx); selfID != "" {
+			pushSelfSnapshot(ctx, m, selfID, selfNames)
+		}
 		return rows, nil
 	}
 }
--- a/cmd/camp/list_remote.go
+++ b/cmd/camp/list_remote.go
@@ -80,6 +80,14 @@ func remoteListArgs(f listFilter) string {
 // the far machine).
 func enumerateRemoteFor(f listFilter) enumerateFunc {
 	args := remoteListArgs(f)
+	// selfSnapshot detects the tailnet name and loads the whole registry, and
+	// returns the same answer for every machine. Computing it inside the
+	// per-machine closure paid that cost N times over a concurrent fan-out;
+	// sync.Once shares one result across the goroutines.
+	var snapshotOnce sync.Once
+	var selfID string
+	var selfNames []string
+
 	return func(ctx context.Context, m *machines.Machine) ([]campaignEntry, error) {
 		out, err := remote.RunCampCommand(ctx, m, args)
 		if err != nil {
@@ -104,7 +112,8 @@ func enumerateRemoteFor(f listFilter) enumerateFunc {
 		writeMachineCacheCampaigns(m.ID, names)
 		// A successful enumerate proves this machine can reach m, so it is also
 		// the moment to tell m about us. Silent on every failure.
-		if selfID, selfNames := selfSnapshot(ctx); selfID != "" {
+		snapshotOnce.Do(func() { selfID, selfNames = selfSnapshot(ctx) })
+		if selfID != "" {
 			pushSelfSnapshot(ctx, m, selfID, selfNames)
 		}
 		return rows, nil
--- a/cmd/camp/machine_cache_put.go
+++ b/cmd/camp/machine_cache_put.go
@@ -1,0 +1,156 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/machines"
+	"github.com/Obedience-Corp/camp/internal/remote"
+)
+
+// snapshotMaxNames caps how many campaign names one push may carry. The payload
+// crosses as an argv, so the bound protects the receiver's command line; the
+// rule is truncate-and-still-write rather than refuse, because a partial
+// completion list is useful and an error is not.
+const snapshotMaxNames = 500
+
+// snapshotMaxBytes bounds the joined names for the same reason, far below any
+// platform ARG_MAX.
+const snapshotMaxBytes = 64 * 1024
+
+var cachePutCampaigns []string
+
+// machineCachePutCmd is machine-to-machine plumbing, not an operator verb, so it
+// is hidden. A human who runs it by hand gets correct behavior; they are simply
+// not advertised it.
+var machineCachePutCmd = &cobra.Command{
+	Use:    "cache-put <machine-id>",
+	Short:  "Record another machine's campaign names in this machine's completion cache (internal)",
+	Hidden: true,
+	Long: `Record another machine's campaign names in this machine's completion cache.
+
+Hosts call this over ssh after hopping here, so this machine can complete
+'<id>:<campaign>' for them without ever connecting back. Names only: no paths, no
+ids, no auth material. The receiver validates everything it is handed.`,
+	Args: cobra.ExactArgs(1),
+	RunE: runMachineCachePut,
+}
+
+func runMachineCachePut(cmd *cobra.Command, args []string) error {
+	id := args[0]
+	// Validated FIRST, and not merely as a naming nicety: the id becomes the
+	// cache file name, so this is what keeps a hostile value inside the cache
+	// directory.
+	if err := validateMachineID(id); err != nil {
+		return err
+	}
+	names, err := sanitizeSnapshotNames(cachePutCampaigns)
+	if err != nil {
+		return err
+	}
+	writeMachineSnapshotCampaigns(id, names)
+	// Stdout stays empty on every path so a caller can distinguish "wrote" from
+	// "said something" without parsing.
+	_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "cached %d campaign names for %s\n", len(names), id)
+	return nil
+}
+
+// sanitizeSnapshotNames validates and bounds the pushed names. Names are display
+// and completion data on the receiver, never resolved as paths, and the
+// separator and control-character rules keep them that way.
+func sanitizeSnapshotNames(raw []string) ([]string, error) {
+	out := make([]string, 0, len(raw))
+	total := 0
+	for _, chunk := range raw {
+		for _, name := range strings.Split(chunk, ",") {
+			name = strings.TrimSpace(name)
+			if name == "" {
+				continue
+			}
+			if len(name) > 255 || strings.ContainsAny(name, "/:\x00\n\r") {
+				return nil, camperrors.New("invalid campaign name in snapshot: " + name)
+			}
+			if len(out) >= snapshotMaxNames || total+len(name) > snapshotMaxBytes {
+				return out, nil
+			}
+			out = append(out, name)
+			total += len(name)
+		}
+	}
+	if len(out) == 0 {
+		return nil, camperrors.New("no campaign names given (use --campaigns a,b,c)")
+	}
+	return out, nil
+}
+
+// snapshotPushTimeout bounds a push. It rides an already-established
+// ControlMaster connection and sends a few KiB, so this is generous; and it sits
+// in front of a hop the operator is waiting on, so the 10s control-plane default
+// would be a visible regression.
+const snapshotPushTimeout = 2 * time.Second
+
+// pushSnapshotRun is the seam that lets the push's silence be tested without ssh.
+var pushSnapshotRun = remote.RunCampCommand
+
+// pushSelfSnapshot tells m about this machine's campaigns, so completion works
+// in the other direction without m ever connecting back (D006).
+//
+// Every failure is silent by design. The operator asked to hop or to list, not
+// to sync caches, so an unreachable machine, an older camp without the verb, a
+// timeout, or a rejected payload must not surface. In particular an old remote
+// exits 127 (POSIX command-not-found, the same signal campNotFoundHint keys on)
+// or with cobra's unknown-command error, and both are swallowed here.
+//
+// It is called AFTER the hop line is written. The line is what the operator is
+// waiting for; the push is bookkeeping, and bookkeeping does not go first.
+func pushSelfSnapshot(ctx context.Context, m *machines.Machine, selfID string, names []string) {
+	if m == nil || selfID == "" || len(names) == 0 {
+		return
+	}
+	if validateMachineID(selfID) != nil {
+		return
+	}
+	safe, err := sanitizeSnapshotNames(names)
+	if err != nil || len(safe) == 0 {
+		return
+	}
+	ctx, cancel := context.WithTimeout(ctx, snapshotPushTimeout)
+	defer cancel()
+	_, _ = pushSnapshotRun(ctx, m, "machine cache-put "+remote.ShellQuote(selfID)+
+		" --campaigns "+remote.ShellQuote(strings.Join(safe, ",")))
+}
+
+// selfSnapshot returns this machine's id and campaign names for a push, or
+// nothing when either is unavailable. Names only: no paths, no ids, no orgs, no
+// auth material, and no list of other machines. Each exclusion is a thing that
+// would otherwise leak to every machine this one hops to.
+func selfSnapshot(ctx context.Context) (string, []string) {
+	host, err := detectReachableName(ctx, runTailscaleStatusForSelf)
+	if err != nil {
+		return "", nil
+	}
+	id := suggestedMachineID(host)
+	if id == "" {
+		return "", nil
+	}
+	reg, err := config.LoadRegistry(ctx)
+	if err != nil {
+		return "", nil
+	}
+	all := reg.ListAll()
+	names := make([]string, 0, len(all))
+	for _, c := range all {
+		if c.Name != "" {
+			names = append(names, c.Name)
+		}
+	}
+	sort.Strings(names)
+	return id, names
+}
--- a/cmd/camp/machine_cache_put.go
+++ b/cmd/camp/machine_cache_put.go
@@ -66,7 +66,13 @@ func runMachineCachePut(cmd *cobra.Command, args []string) error {
 // completion cache wire format. Rejected characters would either escape path
 // boundaries if ever mis-resolved, or break shell/argv framing.
 func invalidSnapshotName(name string) bool {
-	return name == "" || len(name) > 255 || strings.ContainsAny(name, "/:\x00\n\r")
+	// Path separators keep a name from being resolved as a path, and every C0
+	// control (plus DEL) is rejected because these names are emitted as shell
+	// completion candidates and interpolated into error strings. A peer that can
+	// reach cache-put would otherwise park an ANSI escape in a cache with a 24h
+	// TTL and have it replayed into the operator's terminal on every TAB.
+	return name == "" || len(name) > 255 ||
+		strings.ContainsAny(name, "/:\x7f") || hopOriginHasC0(name)
 }
 
 // sanitizeSnapshotNames validates and bounds names on the RECEIVE path.
--- a/cmd/camp/machine_cache_put.go
+++ b/cmd/camp/machine_cache_put.go
@@ -62,32 +62,59 @@ func runMachineCachePut(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-// sanitizeSnapshotNames validates and bounds the pushed names. Names are display
-// and completion data on the receiver, never resolved as paths, and the
-// separator and control-character rules keep them that way.
+// invalidSnapshotName reports whether a campaign name is unsafe for the
+// completion cache wire format. Rejected characters would either escape path
+// boundaries if ever mis-resolved, or break shell/argv framing.
+func invalidSnapshotName(name string) bool {
+	return name == "" || len(name) > 255 || strings.ContainsAny(name, "/:\x00\n\r")
+}
+
+// sanitizeSnapshotNames validates and bounds names on the RECEIVE path.
+// Each element of raw is one campaign name (repeated --campaigns flags via
+// StringArray); commas are allowed inside a name because the wire format no
+// longer joins on them. Invalid names fail the whole put — the receiver must
+// not write a half-trusted cache entry.
 func sanitizeSnapshotNames(raw []string) ([]string, error) {
 	out := make([]string, 0, len(raw))
 	total := 0
-	for _, chunk := range raw {
-		for _, name := range strings.Split(chunk, ",") {
-			name = strings.TrimSpace(name)
-			if name == "" {
-				continue
-			}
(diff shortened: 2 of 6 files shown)


cmd/camp/list_remote.go
```diff
@@ -102,6 +102,11 @@ func enumerateRemoteFor(f listFilter) enumerateFunc {
 		// Warm the completion cache so `csw <id>:<tab>` has campaigns without the
 		// keystroke path ever doing a live ssh.
 		writeMachineCacheCampaigns(m.ID, names)
+		// A successful enumerate proves this machine can reach m, so it is also
+		// the moment to tell m about us. Silent on every failure.
+		if selfID, selfNames := selfSnapshot(ctx); selfID != "" {
```
obey-agent: **Cost note:** this runs inside the concurrent fan-out, so every successful remote pays an extra \`selfSnapshot\` (tailscale + registry load) plus up to \`snapshotPushTimeout\` (2s) of ssh. Parallelism keeps wall clock ~max not sum, which is good — but \`selfSnapshot\` itself is recomputed per machine with no memoization.

Consider computing \`(selfID, names)\` once outside \`enumerateRemoteFor\` / fan-out and closing over it, so N remotes share one detect+registry read. Keeps the silent-on-failure contract; just avoids N× the local work.

obey-agent (commented):
## Verdict: Approve

## Overview
Name-only snapshot push closes the reverse-completion gap without reverse connections (D006). Receiver validates machine id (path-safety) and campaign names (no separators/control bytes), truncates rather than refuses at caps, and keeps stdout empty for callers. Push is silent on every failure including old remotes (exit 127), and runs *after* the hop line so bookkeeping never blocks the operator's line of output conceptually — though it still holds the process briefly.

## Key Findings
- **Cost (non-blocking):** `list --remote` fan-out recomputes `selfSnapshot` per machine. Memoize once outside the enumerate closure. See inline.
- Longer TTL for pushed vs pulled entries is well-justified and tested (30-minute-old push still fresh).
- Pull overwrites push on the same cache file — correct authority ordering.

## What's Done Well
- Hidden internal verb with correct behavior if a human runs it.
- Hostile machine id rejection (path escape) tested.
- "Names only" invariant asserted as *absence* of forbidden substrings in the argv.
- Old camp ignores unknown `source` field and degrades to 60s TTL.
- Atomic cache writes preserved.

## Staff Standard
Yes. Security boundaries and silence contract are solid. Optional memoization is polish, not a blocker.

cmd/camp/machine_cache_put.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// sanitizeSnapshotNames validates and bounds the pushed names. Names are display
+// and completion data on the receiver, never resolved as paths, and the
+// separator and control-character rules keep them that way.
+func sanitizeSnapshotNames(raw []string) ([]string, error) {
+	out := make([]string, 0, len(raw))
+	total := 0
+	for _, chunk := range raw {
+		for _, name := range strings.Split(chunk, ",") {
```
obey-agent: **[bug]** Snapshot names are forced through a comma-delimited wire format that cannot round-trip legal campaign names. `sanitizeSnapshotNames` always `strings.Split`s each chunk on `,`, and `pushSelfSnapshot` re-joins with `strings.Join(safe, ",")` under a single `--campaigns` flag whose `StringSliceVar` also CSV-splits. Campaign config validation (`ValidateCampaignConfig`) allows commas (it only rejects `/\:*?\"<>|`). A registry entry named `foo,bar` is therefore split into two completion candidates (`foo` and `bar`) on both the sender sanitize path and the receiver parse path — silent corruption of reverse-completion data, not a reject.

**Suggestion:** Prefer repeated `--campaigns` flags with `StringArrayVar` (no CSV split), one ShellQuoted name each; or reject `,` in `sanitizeSnapshotNames` and skip (not abort) such names on the push path so the rest of the snapshot still lands. Add a test with a name containing a comma.

cmd/camp/machine_cache_put.go
```diff
@@ -0,0 +1,156 @@
...
+			name = strings.TrimSpace(name)
+			if name == "" {
+				continue
+			}
+			if len(name) > 255 || strings.ContainsAny(name, "/:\x00\n\r") {
+				return nil, camperrors.New("invalid campaign name in snapshot: " + name)
```
obey-agent: **[suggestion]** Name validation rejects `/`, `:`, NUL, `\n`, and `\r`, but allows other control bytes (ESC, TAB, BEL, etc.). Those names are never resolved as paths (good), yet they are stored and later emitted as shell completion candidates and can appear in the `invalid campaign name in snapshot: …` error string. A peer that can invoke `cache-put` (i.e. SSH as this user) can park ANSI/control sequences in the 24h push-TTL completion cache and garble the operator’s terminal on tab-complete.

**Suggestion:** Reject any rune with `unicode.IsControl` (or a explicit denylist of `c < 0x20 || c == 0x7f`) in `sanitizeSnapshotNames`, and cover TAB/ESC in the sanitize table. Optionally skip bad names on the push construction path instead of failing the whole snapshot when one registry entry is weird.

cmd/camp/machine_cache_put_test.go
```diff
@@ -0,0 +1,231 @@
...
+	cmd.SetErr(&errb)
+
+	if err := runMachineCachePut(cmd, []string{"mac-studio"}); err != nil {
+		t.Fatal(err)
+	}
+	if out.Len() != 0 {
+		t.Errorf("stdout must stay empty so callers need not parse it, got %q", out.String())
+	}
+
+	names, ok := readMachineCacheCampaigns("mac-studio")
```
obey-agent, line 123: **[suggestion]** `TestCachePutWritesSnapshotEntry` only checks that names read back while fresh. A just-written entry is fresh under both the 60s pull TTL and the 24h push TTL, so the test would still pass if `writeMachineSnapshotCampaigns` forgot `Source: snapshotSource`. The long-TTL behavior is only exercised by `TestPushedSnapshotSurvivesTheShortCompletionTTL`, which hand-writes JSON and never goes through `runMachineCachePut` / `writeMachineSnapshotCampaigns`.

**Suggestion:** After `runMachineCachePut`, read the on-disk JSON (or age `FetchedAt` and re-read via `readMachineCacheCampaigns`) and assert `Source == snapshotSource` so the verb→long-TTL path cannot regress silently.

cmd/camp/switch_remote.go
```diff
@@ -71,7 +71,15 @@ func emitHopOrRefuse(ctx context.Context, cmd *cobra.Command, m *machines.Machin
...
 	}
 	if shellConnect {
-		return emitShellConnect(cmd.OutOrStdout(), true, root, m, hopOriginForEmit(ctx, cmd))
+		if err := emitShellConnect(cmd.OutOrStdout(), true, root, m, hopOriginForEmit(ctx, cmd)); err != nil {
+			return err
+		}
+		// After the line is written, never before: the operator is waiting on
+		// that line, and this is bookkeeping. It rides the ControlMaster socket
+		// the resolve already opened, so it costs no new connection.
+		selfID, names := selfSnapshot(ctx)
```
obey-agent, line 83: **[suggestion]** After `emitShellConnect`, the hop path always runs `selfSnapshot` (tailscale status + full registry load) and then `pushSelfSnapshot` (up to `snapshotPushTimeout` = 2s). Under the real wrapper (`eval "$(camp switch … --shell-connect)"`), the shell does not eval stdout until camp exits, so emit-before-push does not make the hop line available earlier — it still serializes bookkeeping onto every successful hop. `hopOriginForEmit` already paid a reachable-name detection cost moments earlier.

**Suggestion:** Reuse the already-derived self id/host from hop-origin construction when available, and document that hops may pay up to ~2s of silent bookkeeping. Optional later improvement: fire-and-forget only if a parent wrapper can wait, or accept the bound as the product tradeoff.

cmd/camp/machine_cache_put.go
```diff
@@ -0,0 +1,156 @@
...
+// selfSnapshot returns this machine's id and campaign names for a push, or
+// nothing when either is unavailable. Names only: no paths, no ids, no orgs, no
+// auth material, and no list of other machines. Each exclusion is a thing that
+// would otherwise leak to every machine this one hops to.
+func selfSnapshot(ctx context.Context) (string, []string) {
+	host, err := detectReachableName(ctx, runTailscaleStatusForSelf)
+	if err != nil {
+		return "", nil
+	}
+	id := suggestedMachineID(host)
```
obey-agent, line 182: **[suggestion]** The pushed cache key is `suggestedMachineID` of this host’s reachable name, not “whatever id the remote already registered for us.” Reverse completion only helps when the remote’s `machines.yaml` id matches that suggestion. A hand-named fleet entry (`studio` vs suggested `mac-studio`) gets a warm cache file under the wrong basename and `csw studio:<TAB>` still misses until a pull path warms the correct key.

**Suggestion:** Document the id-matching assumption next to `selfSnapshot` / adopt, or have the pusher also learn “id the remote knows me by” when available (e.g. from a prior adopt/origin). Not a security issue; feature effectiveness gap.

cmd/camp/machine_cache_put.go
```diff
@@ -0,0 +1,156 @@
...
+	if len(out) == 0 {
+		return nil, camperrors.New("no campaign names given (use --campaigns a,b,c)")
+	}
+	return out, nil
+}
+
+// snapshotPushTimeout bounds a push. It rides an already-established
+// ControlMaster connection and sends a few KiB, so this is generous; and it sits
+// in front of a hop the operator is waiting on, so the 10s control-plane default
+// would be a visible regression.
```
obey-agent, line 129: **[nit]** The `snapshotPushTimeout` comment says the push “sits in front of a hop the operator is waiting on,” but `pushSelfSnapshot` is invoked after the hop line is written (`switch_remote.go` and the function’s own later comment). Minor doc drift from the emit-then-push correction.

**Suggestion:** Reword to “still holds the hop process for up to 2s after the line is written” (or similar) so comment and control flow agree.

obey-agent (commented):
## Summary

Re-review of head `ffae2a56` (prior review on `53b69312`). Name-only snapshot push remains a clean D006 solution: hidden `machine cache-put`, receiver-side validation, silent failure (including exit 127), pull-over-push authority, and asymmetric TTLs (24h push / 60s pull). Prior cost note is addressed — `list --remote` now memoizes `selfSnapshot` with `sync.Once` inside `enumerateRemoteFor`.

Security posture on the focus areas is mostly solid:
- **Privacy:** payload is campaign names only; `TestPushCarriesNamesOnly` asserts absence of identity files, hosts, and auth material.
- **Cache write trust:** machine id is validated before it becomes a filename; writes stay atomic `0600` under a `0700` dir. Expanding the attack surface beyond “can SSH as this user” is limited (same user can write the cache files directly).
- **Remote command injection:** both `selfID` and the joined campaigns are `remote.ShellQuote`d before `RunCampCommand` wraps again in `sh -lc` — metacharacters do not break out of the remote argv.
- **Tests:** strong matrix for sanitize, hostile ids, silence, TTL distinction, and decoder tolerance.

Residual issues: the comma-joined wire format collides with legal campaign names that contain commas; control characters beyond `\n\r\x00` are still accepted into completion data; and the critical `Source=push` path is not asserted end-to-end through `cache-put`.

## Issue counts by severity

- bugs: 1
- suggestions: 4
- nits: 2

## Issues outside the diff

- **[nit]** cmd/camp/completion_remote.go:81 -- The pre-existing `writeMachineCacheCampaigns` godoc and the new `writeMachineSnapshotCampaigns` godoc are concatenated immediately above `writeMachineSnapshotCampaigns`, so Go attaches both to the snapshot writer and leaves `writeMachineCacheCampaigns` undocumented.
  - **Suggestion:** Split into two proper doc comments, each immediately above its function.

cmd/camp/machine_cache_put.go
```diff
@@ -0,0 +1,156 @@
...
+	if validateMachineID(selfID) != nil {
+		return
+	}
+	safe, err := sanitizeSnapshotNames(names)
+	if err != nil || len(safe) == 0 {
+		return
+	}
+	ctx, cancel := context.WithTimeout(ctx, snapshotPushTimeout)
+	defer cancel()
+	_, _ = pushSnapshotRun(ctx, m, "machine cache-put "+remote.ShellQuote(selfID)+
```
obey-agent: **Wire-format ambiguity for commas in campaign names**

`--campaigns` is encoded as a single comma-joined argv (`strings.Join(safe, ",")`), and the receiver re-splits on `,` in `sanitizeSnapshotNames`. Campaign names are allowed to contain commas under `ValidateCampaignConfig` (only `/\\:*?\"<>|` are rejected), so a registry name like `foo,bar` would be corrupted on the wire (and worse, look like two campaigns).

Two low-cost options:
1. Emit repeated flags (`--campaigns name` per entry) so commas never act as a delimiter, or
2. Reject `,` in `sanitizeSnapshotNames` (and drop such names on the push path rather than failing the whole batch).

Not a launch blocker if campaign names are effectively alphanumeric today, but the wire format should match the name grammar explicitly.

cmd/camp/machine_cache_put.go
```diff
@@ -0,0 +1,156 @@
...
+// It is called AFTER the hop line is written. The line is what the operator is
+// waiting for; the push is bookkeeping, and bookkeeping does not go first.
+func pushSelfSnapshot(ctx context.Context, m *machines.Machine, selfID string, names []string) {
+	if m == nil || selfID == "" || len(names) == 0 {
+		return
+	}
+	if validateMachineID(selfID) != nil {
+		return
+	}
+	safe, err := sanitizeSnapshotNames(names)
```
obey-agent: **All-or-nothing sanitize can silence a whole push**

On the receive path, failing hard on an invalid name is correct (reject the payload). Here on the *push* path, a single registry entry with a disallowed character causes `sanitizeSnapshotNames` to error and this returns without pushing *any* names.

That is a sharp edge for a best-effort feature: one hand-edited / legacy name kills reverse completion entirely, still silently.

Prefer filter-and-continue on the push path (skip invalid names, push the rest), and keep fail-closed only in `runMachineCachePut`.

cmd/camp/machine_cache_put_test.go
```diff
@@ -0,0 +1,231 @@
...
+	var sent string
+	restore := pushSnapshotRun
+	pushSnapshotRun = func(_ context.Context, _ *machines.Machine, args string) ([]byte, error) {
+		sent = args
+		return nil, nil
+	}
+	t.Cleanup(func() { pushSnapshotRun = restore })
+
+	m := &machines.Machine{ID: "archdtop", Host: "archdtop.ts.net", SSHUser: "lance", IdentityFile: "/keys/secret"}
+	pushSelfSnapshot(context.Background(), m, "mac-studio", []string{"obey-campaign"})
```
obey-agent: **Names-only invariant is under-tested relative to the claim**

This test only exercises `pushSelfSnapshot` with caller-supplied `names` and a pre-baked `selfID`. It never runs `selfSnapshot`, so a future change that appends path/id/org/host material *inside* `selfSnapshot` would not fail this suite.

The design note (`snapshot-verb-spec.md` §6.1) asked for a registry with distinctive sentinels (path, id, org, tags) and an assertion that none appear in the built argv. Strengthening this (or adding a companion test that builds a temp registry and calls `selfSnapshot` → `pushSelfSnapshot`) would make the absence-based invariant actually guard the extraction path, not just the argv assembler.

Also: forbidding the bare substrings `"path"` and `"org"` is brittle if a legitimate campaign is ever named something containing those tokens.

cmd/camp/switch_remote.go
obey-agent, line 83: **Emit-then-push still blocks the hop (documented, residual)**

Spec §4.1 correctly notes that this still delays camp exit (and therefore the hop) by up to `snapshotPushTimeout`. Worth keeping explicit in code comments too: the wrapper is

```bash
line=$(command camp switch "$@" --shell-connect) || return "$?"
eval "$line"
```

so command substitution waits for full process exit — writing the line to stdout first does **not** make the operator see the hop sooner. Ordering is still right for cleanliness and for “never bookkeeping-before-emit,” but the latency budget is the real product constraint.

Not a change request if 2s on ControlMaster stays rare in practice; just do not oversell “operator is waiting on that line” as a latency win. A future improvement is fire-and-forget only if it can outlive process exit (today a bare goroutine dies with camp).

obey-agent (commented):
## Verdict

**Approve with comments** — shippable architecture for D006 reverse name-visibility; a few non-blocking follow-ups on wire format, push-path sanitization, and test depth.

## Overview

PR #507 implements host→remote **name-only** completion-cache snapshots so remotes can tab-complete the host without reverse SSH (stacked on the mesh series). It adds a hidden `camp machine cache-put` verb (receiver validates), push triggers after successful hop emit and after successful `list --remote` enumerate, a longer TTL for pushed entries (`source: "push"`, 24h vs 60s pull), and silent failure including old remotes (exit 127). A follow-up commit hoists `selfSnapshot` behind `sync.Once` in the enumerate fan-out.

Diff is focused: ~457 LOC across six files, with a solid unit-test suite for validation, hostile IDs, TTL asymmetry, decoder tolerance, and push silence.

## Key Findings

### 1. Comma-joined wire format can corrupt legal campaign names (Medium)

`pushSelfSnapshot` joins names with `,` and the receiver re-splits. `ValidateCampaignConfig` does **not** forbid commas in campaign names, so a name containing `,` is ambiguous on the wire.

**Suggestion:** repeated `--campaigns` flags per name, or reject/skip `,` in `sanitizeSnapshotNames` so grammar and transport agree.

### 2. Push path fails closed on one bad name (Medium–Low)

`sanitizeSnapshotNames` returns an error on the first invalid name. On the receive path that is correct; on the push path it aborts the entire silent push. One legacy/hand-edited registry name disables reverse completion for everything.

**Suggestion:** filter-and-continue when building the outbound payload; keep fail-closed only in `runMachineCachePut`.

### 3. `TestPushCarriesNamesOnly` does not cover `selfSnapshot` (Medium)

The PR/description claim the names-only invariant is enforced as an absence so future field additions fail the test. The test only feeds pre-baked names into `pushSelfSnapshot`. Extraction leaks inside `selfSnapshot` would not be caught. Spec §6.1 asked for registry sentinels (path/id/org/tags).

**Suggestion:** temp registry → `selfSnapshot` → capture argv; assert name present and path/id/org/tags/host absent. Avoid bare substrings `"path"` / `"org"` as forbidden tokens.

### 4. Hop latency: emit-first is not a wall-clock win (Low / residual)

Documented in the corrected spec §4.1. The `csw` wrapper uses `line=$(camp switch … --shell-connect)`, so the operator still waits for process exit (including up to 2s push). Ordering is still the right default; just do not frame it as reducing perceived hop latency. Bound is appropriate.

### 5. Cache key = `suggestedMachineID` (Low / residual)

Consistent with hop-origin and adopt defaults. If a remote renames the host’s machine id away from the suggested form, the push lands under a key completion never reads. Acceptable if adopt-default is the common path; worth a diagnose note later (D007 “explain in diagnose” open item already recorded).

## What's Done Well

- **Receiver is authority**: verb instead of scp/file-drop; `validateMachineID` first because it becomes the cache filename — path-escape is correctly treated as a security boundary, not a naming nicety.
- **Asymmetric TTL with one file / one read**: keystroke path stays a single `os.ReadFile`; pull overwrites push with short TTL when reverse reach exists; pushed entries survive the long gap between hops. Backward compatible (`omitempty` + plain unmarshal).
- **Silent by contract**: exit 127, timeout, unreachable, validation — none pollute hop/list UX. `pushSnapshotRun` seam makes that testable without ssh.
- **Emit-then-push + tight 2s bound**: matches the corrected mental model of the shell wrapper; avoids 10s control-plane default on the hop path.
- **Payload minimalism**: names only; exclusions are reasoned and documented.
- **`sync.Once` for self snapshot** in concurrent enumerate: correct fix for N× tailscale+registry cost; race-tested.
- **Atomic 0600 writes** via existing `fsutil.WriteFileAtomically`.
- **Hostile machine id table** and TTL asymmetry test are the right kind of regression guards.

## Staff Standard

| Bar | Assessment |
| --- | --- |
| Read before write / design alignment | Strong — implements D006, corrects a wrong “after = never” claim in-place rather than silently diverging |
| No parallel systems | Extends existing completion cache; no second store |
| Security / trust boundary | Receiver validation, ID-as-filename, names-only payload — staff-grade |
| Errors / silence contract | Intentional and tested; correct for bookkeeping on operator verbs |
| Tests as safety net | Good breadth; one oversold invariant (selfSnapshot) |
| Operability | Hidden verb, empty stdout, stderr status for manual runs |
| Latency / friction | Conscious 2s bound; acceptable for launch with eyes open |

**Bottom line:** Mergeable after optional follow-ups. I would not block launch on (1)–(3), but I would land a small follow-up for wire-format/name-grammar alignment and a real `selfSnapshot` names-only test before calling the invariant “enforced.”

Inline comments posted on:
- `cmd/camp/machine_cache_put.go` (comma wire format; push sanitize)
- `cmd/camp/machine_cache_put_test.go` (invariant coverage)
- `cmd/camp/switch_remote.go` (hop latency framing)

PR #508, opened 2026-07-26
feat(transfer): machine-first grammar and cross-machine transport

lancekrogers:
> **Stacked on #507** (chain: #504 → #505 → #506 → #507). Base is `mesh-snapshot-push`.

Makes `camp transfer` work across machines: `camp transfer archdtop:obey-campaign:docs/x.md docs/`.

Festival MN0001, phase 005 sequence 02. Design: `003_DESIGN/03_transfer_reach/outputs/transfer-grammar-spec.md` and `transfer-transport-spec.md`.

## Backward compatibility was measured, not asserted

The compat matrix was written **first**, against the unchanged parser, and watched to pass. Only then did the parser change. Ten rows are pinned unchanged; exactly the two the design flagged moved.

The behavior change is gated on the leading segment matching a **registered machine id**, so a user with no `~/.obey/machines.yaml` gets byte-identical behavior. On an unconfigured fleet the new grammar is unreachable.

Live, against the real registry:

```
$ camp transfer weird:name.md /tmp/zzz          # unchanged
Error: resolve source: campaign "weird" not found in registry

$ camp transfer archdtop:notes.md /tmp/zzz      # the one change
Error: resolve source: "archdtop" is a machine; use machine:campaign:path (for example archdtop:<campaign>:notes.md)
```

That row used to fail a campaign lookup for a campaign named `archdtop`, sending the operator hunting for the wrong thing. `local:` is also new (it was an error before) and exists so registering a machine can never make a campaign unreachable by name.

## Transport

The far campaign root is resolved by the **remote's own camp**; nothing computes a far path locally. The copy rides the same ssh option set a hop uses, via `internal/peer`'s builder rather than a second construction, which is what makes "no new auth surface" structural.

**This applies the D005 amendment raised during design.** D005 said scp; reading CS0005 showed camp already has a reviewed, ssh-multiplexed **rsync** transport, and rsync brings path quoting through the protocol (`-s`), stall detection, and partial-transfer semantics scp has no vocabulary for. scp stays as the fallback since it ships with OpenSSH. If you'd rather keep D005 as written, the fallback ordering is the only thing that changes.

**Neither side is probed.** An earlier design folded a probe into the resolve round-trip, which would have corrupted every resolved remote path, because `ResolveRoot` treats *all* of stdout as the path. Remote absence is detected from the failed attempt instead (exit 12 + connection-closed + nothing transferred), which is free in the common case and safe to retry precisely because no data stream started.

**No wall-clock deadline on the copy.** A size-scaled bound needs the source size, which for a pull lives on the far machine, and any rate constant turns a slow link into a spurious failure. `rsync --timeout` detects a *stall*, which is the question a timeout is actually asking. scp gets keepalives.

## Live verification (mac-studio ↔ archdtop)

```
$ camp transfer mesh-test.txt archdtop:lance-arch:mesh-test.txt
Transferred mesh-test.txt -> archdtop:lance-arch:mesh-test.txt
$ ssh archdtop 'cat .../mesh-test.txt'
mesh transfer live test 115243

$ camp transfer mesh-test.txt archdtop:lance-arch:mesh-test.txt      # no --force
Error: destination exists on archdtop, not overwritten (use --force)

$ camp transfer ... --force        → overwrote, confirmed by reading the remote file
$ camp transfer archdtop:lance-arch:mesh-test.txt pulled.txt   → pulled, source untouched
```

Test files removed from both machines afterward.

## A defect only the live run could find

Without `--force` the copy correctly declined to overwrite and then printed **"Transferred"**. rsync exits 0 whether `--ignore-existing` copied or skipped, so the command reported success for a file it had left alone, which an operator would act on. Fixed with `--itemize-changes` and an explicit transfer-line check. Every unit test passed before and after the fix.

## Tests

`just gate` green: 3802.

Worth noting: **both fleet machines have rsync**, so the scp fallback can never be exercised naturally. It is forced in a test, which is the only way it ships executed rather than assumed. The fallback signature test also pins that exit 12 *alone* is not enough — it is a general protocol error, and retrying on it would hide real failures.

## Scope limit

Directories are not supported cross-machine here; local-to-local keeps its existing directory support, so nothing that worked stops working. Recursive copy interacts with the no-delete invariant, symlink policy, and partial-transfer reporting in ways that deserve their own pass.

### DIFF
--- a/cmd/camp/transfer.go
+++ b/cmd/camp/transfer.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -10,6 +12,7 @@ import (
 
 	"github.com/Obedience-Corp/camp/internal/campaign"
 	"github.com/Obedience-Corp/camp/internal/config"
+	"github.com/Obedience-Corp/camp/internal/machines"
 	"github.com/Obedience-Corp/camp/internal/transfer"
 	"github.com/spf13/cobra"
 )
@@ -73,8 +76,7 @@ func runTransfer(cmd *cobra.Command, args []string) error {
 		return transfer.BothRemoteError(src, dest)
 	}
 	if src.IsRemote() || dest.IsRemote() {
-		return camperrors.New("cross-machine transfer is not implemented yet; " +
-			"the grammar parses but the transport lands in a follow-up")
+		return runRemoteTransfer(ctx, cmd, root, src, dest, force)
 	}
 
 	srcPath, err := transfer.ResolveCrossCampaignPath(ctx, root, src.Spec)
@@ -204,4 +206,48 @@ func completeCampaignPath(campRoot, pathPrefix, colonPrefix string) ([]string, c
 		completions = append(completions, colonPrefix+prefix+name+suffix)
 	}
 	return completions, cobra.ShellCompDirectiveNoSpace | cobra.ShellCompDirectiveNoFileComp
+}
+
+// runRemoteTransfer moves one file between this machine and a fleet member. The
+// far campaign root is resolved by the remote's own camp, never computed here.
+func runRemoteTransfer(ctx context.Context, cmd *cobra.Command, root string, src, dest transfer.Endpoint, force bool) error {
+	remoteEnd, localSpec, pull := dest, src.Spec, false
+	if src.IsRemote() {
+		remoteEnd, localSpec, pull = src, dest.Spec, true
+	}
+
+	mf, err := machines.Load()
+	if err != nil {
+		return err
+	}
+	m, _, found := mf.Lookup(remoteEnd.Machine)
+	if !found {
+		return camperrors.New("unknown machine \"" + remoteEnd.Machine + "\"; add it to ~/.obey/machines.yaml")
+	}
+
+	localPath := transfer.ResolveCampaignRelative(root, localSpec)
+	if pull && transfer.IsDestDir(localPath) {
+		localPath = filepath.Join(localPath, filepath.Base(remoteEnd.Path))
+	}
+	if !pull {
+		if err := transfer.ValidatePathExists(localPath); err != nil {
+			return camperrors.Wrap(err, "source")
+		}
+	}
+
+	if err := transfer.CopyRemote(ctx, transfer.CopyOptions{
+		Machine:  m,
+		Endpoint: remoteEnd,
+		Local:    localPath,
+		Pull:     pull,
+		Force:    force,
+	}); err != nil {
+		if errors.Is(err, transfer.ErrDestinationExists) {
+			return camperrors.Newf("destination exists on %s, not overwritten (use --force)", m.ID)
+		}
+		return camperrors.Wrapf(err, "run 'camp machine diagnose %s' to check reachability", m.ID)
+	}
+
+	_, err = fmt.Printf("Transferred %s -> %s\n", src.Spec, dest.Spec)
+	return err
 }
--- a/cmd/camp/transfer.go
+++ b/cmd/camp/transfer.go
@@ -13,25 +13,38 @@ import (
 	"github.com/Obedience-Corp/camp/internal/campaign"
 	"github.com/Obedience-Corp/camp/internal/config"
 	"github.com/Obedience-Corp/camp/internal/machines"
+	"github.com/Obedience-Corp/camp/internal/remote"
 	"github.com/Obedience-Corp/camp/internal/transfer"
 	"github.com/spf13/cobra"
 )
 
 var transferCmd = &cobra.Command{
 	Use:   "transfer <src> <dest>",
-	Short: "Copy files between campaigns",
-	Long: `Copy files between different campaigns using campaign:path syntax.
+	Short: "Copy files between campaigns (and machines)",
+	Long: `Copy files between campaigns, and between this machine and a registered
+fleet machine.
 
 Transfer always copies — it never moves or deletes the source.
-Either the source or destination (or both) can use "campaign:path"
-notation to reference a different registered campaign. Paths without
-a campaign prefix resolve relative to the current campaign root.
 
-At least one side must reference a different campaign. For copies
-within the same campaign, use 'camp copy' instead.`,
-	Example: `  camp transfer docs/my-doc.md other-campaign:docs/my-doc.md     # push
-  camp transfer other-campaign:docs/my-doc.md docs/              # pull
-  camp transfer other:festivals/plan.md festivals/planned/       # pull into dir`,
+Local forms:
+  campaign:path     another registered campaign on this machine
+  path              relative to the current campaign root
+  local:campaign:path
+                    force the campaign reading when campaign name collides
+                    with a registered machine id
+
+Machine forms (one side only; both-remote is refused):
+  machine:campaign:path
+                    file on a machine registered in ~/.obey/machines.yaml
+
+At least one side must reference a different campaign or machine. For copies
+within the same campaign on this machine, use 'camp copy' instead.`,
+	Example: `  camp transfer docs/my-doc.md other-campaign:docs/my-doc.md     # local push
+  camp transfer other-campaign:docs/my-doc.md docs/              # local pull
+  camp transfer other:festivals/plan.md festivals/planned/       # pull into dir
+  camp transfer docs/x.md archdtop:obey-campaign:docs/x.md       # push to machine
+  camp transfer archdtop:obey-campaign:docs/x.md docs/x.md       # pull from machine
+  camp transfer local:other:docs/x.md archdtop:camp:docs/x.md    # local: escape hatch`,
 	Args: cobra.ExactArgs(2),
 	RunE: runTransfer,
 	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
@@ -224,14 +237,30 @@ func runRemoteTransfer(ctx context.Context, cmd *cobra.Command, root string, src
 	if !found {
 		return camperrors.New("unknown machine \"" + remoteEnd.Machine + "\"; add it to ~/.obey/machines.yaml")
 	}
+	// Same auth precondition as hop/switch: password-auth machines fail here
+	// with an actionable message instead of a BatchMode transport failure.
+	if err := remote.EnsureKeyAuth(m); err != nil {
+		return err
+	}
 
-	localPath := transfer.ResolveCampaignRelative(root, localSpec)
+	// Cross-campaign local resolution (other-campaign:path, local:other:path →
+	// Spec "other:path") matches the local transfer path. ResolveCampaignRelative
+	// alone would treat the colon as a literal filename under the current root.
+	localPath, err := transfer.ResolveCrossCampaignPath(ctx, root, localSpec)
+	if err != nil {
+		return camperrors.Wrap(err, "resolve local path")
+	}
 	if pull && transfer.IsDestDir(localPath) {
 		localPath = filepath.Join(localPath, filepath.Base(remoteEnd.Path))
 	}
 	if !pull {
 		if err := transfer.ValidatePathExists(localPath); err != nil {
 			return camperrors.Wrap(err, "source")
+		}
+	} else {
+		// Parity with local transfer: create missing parent dirs before the copy.
+		if err := os.MkdirAll(filepath.Dir(localPath), 0o755); err != nil {
+			return camperrors.Wrap(err, "create destination directory")
 		}
 	}
 
@@ -243,6 +272,11 @@ func runRemoteTransfer(ctx context.Context, cmd *cobra.Command, root string, src
 		Force:    force,
 	}); err != nil {
 		if errors.Is(err, transfer.ErrDestinationExists) {
+			// Phrase by where the destination lives — m.ID is always the far
+			// machine, so a pull must not blame the remote for a local path.
+			if pull {
+				return camperrors.Newf("destination %q already exists (use --force to overwrite)", localPath)
+			}
 			return camperrors.Newf("destination exists on %s, not overwritten (use --force)", m.ID)
 		}
 		return camperrors.Wrapf(err, "run 'camp machine diagnose %s' to check reachability", m.ID)
--- a/cmd/camp/transfer.go
+++ b/cmd/camp/transfer.go
@@ -57,11 +57,11 @@ func runTransfer(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	src, err := transfer.ParseEndpointDefault(args[0])
+	src, err := transfer.ParseEndpointDefault(ctx, args[0])
 	if err != nil {
 		return camperrors.Wrap(err, "resolve source")
 	}
-	dest, err := transfer.ParseEndpointDefault(args[1])
+	dest, err := transfer.ParseEndpointDefault(ctx, args[1])
 	if err != nil {
 		return camperrors.Wrap(err, "resolve destination")
 	}
--- a/cmd/camp/transfer.go
+++ b/cmd/camp/transfer.go
@@ -54,7 +54,30 @@ func runTransfer(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	srcPath, err := transfer.ResolveCrossCampaignPath(ctx, root, args[0])
+	src, err := transfer.ParseEndpointDefault(args[0])
+	if err != nil {
+		return camperrors.Wrap(err, "resolve source")
+	}
+	dest, err := transfer.ParseEndpointDefault(args[1])
+	if err != nil {
+		return camperrors.Wrap(err, "resolve destination")
+	}
+	for _, e := range []transfer.Endpoint{src, dest} {
+		if e.Shadowed {
+			_, _ = fmt.Fprintln(cmd.ErrOrStderr(), transfer.ShadowNote(e.Machine))
+		}
+	}
+	if src.IsRemote() && dest.IsRemote() {
+		// Brokering a copy between two far machines would need them to reach
+		// each other, which D003 says camp diagnoses rather than manages.
+		return transfer.BothRemoteError(src, dest)
+	}
+	if src.IsRemote() || dest.IsRemote() {
+		return camperrors.New("cross-machine transfer is not implemented yet; " +
+			"the grammar parses but the transport lands in a follow-up")
+	}
+
+	srcPath, err := transfer.ResolveCrossCampaignPath(ctx, root, src.Spec)
 	if err != nil {
 		return camperrors.Wrap(err, "resolve source")
 	}
@@ -62,13 +85,13 @@ func runTransfer(cmd *cobra.Command, args []string) error {
 		return camperrors.Wrap(err, "source")
 	}
 
-	destPath, err := transfer.ResolveCrossCampaignPath(ctx, root, args[1])
+	destPath, err := transfer.ResolveCrossCampaignPath(ctx, root, dest.Spec)
 	if err != nil {
 		return camperrors.Wrap(err, "resolve destination")
 	}
 
 	// If dest is a directory or ends with /, place source inside it
-	destArg := args[1]
+	destArg := dest.Spec
 	// Strip campaign prefix for trailing slash check
 	if idx := strings.Index(destArg, ":"); idx >= 0 {
 		destArg = destArg[idx+1:]
--- a/internal/transfer/endpoint.go
+++ b/internal/transfer/endpoint.go
@@ -1,6 +1,11 @@
 package transfer
 
-import "github.com/Obedience-Corp/camp/internal/machines"
+import (
+	"context"
+
+	"github.com/Obedience-Corp/camp/internal/config"
+	"github.com/Obedience-Corp/camp/internal/machines"
+)
 
 // LocalPrefix forces the campaign reading of an endpoint whose leading segment
 // would otherwise be read as a machine id. It is spelled "local" to match
@@ -101,6 +106,23 @@ func BothRemoteError(src, dest Endpoint) error {
 // campaign check that only runs when a machine already matched, so the common
 // case (no fleet configured) costs one cheap file read that Load already caches
 // behind an absent-file fast path.
-func ParseEndpointDefault(spec string) (Endpoint, error) {
-	return ParseEndpoint(spec, registeredMachine, nil)
+func ParseEndpointDefault(ctx context.Context, spec string) (Endpoint, error) {
+	return ParseEndpoint(spec, registeredMachine, registeredCampaign(ctx))
+}
+
+// registeredCampaign reports whether head also names a campaign, which is what
+// makes a spec ambiguous and worth a note. The registry is only read after a
+// machine id already matched, so an unconfigured fleet never pays for it.
+func registeredCampaign(ctx context.Context) campaignLookupFunc {
+	return func(head string) bool {
+		if head == "" {
+			return false
+		}
+		reg, err := config.LoadRegistry(ctx)
+		if err != nil {
+			return false
+		}
+		_, found := reg.GetByName(head)
+		return found
+	}
 }
--- a/internal/transfer/endpoint.go
+++ b/internal/transfer/endpoint.go
@@ -1,0 +1,106 @@
+package transfer
+
+import "github.com/Obedience-Corp/camp/internal/machines"
+
+// LocalPrefix forces the campaign reading of an endpoint whose leading segment
+// would otherwise be read as a machine id. It is spelled "local" to match
+// machines.LocalMachineID and `camp switch local:<campaign>`, so one reserved
+// word means the same thing in both commands.
+const LocalPrefix = machines.LocalMachineID
+
+// Endpoint is a parsed transfer endpoint. Machine is empty for every form that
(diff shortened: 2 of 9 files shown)


internal/transfer/remote.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+		// Safe to retry: rsync never started a data stream, so nothing was
+		// written on either side.
+	}
+
+	if _, err := opts.LookPath("scp"); err != nil {
+		return camperrors.New("rsync not found on " + opts.Machine.ID + " and scp not found locally; " +
+			"install rsync on both machines, or scp on this one")
+	}
+	out, runErr := opts.Run(ctx, "scp", opts.scpArgs(src, dest)...)
```
obey-agent, line 133: **Bug: no-clobber contract is dropped on the scp fallback.**

rsync honors \`Force\` via \`--ignore-existing\` + \`rsyncCopied\`. The scp path always overwrites, regardless of \`Force\`. OpenSSH scp has no portable no-clobber flag, so this is not free — but silent clobber on the fallback path violates the user-visible contract already documented and tested for rsync:

- without \`--force\`: destination must not be overwritten
- operator sees "Transferred …" after a clobber they never consented to

Options (pick one, pin with a test):
1. When \`!Force\`, refuse the scp fallback with a clear error ("rsync missing on <id>; install rsync, or re-run with --force to allow scp overwrite")
2. Preflight existence with one \`ssh … test ! -e <path>\` using the same option set, then refuse/copy
3. For local destinations only, \`os.Stat\` before scp (covers pull; still need (1) or (2) for push)

This is the production-confidence item on this PR.

internal/transfer/endpoint.go
```diff
@@ -0,0 +1,106 @@
...
+func BothRemoteError(src, dest Endpoint) error {
+	return newBothRemoteError(src.Machine, dest.Machine)
+}
+
+// ParseEndpointDefault is ParseEndpoint wired to the real machines.yaml and to a
+// campaign check that only runs when a machine already matched, so the common
+// case (no fleet configured) costs one cheap file read that Load already caches
+// behind an absent-file fast path.
+func ParseEndpointDefault(spec string) (Endpoint, error) {
+	return ParseEndpoint(spec, registeredMachine, nil)
```
obey-agent: **Dead path: shadow notes never fire in production.**

Comment claims ParseEndpointDefault is "wired … to a campaign check", but the third arg is \`nil\`. With \`isCampaign == nil\`, \`Shadowed\` is always false, so the stderr note in \`runTransfer\` is unreachable.

Either wire a real campaign lookup (cheap after machine match, as the comment describes) or drop the production call site / adjust the comment so the design claim matches the binary. The unit tests cover the feature only because they inject \`isCampaign\`.

obey-agent (commented):
## Verdict: Request Changes

## Overview
Machine-first transfer grammar and cross-machine transport are thoughtfully designed: compat matrix written before the parser change, behavior gated on registered machine ids, remote roots resolved by the far camp, rsync-first over shared peer ssh options, stall timeout instead of wall-clock size scaling, and no far-side probe that would corrupt `ResolveRoot` stdout. That is the right architecture. Two production gaps keep this from a clean approve.

## Key Findings
1. **No-clobber broken on scp fallback (must fix):** rsync honors `--force` / `--ignore-existing`; scp always overwrites. Without `--force`, a missing-remote-rsync path can clobber a destination and still print "Transferred …". Pin a test. See inline.
2. **Shadow notes dead in production:** `ParseEndpointDefault` passes `nil` for `isCampaign` despite the comment claiming a campaign check is wired. Either wire it or stop claiming the note fires. See inline.
3. (Non-blocking) Both-remote rejection and `local:` escape hatch are correct and tested.

## What's Done Well
- Compat matrix discipline (measure before change) is exemplary.
- Shared `peer.SSHCommandFor` makes "no new auth surface" structural.
- rsync `-s` + itemize-changes for honest skip detection.
- Remote-lacks-rsync signature is narrow (exit 12 + closed/not-found) so real failures do not silently retry.
- Live verification against the real fleet.

## Staff Standard
Not yet — not while the no-clobber contract depends on which binary won. Fix the scp/`Force` path (and preferably wire or drop shadowing), then this is merge-ready and strong.

obey-agent (commented):
## Summary

This head adds a machine-first endpoint grammar (`machine:campaign:path`), `local:` disambiguation, shadow notes when a head is both a machine and a campaign, and cross-machine copy via rsync (scp fallback) that reuses the shared hop ssh option set through `peer.SSHCommandFor`.

Strengths:
- Backward-compat story is evidence-based: compat matrix written first, gated on **registered** machine ids, unconfigured fleet stays on the old reading.
- Shadow note is reachable in production now (`ParseEndpointDefault` wires a real campaign lookup; covered by `endpoint_shadow_test.go`).
- Transport design is careful: remote root via far `camp switch --print`, no second auth construction, rsync stall timeout instead of wall-clock deadline, scp no-clobber probe, itemize-based detection of `--ignore-existing` skips (the live “Transferred but did nothing” defect).
- Fallback signature is tight (exit 12 alone is not enough) and unit-tested; both-remote is rejected early.

Main gaps: the local side of a remote transfer does not run the campaign:path resolver (breaks `local:` and cross-campaign local endpoints when the other side is remote), remote path join does not keep the path inside the campaign root, and a few parity/error-message issues around dest-dir placement, parent `MkdirAll`, and the “rsync missing” message.

## Issue counts by severity

- bugs: 4
- suggestions: 4
- nits: 3

## Issues outside the diff

- **[nit]** internal/transfer/endpoint.go:267 -- Shadow detection uses `reg.GetByName(head)` only. Campaign resolution elsewhere uses `Get` (id, id-prefix, then name). A head that matches a campaign id/prefix but not its name will not set `Shadowed`, so the note can stay silent in odd id-collision cases. Name collision is the common case; this is incomplete relative to `Get`.
  - **Suggestion:** Use the same lookup as campaign resolution (`Get`) for the shadow bit, or document that shadowing is name-only by design.
- **[nit]** internal/transfer/endpoint_compat_test.go:327 -- Compat matrix includes B6 (`8deed8b4-aaaa:docs/x.md`) in `TestEndpointCompatMatrix` (parseSpec only) but not in `TestParseEndpointCompatAfterChange` unchanged rows. Low risk, but the “only the intended rows move” claim is slightly under-pinned for UUID-shaped campaign prefixes.
  - **Suggestion:** Add B6 to the post-change unchanged table (and any other B-rows still missing).
- **[nit]** internal/transfer/endpoint.go:182 -- `registeredMachine` returns false on `machines.Load()` error (corrupt YAML, permission). The head then falls through to the campaign reading instead of failing closed. Absent file is fine (empty fleet); a corrupt file silently changes grammar behavior.
  - **Suggestion:** Prefer surfacing load errors from `ParseEndpointDefault` (or treat load failure as hard error when a colon form is present), so misconfiguration is not misread as “not a machine.”

cmd/camp/transfer.go
```diff
@@ -182,3 +207,47 @@ func completeCampaignPath(campRoot, pathPrefix, colonPrefix string) ([]string, c
...
+	mf, err := machines.Load()
+	if err != nil {
+		return err
+	}
+	m, _, found := mf.Lookup(remoteEnd.Machine)
+	if !found {
+		return camperrors.New("unknown machine \"" + remoteEnd.Machine + "\"; add it to ~/.obey/machines.yaml")
+	}
+
+	localPath := transfer.ResolveCampaignRelative(root, localSpec)
```
obey-agent: **[bug]** `runRemoteTransfer` resolves the local endpoint with `ResolveCampaignRelative(root, localSpec)` instead of `ResolveCrossCampaignPath`. After parsing, a non-remote local endpoint may still be `campaign:path` (or `local:campaign:path` rewritten to `campaign:path` in `Spec`). That string is then treated as a single campaign-relative path component, so e.g. `camp transfer other:docs/x.md archdtop:camp:docs/x.md` and `camp transfer local:other:docs/x.md archdtop:camp:docs/x.md` look for a literal file named with a colon under the current campaign rather than resolving the registry. The `local:` escape hatch is documented as how you reach a campaign when its name collides with a machine id; it is unusable for the local side of any remote transfer.

**Suggestion:** Resolve the non-remote side with `ResolveCrossCampaignPath(ctx, root, localSpec)` (same as the pure-local path). Keep `IsDestDir` / basename placement on the resolved absolute path. Add a regression test for `local:<campaign>:<path>` + remote and for plain `other-campaign:path` + remote.

internal/transfer/remote.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// endpoints resolves the remote campaign root through the remote's OWN camp, so
+// the far registry decides the path and this machine never computes it.
+func (o CopyOptions) endpoints(ctx context.Context) (src, dest string, err error) {
+	root, err := remote.ResolveRoot(ctx, o.Machine, o.Endpoint.Campaign)
+	if err != nil {
+		return "", "", err
+	}
+	remotePath := path.Join(root, o.Endpoint.Path)
```
obey-agent: **[bug]** `remotePath := path.Join(root, o.Endpoint.Path)` does not enforce “path within the campaign.” Go’s `path.Join` discards prior segments when an element is absolute, so `archdtop:camp:/etc/passwd` becomes `/etc/passwd` on the far host. `..` segments also walk out of the campaign root (`path.Join("/srv/camp", "../../.ssh/id_rsa")`). Endpoint documents `Path` as within the campaign; artifact pull already uses `EnsureRootWithin`-style checks for the same class of escape. Operator ssh access does not make the product claim true.

**Suggestion:** Reject absolute `Endpoint.Path` and cleaned paths that are not strictly under `root` (e.g. `path.Clean`, then require relative + no `..` escape via `path.Rel` / prefix check). Prefer failing closed before rsync/scp. Add tests for `/abs`, `..`, and `foo/../../outside`.

cmd/camp/transfer.go
obey-agent: **[bug]** Local-to-local transfer `MkdirAll`s the destination parent before copy. Remote pull never creates missing local parents; rsync/scp then fail with a low-level missing-directory error. Same gap on push for missing remote parents (harder to fix without a remote mkdir policy), but local parent creation on pull is straightforward parity.

**Suggestion:** On pull (and ideally before any copy), `os.MkdirAll(filepath.Dir(localPath), 0o755)` when the dest is a file path, matching the local branch. Document whether remote parents are created or must exist.

internal/transfer/remote.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+		if !remoteLacksRsync(runErr, out) {
+			return camperrors.Wrapf(runErr, "%s %s: %s", opts.direction(), opts.Machine.ID, trimOutput(out))
+		}
+		// Safe to retry: rsync never started a data stream, so nothing was
+		// written on either side.
+	}
+
+	if _, err := opts.LookPath("scp"); err != nil {
+		return camperrors.New("rsync not found on " + opts.Machine.ID + " and scp not found locally; " +
```
obey-agent, line 112: **[bug]** When local `rsync` is absent, `LookPath("rsync")` fails and the code never attempts rsync; if `scp` is also missing it returns `rsync not found on <machine> and scp not found locally`. That blames the remote for a local PATH problem and never verified remote rsync. The message is only accurate when local rsync ran and hit the “remote lacks rsync” signature, then scp was also missing.

**Suggestion:** Branch the error: if local rsync was never found, say so; if local rsync failed with the remote-missing signature and scp is missing, name the remote. Avoid implying a remote diagnosis that was not performed.

cmd/camp/transfer.go
```diff
@@ -182,3 +207,47 @@ func completeCampaignPath(campRoot, pathPrefix, colonPrefix string) ([]string, c
...
+	if pull && transfer.IsDestDir(localPath) {
```
obey-agent, line 253: **[suggestion]** Dest-as-directory / trailing-slash placement (`IsDestDir` + `filepath.Base`) runs only for **pull** into a local dir. Push has no analogue: `archdtop:camp:docs/` does not append the source basename. Additionally, `path.Join` in `endpoints` strips a trailing slash from `Endpoint.Path`, so the trailing-slash signal is lost before rsync/scp see it. Local transfer still handles dir dests; remote behavior is inconsistent and surprising.

**Suggestion:** Mirror local rules: if remote path is a dir intent (trailing `/` on the raw path segment, or a remote `test -d` when needed), join `filepath.Base(localSource)`. Preserve dir intent before `path.Join` cleans it, or apply basename placement in `runRemoteTransfer` before `CopyRemote`.

internal/transfer/remote.go
```diff
@@ -0,0 +1,248 @@
...
+
+// rsyncArgs mirrors the artifact-pull argv minus the pieces that exist for a
+// no-clobber directory sync: no --compare-dest, no exclude file, no staging
+// tree. -s sends paths through the rsync protocol so spaces and metacharacters
+// do not depend on remote-shell parsing.
+func (o CopyOptions) rsyncArgs(src, dest string) []string {
+	// --itemize-changes makes the skip observable: with --ignore-existing rsync
+	// exits 0 whether it copied or declined, and reporting "Transferred" for a
+	// file that was left alone is a lie the operator acts on.
+	args := []string{"-a", "-s", "--no-links", "--itemize-changes", "--timeout=" + rsyncStallTimeout}
```
obey-agent, line 195: **[suggestion]** PR scope says directories are not supported cross-machine, but rsync args use `-a` (recursive). A directory source will be copied by rsync; scp fallback (no `-r`) will fail. Local path still has explicit `CopyDir` vs file branching. Operators get different outcomes depending on which transport wins.

**Suggestion:** Before copy, `Stat` the local source (push) or reject remote dirs once detected; return a clear “cross-machine transfer is file-only” error. Or drop recursion from the argv if file-only is the contract (`-a` without recursive is awkward—prefer an explicit non-recursive flag set).

cmd/camp/transfer.go
```diff
@@ -182,3 +207,47 @@ func completeCampaignPath(campRoot, pathPrefix, colonPrefix string) ([]string, c
...
+	remoteEnd, localSpec, pull := dest, src.Spec, false
+	if src.IsRemote() {
+		remoteEnd, localSpec, pull = src, dest.Spec, true
+	}
+
+	mf, err := machines.Load()
+	if err != nil {
+		return err
+	}
+	m, _, found := mf.Lookup(remoteEnd.Machine)
```
obey-agent, line 236: **[suggestion]** `runRemoteTransfer` loads the machine and copies without `remote.EnsureKeyAuth`. Peer/switch paths reject password-auth machines with an actionable error; transfer would proceed to BatchMode ssh/rsync/scp and surface a weaker transport failure. Same “no new auth surface” goal should include the same auth precondition.

**Suggestion:** Call `remote.EnsureKeyAuth(m)` after lookup (or factor a shared “machine ready for hop” helper) and return the existing password-auth message.

cmd/camp/transfer.go
```diff
@@ -10,6 +12,7 @@ import (
 
 	"github.com/Obedience-Corp/camp/internal/campaign"
 	"github.com/Obedience-Corp/camp/internal/config"
+	"github.com/Obedience-Corp/camp/internal/machines"
 	"github.com/Obedience-Corp/camp/internal/transfer"
 	"github.com/spf13/cobra"
```
obey-agent, line 18: **[suggestion]** Command `Long` / examples / docs still describe only `campaign:path`. Users will not discover `machine:campaign:path`, both-remote rejection, `local:`, or shadow notes from `--help`.

**Suggestion:** Update `Use`/`Long`/`Example` (and `docs/cli-reference/camp_transfer.md` if that ships with the PR series) to document the machine-first form, one-remote-only rule, and `local:` escape hatch.

cmd/camp/transfer.go
obey-agent: **High — local side drops campaign:path resolution**

`ResolveCampaignRelative` only joins against the *current* campaign root. The local transfer path uses `ResolveCrossCampaignPath`, which understands `other-campaign:rel` (and therefore also the post-`local:` Spec form `other:rel`).

Cross-machine forms that still have a campaign-qualified local endpoint currently break:

```text
camp transfer archdtop:obey-campaign:docs/x.md other-campaign:docs/x.md
camp transfer other-campaign:docs/x.md archdtop:obey-campaign:docs/x.md
camp transfer archdtop:obey-campaign:x.md local:other:x.md   # Spec becomes other:x.md
```

Those resolve to a literal `…/other-campaign:docs/x.md` path under the current root instead of a registry lookup. Live verification only exercised bare relative paths, so this would not have shown up.

**Fix:** use `ResolveCrossCampaignPath(ctx, root, localSpec)` (and surface the error), matching the local path. Worth a unit/integration test that pins the three forms above.

cmd/camp/transfer.go
```diff
@@ -182,3 +207,47 @@ func completeCampaignPath(campRoot, pathPrefix, colonPrefix string) ([]string, c
...
+
+	if err := transfer.CopyRemote(ctx, transfer.CopyOptions{
+		Machine:  m,
+		Endpoint: remoteEnd,
+		Local:    localPath,
+		Pull:     pull,
+		Force:    force,
+	}); err != nil {
+		if errors.Is(err, transfer.ErrDestinationExists) {
+			return camperrors.Newf("destination exists on %s, not overwritten (use --force)", m.ID)
```
obey-agent, line 280: **High — no-clobber error always blames the remote machine**

`m.ID` is the far machine for both push and pull. On a pull the destination is local, so:

```text
camp transfer archdtop:camp:x.md ./already-exists.md
# → "destination exists on archdtop, not overwritten"
```

is false: the occupied path is local. Operators (and agents) will dig on the wrong host.

Phrase by direction, e.g. local dest → `"destination %q already exists (use --force)"`, remote dest → keep `"on %s"`. Same honesty bar as the rsync `--itemize-changes` fix in this PR.

cmd/camp/transfer.go
```diff
@@ -182,3 +207,47 @@ func completeCampaignPath(campRoot, pathPrefix, colonPrefix string) ([]string, c
...
+	if pull && transfer.IsDestDir(localPath) {
+		localPath = filepath.Join(localPath, filepath.Base(remoteEnd.Path))
+	}
+	if !pull {
+		if err := transfer.ValidatePathExists(localPath); err != nil {
+			return camperrors.Wrap(err, "source")
+		}
+	}
+
+	if err := transfer.CopyRemote(ctx, transfer.CopyOptions{
```
obey-agent, line 267: **Medium — pull path skips parent `MkdirAll` that local transfer always does**

Local branch (above) does `os.MkdirAll(filepath.Dir(destPath), 0o755)` before `CopyFile`. The remote path hands `localPath` straight to rsync/scp.

`camp transfer archdtop:camp:x.md nested/new/x.md` therefore fails if `nested/new/` does not exist, while the equivalent local transfer would create it. That is an avoidable cross-path inconsistency.

On pull, after resolving `localPath` (and after dir/basename join), `MkdirAll` the parent — or document/refuse nested missing parents explicitly. Prefer matching local behavior.

internal/transfer/remote.go
```diff
@@ -0,0 +1,248 @@
...
+	if opts.Run == nil {
+		opts.Run = execRunner
+	}
+	if opts.LookPath == nil {
+		opts.LookPath = exec.LookPath
+	}
+	if _, err := opts.LookPath("rsync"); err == nil {
+		out, runErr := opts.Run(ctx, "rsync", opts.rsyncArgs(src, dest)...)
+		if runErr == nil {
+			if !opts.Force && !rsyncCopied(out) {
```
obey-agent, line 95: **Medium — empty itemize ⇒ `ErrDestinationExists` is an inference, not a detection**

The live defect fix is real and welcome: rsync exits 0 on `--ignore-existing` skip, so success must be observed. Mapping *any* empty itemize run to "destination exists" is still a bit loose:

- itemize format varies slightly by rsync version/locale
- other no-op conditions (if flags grow later) would get the same operator-facing error
- callers rephrase it as deliberate no-clobber, which should only fire when that is actually why nothing moved

Not a merge-blocker given current argv (`-a -s --no-links --ignore-existing` only), but consider treating empty itemize as a distinct internal outcome (`ErrNothingTransferred`) and mapping to the no-clobber message only when `--ignore-existing` was set — or assert a positive skip signal if you can find one that is stable across the rsync versions you support.

internal/transfer/remote.go
obey-agent, line 112: **Medium — missing-binary message can blame the wrong machine**

`LookPath("rsync")` is **local**. This branch is also hit when *this* machine has no rsync (never attempted) and no scp. The message always says `rsync not found on <remote>`, which sends the operator to install on the wrong host.

Track whether rsync was skipped for "not on PATH here" vs "remote signature said missing", and phrase accordingly:

- local missing: `rsync not found locally…`
- remote signature: `rsync not found on <id>…`

obey-agent (commented):
## Verdict: **Request changes**

Cross-machine `camp transfer` is close: the machine-first grammar is measured (compat matrix first), transport reuses the existing hop/ssh option surface, rsync-first with a tightly gated scp fallback is the right amendment to D005, and the live-found no-clobber honesty bug was fixed for the rsync path and then correctly extended to scp. What blocks a clean ship is local-side path resolution and operator-facing error honesty on the pull path—both integration gaps that unit tests of the transport layer cannot see.

## Overview

PR #508 (stacked on #507) adds:

1. **Grammar** — `machine:campaign:path`, gated on a *registered* machine id; `local:` escape hatch; shadow note when head is both machine and campaign.
2. **Transport** — far root via remote's own `camp` (`remote.ResolveRoot`); rsync over `peer.SSHCommandFor` / shared `remote.Opts`; scp fallback only on exit-12 + signature; stall timeout not wall-clock; no probe folded into resolve stdout.
3. **Honesty** — `--itemize-changes` so skipped no-clobber is not reported as `Transferred`; scp check-then-copy for the same contract.

Scope limit (no recursive cross-machine dirs) is explicit and acceptable.

## Key Findings

### High

1. **`runRemoteTransfer` resolves the local endpoint with `ResolveCampaignRelative` instead of `ResolveCrossCampaignPath`** (`cmd/camp/transfer.go:228`). Campaign-qualified local sides (`other-campaign:…`, and Spec after `local:other:…`) become literal paths under the current campaign root. Live tests used bare relative paths only, so this slipped. Same resolver as the local branch, please, plus a pin test.

2. **`ErrDestinationExists` is always rephrased as `destination exists on <remote>`** (`cmd/camp/transfer.go:246`). On pull the destination is local; blaming the remote sends operators (and agents) to the wrong host. Phrase by direction.

### Medium

3. **Pull skips the parent `MkdirAll` local transfer always does** — nested local dests that local transfer would create fail under remote pull. Match local behavior after path resolution.

4. **Empty rsync itemize ⇒ `ErrDestinationExists`** is a good fix for the known lie, but still an inference. Fine with today's argv; consider a distinct internal outcome if the flag set grows.

5. **Missing-binary error can blame the remote for a local `LookPath("rsync")` miss** when scp is also absent (`remote.go` ~100). Track why rsync was not used.

6. **Help / examples / shell completion still teach only `campaign:path`.** Parser and product moved; discovery did not. Complete machine ids and add one remote example (noted in summary only—line not in the diff hunk for inline).

### Low

7. Non-reachability failures always wrapped with `camp machine diagnose …` — helpful default, occasionally misleading (perm denied / missing file).
8. Success glyph: local `→` vs remote `->` (and remote prints `src.Spec`/`dest.Spec`, local prints raw args)—cosmetic.
9. Shadow check uses `GetByName` only while path resolution uses `Get` (id/prefix/name)—edge case if a machine id collides with a campaign id/prefix rather than name.
10. Cross-machine directory sources are "not supported" by scope but not refused; rsync `-a` may still move tree-shaped data. An early `IsDir` refuse would make the scope limit enforceable.

## What's Done Well

- **Compat matrix written first, then parser changed** — "nothing else moved" is measured. That is the right discipline for a grammar change.
- **Machine-gated additive behavior** — no `machines.yaml` ⇒ old semantics. Correct product instinct.
- **No second auth surface** — `SSHCommandFor` / `remote.Opts` / ControlMaster reuse is structural, not aspirational.
- **ResolveRoot ownership** — far path never computed locally; avoids the probe-corrupts-stdout trap called out in the design.
- **Fallback signature discipline** — exit 12 alone is not enough; tests pin that so scp does not hide real failures.
- **Scp no-clobber follow-up** — probe failures ≠ permission to overwrite; `--force` skips the probe; tests cover push/pull/force/failed-probe. That is how you close a review finding.
- **Shadow note wired in production** (`ParseEndpointDefault` + stderr), with a regression test that `Shadowed` is reachable—not only injectable in unit tests.
- **Comments explain *why*** (stall vs wall-clock, no-delete invariant, check-then-copy not exclusion). That ages well.

## Staff Standard

This is staff-quality design and transport work with review-grade test attention on the hard edges (fallback signature, scp clobber, shadow wiring). It is not yet staff-complete on the **CLI integration boundary**: the remote path reimplemented local endpoint resolution with a weaker helper and a direction-blind error string. Fix the two Highs (and ideally parent `MkdirAll` + binary-missing phrasing), add a couple of pins that go through `runRemoteTransfer`'s local path resolution without needing a live fleet, and this is merge-ready from a staff bar.

**Suggested retest after fix**

```text
# same machine, two local campaigns + one remote form (mock ResolveRoot / Run)
camp transfer archdtop:camp:x.md other-campaign:x.md   # must registry-resolve local dest
camp transfer archdtop:camp:x.md ./exists.md           # error must NOT say "on archdtop"
camp transfer archdtop:camp:x.md nested/new/x.md       # parent created or clear error
```

PR #509, opened 2026-07-26
feat(machine): report reverse reachability in diagnose

lancekrogers:
> **Stacked on #508** (chain: #504 → #505 → #506 → #507 → #508). Base is `mesh-transfer-machine-first`.

Adds a REVERSE row to `camp machine diagnose`, so when a hop-back fails you can tell whether the problem is on this machine.

Festival MN0001, phase 005 sequence 03.

## The honest-reporting choice

Whether archdtop can open a connection to this mac depends on three things: archdtop's client config, the tailnet ACL, and this machine's sshd. Camp can only observe the **last one** without executing on the far side, which D003 rules out.

So this reports **acceptability** ("does this machine accept inbound ssh") and says so, rather than claiming a reachability verdict it has not earned. Getting that framing wrong would produce a green check that means nothing.

## Causes are separated because their fixes are

| Platform | State | Reported |
| --- | --- | --- |
| darwin | Remote Login readable and off | names Remote Login and where to enable it |
| darwin | state unreadable, no sshd | "no sshd is listening" — deliberately **not** "Remote Login is off" |
| linux | no sshd | names `systemctl enable --now sshd` |
| any | sshd up, tailscale down | capable, names `tailscale up` |
| any | sshd up, tailscale up | capable, and says the rest is keys plus ACL, which camp does not manage |

Row two is the one worth having: unreadable state must not become a confident wrong claim. It is asserted as a *forbidden* substring rather than a hoped-for one.

## D003 is enforced by a test, not by intention

`TestReverseHintsNeverOfferToFixAnything` sweeps every hint for offer-to-manage phrasing ("camp will", "automatically", …). Every hint names exactly one action for the operator to take, and camp never takes it.

## Live output

```
$ camp machine diagnose archdtop
ID  archdtop
HOST  archdtop.tail37114b.ts.net
AUTH  Tailscale SSH (identity) (tailscale-ssh)
SOCKET  none · ~/.obey/ssh-ctl/archdtop.sock
PROBE  ssh -o BatchMode=yes lance@archdtop.tail37114b.ts.net true
CAMP  unavailable (unreachable, or camp missing / too old for 'version --json')
REVERSE  ✓ sshd is listening here. Whether a given machine may connect is that machine's keys plus your tailnet ACL, which camp does not manage
HINT  Tailscale SSH: if hop fails, look for a login.tailscale.com check URL ...
```

## Notes

`just gate` green: 3805. Probes are local and non-privileged (a loopback dial rather than a process scan, because what matters is whether something *accepts* connections, which is the question a peer's ssh client asks), bounded at 2s, with cancellation tested. JSON fields are additive with `omitempty` on the hint.

One thing you might push on: the REVERSE row is identical for every machine, since it is a fact about *this* machine. It repeats per row anyway because that is where you are already looking when a hop-back to a specific machine fails. A single summary line would be less repetitive but would separate the answer from the question.

### DIFF
--- a/cmd/camp/machine.go
+++ b/cmd/camp/machine.go
@@ -348,6 +348,12 @@ type machineDiagnoseRow struct {
 	CampVersion  string `json:"camp_version,omitempty"`
 	CampCommit   string `json:"camp_commit,omitempty"`
 	VersionSkew  bool   `json:"version_skew"`
+	// ReverseCapable reports whether THIS machine accepts inbound ssh, which is
+	// the only half of reverse reachability camp can observe without executing
+	// on the far side. Additive with omitempty on the hint so existing --json
+	// consumers are unaffected.
+	ReverseCapable bool   `json:"reverse_capable"`
+	ReverseHint    string `json:"reverse_hint,omitempty"`
 }
 
 // probeRemoteCampVersion asks the machine's own camp for its version. It is
@@ -414,6 +420,7 @@ func runMachineDiagnose(cmd *cobra.Command, args []string) error {
 	for i := range targets {
 		m := &targets[i]
 		d := remote.CheckControlMaster(ctx, m)
+		reverse := checkReverseReachability(ctx, defaultReverseProbes())
 		remoteVersion, remoteCommit := probeRemoteCampVersion(ctx, m)
 		// Diagnose is the only surface that pays for a live probe, so it is the
 		// only one that can warm the cache the hop path reads.
@@ -432,6 +439,12 @@ func runMachineDiagnose(cmd *cobra.Command, args []string) error {
 			CampVersion:  remoteVersion,
 			CampCommit:   remoteCommit,
 			VersionSkew:  campVersionSkew(localInfo, remoteVersion, remoteCommit),
+			// What this machine can honestly say about being reachable FROM
+			// that one. Identical for every row (it is a fact about here, not
+			// about them), and reported per row because that is where an
+			// operator is already looking when a hop-back fails.
+			ReverseCapable: reverse.Capable,
+			ReverseHint:    reverse.Hint,
 		}
 		if machineDiagnoseReset && d.State == remote.ControlStale {
 			if err := remote.ResetControlMaster(ctx, m); err != nil {
@@ -496,6 +509,13 @@ func renderMachineDiagnoseTable(w io.Writer, rows []machineDiagnoseRow) error {
 				campVersionDisplay(r)+"  ⚠ differs from this machine ("+campLocalVersionDisplay()+"); remote errors may not match current behavior"))
 		default:
 			lines = append(lines, fmt.Sprintf("%s  %s", ui.Label("CAMP"), campVersionDisplay(r)))
+		}
+		if r.ReverseHint != "" {
+			mark := "✗"
+			if r.ReverseCapable {
+				mark = "✓"
+			}
+			lines = append(lines, fmt.Sprintf("%s  %s %s", ui.Label("REVERSE"), mark, r.ReverseHint))
 		}
 		if r.Hint != "" {
 			lines = append(lines, fmt.Sprintf("%s  %s", ui.Label("HINT"), r.Hint))
--- a/cmd/camp/machine.go
+++ b/cmd/camp/machine.go
@@ -416,11 +416,14 @@ func runMachineDiagnose(cmd *cobra.Command, args []string) error {
 	}
 
 	localInfo := version.Get()
+	// Reverse reachability is a fact about this machine, identical for every
+	// row. Probing it inside the loop cost N loopback dials and N tailscale
+	// probes against a 2s budget each to learn the same answer.
+	reverse := checkReverseReachability(ctx, defaultReverseProbes())
 	rows := make([]machineDiagnoseRow, 0, len(targets))
 	for i := range targets {
 		m := &targets[i]
 		d := remote.CheckControlMaster(ctx, m)
-		reverse := checkReverseReachability(ctx, defaultReverseProbes())
 		remoteVersion, remoteCommit := probeRemoteCampVersion(ctx, m)
 		// Diagnose is the only surface that pays for a live probe, so it is the
 		// only one that can warm the cache the hop path reads.
--- a/cmd/camp/machine_reverse.go
+++ b/cmd/camp/machine_reverse.go
@@ -97,12 +97,34 @@ func remoteLoginState(ctx context.Context, p reverseProbes) (on, knowable bool) 
 	return p.RemoteLoginOn(ctx)
 }
 
-// sshdListening dials the loopback ssh port. Deliberately non-privileged and
-// deliberately not a process scan: what matters is whether something is
-// accepting connections, which is the question a peer's ssh client asks.
+// sshdListening reports whether something is accepting ssh here. Deliberately
+// non-privileged and deliberately not a process scan.
+//
+// It dials this machine's own reachable name first, because that is the address
+// a peer's ssh client actually uses. Loopback alone answers a different question
+// and gets both directions wrong: an sshd (or Tailscale SSH) bound only to the
+// tailnet address reads as "not listening", and a sshd_config with
+// ListenAddress 127.0.0.1 reads as "listening" when no peer can reach it.
+//
+// Loopback remains the fallback for when no reachable name can be determined,
+// which is a weaker observation than the caller would like but still better
+// than reporting nothing.
 func sshdListening(ctx context.Context) bool {
+	if host, err := detectReachableName(ctx, runTailscaleStatusForSelf); err == nil && host != "" {
+		if dialSSH(ctx, net.JoinHostPort(host, "22")) {
+			return true
+		}
+		// A reachable name that refuses the connection is the answer: a peer
+		// dialing the same address would be refused too. Do not fall back to
+		// loopback and turn that into a yes.
+		return false
+	}
+	return dialSSH(ctx, "127.0.0.1:22")
+}
+
+func dialSSH(ctx context.Context, addr string) bool {
 	d := net.Dialer{Timeout: time.Second}
-	conn, err := d.DialContext(ctx, "tcp", "127.0.0.1:22")
+	conn, err := d.DialContext(ctx, "tcp", addr)
 	if err != nil {
 		return false
 	}
--- a/cmd/camp/machine_reverse.go
+++ b/cmd/camp/machine_reverse.go
@@ -1,0 +1,122 @@
+package main
+
+import (
+	"context"
+	"net"
+	"os/exec"
+	"runtime"
+	"strings"
+	"time"
+)
+
+// reverseProbeTimeout bounds the local reachability checks. They are local
+// (a loopback dial, one CLI call), so this only guards a wedged binary.
+const reverseProbeTimeout = 2 * time.Second
+
+// reverseReport is what diagnose can honestly say about whether OTHER machines
+// can reach back to this one.
+type reverseReport struct {
+	Capable bool
+	Hint    string
+}
+
+// reverseProbes are the seams the cause matrix is tested through.
+type reverseProbes struct {
+	SSHDListening func(ctx context.Context) bool
+	RemoteLoginOn func(ctx context.Context) (bool, bool) // (on, knowable)
+	TailscaleUp   func(ctx context.Context) bool
+	GOOS          string
+}
+
+// checkReverseReachability reports what this machine can know about being
+// reachable FROM another machine.
+//
+// The honest framing matters, and it is why this reports "acceptability" rather
+// than "reachability". Whether archdtop can actually open a connection to this
+// mac depends on archdtop's client config, the tailnet ACL, and this machine's
+// sshd, and camp can only observe the last one without executing on the far
+// side. So it reports the half it can see and labels it, instead of claiming a
+// verdict it has not earned. D003 also forbids fixing any of it: every hint
+// names one action for the operator to take, and camp never takes it.
+func checkReverseReachability(ctx context.Context, p reverseProbes) reverseReport {
+	ctx, cancel := context.WithTimeout(ctx, reverseProbeTimeout)
+	defer cancel()
+
+	goos := p.GOOS
+	if goos == "" {
+		goos = runtime.GOOS
+	}
+
+	listening := p.SSHDListening != nil && p.SSHDListening(ctx)
+	if !listening {
+		if goos == "darwin" {
+			if on, knowable := remoteLoginState(ctx, p); knowable && !on {
+				return reverseReport{Hint: "macOS Remote Login is off, so nothing can ssh in; " +
+					"enable it in System Settings > General > Sharing > Remote Login"}
+			}
+			return reverseReport{Hint: "no sshd is listening on this machine; enable " +
+				"System Settings > General > Sharing > Remote Login so other machines can hop here"}
+		}
+		return reverseReport{Hint: "no sshd is listening on this machine; start it " +
+			"(for example 'sudo systemctl enable --now sshd') so other machines can hop here"}
+	}
+
+	// sshd answers, so the local half is satisfied. Everything left is policy on
+	// the other side, which is exactly the part camp must not manage.
+	if p.TailscaleUp != nil && !p.TailscaleUp(ctx) {
+		return reverseReport{Capable: true, Hint: "sshd is listening, but tailscale is not up here, " +
+			"so tailnet peers cannot route to this machine; run 'tailscale up'"}
+	}
+	return reverseReport{Capable: true, Hint: "sshd is listening here. Whether a given machine may " +
+		"connect is that machine's keys plus your tailnet ACL, which camp does not manage"}
+}
+
+func remoteLoginState(ctx context.Context, p reverseProbes) (on, knowable bool) {
+	if p.RemoteLoginOn == nil {
+		return false, false
+	}
+	return p.RemoteLoginOn(ctx)
+}
+
+// sshdListening dials the loopback ssh port. Deliberately non-privileged and
+// deliberately not a process scan: what matters is whether something is
+// accepting connections, which is the question a peer's ssh client asks.
+func sshdListening(ctx context.Context) bool {
+	d := net.Dialer{Timeout: time.Second}
+	conn, err := d.DialContext(ctx, "tcp", "127.0.0.1:22")
+	if err != nil {
+		return false
+	}
+	_ = conn.Close()
+	return true
+}
+
+// remoteLoginOn reads macOS Remote Login state. `systemsetup -getremotelogin`
+// needs no privileges for a read, and a failure reports "not knowable" rather
+// than "off", so an unreadable state never produces a confident wrong hint.
+func remoteLoginOn(ctx context.Context) (bool, bool) {
+	out, err := exec.CommandContext(ctx, "systemsetup", "-getremotelogin").Output()
+	if err != nil {
+		return false, false
+	}
+	return strings.Contains(strings.ToLower(string(out)), "on"), true
+}
+
+// tailscaleUpLocally reports whether this machine's tailscale backend is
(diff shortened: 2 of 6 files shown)


cmd/camp/machine.go
```diff
@@ -407,6 +413,7 @@ func runMachineDiagnose(cmd *cobra.Command, args []string) error {
 	for i := range targets {
 		m := &targets[i]
 		d := remote.CheckControlMaster(ctx, m)
+		reverse := checkReverseReachability(ctx, defaultReverseProbes())
```
obey-agent: **Perf nit:** \`checkReverseReachability\` is identical for every row (local fact), but it runs inside the per-machine loop. With a 2s timeout budget and N targets that is N loopback dials + N tailscale/status probes for the same answer.

Hoist once before the loop:

```go
reverse := checkReverseReachability(ctx, defaultReverseProbes())
for i := range targets {
    …
}
```

Matches the PR description ("identical for every row") and keeps diagnose snappy on larger fleets.

obey-agent (commented):
## Verdict: Approve

## Overview
Reverse row on `camp machine diagnose` reports *acceptability* (what this machine can know about inbound ssh) rather than a reachability verdict it has not earned. Cause matrix separates darwin Remote Login, unreadable state, linux sshd, and tailscale-down paths with distinct fixes, and D003 is enforced by a forbidden-phrase test rather than intention. Local non-privileged loopback dial is the right probe.

## Key Findings
- **Perf nit:** reverse check is identical per row but runs inside the machine loop — hoist once. See inline.
- Unreadable Remote Login deliberately does *not* claim "off"; tested as a forbidden substring. Excellent.

## What's Done Well
- Honest framing in comments and UX copy ("camp does not manage").
- Context cancellation covered.
- Additive JSON fields with `omitempty` on the hint.
- Probe seams for full cause matrix without privileged process scans.

## Staff Standard
Yes. Small, careful, and consistent with the mesh diagnostic philosophy. Ship after optional hoist.

cmd/camp/machine_reverse.go
```diff
@@ -0,0 +1,122 @@
...
+
+	goos := p.GOOS
+	if goos == "" {
+		goos = runtime.GOOS
+	}
+
+	listening := p.SSHDListening != nil && p.SSHDListening(ctx)
+	if !listening {
+		if goos == "darwin" {
+			if on, knowable := remoteLoginState(ctx, p); knowable && !on {
```
obey-agent: **[bug]** When Remote Login is knowably off, the hint says “macOS Remote Login is off, **so nothing can ssh in**.” That absolute claim is false on hosts that accept inbound ssh only via Tailscale SSH: `tailscaled` serves port 22 on Tailscale IPs without system `sshd` / Remote Login, and camp’s hop path is deliberately auth-mode agnostic for transient origins (empty `AuthMethod`; argv converges). The production probe only dials `127.0.0.1:22`, so a Tailscale-SSH-only Mac reports ✗ with this confident wording even though peers can still `ssh user@host.ts.net`. The linux no-sshd arm (“start sshd … so other machines can hop here”) has the same implication. That is exactly the false negative the honest-reporting design set out to avoid: hop-back fails for ACL/keys, diagnose blames local sshd, operator enables Remote Login for the wrong reason (or thinks local half is broken when it is not).

**Suggestion:** Soften absolute language to match what was measured (“no process is accepting on loopback :22”; “classic sshd / Remote Login is off”) and/or treat Tailscale SSH as a second local acceptance signal (e.g. dial Self’s Tailscale IP:22 when `selfDNSName` is non-empty, or read the Tailscale SSH pref). At minimum, when Tailscale is up, do not claim “nothing can ssh in.”

cmd/camp/machine_reverse.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	return p.RemoteLoginOn(ctx)
+}
+
+// sshdListening dials the loopback ssh port. Deliberately non-privileged and
+// deliberately not a process scan: what matters is whether something is
+// accepting connections, which is the question a peer's ssh client asks.
+func sshdListening(ctx context.Context) bool {
+	d := net.Dialer{Timeout: time.Second}
+	conn, err := d.DialContext(ctx, "tcp", "127.0.0.1:22")
```
obey-agent: **[bug]** `sshdListening` dials only `127.0.0.1:22`. Comment says this is “the question a peer’s ssh client asks,” but peers dial the MagicDNS / Tailscale IP (or a LAN address), not loopback. Two concrete misreports follow: (1) **false negative** — Tailscale SSH or `sshd` bound only on non-loopback addresses → not listening; (2) **false positive** — `ListenAddress 127.0.0.1` (or any loopback-only acceptor on :22) → `Capable: true` and green REVERSE while no peer can connect. Non-22 `Port` in sshd_config is also an unhandled false negative (machines have no port field today, so less urgent).

**Suggestion:** Prefer probing a peer-visible address when known (Self Tailscale IP from the same status parse already used by `tailscaleUpLocally`), fall back to loopback, and phrase success as “something accepts TCP on …,” not full reverse verdict. Optionally try both and only mark capable when a non-loopback dial succeeds, or when loopback succeeds *and* Tailscale is up (still imperfect, but closer to hop-back reality).

cmd/camp/machine_reverse.go
```diff
@@ -0,0 +1,122 @@
...
+				"System Settings > General > Sharing > Remote Login so other machines can hop here"}
+		}
+		return reverseReport{Hint: "no sshd is listening on this machine; start it " +
+			"(for example 'sudo systemctl enable --now sshd') so other machines can hop here"}
+	}
+
+	// sshd answers, so the local half is satisfied. Everything left is policy on
+	// the other side, which is exactly the part camp must not manage.
+	if p.TailscaleUp != nil && !p.TailscaleUp(ctx) {
+		return reverseReport{Capable: true, Hint: "sshd is listening, but tailscale is not up here, " +
```
obey-agent, line 86: **[suggestion]** When sshd answers but Tailscale is down, the code returns `Capable: true` with a problem-oriented hint. Render then prints `REVERSE  ✓ … tailnet peers cannot route to this machine`. The PR body owns “capable = local sshd half,” but the row label is REVERSE, the mark is a binary ✓/✗, and JSON exposes `reverse_capable: true` with no separate “tailscale_up” field. Agents or operators that only look at the mark/bool will treat reverse as healthy while hop-back over the tailnet cannot work. That undercuts the “getting that framing wrong would produce a green check that means nothing” claim in the PR description.

**Suggestion:** Use ✗ (or a third mark / neutral `·`) when Tailscale is down; keep `Capable` for pure sshd if needed but rename JSON toward `sshd_accepts` / split `reverse_capable` vs `tailscale_up`; or leave Capable false whenever any local prerequisite for tailnet hop-back fails.

cmd/camp/machine_reverse.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	listening := p.SSHDListening != nil && p.SSHDListening(ctx)
+	if !listening {
+		if goos == "darwin" {
+			if on, knowable := remoteLoginState(ctx, p); knowable && !on {
+				return reverseReport{Hint: "macOS Remote Login is off, so nothing can ssh in; " +
+					"enable it in System Settings > General > Sharing > Remote Login"}
+			}
+			return reverseReport{Hint: "no sshd is listening on this machine; enable " +
```
obey-agent: **[suggestion]** On darwin, if Remote Login is knowably **on** (or the sshd-not-listening path is taken for any reason other than knowably-off), the fallback hint still says “enable … Remote Login.” That is the wrong action when Remote Login is already enabled but nothing answers on :22 (crash, wrong port, startup race, firewall). The matrix tests off and unknowable, but not “on + not listening,” so this wording is unpinned.

**Suggestion:** Branch knowable&&on to a distinct hint (“Remote Login is on but nothing is accepting on :22; check sshd / Sharing / firewall”) and add that case to `TestCheckReverseReachabilityCauseMatrix` with a forbidden “enable … Remote Login” substring.

cmd/camp/machine_reverse.go
```diff
@@ -0,0 +1,122 @@
...
+
+// remoteLoginOn reads macOS Remote Login state. `systemsetup -getremotelogin`
+// needs no privileges for a read, and a failure reports "not knowable" rather
+// than "off", so an unreadable state never produces a confident wrong hint.
+func remoteLoginOn(ctx context.Context) (bool, bool) {
+	out, err := exec.CommandContext(ctx, "systemsetup", "-getremotelogin").Output()
+	if err != nil {
+		return false, false
+	}
+	return strings.Contains(strings.ToLower(string(out)), "on"), true
```
obey-agent: **[suggestion]** `remoteLoginOn` treats any successful `systemsetup -getremotelogin` stdout containing the substring `"on"` as On. That is probably fine for the usual `Remote Login: On/Off` English lines (`"off"` does not contain `"on"`), but it is brittle against localized or multi-line output and does not require a word boundary / known prefix. Failure → unknowable is correct and well tested.

**Suggestion:** Match a tighter pattern (e.g. case-insensitive `remote login:\s*on` as a full result line) and unit-test fixture strings for On, Off, and garbage-success output.

cmd/camp/machine.go
```diff
@@ -490,6 +506,13 @@ func renderMachineDiagnoseTable(w io.Writer, rows []machineDiagnoseRow) error {
 		default:
 			lines = append(lines, fmt.Sprintf("%s  %s", ui.Label("CAMP"), campVersionDisplay(r)))
 		}
+		if r.ReverseHint != "" {
```
obey-agent, line 516: **[suggestion]** REVERSE is omitted from human output only when `ReverseHint == ""`, but every production path of `checkReverseReachability` sets a hint, and there is no unit test that `renderMachineDiagnoseTable` / JSON encoding actually emit REVERSE / `reverse_capable` / `reverse_hint`. Integration `machineDiagnoseRow` in tests/integration still only mirrors id/socket/state/reset, so a regression that drops the fields would not fail CI.

**Suggestion:** Add a small render/JSON test with a fixed `machineDiagnoseRow{ReverseCapable, ReverseHint}` (no live dials), and optionally extend the integration mirror struct if diagnose JSON is considered a contract.

cmd/camp/main.go
```diff
@@ -19,10 +19,20 @@ func main() {
...
+		// redundant.
+		//
+		// This is a type assertion, not errors.As, and the difference is
+		// load-bearing. Remote failures create a CommandError deep in the chain
+		// (remote.Run wraps a non-zero ssh exit), and camp then wraps THAT with
+		// the sentence the operator actually needs: which machine, which
+		// campaign, and which diagnostic to run next. Matching anywhere in the
+		// chain discarded all of it and exited silently with ssh's code, so a
+		// failed hop printed nothing at all.
 		var cmdErr *camperrors.CommandError
```
obey-agent: **[nit]** Comment says the silent-exit rule “is a type assertion, not errors.As,” but the code still uses `errors.As` and then requires `error(cmdErr) == err` so only a *directly returned* `*CommandError` is silent. Behavior is right and well covered by `main_exit_test.go`; the comment disagrees with the implementation and will confuse the next reader. The duplicated `silentExit` helper can also drift from `main`.

**Suggestion:** Prefer `cmdErr, ok := err.(*camperrors.CommandError)` to match the comment, or rewrite the comment to describe “As + top-level identity.” Consider exporting one shared predicate used by both `main` and the test.

cmd/camp/machine_reverse.go
```diff
@@ -0,0 +1,122 @@
...
+// reachable FROM another machine.
+//
+// The honest framing matters, and it is why this reports "acceptability" rather
+// than "reachability". Whether archdtop can actually open a connection to this
+// mac depends on archdtop's client config, the tailnet ACL, and this machine's
+// sshd, and camp can only observe the last one without executing on the far
+// side. So it reports the half it can see and labels it, instead of claiming a
+// verdict it has not earned. D003 also forbids fixing any of it: every hint
+// names one action for the operator to take, and camp never takes it.
+func checkReverseReachability(ctx context.Context, p reverseProbes) reverseReport {
```
obey-agent, line 42: **[nit]** Probe order reports only the first local failure (no sshd → never mentions Tailscale). That matches “one action for the operator,” but when both sshd and Tailscale are down the operator may enable sshd, re-run diagnose, and only then learn about Tailscale. Acceptable for D003, slightly chatty as a multi-step loop.

**Suggestion:** Optional: append a secondary clause when both probes fail (“and tailscale does not look up”) without offering to fix either. Not required if single-action hints stay policy.

obey-agent (commented):
## Summary

Re-review of head `c6815fc` (stacked on #508; prior review was on `a2b266ed`). The PR adds a local-only REVERSE row to `camp machine diagnose` via injectible probes, a cause matrix with D003 “never manage” coverage, and an orthogonal but high-value silent-exit fix in `main` so wrapped remote `CommandError`s print again. Framing is mostly careful: hints name operator actions, unreadable macOS Remote Login does not become “off,” and probes are noninteractive (loopback dial, `systemsetup -getremotelogin`, `tailscale status --json`) with a 2s bound and a cancellation smoke test.

Dominant residual risk is **over-claiming the negative path**: several hints and the REVERSE mark/bool assert “inbound ssh is impossible / the local half is fine” based only on classic `127.0.0.1:22`, which misses Tailscale SSH (a first-class camp auth mode) and can green-check a loopback-only listener. JSON `reverse_capable: true` while Tailscale is down is intentional per the PR body but still easy for both humans (✓) and machines (bool alone) to misread.

## Issue counts by severity

- bugs: 2
- suggestions: 4
- nits: 2

cmd/camp/machine_reverse.go
obey-agent: **Medium — misleading fix when Remote Login is already on**

This branch is taken for every non-listening darwin path that is *not* `knowable && !on` — including `knowable && on` (Remote Login reports on, but nothing accepts on the probed port).

In that case the operator is told to *enable* Remote Login even though the probe already established it is on. The careful “unreadable must not claim off” distinction is right; this sibling case still steers them at the wrong fix.

Suggested matrix extension (and a forbidden substring on that row):

```go
// knowable && on && !listening
Hint: "no sshd is listening on this machine (Remote Login reports on); check that sshd is running and accepting on port 22"
// must not contain: "enable" … "Remote Login"
```

Also worth a test row: `darwin, Remote Login on, no sshd` — currently uncovered while the off/unreadable cases are pinned.

cmd/camp/machine_reverse.go
obey-agent, line 86: **Suggestion — green check with a remaining local problem**

`Capable: true` here is intentional (sshd is the half camp claims to observe), but the human row is driven purely by that flag:

```text
REVERSE  ✓ sshd is listening, but tailscale is not up here…; run 'tailscale up'
```

Operators scanning for ✓/✗ will read green as “reverse is fine” and skip the sentence. For the partial-local-failure case, consider either:

1. `Capable: false` when tailscale is down (capability for *tailnet* reverse hop), or
2. a third mark (`⚠`) when sshd is up but the tailnet path is not,

so the scan path and the prose agree. As written, the prose is honest; the glyph undercuts it.

cmd/camp/machine_reverse.go
obey-agent: **Nit — fragile Remote Login parse**

`strings.Contains(..., "on")` works for the classic `Remote Login: On` / `Off` strings today (Off does not contain `on`), but it is a substring match against the entire command output. Prefer an exact-state parse, e.g. lower/trim and check for `"remote login: on"` / `"remote login: off"`, and treat anything else as `knowable=false`. That keeps the same “never confidently claim off” contract without relying on accidental non-overlap with `off`.

cmd/camp/machine_reverse.go
obey-agent: **Suggestion — port 22 / loopback-only probe is narrower than the hop path**

Dialing only `127.0.0.1:22` is a deliberate non-privileged “does something accept” check, and it matches the default hop-back path (`ssh-agent` → classic sshd). Two residual edges worth naming in the hint or a comment so a future reader does not treat this as full reverse reachability:

1. **Non-22 `Port` / IPv6-only bind** — sshd up but not on this exact endpoint reports not-listening and points at Remote Login / systemctl, which may already be correct.
2. **Tailscale SSH without system sshd** — inbound can work via `tailscaled` with nothing on `:22`. Camp’s default adopt/hop-back is still ssh-agent, so this is out of scope for the default reverse path, but the absolute phrasing (“nothing can ssh in”) overclaims for a Tailscale-SSH-only host.

Not a merge blocker given the default auth path; just keep the wording at the same honesty bar the PR already holds for ACL/keys.

obey-agent (commented):
## Verdict: Approve with comments

## Overview

PR #509 (head `775d0eab`, base `mesh-transfer-machine-first`, stacked on #508) adds a **REVERSE** row to `camp machine diagnose` that reports the only half of reverse reachability camp can honestly observe without executing on the far side: whether **this** machine accepts inbound SSH. It also carries the #510 silent-exit fix (direct `*CommandError` only) extracted into production `silentExit` in `cmd/camp/exit.go`, so production and tests cannot drift.

The design choice is the right one for a diagnostic surface: **acceptability, not reachability**; cause matrix separated by fix; D003 (“diagnose, never manage”) enforced by a forbidden-phrase test; reverse probed **once** per diagnose, not per row; JSON additive with `omitempty` on the hint.

## Key Findings

### Medium
1. **Darwin “enable Remote Login” when it is already on** (`machine_reverse.go` ~57–58) — the fallback non-listening hint always tells the operator to enable Remote Login for every path that is not `knowable && !on`, including `knowable && on`. That steers them at the wrong fix when Remote Login reports on but nothing is accepting on the probed port. Add a matrix row + distinct hint; pin a forbidden “enable … Remote Login” substring on that row. (inline)

### Suggestions
2. **✓ glyph vs partial failure** (`Capable: true` when tailscale is down) — prose is honest; the green check undercuts it for scan readers. Consider `Capable: false` or a `⚠` third state so mark and sentence agree. (inline)
3. **Loopback `:22` only** — correct for default hop-back (`ssh-agent` → classic sshd); residual edges are non-22 `Port`, IPv6-only bind, and Tailscale-SSH-without-sshd. Absolute “nothing can ssh in” overclaims slightly in those edges; keep wording at the same honesty bar already used for ACL/keys. (inline)

### Nits
4. **`remoteLoginOn` substring `"on"`** — works for classic On/Off output today; prefer exact-state parse and `knowable=false` on anything else. (inline)
5. **Test gap** — no row for `darwin, Remote Login on, no sshd`; cancel test only asserts non-empty hint (acceptable but weak).

## What's Done Well

- **Honest product framing** written into code comments, PR body, and operator-facing copy: camp does not pretend to know ACL/keys/peer config.
- **Cause matrix is test-driven** with a *forbidden* substring for the unreadable-darwin case — that is the difference between “we hope” and “we pin.”
- **D003 is a test, not a promise** (`TestReverseHintsNeverOfferToFixAnything`).
- **Perf fix already landed in the PR**: reverse probed once outside the per-machine loop after earlier review feedback.
- **`silentExit` extraction** is staff-grade: type assertion is load-bearing, documented, production-owned, and covered by five cases including wrapped remote failures that previously printed nothing.
- **Context**: 2s bound, `DialContext`, cancel-safe path; non-privileged loopback dial rather than process scan — right question for a peer SSH client.
- **Integration**: reuses `runTailscaleStatusForSelf` / `selfDNSName` from hop-origin rather than inventing a parallel tailscale parse.
- Stack hygiene: merge from main brought #510 in cleanly; reverse fields remain additive for `--json` consumers.

## Staff Standard

**Meets staff standard with minor follow-ups.** The core judgment (report only what this machine can know, label it, never manage it) is production-grade and better documented than most diagnose features. The one real correctness gap in operator guidance is the darwin on-but-not-listening mis-hint; the glyph/partial and port-scope notes are consistency polish, not architectural rewrites. Ship after addressing or consciously accepting finding #1; #2–#4 can land as follow-ups without blocking the stacked MN0001 chain.

**Inline comments:** 4 on `cmd/camp/machine_reverse.go` (lines 57, 67, 86, 102).

PR #510, opened 2026-07-26
fix(cli): stop discarding remote error messages

lancekrogers:
> **Independent of the MN0001 PR chain** (#504–#509). Based directly on `main`, small, and mergeable on its own. Worth taking first: it affects every remote command today.

A failed remote hop prints **nothing**. Not a vague message, nothing: exit 127 or 255 with empty stdout and empty stderr.

```
$ camp switch archdtop:lance-arch
$ echo $?
127
```

## Cause

`main()` propagates a child process's exit code without printing a redundant message. That is right for `camp sync`, `camp doctor`, `camp clone`, and `camp run`, which return a `CommandError` **directly** and have already said their piece.

But it matched with `errors.As`, which searches the whole chain. Remote failures create a `CommandError` deep inside (`remote.Run` wraps a non-zero ssh exit), and camp then wraps *that* with the sentence the operator actually needs: which campaign, which machine, and the escape hatch. `errors.As` found the inner error and threw the outer message away.

**The message was being built correctly all along and discarded before anyone could read it.**

## Fix

Scope the silent path to a `CommandError` that IS the returned error, rather than one anywhere in its chain. One condition, and the direct-return commands are unaffected.

```
before:  $ camp switch archdtop:lance-arch
         (nothing, exit 127)

after:   $ camp switch archdtop:lance-arch
         Error: could not resolve "lance-arch" on archdtop: remote camp not found on
         archdtop (tried "camp" via sh -lc, i.e. the machine's login-shell PATH); if
         camp lives outside that PATH, set CAMP_REMOTE_CAMP_PATH to its exact path on
         that machine: command "ssh lance@archdtop..." exited with code 127: sh: line 1:
         camp: command not found
```

That `CAMP_REMOTE_CAMP_PATH` hint has existed in the code the whole time. No user could ever have seen it.

## Tests

`TestSilentExitOnlyForDirectlyReturnedCommandErrors`, five rows: direct return still propagates silently; a wrapped remote failure prints; double-wrapping does not re-enable the silent path; a zero exit code (launch failure) prints; a plain error prints.

`just gate` green: 3740.

## How it was found

The MN0001 live fleet gate, which exists precisely to catch what unit suites cannot. Every test in the repo passed with these messages being discarded, because no test asserted that a human ever sees them. It surfaced the moment a real hop failed on a real second machine.

### DIFF
--- a/cmd/camp/main.go
+++ b/cmd/camp/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -11,7 +10,6 @@ import (
 	// bginit must initialize before the bubbletea subtree under the command
 	// packages; its path keeps it first under gofmt.
 	_ "github.com/Obedience-Corp/camp/internal/bginit"
-	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 )
 
 func main() {
@@ -19,21 +17,8 @@ func main() {
 	defer stop()
 
 	if err := Execute(ctx); err != nil {
-		// A command that returns a CommandError DIRECTLY is propagating a
-		// child process's exit code (camp sync, doctor, clone, run), and has
-		// already said whatever needed saying; printing again would be
-		// redundant.
-		//
-		// This is a type assertion, not errors.As, and the difference is
-		// load-bearing. Remote failures create a CommandError deep in the chain
-		// (remote.Run wraps a non-zero ssh exit), and camp then wraps THAT with
-		// the sentence the operator actually needs: which machine, which
-		// campaign, and which diagnostic to run next. Matching anywhere in the
-		// chain discarded all of it and exited silently with ssh's code, so a
-		// failed hop printed nothing at all.
-		var cmdErr *camperrors.CommandError
-		if ok := errors.As(err, &cmdErr); ok && cmdErr.ExitCode != 0 && error(cmdErr) == err {
-			os.Exit(cmdErr.ExitCode)
+		if code, silent := silentExit(err); silent {
+			os.Exit(code)
 		}
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 		os.Exit(1)
--- a/cmd/camp/main.go
+++ b/cmd/camp/main.go
@@ -19,10 +19,20 @@ func main() {
 	defer stop()
 
 	if err := Execute(ctx); err != nil {
-		// If a child process exited with a specific code (e.g. camp run),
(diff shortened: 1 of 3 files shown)


cmd/camp/main.go
```diff
@@ -19,10 +19,20 @@ func main() {
...
+		// This is a type assertion, not errors.As, and the difference is
+		// load-bearing. Remote failures create a CommandError deep in the chain
+		// (remote.Run wraps a non-zero ssh exit), and camp then wraps THAT with
+		// the sentence the operator actually needs: which machine, which
+		// campaign, and which diagnostic to run next. Matching anywhere in the
+		// chain discarded all of it and exited silently with ssh's code, so a
+		// failed hop printed nothing at all.
 		var cmdErr *camperrors.CommandError
-		if errors.As(err, &cmdErr) && cmdErr.ExitCode != 0 {
+		if ok := errors.As(err, &cmdErr); ok && cmdErr.ExitCode != 0 && error(cmdErr) == err {
```
obey-agent: **Nit (clarity):** the comment says this is a type assertion, not \`errors.As\`, but the code still uses \`errors.As\` plus pointer identity. Functionally correct today, but a future reader will trust the comment over the code.

Prefer the form that matches the comment (and cannot drift):

```go
if cmdErr, ok := err.(*camperrors.CommandError); ok && cmdErr.ExitCode != 0 {
    os.Exit(cmdErr.ExitCode)
}
```

Also consider calling the shared \`silentExit\` helper from \`main\` so the test cannot drift from production.

obey-agent (commented):
## Verdict: Approve (with a small clarity nit)

## Overview
Correct, minimal fix for a real operator-facing defect: remote failures were building the right error message and then discarding it because `main` treated any `CommandError` in the chain as a silent exit. The condition is now "direct return only," which matches the original intent for `sync`/`doctor`/`clone`/`run` and restores messages for wrapped remote failures. The five-row decision table is exactly the regression suite this needed.

## Key Findings
- **Nit:** Comment claims a type assertion but the code still uses `errors.As` + pointer identity. Prefer `err.(*camperrors.CommandError)` (or call a shared `silentExit` from `main`) so comment, test, and production cannot diverge. See inline.

## What's Done Well
- Root-cause analysis is precise and production-shaped (found by live fleet gate, not unit suite).
- Fix is one condition change with clear comments explaining *why* `errors.As` was wrong here.
- Tests pin direct silent exit, wrapped print, double-wrap, zero exit, and plain errors.

## Staff Standard
Yes — I would merge this after the tiny clarity cleanup (or even as-is; the behavior is right). Worth landing first on `main` independent of the mesh stack.

PR #511, opened 2026-07-26
feat(stageguard): size and bulk staging guards, artifact root gitignore handling (CA0004 phase 001)

lancekrogers:
Implements phase 001 of festival `camp-artifact-commit-guardrails-CA0004`, from the WI-9ad89e design package (`workflow/design/camp-artifact-commit-updates-2026-07-25/`).

**Draft:** sequences 01 and 02 of 4 are complete. 03 (wiring the guard into `internal/git.Stage`) and 04 (doctor checks and notices) land on this branch as work continues.

## What this adds

**`internal/stageguard`** — the guard's detection core, as a leaf package.

- `ResolveLimits(ctx, repoPath)` self-loads thresholds, modes, and the allowlist from campaign and per-project config, so the staging chokepoint needs no config knowledge. This is what lets fest inherit the guard from a version bump with no fest code change.
- `CheckStaging(ctx, repoPath, limits)` enumerates what a stage-everything operation would add and returns typed violations. Stat-only: nothing is opened, read, or hashed.
- Per-file threshold (10 MiB campaign root / 25 MiB project), bulk detection (1000+ untracked files under a dominant prefix, deliberately **no** total-bytes leg), LFS exemption in every mode, allowlist globs.
- Per-file guard runs first; bulk counts only what remains, so one auto-handled 512 MB video cannot swell the batch back over the bulk trigger.
- Tracked files are never violations of the exclusion kind. Over-threshold tracked growth is a distinct `ReportOnly` kind: splitting one file's history between git and an artifact root is never correct.

It imports only `internal/errors` and `internal/config` from camp, and never `internal/git` — `internal/git` will import it in sequence 03, so the reverse edge would be a cycle. That constraint is enforced by a test rather than a one-time manual check.

**`pkg/commitkit`** — the guard surface re-exported for consumers outside the module. Type aliases, not new named types, so fest sees exactly the types the guard returns and camp never converts. `commitkit.StageAll` already existed at the shape the design specified; its godoc now states what a caller inherits for free.

**`camp artifacts add`** — acts on what it already knew instead of warning.

- Clean root: appends the ignore rule and states the consequence (content there lives outside git and syncs with `camp sync`).
- Mixed root: declaration saved, `.gitignore` never touched, tracked/untracked split reported.
- `--dry-run` previews file count, byte total, and the three-way split without writing anything.
- `--no-gitignore` opts out and retains the existing warning.

**Consolidation.** `isGitignored`/`hasTrackedFiles` moved into `internal/artifacts` so the command, doctor, and staging exclusion share one implementation. Gitignore rule editing moved to `internal/fsutil` so camp has one writer; `internal/scaffold` now delegates to it. `ui.FormatBytes`/`ui.FormatCount` added and `cmd/camp/cache` pointed at the former, which also fixes a 5 GB root rendering as "5632.0 MB".

## Notes for review

Three bugs were caught and fixed during the festival's own review gates; each is recorded with its reasoning in the sequence `05_iterate.md` documents.

- **Hardcoded `projects/` and `worktrees/`** in scope resolution. Both are campaign-configurable via `.campaign/settings/jumps.yaml`, so a campaign that renames them would have had every per-project guard override silently keyed to the wrong name. Now reads `config.CampaignPaths`.
- **`--dry-run` skipped validation.** Surveying before declaring moved the work ahead of `File.Add`, where path and policy validation live, so `camp artifacts add ../outside --dry-run` would have printed a confident report for a declaration that could never succeed.
- **Mixed-root report overstated the artifact set.** The design's `(5.5 GB)` qualifies the untracked files; the implementation printed the root's total, including tracked bytes.

One deviation from the design worth flagging: design doc 04 states `commitkit.StageAll` "does not exist yet". It does, at `pkg/commitkit/commitkit.go:164`, in exactly the specified shape — the repo moved since the doc was written. No duplicate was added.

## Verification

- `just gate` green: 3,820 tests (dev profile), 3,781 (stable), 0 lint issues in both profiles.
- Containerized integration suite: 570/570.
- 6 new containerized tests covering acceptance criteria 22, 23, and 24, with fixtures built in-container.
- Porcelain parsing verified against real `git status --porcelain=v1 -z -uall` output rather than only against authored strings — `-z` emits paths verbatim where the default format escapes them, which is why the parser asks for it.

Guard coverage is currently pure-logic only. `Enumerate`, `CheckStaging`, `ResolveLimits`, and the LFS filter have no reachable caller until sequence 03 wires them into `internal/git.Stage`; they get end-to-end coverage through `camp commit` there.

## Known unrelated issue

`internal/fsutil` flakes under full-gate load (1/3781, passes standalone and at `-count=20`). `lock_test.go` asserts on wall-clock budgets while the gate runs the stable suite at 457s CPU in 41s wall. Pre-existing and not caused by this branch — tracked separately.

### DIFF
--- a/cmd/camp/commit.go
+++ b/cmd/camp/commit.go
@@ -59,6 +59,7 @@ var (
 	commitWorkitem    string
 	commitNoEdit      bool
 	commitLarge       bool
+	commitJSONOut     bool
 )
 
 func init() {
@@ -72,6 +73,7 @@ func init() {
 	commitCmd.Flags().BoolVar(&commitAutoWrite, "auto-write", false, "Run configured commit message writer")
 	commitCmd.Flags().StringVar(&commitWorkitem, "workitem", "", "explicit workitem selector for the commit tag (overrides cwd-based resolution)")
 	commitCmd.Flags().BoolVar(&commitLarge, "commit-large", false, "Commit over-threshold files instead of keeping them out of git")
+	commitCmd.Flags().BoolVar(&commitJSONOut, "json", false, "Emit a JSON result on stdout; human output goes to stderr")
 
 	rootCmd.AddCommand(commitCmd)
 	commitCmd.GroupID = "git"
@@ -100,6 +102,16 @@ func completeProjectFlag(cmd *cobra.Command, args []string, toComplete string) (
 func runCommit(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
 
+	// Under --json, stdout carries exactly one document and nothing else.
+	// Human lines go to stderr rather than being dropped, so a person watching
+	// a scripted run still sees what happened. fest has the interleave bug
+	// this avoids; the point of doing it here is not to recreate it.
+	humanOut := cmd.OutOrStdout()
+	if commitJSONOut {
+		humanOut = cmd.ErrOrStderr()
+	}
+	jsonResult := newCommitJSONResult("")
+
 	// Join repeated -m values git-style before any tag prepending so the tag
 	// lands on the subject line.
 	commitMessage := commitkit.JoinMessages(commitMessages)
@@ -117,7 +129,7 @@ func runCommit(cmd *cobra.Command, args []string) error {
 	}
 
 	if target.IsSubmodule {
-		fmt.Println(ui.Info(fmt.Sprintf("Operating on submodule: %s", target.Name)))
+		_, _ = fmt.Fprintln(humanOut, ui.Info(fmt.Sprintf("Operating on submodule: %s", target.Name)))
 	}
 
 	if commitAutoWrite && commitMessage != "" {
@@ -153,7 +165,7 @@ func runCommit(cmd *cobra.Command, args []string) error {
 
 	// Stage if requested
 	if stageAll {
-		fmt.Println(ui.Info("Staging changes..."))
+		_, _ = fmt.Fprintln(humanOut, ui.Info("Staging changes..."))
 		var guardOutcome *git.StageOutcome
 		if target.IsSubmodule || commitIncludeRefs {
 			outcome, err := git.StageWithGuardOptions(ctx, target.Path, nil, git.StageOptions{CommitLarge: commitLarge})
@@ -185,10 +197,11 @@ func runCommit(cmd *cobra.Command, args []string) error {
 
 		// Act on what the guard decided and tell the user. Runs after staging
 		// so the declaration it may write lands in this same commit.
-		handling, err := cmdutil.HandleStageOutcome(ctx, cmd.OutOrStdout(), campRoot, target.Path, guardOutcome)
+		handling, err := cmdutil.HandleStageOutcome(ctx, humanOut, campRoot, target.Path, guardOutcome)
 		if err != nil {
 			return err
 		}
+		jsonResult.applyGuardHandling(handling)
 		// The excluded file is still an unstaged worktree change, so the
 		// ordinary has-changes check would say yes and git would then reject
 		// the empty commit. Report the no-op instead of failing.
@@ -197,13 +210,13 @@ func runCommit(cmd *cobra.Command, args []string) error {
 			return err
 		}
 		if empty && !commitAmend {
-			cmdutil.ReportNothingLeftToCommit(cmd.OutOrStdout())
-			return nil
+			cmdutil.ReportNothingLeftToCommit(humanOut)
+			return commitJSONNoop(cmd, jsonResult)
 		}
 	}
 
 	// Show what will be committed
-	cmdutil.ShowStagedSummary(ctx, target.Path)
+	cmdutil.ShowStagedSummaryTo(humanOut, ctx, target.Path)
 
 	// Refuse root content commits that would accidentally sweep pre-staged
 	// submodule gitlinks into this commit's message/tag context. The user can
@@ -224,12 +237,12 @@ func runCommit(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	if !hasChanges && !commitAmend {
-		fmt.Println(ui.Success("Nothing to commit, working tree clean"))
-		return nil
+		_, _ = fmt.Fprintln(humanOut, ui.Success("Nothing to commit, working tree clean"))
+		return commitJSONNoop(cmd, jsonResult)
 	}
 
 	if commitAutoWrite {
-		fmt.Println(ui.Info("Writing commit message..."))
+		_, _ = fmt.Fprintln(humanOut, ui.Info("Writing commit message..."))
 		var hookErr error
 		extraEnv := workitemEnvForCommit(ctx, campRoot, commitWorkitem)
 		extraEnv = commitkit.WithCommitAmendEnv(extraEnv, commitAmend)
@@ -255,7 +268,7 @@ func runCommit(cmd *cobra.Command, args []string) error {
 	}
 
 	// Perform commit
-	fmt.Println(ui.Info("Committing changes..."))
+	_, _ = fmt.Fprintln(humanOut, ui.Info("Committing changes..."))
 	opts := &git.CommitOptions{
 		Message: message,
 		Amend:   commitAmend,
@@ -270,13 +283,13 @@ func runCommit(cmd *cobra.Command, args []string) error {
 					return camperrors.Wrap(driftErr, "check unstaged submodule refs")
 				}
 				if len(driftRefs) > 0 {
-					fmt.Println(ui.Warning("Nothing to commit (submodule ref changes are excluded by default)"))
-					fmt.Println(ui.Dim("  Use 'camp refs-sync' to commit only the submodule pointers."))
-					fmt.Println(ui.Dim("  Use 'camp commit --include-refs -m \"...\"' to include them in this commit."))
+					_, _ = fmt.Fprintln(humanOut, ui.Warning("Nothing to commit (submodule ref changes are excluded by default)"))
+					_, _ = fmt.Fprintln(humanOut, ui.Dim("  Use 'camp refs-sync' to commit only the submodule pointers."))
+					_, _ = fmt.Fprintln(humanOut, ui.Dim("  Use 'camp commit --include-refs -m \"...\"' to include them in this commit."))
 					return nil
 				}
 			}
-			fmt.Println(ui.Success("Nothing to commit"))
+			_, _ = fmt.Fprintln(humanOut, ui.Success("Nothing to commit"))
 			return nil
 		}
 		return err
@@ -290,8 +303,26 @@ func runCommit(cmd *cobra.Command, args []string) error {
 			CommitEvidence(ctx, ledgerkit.Scope{Workitem: workitemRef, Festival: festivalRef}, campRoot, target.Path, sha, message)
 	}
 
-	fmt.Println(ui.Success("Changes committed successfully"))
+	_, _ = fmt.Fprintln(humanOut, ui.Success("Changes committed successfully"))
+
+	if commitJSONOut {
+		jsonResult.Repo = target.Path
+		if err := jsonResult.finalize(ctx, target.Path); err != nil {
+			return err
+		}
+		return jsonResult.emit(cmd.OutOrStdout())
+	}
 	return nil
+}
+
+// commitJSONNoop emits the document for a run that legitimately produced no
+// commit, so a machine caller always receives one document per invocation
+// rather than having to treat "no output" as a distinct outcome.
+func commitJSONNoop(cmd *cobra.Command, result *commitJSONResult) error {
+	if !commitJSONOut {
+		return nil
+	}
+	return result.emit(cmd.OutOrStdout())
 }
 
 func worktreesRelPath(ctx context.Context, campRoot string) string {
--- a/cmd/camp/commit.go
+++ b/cmd/camp/commit.go
@@ -58,6 +58,7 @@ var (
 	commitAutoWrite   bool
 	commitWorkitem    string
 	commitNoEdit      bool
+	commitLarge       bool
 )
 
 func init() {
@@ -70,6 +71,7 @@ func init() {
 	commitCmd.Flags().BoolVar(&commitIncludeRefs, "include-refs", false, "Include submodule ref changes when staging at campaign root")
 	commitCmd.Flags().BoolVar(&commitAutoWrite, "auto-write", false, "Run configured commit message writer")
 	commitCmd.Flags().StringVar(&commitWorkitem, "workitem", "", "explicit workitem selector for the commit tag (overrides cwd-based resolution)")
+	commitCmd.Flags().BoolVar(&commitLarge, "commit-large", false, "Commit over-threshold files instead of keeping them out of git")
 
 	rootCmd.AddCommand(commitCmd)
 	commitCmd.GroupID = "git"
@@ -154,7 +156,7 @@ func runCommit(cmd *cobra.Command, args []string) error {
 		fmt.Println(ui.Info("Staging changes..."))
 		var guardOutcome *git.StageOutcome
 		if target.IsSubmodule || commitIncludeRefs {
-			outcome, err := git.StageWithGuard(ctx, target.Path, nil)
+			outcome, err := git.StageWithGuardOptions(ctx, target.Path, nil, git.StageOptions{CommitLarge: commitLarge})
 			if err != nil {
 				return err
 			}
@@ -166,7 +168,7 @@ func runCommit(cmd *cobra.Command, args []string) error {
 			if pathErr != nil {
 				return pathErr
 			}
-			outcome, err := git.StageAllExcludingWithGuard(ctx, target.Path, paths)
+			outcome, err := git.StageAllExcludingWithGuardOptions(ctx, target.Path, paths, git.StageOptions{CommitLarge: commitLarge})
 			if err != nil {
 				return err
 			}
--- a/cmd/camp/commit.go
+++ b/cmd/camp/commit.go
@@ -152,10 +152,13 @@ func runCommit(cmd *cobra.Command, args []string) error {
 	// Stage if requested
 	if stageAll {
 		fmt.Println(ui.Info("Staging changes..."))
+		var guardOutcome *git.StageOutcome
 		if target.IsSubmodule || commitIncludeRefs {
-			if err := executor.StageAll(ctx); err != nil {
+			outcome, err := git.StageWithGuard(ctx, target.Path, nil)
+			if err != nil {
 				return err
 			}
+			guardOutcome = outcome
 		} else {
 			// Campaign root: exclude submodule refs to prevent accidental
 			// ref changes from polluting content commits.
@@ -163,9 +166,11 @@ func runCommit(cmd *cobra.Command, args []string) error {
 			if pathErr != nil {
 				return pathErr
 			}
-			if err := git.StageAllExcluding(ctx, target.Path, paths); err != nil {
+			outcome, err := git.StageAllExcludingWithGuard(ctx, target.Path, paths)
+			if err != nil {
 				return err
 			}
+			guardOutcome = outcome
 			// git add rejects exclude pathspecs whose target contains
 			// gitignored entries, so worktrees are unstaged after staging
 			// instead of excluded up front.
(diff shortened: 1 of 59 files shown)


obey-agent:
Two issues were found while verifying this branch. Both are pre-existing and neither is caused by these changes, but both affected how this branch was verified, so recording them here.

- #512 — `internal/fsutil` lock tests flake under `just gate` load. Seen twice while gating this branch: 1/3781 failed in the stable profile, while the same package passes standalone and at `-count=20`. The tests assert on wall-clock budgets (200ms, 500ms) and the gate runs the stable suite at 457s CPU in 41s wall. This branch's only addition to that package is pure string comparison with no timing.
- #513 — the containerized suite reported 474/572 failures on its first run here, then 570/570 on an identical re-run. The failing run bailed at 142s where the passing run took 471s, which is container-pool exhaustion rather than a code fault.

The three defects found in this branch's own code were fixed before push rather than filed: hardcoded `projects/`/`worktrees/` in scope resolution, `--dry-run` bypassing path and policy validation, and the mixed-root report counting tracked bytes as part of the artifact set. Each is described with its reasoning in the corresponding sequence's `05_iterate.md` in the festival, and summarized in the PR description.

pkg/commitkit/guard.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// The staging guard's public surface, re-exported for consumers outside this
+// module. fest is a separate module and cannot import camp's internal/
+// packages, so this file is the contract it codes against; only the
+// implementation lives behind it.
+//
+// Almost none of this needs calling. The guard runs inside git.Stage, which
+// every camp staging path funnels through, and it resolves its own thresholds
+// from repoPath. A consumer that stages through Stage, StageAll, StageFiles,
```
obey-agent, line 17: **commitkit inherits silent auto-exclusion.**

`StageAll` / `Stage` discard `StageOutcome` and return only `error`. Under default `large_files: auto`, over-threshold untracked files are excluded with `nil` error — a fest/consumer caller has no typed signal that bytes never entered the index.

That undercuts the package contract above ("inherits the guard … with no code change") and the design rule that automatic handling is only safe when it reports what it did. `CheckStaging` preflight is insufficient for race-free reporting after the fact, and `GuardBlockedError` (block mode) also lives only in `internal/git`, so fest cannot `errors.As` it either.

Before this leaves draft: export a staging outcome / blocked-error surface through commitkit (`StageAllWithOutcome`, typed `GuardBlockedError` alias, or equivalent), or document that consumers must never use bare `StageAll` if they care about exclusion visibility — and update the godoc accordingly.

pkg/commitkit/commitkit.go
```diff
@@ -161,6 +161,11 @@ func LoadCampaignName(ctx context.Context, campaignRoot string) (string, error)
 
 // StageAll stages all changes in the repository at repoPath.
 // Uses automatic lock retry with stale lock cleanup.
+//
+// It takes no limits by design. The staging guard runs inside the staging
+// chokepoint and resolves its own thresholds from repoPath, so a caller
+// inherits large-file and bulk protection here without passing anything and
+// without a code change. Use CheckStaging only to preflight.
 func StageAll(ctx context.Context, repoPath string) error {
```
obey-agent, line 175: **Same gap at the StageAll entry point.**

Godoc says callers inherit large-file protection for free, but they also inherit silent exclusion: this function never returns what was kept out. Pair any "inherits protection" claim with a non-silent API (or an explicit "reporting is the caller's preflight" contract and a fest migration note).

internal/stageguard/check.go
```diff
@@ -0,0 +1,273 @@
...
+		TotalBytes: total,
+	}, true
+}
+
+// dominantPrefix finds the deepest directory holding at least
+// dominantPrefixShare of the batch, and returns it with its file count and
+// total size. It returns an empty prefix when the batch is spread across
+// directories, which is the ordinary case that must not block.
+func dominantPrefix(candidates []Candidate) (prefix string, count int, total int64) {
+	threshold := int(float64(len(candidates)) * dominantPrefixShare)
```
obey-agent: **Bulk false-negative under mixed untracked work.**

`detectBulk` requires both (a) total untracked ≥ MaxAddedFiles and (b) a dominant prefix holding ≥ 90% of the *batch*. A realistic case fails:

- `vendor_tree/` 1000 files + 120 other untracked → total 1120, share threshold 1008, vendor has 1000 → no block.

Users often drop a vendor tree while also having misc untracked files. Prefer triggering when **any directory prefix contains ≥ MaxAddedFiles untracked files** (absolute count under that prefix). Keep the 90% / deepest-prefix logic for *naming* the reported path if useful, not as a second gate that can cancel a clear vendor dump.

Add a unit case: 1000 under one prefix + ~12% stragglers still blocks.

internal/git/stageguard.go
```diff
@@ -0,0 +1,273 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// runStageGuard evaluates the guard for a staging operation and returns both
+// the outcome and the pathspec exclusions to apply.
+//
+// It is deliberately fail-closed on refusals and fail-open on its own errors:
```
obey-agent: **Comment/code mismatch: fail-open vs fail-closed.**

Comment claims fail-open on the guard's own errors so a unreadable repo remains stageable. Implementation returns wrapped errors from both `ResolveLimits` and `CheckStaging` (lines 157–158, 165–166), and `StageWithGuardOptions` aborts staging on any non-nil error.

Either implement fail-open for non-refusal errors (refuse only `GuardBlockedError`; log/skip other guard failures and stage unguarded), or rewrite the comment to match fail-closed. As written, the safety-net rationale is aspirational and will confuse the next reader (and fest) about intended behavior under config/git failures.

Note: fail-closed on *malformed* guard config is reasonable and already argued in `ResolveLimits`; the mismatch is specifically "git unavailable / not a repo / check-attr failed".

cmd/camp/commit.go
```diff
@@ -151,21 +165,26 @@ func runCommit(cmd *cobra.Command, args []string) error {
...
 	// Stage if requested
 	if stageAll {
-		fmt.Println(ui.Info("Staging changes..."))
+		_, _ = fmt.Fprintln(humanOut, ui.Info("Staging changes..."))
+		var guardOutcome *git.StageOutcome
 		if target.IsSubmodule || commitIncludeRefs {
-			if err := executor.StageAll(ctx); err != nil {
+			outcome, err := git.StageWithGuardOptions(ctx, target.Path, nil, git.StageOptions{CommitLarge: commitLarge})
+			if err != nil {
 				return err
```
obey-agent: **`--json` refusal path only on campaign-root branch.**

Submodule / `--include-refs` returns `err` raw. Campaign-root path calls `reportGuardRefusalJSON`. A machine caller doing `camp commit --json` inside a project submodule on a bulk/size block gets empty stdout + non-zero exit — exactly the interleave/empty-stdout bug this PR fixed for the root path.

Route both branches through `reportGuardRefusalJSON` (or hoist the call after the if/else).

cmd/camp/main.go
```diff
@@ -25,6 +26,13 @@ func main() {
 		if errors.As(err, &cmdErr) && cmdErr.ExitCode != 0 {
 			os.Exit(cmdErr.ExitCode)
 		}
+		// A staging-guard refusal is the one moment this package spends the
+		// user's attention, so it prints the full report (what was found,
+		// that nothing was staged, and the ways forward) in place of the
+		// one-line error, rather than in addition to it.
+		if cmdutil.RenderGuardRefusal(os.Stderr, err, "camp commit") {
```
obey-agent: **Hardcoded remedy command name.**

Always renders `camp commit --commit-large` even when the user ran `camp p commit` or `camp worktrees commit` (those paths wrap and still hit this). The remedy then names a command that may not match flags/context the user was already in.

Plumb the actual invocation (cobra command path, or per-entry-point constant) into `RenderGuardRefusal`.

cmd/camp/worktrees/commit.go
```diff
@@ -178,3 +182,20 @@ func runWorktreesCommit(cmd *cobra.Command, args []string) error {
...
+// it excludes, says what it left out, and offers the three ways forward. The
+// reporting is shared with `camp commit` and `camp p commit` so the same
+// situation never reads differently depending on which command found it.
+func stageWorktreeWithGuard(ctx context.Context, cmd *cobra.Command, campRoot, worktreePath string) error {
+	outcome, err := git.StageWithGuardOptions(ctx, worktreePath, nil,
+		git.StageOptions{CommitLarge: wtCommitLarge})
+	if err != nil {
+		return err
+	}
+	_, err = cmdutil.HandleStageOutcome(ctx, cmd.OutOrStdout(), campRoot, worktreePath, outcome)
```
obey-agent: **Path parity: missing empty-after-exclude handling.**

`camp commit` and `camp p commit` call `GuardExcludedEverything` + `ReportNothingLeftToCommit` so a worktree that only held over-threshold files does not fall through to a generic "Nothing to commit". This helper reports the exclusion then returns, leaving the caller to hit `HasChanges` / `ErrNoChanges` with weaker copy — and it drops the handling value entirely (`_`).

Mirror the project-commit path: return handling (or check empty here) so worktree commit messaging stays consistent with the rest of the guard surface.

cmd/camp/settings_cli.go
```diff
@@ -495,11 +523,66 @@ func applyCampaignScalarKey(cfg *config.CampaignConfig, key, value string) (stri
...
+		cfg.Commit.Guards.MaxProjectFileSize = trimmed
+	}
+	return trimmed, nil
+}
+
+// applyGuardCountKey validates the bulk trigger count.
+func applyGuardCountKey(cfg *config.CampaignConfig, key, value string) (string, error) {
+	trimmed := strings.TrimSpace(value)
+	n, err := strconv.Atoi(trimmed)
+	if err != nil || n < 0 {
```
obey-agent, line 563: **`max_added_files: 0` is accepted then silently ignored.**

`applyGuardCountKey` allows `n == 0`, but `applyGuardsConfig` only overlays when `n > 0`, so the shipped default (1000) remains. A user who sets 0 intending "disable" or "always fire" gets neither — they need `bulk: off`.

Reject 0 with a validation message pointing at `bulk: off`, or treat 0 as an explicit disable. Silent no-ops on safety knobs are the kind of config bug this PR is careful to avoid elsewhere (invalid modes error rather than fall back).

obey-agent (commented):
## Verdict: **Request changes** (draft — not merge-ready)

**PR:** #511 · **Draft:** yes · **Head:** `1a1b216da13fb574a2cf9e0cf1a91d822b0369c0` · **Branch:** `feat/stage-guards` → `main`

Staff-level review of stageguard / bulk+size guards / artifact-root gitignore handling (CA0004 phase 001). Draft status noted; full review applied anyway. Body says sequences 01–02 done and 03–04 still landing — the tree already contains Stage chokepoint wiring, doctor checks, notices, and `--json`, so the description is behind the branch.

---

## Overview

This is a large, carefully designed safety system for camp's stage-everything default: a leaf `internal/stageguard` package (stat-only enumeration, LFS + allowlist exemption, per-file before bulk), chokepoint attachment in `internal/git.Stage*`, campaign-root auto-declare + project exclude-and-ask, `camp artifacts add` gitignore/dry-run behavior, doctor artifacts/bigfiles, status notices with committed dismissals, and `camp commit --json` with excluded/declared fields.

Architecture choices are generally staff-grade: import-cycle avoidance enforced by test, funnel test against bare `git add`, typed violations instead of string parsing, bookkeeping staged into the same commit, mixed-root never blanket-gitignored, tracked growth report-only. Several bugs the festival already caught (jumps-aware scope, dry-run validation order, mixed-root byte accounting) show real discipline.

The draft is **not** just missing polish — there are correctness and consumer-contract gaps that should land before this is treated as launch-ready for fest inheritance.

---

## Key Findings

### Critical / high

1. **`commitkit` silently inherits auto-exclusion (consumer contract hole).**  
   `StageAll`/`Stage` discard `StageOutcome`. Under default `large_files: auto`, over-threshold files are excluded with a nil error. Fest cannot see what was kept out without a preflight race, and `GuardBlockedError` is not re-exported through `pkg/commitkit` so block mode is opaque too. The godoc promise of "inherit the guard with no code change" is only half true: protection yes, reporting no — and the design's own rule is that automatic handling is only safe when it reports. Export `StageAllWithOutcome` / typed blocked error (or rewrite the contract and give fest an explicit migration).

2. **Bulk detector false-negative when other untracked work coexists.**  
   Requiring a prefix to hold ≥90% of the *entire* untracked batch means `vendor/` with 1000 files + ~12% miscellaneous untracked does **not** block. Real agent/human sessions often have stragglers. Trigger on absolute count under any prefix (≥ `MaxAddedFiles`); use dominance only for which path to name. Missing unit case.

### Medium

3. **`runStageGuard` documents fail-open, implements fail-closed.**  
   Resolve/CheckStaging errors abort staging. Align comment with code, or actually fail-open for non-refusal errors while keeping malformed config fail-closed.

4. **`camp commit --json` refusal document only on campaign-root branch.**  
   Submodule / `--include-refs` returns raw error → empty stdout on block. Hoist `reportGuardRefusalJSON`.

5. **`RenderGuardRefusal` hardcodes `"camp commit"`.**  
   `camp p commit` / `camp worktrees commit` remedies name the wrong command.

6. **Worktree commit path incomplete vs project/root.**  
   No `GuardExcludedEverything` / tailored empty message; drops handling.

### Low

7. **`local.commit.guards.max_added_files 0` validates then silently keeps default** (`n > 0` overlay). Reject 0 or map to disable; point users at `bulk: off`.

8. **Draft body drift** — sequences 03–04 appear largely present; refresh PR description before review handoff.

---

## What's Done Well

- **Leaf package + dependency tests** (`deps_test.go`) and **funnel test** (`funnel_test.go`) make the chokepoint claim enforceable rather than documentary.
- **Per-file before bulk** so one auto-handled video cannot re-inflate the bulk batch — correct product logic.
- **Tracked growth never excluded** — avoids bifurcating history; rare clarity about a hard invariant.
- **Mixed vs clean artifact roots** (no blanket gitignore on tracked content) and consolidation of `IsGitignored`/`HasTrackedFiles` into `internal/artifacts`.
- **`--json` schema care** (empty slices not null, real hash, excluded[] visibility) and human output on stderr under `--json`.
- **Notices**: committed per-signature dismissals, never-synced as the justifying case, bigfiles deliberately *not* on every status.
- **bigfiles opt-in** with an explicit cost rationale — correct product judgment.
- **Allowlist / LFS exemptions in every mode**, pathspec composition with submodule excludes, `--commit-large` as inverse of "skip safety".
- Strong unit + containerized integration coverage for the paths that exist; porcelain `-z` parsing justified and tested against real git shape.
- Festival-caught bug fixes recorded in the PR body with reasoning.

---

## Staff Standard

| Bar | Assessment |
| --- | --- |
| Read real execution paths | Guard → Stage → HandleStageOutcome → commit JSON/human; commitkit discard path checked |
| No parallel systems | Consolidates gitignore writer, artifact git state helpers; routes `project/new` through funnel |
| Context / errors / operability | ctx checks present; error typing strong inside camp; **weak across module boundary** |
| Integration over sprawl | Right chokepoint; doctor/notices scoped to what commit cannot see |
| Explicit formats over heuristics | Size/shape decide, not extension lists; allow globs validated |
| Tests | Excellent for logic; still need bulk-straggler case + commitkit outcome API tests |

**Required before ready (even as non-draft):** fix or explicitly scope the commitkit reporting contract; fix bulk absolute-count trigger; close the `--json` submodule refusal hole. Medium items should land in the same phase so all commit entry points behave the same.

**Not blocking conceptually:** phase-002 stubs (`ArtifactRootDrift`, manifest drift skip) are honest and documented.

---

## Recommendation

Keep as **draft**. Address high findings (especially commitkit silence and bulk share gate) before marking ready-for-review. Remaining work is concentrated and well-bounded — the foundation is strong enough that fixing these will leave a mergeable staff-level change.

— obey-agent staff review

obey-agent (changes requested):
Requesting changes on the current ready-for-review head. The eight unresolved inline findings remain current: library callers can silently lose auto-excluded files, the bulk detector misses a qualifying subtree when unrelated untracked files dilute its global share, guard setup/read failures contradict the documented fail-open behavior, JSON refusal output is absent on several paths and hard-codes the command hint, worktree commit does not handle an empty index after exclusion, and max_added_files accepts a zero value that is then ignored. The commits added since the earlier review only merge main and change unrelated CLI exit handling; they do not touch these findings. Focused tests pass: go test ./internal/stageguard ./pkg/commitkit ./cmd/camp/....

obey-agent:
All eight findings addressed. Head is now `bcf2097f`.

**1. Fail-open/fail-closed contradiction** (`internal/git/stageguard.go:140`)
Kept the documented contract and made the code match: the guard is a safety net, not a precondition, so a repository it cannot read is still stageable. Degrading is not silent — `StageOutcome.Unavailable` carries the cause and `HandleStageOutcome` prints it. A quiet fail-open is the genuinely dangerous version, because the user assumes the protection ran.

**2. Bulk false negative** (`internal/stageguard/check.go:124`)
The trigger is now an absolute count rather than a share of the batch. `dominantPrefix` became `heaviestPrefix(candidates, minCount)`; the descent survives only to name the reported path. Whether a directory is being dumped into git does not depend on what else the user left lying around.
Tests: `TestDetectBulkIsNotDilutedByUnrelatedFiles` (the review's 1000 + 120 case) and `TestDetectBulkIgnoresAScatteredBatch` for the inverse.

**3. commitkit contract** (`pkg/commitkit/guard.go`, `commitkit.go`)
Added `StageAllWithOutcome` and `StageFilesWithOutcome`, plus `StageOutcome` and `GuardBlockedError` aliases. The godoc on the plain forms now says what they discard. `StageFilesWithOutcome` keeps the empty-list error rather than letting it become stage-everything.
Test: `TestStageWithOutcomeIsAvailableToConsumers`.

**4. Guard error not routed under --json** (`cmd/camp/commit.go:173`)
Both branches now return through `reportGuardRefusalJSON`, hoisted after the if/else. A machine caller previously got an actionable document from one branch and a bare non-zero exit from the other.

**5. Worktree commit dropped the excluded-everything value** (`cmd/camp/worktrees/commit.go:199`)
`stageWorktreeWithGuard` returns it and the caller reports "nothing left to commit". Previously a worktree holding only over-threshold files was told its tree was clean while the user looked at files they had just created.

**6. Refusal named the wrong command** (`cmd/camp/main.go:28`)
`Execute` uses `ExecuteContextC`, which returns the dispatched command, and `main` renders `ExecutedCommandPath()`. Someone who ran `camp p commit` is no longer told to rerun `camp commit --commit-large`, which points at a different repository.

**7. Bulk count of 0 silently kept the default** (`cmd/camp/settings_cli.go:557`)
Rejected, with the message pointing at `commit.guards.bulk: off`. Storing 0 reported success while the overlay left 1000 in force, so the setting read as applied and was not.

**8.** Inline threads read before each fix; wording followed.

## Verification

- `just gate` green: 3856 tests (gate-fast), 3811 (gate).
- `GOOS=windows GOARCH=amd64 go build ./...` clean.
- Findings 1, 2 and 3 verified by negative control — the invariant deleted, the test re-run, confirmed failing. Worth flagging: my first version of the dilution test passed with the share rule restored, because the straggler fixtures did not set `Untracked` and `detectBulk` filters on it, so the dilution never reached the code under test. Fixed, and the control now fails correctly.

Not approving; over to you for the merge. Remember `#518` is stacked on this branch — retarget it to `main` before deleting `feat/stage-guards`.

obey-agent:
Verification complete on head bcf2097f. All eight findings from the review are confirmed fixed in code: the guard fail-opens with a reported degradation (StageOutcome.Unavailable) instead of a lying comment, the bulk trigger is an absolute per-prefix count so the dilution case blocks, commitkit exports StageAllWithOutcome and StageFilesWithOutcome plus a typed GuardBlockedError alias, the JSON refusal path now covers both staging branches, worktree commit mirrors the GuardExcludedEverything handling, the refusal remedy carries the real command name, and max_added_files: 0 is rejected with a pointer at bulk: off.

Independently re-ran the gate (green) and the Windows build (clean) on the stacked tree containing this head.

Merge-ready pending Lance. Merge order: this PR first, and do not delete feat/stage-guards until #518 has been retargeted to main, or GitHub will close it.

Issue #513, opened 2026-07-26
Integration suite reports mass false failures when the container pool is exhausted


lancekrogers:
Running the full containerized suite on a loaded machine can fail the large majority of tests for reasons unrelated to the code under test. The failure is loud and total, which makes it easy to misread as a real regression in whatever branch happens to be checked out.

## Observed

First run, on `feat/stage-guards` (#511):

```
║ 1 suites                        98/572 tests passed  142.68s ║
║                    ✗ 474/572 TESTS FAILED                    ║
```

Individual failures pass in isolation:

```
$ just test integration-run TestCp_NormalDirCopyWorks
--- PASS: TestCp_NormalDirCopyWorks (3.48s)
ok      github.com/Obedience-Corp/camp/tests/integration  72.487s
```

An immediate re-run of the identical suite, unchanged:

```
║                     ✓ ALL 570 TESTS PASSED                   ║
║                       Total Time: 471.44s                    ║
```

The signal is the timing. The failing run bailed at **142s** where the passing run took **471s** — it did not run the tests slowly, it gave up on acquiring containers. The failure set is broad and unrelated (dungeon moves, commit tags, project remove, cp guards, shell init), which is the shape of a harness fault rather than a code fault.

## Why it matters

The failure mode inverts the normal debugging instinct. A branch author sees 474 failures, assumes their change broke something fundamental, and starts bisecting. Ruling it out took a re-run plus separately disproving the one plausible mechanism by which that branch could have touched many tests at once (a new `CampaignConfig` field altering every serialized `campaign.yaml` — it does not; yaml.v3 omits the zero struct).

## Suggested direction

Distinguish "container could not be acquired" from "test assertion failed", and report the former as a harness error that fails the run with a distinct message rather than as N test failures. Failing fast with `container pool exhausted` would have made the first run self-explanatory. Bounding suite concurrency against available resources, or retrying acquisition with backoff, would reduce how often it happens at all.

## Environment

macOS, colima. Suite was run immediately after several consecutive `just gate` invocations, so the machine had little headroom.

lancekrogers:
Closing as completed. This was fixed in #514 (`test: make lock tests load-tolerant and container faults legible (fixes #512, #513)`), merged in commit 43e1d477 on 2026-07-28 and confirmed on origin/main.

The fix shipped exactly the direction this issue suggested:

[...]

PR #514, opened 2026-07-27
test: make lock tests load-tolerant and container faults legible (fixes #512, #513)

### LINKED ISSUE #512: internal/fsutil lock tests flake under just gate load, making gate-green unreliable
`just gate` intermittently reports one failure in `internal/fsutil` while the same package passes standalone and under repeat runs. Because gate-green is the precondition for cutting a camp release, an unreliable gate is worse than a slow one: the honest response to a red gate is to re-run it, and that trains everyone to re-run rather than read.

## Observed

Seen twice in one session, in the stable profile:

```
║ internal/fsutil  ✗ 1 failed              3.42s                               ║
║                            ✗ 1/3781 TESTS FAILED                             ║
```

Not reproducible in isolation:

```
$ go test ./... -count=1                       # clean
$ go test ./internal/fsutil/ -count=20         # ok, 5.854s
$ go test ./internal/fsutil/ -count=10 -cpu=1,2,8   # ok, 8.693s
$ just test unit                               # clean on 3 of 4 runs
```

## Cause

`internal/fsutil/lock_test.go` asserts against wall-clock budgets:

- `:38` — 200ms context timeout
- `:156` — 50ms sleep before asserting a waiter is still blocked
- `:163` — 500ms channel wait
- `:109` — 2s timeout

`just gate` runs both-profile builds, both lint profiles, the dev suite, and the stable suite back to back. The stable run measured **457s CPU in 41s wall**, so the machine is saturated when these budgets are being measured. A 200-500ms budget is not a safe margin under that load, and the failure rate should rise on slower or busier machines (and in any future CI runner).

## Suggested direction

Replace wall-clock assertions with synchronization where the test is really about ordering — a channel or barrier proving the waiter blocked, rather than "it was still blocked 50ms later". Where a timeout genuinely is the behavior under test (`TestAcquireFileLock_TimeoutIsCategorized`), the assertion should be on the error category, which it already is; the budget just needs enough headroom that load cannot flip it.

## Not caused by

Found while running the gate on `feat/stage-guards` (#511). Not introduced there: that branch's only addition to the package is `gitignore_test.go`, pure string comparison with no timing, goroutines, or filesystem access.

lancekrogers:
Fixes #512 and #513. Both are test-infrastructure faults that made `just gate` and the integration suite unreliable signals, found while gating CA0004 phase 001.

Test-only changes; no production code is touched.

## #512 — lock tests measured the scheduler, not the lock

`just gate` intermittently reported one failure in `internal/fsutil` while the same package passed standalone and at `-count=20`. Reproduced under CPU contention:

```
$ for i in $(seq 1 12); do (yes > /dev/null &) ; done
$ go test ./internal/fsutil/ -count=60 -cpu=1,2
--- FAIL: TestAcquireFileLock_StaleLockStealRaceAcquiresSerially (2.00s)
```

The failing test races two goroutines to steal one stale lock and requires both to acquire, each under a 2-second context. But the property under test is that they acquire **serially** — so the loser's wait is as long as the winner's entire critical section. The budget was standing in for "eventually" while actually asserting that the scheduler ran both goroutines promptly. `just gate` runs the stable suite at roughly 450s of CPU in 40s of wall time, which is exactly when that assumption breaks.

`TestAcquireFileLock_ContextCancellationWhileWaiting` had the same shape: a 50ms sleep hoping its goroutine had reached the wait loop, then a 500ms budget for the return.

Changes:

- The steal race and the cancellation test now use a shared `lockLivenessBudget` (60s), documented as a bound against hanging rather than an assertion about speed. It costs nothing in the normal case — both complete in milliseconds — and reaching it means genuinely stuck.
- The cancellation test signals from inside its goroutine instead of sleeping in the parent, so the race is removed rather than out-waited.

Verified with the same load that reproduced the failure: 60 iterations across two CPU profiles, clean.

## #513 — infrastructure faults reported as hundreds of test failures

A run on an unrelated branch reported 474 of 572 tests failing. The same suite, unchanged, then passed 570/570. The tell was timing: the failing run gave up at 142s where the passing run took 471s. It had not run tests slowly, it had failed to acquire containers.

The mechanism: when the Docker daemon is out of headroom, `Reset()` fails on every pooled checkout, and each failure surfaces through `t.Fatalf` as an ordinary test failure. The natural reading of 474 failures is that the branch broke something fundamental, and ruling that out costs a full re-run plus separate investigation.

Changes:

- `Reset()` gets one retry, which absorbs a transient blip.
- A second failure records a run-level infrastructure fault and every subsequent checkout fails immediately against it, so the run stops early instead of producing the same fault N more times.
- The message is unmistakable and names the likely cause and remedy:

```
INFRASTRUCTURE FAILURE (not a test failure): could not reset a pooled container: ...

The container pool could not be used. Common cause: the Docker daemon is out of
headroom, often because several suites or gates are running at once. Re-run the
suite on an idle machine, or lower CAMP_TEST_POOL_SIZE.
```

This does not make acquisition more reliable, which is not in the harness's gift. It makes the failure legible: one clear message instead of hundreds of plausible ones.

## Why this is worth landing on its own

Both defects cost real time during CA0004 phase 001 — #512 was hit three times, each costing a gate cycle to disambiguate from a real regression, and #513 cost roughly twenty minutes of investigation on a branch where nothing was wrong. A flaky gate is worse than a slow one, because the honest response to a red gate becomes "run it again" rather than "read it", and gate-green is the precondition for cutting a camp release.

Separate from #511 so it can merge independently.

## Verification

- `just gate` green: both build profiles, 0 lint issues in each, both unit suites.
- `internal/fsutil` at `-count=60 -cpu=1,2` under twelve competing CPU hogs: clean.
- Integration spot-check green; the harness change is inert on a healthy run.

### DIFF
--- a/internal/fsutil/lock_test.go
+++ b/internal/fsutil/lock_test.go
@@ -12,6 +12,17 @@ import (
 	camperrors "github.com/Obedience-Corp/camp/internal/errors"
 )
 
+// lockLivenessBudget bounds tests that must not hang, without asserting how
+// fast the scheduler is.
+//
+// These tests previously used budgets in the 200ms-2s range, which made them
+// fail on a loaded machine while passing at idle: `just gate` runs the stable
+// suite at roughly 450s of CPU in 40s of wall time, and a goroutine can easily
+// wait longer than that for a slot. A flaky gate is worse than a slow one,
+// because the honest response to a red gate becomes "run it again" rather than
+// "read it".
+const lockLivenessBudget = 60 * time.Second
+
 func TestAcquireFileLock_RemovesStaleLock(t *testing.T) {
 	lockPath := filepath.Join(t.TempDir(), "links.yaml.lock")
 	if err := os.WriteFile(lockPath, []byte("orphaned"), 0o644); err != nil {
@@ -106,7 +117,16 @@ func TestAcquireFileLock_StaleLockStealRaceAcquiresSerially(t *testing.T) {
 		i := i
 		go func() {
 			defer wg.Done()
-			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+			// The budget here is a liveness bound, not a performance
+			// assertion. What is under test is that both acquirers succeed
+			// *serially*: one steals the stale lock, and the other waits for
+			// it rather than stealing concurrently. The loser's wait is
+			// therefore as long as the winner's whole critical section, so a
+			// tight budget measures the scheduler rather than the lock. It is
+			// set far above any plausible scheduling delay so that reaching it
+			// means genuinely stuck, and costs nothing in the normal case
+			// where both finish in milliseconds.
+			ctx, cancel := context.WithTimeout(context.Background(), lockLivenessBudget)
 			defer cancel()
 			release, err := AcquireFileLock(ctx, lockPath)
 			errs[i] = err
@@ -148,20 +168,26 @@ func TestAcquireFileLock_ContextCancellationWhileWaiting(t *testing.T) {
 
 	ctx, cancel := context.WithCancel(context.Background())
 	done := make(chan error, 1)
+	waiting := make(chan struct{})
 	go func() {
+		// Signal that the goroutine is scheduled and about to block, rather
+		// than sleeping in the parent and hoping. A sleep makes the test
+		// assert that the scheduler ran this goroutine within 50ms, which is
+		// not the property under test and is exactly what fails under load.
+		close(waiting)
 		_, err := AcquireFileLock(ctx, lockPath)
 		done <- err
 	}()
 
-	time.Sleep(50 * time.Millisecond)
+	<-waiting
 	cancel()
 	select {
 	case err := <-done:
 		if !errors.Is(err, context.Canceled) {
 			t.Fatalf("AcquireFileLock canceled error = %v, want context.Canceled", err)
 		}
-	case <-time.After(500 * time.Millisecond):
-		t.Fatal("AcquireFileLock did not return promptly after context cancellation")
+	case <-time.After(lockLivenessBudget):
+		t.Fatal("AcquireFileLock did not return after context cancellation")
 	}
 }
 
--- a/internal/fsutil/lock_test.go
+++ b/internal/fsutil/lock_test.go
@@ -172,20 +172,30 @@ func TestAcquireFileLock_ContextCancellationWhileWaiting(t *testing.T) {
 	}
 	defer release()
 
+	// Signal from inside the wait path (after a failed tryAcquire, before
+	// select on ctx.Done), not before AcquireFileLock returns. Closing early
+	// let cancel win between schedule and first acquire and collapsed this
+	// into the already-canceled path covered by TestAcquireFileLock_ContextCancellation.
+	waiting := make(chan struct{})
+	var once sync.Once
+	prev := onContendedLock
+	onContendedLock = func() {
+		once.Do(func() { close(waiting) })
+	}
+	t.Cleanup(func() { onContendedLock = prev })
+
 	ctx, cancel := context.WithCancel(context.Background())
 	done := make(chan error, 1)
-	waiting := make(chan struct{})
 	go func() {
-		// Signal that the goroutine is scheduled and about to block, rather
-		// than sleeping in the parent and hoping. A sleep makes the test
-		// assert that the scheduler ran this goroutine within 50ms, which is
-		// not the property under test and is exactly what fails under load.
-		close(waiting)
 		_, err := AcquireFileLock(ctx, lockPath)
 		done <- err
 	}()
 
-	<-waiting
+	select {
+	case <-waiting:
+	case <-time.After(lockLivenessBudget):
+		t.Fatal("waiter never entered the contended wait path")
+	}
 	cancel()
 	select {
 	case err := <-done:
--- a/internal/fsutil/lock_test.go
+++ b/internal/fsutil/lock_test.go
@@ -21,6 +21,12 @@ import (
 // wait longer than that for a slot. A flaky gate is worse than a slow one,
 // because the honest response to a red gate becomes "run it again" rather than
 // "read it".
+//
+// This outer budget coexists with AcquireFileLock's independent
+// defaultLockTimeout (5s in lock.go). A contending waiter can still fail with
+// ErrTimeout from that internal bound while the test context has time left;
+// lengthening only this budget does not extend the production wait. "Stuck"
+// may therefore surface as the 5s timeout rather than this 60s deadline.
 const lockLivenessBudget = 60 * time.Second
 
 func TestAcquireFileLock_RemovesStaleLock(t *testing.T) {
--- a/tests/integration/main_test.go
+++ b/tests/integration/main_test.go
@@ -173,9 +173,12 @@ func GetSharedContainer(t *testing.T) *TestContainer {
 	// A run whose infrastructure has already collapsed should say so once and
 	// stop, rather than reporting the same fault as N test failures. See
 	// failInfrastructure for why this distinction matters.
+	//
+	// The first double-Reset failure uses t.Fatalf (the real infra death).
+	// Later checkouts t.Skip so the summary is 1 fail + N skips, not hundreds
+	// of identical hard fails with a contradictory "Skipped:" banner.
 	if msg := infraFailure(); msg != "" {
-		t.Fatalf("%s\n\nSkipped: the container infrastructure failed earlier in "+
-			"this run. This is not a failure of this test.", msg)
+		t.Skip(msg)
 	}
 
 	c := <-containerPool
--- a/tests/integration/main_test.go
+++ b/tests/integration/main_test.go
@@ -5,6 +5,7 @@ package integration
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"runtime"
 	"strconv"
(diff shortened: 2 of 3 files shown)


internal/fsutil/lock_test.go
```diff
@@ -148,20 +168,26 @@ func TestAcquireFileLock_ContextCancellationWhileWaiting(t *testing.T) {
 
 	ctx, cancel := context.WithCancel(context.Background())
 	done := make(chan error, 1)
+	waiting := make(chan struct{})
 	go func() {
+		// Signal that the goroutine is scheduled and about to block, rather
+		// than sleeping in the parent and hoping. A sleep makes the test
+		// assert that the scheduler ran this goroutine within 50ms, which is
+		// not the property under test and is exactly what fails under load.
+		close(waiting)
```
obey-agent: **Residual scheduling window (non-blocking).** `close(waiting)` runs *before* `AcquireFileLock`, so the parent can cancel while this goroutine is still between the signal and the call. In that path the test collapses into the already-canceled case covered by `TestAcquireFileLock_ContextCancellation`, and never exercises the wait-loop `select` on `ctx.Done()`.

The assertion still holds (returns `context.Canceled`), and this is strictly better than `time.Sleep(50ms)`. If you ever want the name "WhileWaiting" to be load-proof and precise, signal from *inside* the wait path (e.g. a test-only hook/failpoint after the first failed acquire, or hold the lock and only cancel after observing retry), or accept that this test is "cancel is observed promptly once the goroutine is scheduled." Either is fine for this PR — flagging so the residual is intentional.

internal/fsutil/lock_test.go
```diff
@@ -106,7 +117,16 @@ func TestAcquireFileLock_StaleLockStealRaceAcquiresSerially(t *testing.T) {
...
+			// The budget here is a liveness bound, not a performance
+			// assertion. What is under test is that both acquirers succeed
+			// *serially*: one steals the stale lock, and the other waits for
+			// it rather than stealing concurrently. The loser's wait is
+			// therefore as long as the winner's whole critical section, so a
+			// tight budget measures the scheduler rather than the lock. It is
+			// set far above any plausible scheduling delay so that reaching it
+			// means genuinely stuck, and costs nothing in the normal case
+			// where both finish in milliseconds.
+			ctx, cancel := context.WithTimeout(context.Background(), lockLivenessBudget)
```
obey-agent, line 135: **Liveness budget vs internal timeout.** The outer context is now `lockLivenessBudget` (60s), but `AcquireFileLock` still enforces an independent `defaultLockTimeout` of 5s (`lock.go`). Under pathological descheduling, the *loser* can still fail with `ErrTimeout` while the winner holds the lock, even though the context has ~55s left. That is a different failure mode than the original 2s context flake, and your `-count=60` under load suggests it is not biting in practice.

Worth a one-line note in the comment above that "stuck" may surface as the internal 5s timeout rather than this context deadline — otherwise a future reader may lengthen only the context and wonder why flakes at 5s remain. Not a merge blocker.

tests/integration/main_test.go
```diff
@@ -131,12 +170,25 @@ func GetSharedContainer(t *testing.T) *TestContainer {
 		t.Fatal("container pool not initialized - TestMain not called?")
 	}
 
+	// A run whose infrastructure has already collapsed should say so once and
+	// stop, rather than reporting the same fault as N test failures. See
+	// failInfrastructure for why this distinction matters.
+	if msg := infraFailure(); msg != "" {
+		t.Fatalf("%s\n\nSkipped: the container infrastructure failed earlier in "+
```
obey-agent: **Wording nit:** this path uses `t.Fatalf` but the message says `Skipped:`. Operators grepping for skip vs fail, or reading the failure as a true `t.Skip`, will be misled. Prefer something like:

```go
t.Fatalf("%s

Not executed: container infrastructure failed earlier in this run. This is not a product regression in this test.", msg)
```

Or, if you want the go test summary to show 1 fail + N skips instead of hundreds of fails with the same banner, use `t.Skip(msg)` here after the first `failInfrastructure` (keep `t.Fatalf` only at the original Reset double-fault). The current approach is still a large UX win over 474 opaque Reset errors.

tests/integration/main_test.go
```diff
@@ -131,12 +170,25 @@ func GetSharedContainer(t *testing.T) *TestContainer {
...
+
 	c := <-containerPool
 
 	if err := c.Reset(); err != nil {
-		// Return the container so the pool does not leak a slot, then fail.
-		containerPool <- c
-		t.Fatalf("failed to reset container: %v", err)
+		// One retry absorbs a transient blip; a second failure means the
+		// daemon or the container is genuinely gone.
+		if retryErr := c.Reset(); retryErr != nil {
```
obey-agent, line 226: **Design tradeoff (ack, not a blocker).** Double-`Reset` failure on *one* pool member is treated as run-level infrastructure death. That is correct for the Docker-headroom cascade that motivated #513. The residual risk is a single wedged container (corrupt FS, stuck exec) poisoning otherwise-healthy siblings: every later checkout short-circuits with the same INFRASTRUCTURE banner.

Given pool sizes of 2–6 and the documented common cause, this is the right bias. If it ever false-positives in CI, a cheap hardening would be "N distinct members fail Reset before arming `infraReason`" or "only arm after retry fails *and* daemon ping fails." No change required for this PR — the comment block above already explains the intent well; consider one sentence on the single-bad-member case so the next reader does not "fix" it back into per-test silence.

obey-agent (commented):
## Verdict: **Approve**

Staff-level review of test-reliability fixes for #512 / #513. Test-only; production lock and container code paths are untouched. Ship it.

---

## Overview

PR #514 fixes two real signal-quality failures that burned gate cycles during CA0004 phase 001:

1. **#512** — `internal/fsutil` lock tests used short context budgets (2s / 500ms) and a 50ms sleep, which asserted scheduler latency under `just gate` CPU oversubscription rather than lock correctness. Fixed with a shared 60s `lockLivenessBudget` and a channel signal instead of sleep.
2. **#513** — Pooled container `Reset()` failures under Docker headroom pressure cascaded into hundreds of ordinary `t.Fatalf`s. Fixed with one retry, a run-level `infraReason`, and an unmistakable `INFRASTRUCTURE FAILURE` banner so operators stop re-litigating product regressions that never happened.

Both changes attack the right layer (test harness semantics, not production hardening theater) and the PR description is exemplary: reproduction, mechanism, and why legibility matters more than making Docker "more reliable."

---

## Key Findings

### Non-blocking / residual

1. **`waiting` signals before enter, not "in wait loop"** (`lock_test.go:177`)  
   `close(waiting)` then `AcquireFileLock` still allows cancel-before-enter, collapsing into the already-canceled path. Assertion remains correct; name "WhileWaiting" is slightly stronger than the sync. Optional future: failpoint after first failed acquire.

2. **Outer 60s budget ≠ real contending wait bound** (`lock_test.go:129` + `lock.go` `defaultLockTimeout=5s`)  
   A loser waiting on a held lock still dies at the internal 5s timeout. Document that residual so future flake hunts don't only extend the context. Load verification at `-count=60` suggests this is not the current failure mode.

3. **`"Skipped:"` + `t.Fatalf`** (`main_test.go:177`)  
   Contradictory wording. Prefer "Not executed" or switch subsequent checkouts to `t.Skip` so the summary is `1 fail + N skips` instead of hundreds of identical fails. Banner text is still a large win over opaque Reset errors.

4. **Single bad pool member → run-level kill** (`main_test.go:186`)  
   Correct bias for Docker-headroom cascades; residual false-positive if one container is wedged and siblings are healthy. Pool size 2–6 makes this acceptable; optional future threshold is "N members" or daemon ping.

### No bugs found

- Mutex on `infraReason` is correct; first writer wins; double-fail returns the slot to the pool.
- Cleanup registration only after successful checkout — no double-return / leak on the failure path.
- Retry-then-arm is the right order; transient single Reset blip does not poison the run.
- Serial steal test still asserts both succeed (correctness), not latency.

---

## What's Done Well

- **Root-cause quality:** Distinguished "scheduler measurement" from "lock property" and "infra cascade" from "product regression" — the rare diagnosis that pays for itself on the next red gate.
- **Cost model of the 60s budget:** Explicitly "free when healthy, only hits if stuck." That is the correct shape for liveness bounds in unit tests.
- **Operator message design:** Names the class of failure, the likely cause, and the remedy (`CAMP_TEST_POOL_SIZE` / idle machine). Actionable failure text is production UX for the harness.
- **Scope discipline:** Test-only, separated from #511, independent merge. No production code churn for a test signal problem.
- **Comments that encode tribal knowledge** rather than restating the code — future agents/humans will not re-introduce 2s budgets without reading why they failed.

---

## Staff Standard

| Bar | Assessment |
| --- | --- |
| Correctness | Pass — both fixes preserve the property under test |
| Test reliability | Pass — removes known flake classes; residuals documented inline |
| Maintainability | Pass — small, commented, local |
| Production confidence | Pass — no production surface; gate signal quality improves release confidence |
| Over-engineering | Pass — no new framework, one mutex + one const |
| Merge readiness | **Approve** — nits are optional follow-ups, not gates |

**Recommendation:** Merge as-is. Optional drive-bys (wording, one-line defaultLockTimeout note, `t.Skip` for subsequent infra checkouts) can land later if the banner still produces noisy summaries in CI.

---

*Reviewed as obey-agent via campaign `just gh pr` recipes. Head: `1fef30105f9a7f0724cb430eaafec5e4dba62548`.*

PR #515, opened 2026-07-27
docs(machine): the mesh model, reachability matrix, and transfer grammar

lancekrogers:
Closes the documentation half of MN0001's 007_DOCS_CLOSE. Stacked on #509.

## What

- `docs/machine-mesh.md` — origin payload, the consent model behind `camp machine adopt`, the hop-back gesture, self-reference, visibility-without-reachability, and the reachability matrix with a degradation story per cell.
- `docs/transfer.md` — machine-first grammar, the shadowing note and `local:` escape, copy-only semantics, rsync/scp transport and the no-clobber contract, both-remote refusal, skew guidance.
- `camp machine` long help — the model in one paragraph, pointing at both.

## Written against observed behavior, not the specs

Every claim comes from the 006_FLEET_GATE run on a real two-machine fleet at `mesh-gate 775d0eab`. Two findings the specs did not predict are now documented:

**The Tailscale SSH server does not run in sandboxed macOS GUI builds.**

```
$ tailscale set --ssh
The Tailscale SSH server does not run in sandboxed Tailscale GUI builds.
```

So a mac Tailscale-SSHes *out* but cannot accept it *in*; reaching a mac means an OpenSSH key. That cell is in the matrix rather than left for the next person to rediscover.

**A stock `go install` leaves camp off the non-interactive login PATH** on both macOS zsh and Arch bash, so every cross-machine call fails until `CAMP_REMOTE_CAMP_PATH` is set. Documented under degradation, with camp's own error quoted since it names the fix.

## Conventions

No real fleet ids or personal campaigns; examples use `devbox`/`notes` per campaign standards.

## Verification

`just lint` 0 issues, `cmd/camp` tests pass, `camp machine --help` renders correctly.

### DIFF
--- a/cmd/camp/machine.go
+++ b/cmd/camp/machine.go
@@ -72,6 +72,21 @@ hops always use BatchMode (agents never hang on password prompts).
 'camp machine diagnose' reports auth mode, a copy-paste ssh probe, and
 ControlMaster socket state (and can clear a stale socket with --reset).
 
+The mesh model, in one paragraph: hopping exports CAMP_HOP_ORIGIN into the
+remote login shell, a single v1 line naming where the shell came from, which is
+what lets 'camp switch -' return without either machine storing state about the
+other. A payload does not register anything: if the origin is unknown here,
+camp names 'camp machine adopt', which previews and asks, requires a TTY, and
+remembers a decline. Reachability need not be symmetric -- a machine you cannot
+dial still appears in completion, because visibility travels as a snapshot
+pushed during a successful enumerate rather than as a live query. Camp will
+never install a key, register a machine unattended, or claim a host is reachable
+on the strength of tailnet membership alone.
+
+See docs/machine-mesh.md for the reachability matrix (notably: the Tailscale SSH
+server does not run in sandboxed macOS GUI builds, so a mac accepts OpenSSH keys
+instead) and docs/transfer.md for the machine-first transfer grammar.
+
 Run without a subcommand in a terminal to manage the fleet interactively: add,
 discover, edit, and remove machines, and see each one's socket state. The
 subcommands stay the interface for scripts and agents, and remain what a
--- a/docs/machine-mesh.md
+++ b/docs/machine-mesh.md
@@ -196,11 +196,15 @@ hangs:
   This is common on a fresh fleet member: a stock `go install` puts camp in `~/go/bin`,
   which a non-interactive login shell does not have on its PATH.
 - **Version skew.** If a previous `camp machine diagnose` observed a different camp version
-  on the target, hops warn once, without paying for a probe:
+  on the target, every hop warns, reading that cached result rather than paying for a
+  probe:
   ```
-  camp: camp on devbox is v0.9.1, this machine is v0.10.0; features may not match
-  (run 'camp machine diagnose devbox' to re-check)
+  camp: camp on devbox is v0.9.1, this machine is v0.10.0; features may not match (run 'camp machine diagnose devbox' to re-check)
   ```
+  It repeats on each hop by design — the condition is still true, and a warning that
+  appeared once would be missed by whoever hits the mismatch later. It stops when the
+  cache entry expires (12h) or a fresh `camp machine diagnose` finds the versions agree.
+
   The warning says only that the versions *differ*. Camp does not order them, because the
   remote is as likely to be ahead as behind.
 - **Stale visibility.** A pushed snapshot past its TTL is dropped rather than shown, so
--- a/docs/machine-mesh.md
+++ b/docs/machine-mesh.md
@@ -1,0 +1,221 @@
+# The Machine Mesh
+
+Camp can move you between machines the same way it moves you between campaigns:
+`csw devbox:notes` opens a shell in the `notes` campaign on `devbox`. This document
+describes the model that makes that safe, and what each piece degrades to when the
+network or the far machine does not cooperate.
+
+The short version: **camp never registers a machine for you, never installs a key, and
+never claims reachability it has not observed.** Everything below follows from that.
+
+## The fleet file
+
+Machines live in `~/.obey/machines.yaml`:
+
+```yaml
+version: 1
+machines:
+    - id: devbox
+      label: Dev Box
+      host: devbox.example.ts.net
+      auth_method: tailscale-ssh      # or ssh-agent
+      ssh_user: alex
+```
+
+`id` is what you type. `host` is what ssh dials. They are separate so a machine can be
+renamed without rewriting every command you have in muscle memory.
+
+## The origin payload
+
+When you hop, camp exports one environment variable into the remote login shell:
+
+```
+CAMP_HOP_ORIGIN=v1;host=devbox.example.ts.net;user=alex;campaign=notes;id=devbox
+```
+
+A single line, percent-encoded values, `v1` prefix. It answers exactly one question on
+the far side: *where did this shell come from?* That is what makes `csw -` possible
+without either machine keeping state about the other.
+
+Fields:
+
+| Field | Meaning | Absent when |
+| --- | --- | --- |
+| `host` | the origin's reachable name | never |
+| `user` | ssh user on the origin | never |
+| `campaign` | campaign the hop started in | the hop did not start inside a campaign |
+| `id` | the id derived from the origin's own reachable name | the name is empty, or the field was dropped for size |
+
+`campaign` is genuinely optional, and its absence is meaningful. Hop from your home
+directory and `csw -` will refuse, because there is no campaign to return to:
+
+```
+$ csw -
+Error: camp switch -: origin campaign unknown (the outbound hop did not start inside a
+campaign); hop back with 'camp switch <machine>:<campaign>'
+```
+
+That is the payload being honest rather than guessing.
+
+## Consent: adopt is always yours
+
+A payload tells the far machine where you came from. It does **not** register you
+there, and hop-back does not require registration — an unknown origin still works,
+built from the payload.
+
+Registration matters when the hop *fails*. Camp phrases the failure by whether it
+has a fleet row to point you at:
+
+```
+# origin is registered here
+Error: run 'camp machine diagnose devbox' to check reachability: <cause>
+
+# origin is not registered here
+Error: the origin is not registered here, probe it with: ssh -o BatchMode=yes alex@devbox.example.ts.net true; 'camp machine adopt' registers it so future hops know the way back: <cause>
+```
+
+`camp machine adopt` reads the payload, shows a preview, and asks. It requires a TTY:
+the confirm-and-write path refuses non-interactive stdin on purpose, so no script and no
+agent can add a machine to your fleet on your behalf. Declining is remembered, and the
+hint stops.
+
+Adopting a machine you already have is a no-op that says so:
+
+```
+$ camp machine adopt
+devbox.example.ts.net is already in your fleet as "devbox"; nothing to adopt
+```
+
+## The hop-back gesture
+
+`csw -` returns to the origin campaign. There is no history file and no daemon: the
+gesture is stateless by construction, which is why it survives a machine reboot on
+either end and why it cannot drift.
+
+It is registration-*independent*, not fleet-blind. Camp looks for a `machines.yaml` row
+whose host matches the payload's, so your own `ssh_user`, `identity_file`, and
+`auth_method` win when you have registered that machine; with no match it builds a
+transient machine from the payload alone.
+
+If the origin is unreachable, the failure explains and points at the diagnostic rather
+than reporting a bare exit code:
+
+```
+$ csw -
+Error: run 'camp machine diagnose devbox' to check reachability: could not resolve
+"notes" on devbox: ... SSH permission denied (publickey) — check ssh-agent keys
+(`ssh-add -l`), identity_file, remote authorized_keys, and ssh_user …
+```
+
+(Abbreviated; the live classification continues with Tailscale SSH guidance.)
+
+## Self-reference resolves locally
+
+`csw devbox:notes` *on devbox* does not ssh to itself. Camp derives an id from this
+machine's own reachable name — character for character the derivation that fills the
+payload's `id` field — and resolves locally when it matches the selector:
+
+```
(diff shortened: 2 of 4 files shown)


docs/machine-mesh.md
```diff
@@ -0,0 +1,204 @@
...
+That is the payload being honest rather than guessing.
+
+## Consent: adopt is always yours
+
+A payload tells the far machine where you came from. It does **not** register you there.
+If the origin is unknown, camp says so and names the command:
+
+```
+$ csw -
+Error: the origin is not registered here, probe it with:
```
obey-agent: **[bug]** The “Consent: adopt is always yours” example presents `csw -` failing with “the origin is not registered here…” as the normal response when the origin is unknown. Hop-back is registration-independent (`originTarget` builds a transient in-memory machine from the payload and only prefers a fleet row when host matches). The quoted error is produced by `hopBackFailure` only when remote resolve fails *and* the origin is unregistered; if SSH works, hop-back succeeds without adopt. Adopt is for fleet membership / diagnose / future reverse hops, not a precondition for `csw -`.

**Suggestion:** Split the stories: (1) hop-back works from the payload alone; (2) soft hint / failed-unregistered probe text when registration would help; (3) `camp machine adopt` as the explicit consent path. Quote the real unregistered failure shape (`probe it with: <BatchMode line>; 'camp machine adopt' registers it…` as a suffix on a failed resolve), not as the default `csw -` outcome for an unknown origin.

docs/machine-mesh.md
```diff
@@ -0,0 +1,204 @@
...
+```
+$ csw -
+Error: run 'camp machine diagnose devbox' to check reachability: could not resolve
+"notes" on devbox: ... SSH permission denied (publickey) - check ssh-agent keys
+(`ssh-add -l`), identity_file, remote authorized_keys, and ssh_user
+```
+
+## Self-reference resolves locally
+
+`csw devbox:notes` *on devbox* does not ssh to itself. Camp compares both the machine's
```
obey-agent: **[bug]** Self-reference is described as comparing “both the machine's configured host and its detected reachable name against the local machine.” Code (`isSelfMachine` from the origin-hop stack) does one check: `suggestedMachineID(detectReachableName(...)) == selector id`, and only when that id is **not** already in `machines.yaml`. A registered entry deliberately wins and forces a real hop (including a misconfigured self-row), which is tested. The dual host comparison is closer to `isSelfOrigin` (adopt), not switch self-ref.

**Suggestion:** Document the actual rule: derived id from this machine’s reachable name matches the selector id, same derivation as the payload’s `id`; if the id is present in the local fleet file, that entry wins and camp hops. Drop “configured host + reachable name” unless the code is changed to match.

docs/transfer.md
```diff
@@ -0,0 +1,128 @@
...
+Camp reads the leading segment as a **machine id first**, and only if that id is in your
+`~/.obey/machines.yaml`. On a machine with no fleet configured the machine reading is
+unreachable, so every pre-existing form keeps its old meaning exactly.
+
+A machine endpoint requires all three parts. Naming a machine without a campaign is an
+error rather than a guess:
+
+```
+$ camp transfer devbox:notes.md .
+Error: devbox is a machine; name the campaign too: devbox:<campaign>:<path>
```
obey-agent: **[bug]** Quoted machine-without-campaign error does not match production text. Doc: `devbox is a machine; name the campaign too: devbox:<campaign>:<path>`. Code (`newEndpointError`): `"devbox" is a machine; use machine:campaign:path (for example devbox:<campaign>:notes.md)`.

**Suggestion:** Paste the live `newEndpointError` string (or regenerate the example from a real `camp transfer` run against a registered machine id).

docs/transfer.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+### Shadowing, and the `local:` escape
+
+If a campaign happens to share a name with a registered machine, the machine wins — and
+camp tells you, every time, because the ambiguity is in the command you just typed:
+
+```
+$ camp transfer devbox:notes:file.md .
+camp: "devbox" is both a machine and a campaign here; reading it as the machine.
```
obey-agent: **[bug]** Shadow note text is invented. Doc: `camp: "devbox" is both a machine and a campaign here; reading it as the machine. Use "local:devbox" to mean the campaign.` Code (`ShadowNote`): `camp: devbox is a registered machine; reading it as machine:campaign:path (use local:devbox:... for the campaign)`.

**Suggestion:** Quote `ShadowNote` verbatim; keep the every-invocation firing claim (that part is correct).

docs/transfer.md
```diff
@@ -0,0 +1,128 @@
...
+a large file is still making progress; a dead connection is not, and that is the distinction
+worth acting on.
+
+## Both endpoints remote
+
+Refused:
+
+```
+$ camp transfer devbox:a:f.md buildhost:b:f.md
+Error: camp cannot broker a copy between two remote machines; it cannot promise they can
```
obey-agent: **[bug]** Both-remote refusal message does not match code. Doc paraphrases policy (“camp cannot broker a copy between two remote machines…”). Code (`newBothRemoteError`): `at most one endpoint may be on another machine (got <src> and <dest>)`.

**Suggestion:** Quote the real error; optional second sentence may keep the D003 rationale, but the example block should be what operators see.

docs/machine-mesh.md
```diff
@@ -0,0 +1,204 @@
...
+without either machine keeping state about the other.
+
+Fields:
+
+| Field | Meaning | Absent when |
+| --- | --- | --- |
+| `host` | the origin's reachable name | never |
+| `user` | ssh user on the origin | never |
+| `campaign` | campaign the hop started in | the hop did not start inside a campaign |
+| `id` | the origin's machine id, if it has one | the origin is not in the far machine's fleet |
```
obey-agent: **[bug]** Origin payload table says field `id` is “Absent when: the origin is not in the far machine's fleet.” That is not how encoding works. The origin always attempts `id=suggestedMachineID(host)` in `buildHopOrigin` / `encodeHopOrigin`; absence is for empty/over-cap/size-drop of optional fields. Far-fleet membership is a *consumer* concern (lookup / adopt), not a reason the payload omits `id`.

**Suggestion:** Change “Absent when” to empty/unusable derived id or size-bound drop of optional fields; separately note that an id that is not in the far fleet does not auto-register.

cmd/camp/machine.go
```diff
@@ -72,6 +72,21 @@ hops always use BatchMode (agents never hang on password prompts).
...
+what lets 'camp switch -' return without either machine storing state about the
+other. A payload does not register anything: if the origin is unknown here,
+camp names 'camp machine adopt', which previews and asks, requires a TTY, and
+remembers a decline. Reachability need not be symmetric -- a machine you cannot
+dial still appears in completion, because visibility travels as a snapshot
+pushed during a successful enumerate rather than as a live query. Camp will
+never install a key, register a machine unattended, or claim a host is reachable
+on the strength of tailnet membership alone.
+
+See docs/machine-mesh.md for the reachability matrix (notably: the Tailscale SSH
```
obey-agent, line 86: **[suggestion]** `camp machine --help` now points at `docs/transfer.md` for “machine-first transfer grammar,” but this PR does not update `camp transfer`’s own Long help, which still describes only local `campaign:path` copies (“Either the source or destination (or both) can use campaign:path…”). Operators who learn mesh transfer from machine help and then run `camp transfer --help` get a stale surface.

**Suggestion:** In this docs-close (or a tiny follow-up), extend `transferCmd.Long` / Example with `machine:campaign:path`, `local:`, both-remote refusal, and a pointer to `docs/transfer.md`, consistent with the new machine help paragraph.

docs/machine-mesh.md
```diff
@@ -0,0 +1,204 @@
...
+Matched: no -> notes
+/home/alex/campaigns/notes
+```
+
+This matters more than it looks. Without it, a command that is correct everywhere becomes
+a hang or a self-ssh on exactly one machine in your fleet, which is the machine you are
+most likely to be sitting at.
+
+## Visibility without reachability
+
```
obey-agent, line 134: **[suggestion]** Hop-back is described as “using the payload and nothing else.” Steady-state hop-back prefers a `machines.yaml` row matching the origin host so `identity_file` / `ssh_user` / `auth_method` from the operator’s fleet win over the transient payload machine. The gesture is registration-*independent*, not fleet-blind when a match exists.

**Suggestion:** Soften to “payload is enough; a registered host match supplies the operator’s auth settings if present.”

docs/machine-mesh.md
```diff
@@ -0,0 +1,204 @@
...
+reach still shows up usefully, because visibility travels as a **pushed snapshot** rather
+than a live query.
+
+When machine A successfully enumerates machine B, it also pushes its own campaign-name
+list to B. B can then complete `csw A:<tab>` without ever dialing A:
+
+```
+$ camp __complete switch devbox:no
+devbox:notes
+devbox:notebook
```
obey-agent, line 145: **[nit]** Diagnose/publickey example uses a hyphenated “SSH permission denied (publickey) - check ssh-agent…” while `formatPermissionDenied` uses an em dash and continues with Tailscale SSH guidance. Fine as ellipsis, but not a faithful paste of camp’s classified text.

**Suggestion:** Either mark the suffix as abbreviated (`…`) or paste the live classification string.

docs/transfer.md
```diff
@@ -0,0 +1,128 @@
...
+
+Camp only claims reachability it can observe from where it is running.
+
+## Version skew
+
+If a previous `camp machine diagnose` saw a different camp version on the far machine,
+transfers and hops warn:
+
+```
+camp: camp on devbox is v0.9.1, this machine is v0.10.0; features may not match
```
obey-agent: **[nit]** Version-skew example wraps the diagnose clause onto a second line; `hopSkewWarning` emits a single line (`camp on <id> is …; features may not match (run 'camp machine diagnose <id>' to re-check)`), printed as `camp: ` + that line on stderr.

**Suggestion:** Show one line to match stderr, or note the wrap is for readability only.

obey-agent (changes requested):
## Verdict: Request Changes

## Overview

Staff review of **docs(machine): the mesh model, reachability matrix, and transfer grammar** (PR #515, stacked on #509). Scope is three surfaces: `docs/machine-mesh.md`, `docs/transfer.md`, and a one-paragraph mesh summary in `camp machine` Long help. Intent is correct and high-value — this is the right place to freeze fleet-gate learnings (macOS Tailscale SSH sandbox, `CAMP_REMOTE_CAMP_PATH`) so the next operator does not rediscover them. The prose quality, degradation framing, and "what camp will not do" contract are staff-level.

The bar this PR set for itself is **"written against observed behavior, not the specs."** Against that bar, several quoted error strings and a few model claims do not match the implementation already on this stack (`hop_origin.go`, `machine_adopt.go`, `completion_remote.go`, `internal/transfer/*`, `version_cache.go`). Operators and agents who debug from the examples will not find those messages. That is the change gate.

## Key Findings

### Blocking (accuracy / trust)

1. **Unregistered-origin / `csw -` consent story is inverted** (`docs/machine-mesh.md` Consent section). Hop-back is registration-independent: `originTarget` builds a transient machine from the payload. The quoted "origin is not registered here" string is only `hopBackFailure` when *resolve fails* and the origin is unregistered — not the default outcome for an unknown origin. This also contradicts the hop-back section below it.
2. **Payload `id` "Absent when" is wrong.** `buildHopOrigin` always derives `id=suggestedMachineID(host)` on the origin; absence is empty/unusable id or size-bound drop of optional fields — not "not in the far fleet."
3. **Self-reference rule is overstated.** Code is `suggestedMachineID(detectReachableName) == id` and only when the id is *not* already in local `machines.yaml`. Registered entry wins (including a misconfigured self-row). Not "configured host + reachable name."
4. **Transfer quoted errors do not match production** (`docs/transfer.md`):
   - machine-without-campaign → `newEndpointError` (`"…" is a machine; use machine:campaign:path (for example …)`)
   - shadow note → `ShadowNote` (`… is a registered machine; reading it as machine:campaign:path (use local:…:… for the campaign)`)
   - both-remote → `newBothRemoteError` (`at most one endpoint may be on another machine (got A and B)`)
5. **"hops warn once"** is false; skew warns on every hop with a fresh cache entry, without re-probing.

### Non-blocking (should fix or follow up)

6. **`camp transfer` Long help still describes only local `campaign:path`** while machine help now points at `docs/transfer.md` for machine-first grammar — stale secondary surface.
7. Hop-back "payload and nothing else" should note that a registered host match supplies the operator's auth settings when present.
8. Minor paste fidelity on publickey / skew line wrapping (nits).

## What's Done Well

- **Model-first writing**: origin payload, consent boundary, visibility-without-reachability, and degradation cells form a coherent mental model operators can keep.
- **Fleet-gate surprises are named**: macOS sandboxed Tailscale SSH server, and stock `go install` off non-interactive PATH with `CAMP_REMOTE_CAMP_PATH` — both quote real surfaces and save real rediscovery cost.
- **Safety contract is explicit and matches code intent**: no key install, no unattended register, no reachability claimed from tailnet membership alone; adopt requires TTY; decline is remembered.
- **Transfer grammar section** correctly captures machine-first gating, `local:` escape, copy-only / no-clobber, rsync-then-scp, stall vs wall-clock, and both-remote refusal *policy* (even where the example strings need a paste pass).
- **Help surface**: the `camp machine` Long paragraph is the right density for `--help` and correctly points at both docs.
- **Example hygiene**: `devbox` / `notes` / no personal fleet ids.

## Staff Standard

For a docs-close that claims observational fidelity:

1. Every fenced `Error:` / `camp:` line should be copy-pasted from the code path or a real run on this stack — not paraphrased into a fence.
2. Core design invariants (registration-independent hop-back, self-ref rule, payload field semantics) must not be contradicted across sections of the same doc.
3. Secondary surfaces that the new help points at (`camp transfer --help`, generated CLI ref) should not teach the pre-mesh grammar without a link or update — either in this PR or an immediate follow-up.

**Verdict stands at Request Changes** until the blocking accuracy items above are fixed. After a string-and-model pass, this is an easy approve — structure and intent are already there.

---
Reviewed as **obey-agent** via campaign `just gh pr` recipes (token injected per-command; active login remains personal).

PR #516, opened 2026-07-28
fix: preserve divergent default branch during fresh

lancekrogers:
## Summary

- fetch `origin` before synchronizing the default branch
- preserve local-only commits on a deterministic `camp-fresh-recovery-<sha>` branch before realigning to `origin/<default>`
- fast-forward behind-only branches and keep retries idempotent, with collision-safe recovery names
- make dry-run report divergence without moving local branches or files
- document the recovery behavior and explicit undo command

## Root cause

`camp fresh` used `git pull --ff-only` after checking out the default branch. A clean branch with both local-only and remote-only commits therefore stopped at Git's divergence error and required the user to choose a merge or rebase policy manually. That contradicts Fresh's post-merge cleanup contract and adds a decision point even though Camp can preserve the local tip before safely realigning.

## User impact

A divergent clean default branch now completes automatically without losing work. Camp anchors the old tip locally, reports exactly what it did, and prints the command that restores it. Recovery branches are never pushed by Fresh.

## Validation

- `just gate`
- full container integration suite: 576/576 tests passed
- real `camp fresh obey --dry-run --no-prune --no-follow-up --no-branch --no-push` smoke test

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ camp pull all              # Pull all submodules
 camp status                # Show git status of the campaign
 camp status all            # Dashboard of all submodules (branch, dirty/clean, push status, unmerged branches)
 camp status all --view     # Interactive TUI viewer with per-repo detail
-camp fresh                 # Post-merge branch cycling: checkout default, pull, prune, optional new branch
+camp fresh                 # Post-merge branch cycling: fetch, safely sync default, prune, optional new branch
 camp fresh all             # Same cycle across every project in the campaign
 ```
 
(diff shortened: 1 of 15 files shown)


obey-agent (approved):
Reviewed the fresh reconciliation path end to end. The implementation fetches before comparison, fast-forwards behind-only branches, anchors ahead/diverged tips before the hard reset, reuses matching recovery refs on retry, handles collisions, and keeps dry-run local refs/worktree untouched while being explicit that remote-tracking refs refresh.

Focused validation: `go test ./internal/commands/fresh` passes. I found no blocking correctness issue in this change.

PR #517, opened 2026-07-29
test: label Docker transport faults so they stop reading as test failures

lancekrogers:
## What

A dropped Docker exec stream surfaced as an ordinary test failure:

```
--- FAIL: TestFresh_ListRejectsPositionalArg
    failed to init git repo: container exec attach: unexpected EOF
```

That names a test about positional arguments, so the reader goes and studies argument handling. Nothing was wrong with the code under test — the daemon dropped the exec stream, in this case because a gate and the integration suite were running at once.

## Change

Every container command in `helpers_container.go` (15 call sites) now goes through one wrapper that recognizes the transport signatures and labels them `INFRASTRUCTURE FAILURE (not a test failure)`, with the likely cause and what to do.

This is the same distinction #514's Reset latch draws. That one is scoped to pool checkout, so every other exec in the suite could still produce the misleading form. Found while running the suite for CA0004.

## Not relabelled

Real command failures (`exit status 1`, `fatal: not a git repository`) pass through unchanged, asserted by test. Labelling those would hide genuine failures behind a re-run suggestion, which is worse than the noise it removes.

## Verification

- `TestClassifyExecError` covers both directions, including that the original error stays wrapped so the cause survives.
- `TestFresh_*` and `TestWorktreesList_*` (the tests that surfaced the fault) pass.

### DIFF
--- a/tests/integration/exec_classify_test.go
+++ b/tests/integration/exec_classify_test.go
@@ -5,6 +5,7 @@ package integration
 
 import (
 	"errors"
+	"io"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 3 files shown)


obey-agent (changes requested):
Blocking finding:

- `TestContainer.exec` classifies only the error returned directly by `container.Exec`. In testcontainers-go, `Exec` returns a live hijacked `io.Reader`; transport loss can therefore surface later from `io.ReadAll(reader)` as `unexpected EOF`, `connection reset by peer`, or a similar stream error after `Exec` itself returned nil. `RunCamp`, `RunCampInDir`, `ReadFile`, `ListDirectory`, `ExecCommand`, and the other consumers still wrap that path as plain `failed to read output`, recreating the misleading test failure this PR is intended to eliminate.

Please route reader consumption through a helper that applies `classifyExecError` to read failures as well, and add a focused test using a reader that returns `io.ErrUnexpectedEOF`. The direct classifier tests pass by inspection, but they do not exercise this second error channel.

obey-agent:
The blocking finding is addressed in 834c3c62. Every reader consumption in helpers_container.go and helpers.go now routes through readExecOutput, which applies classifyExecError to failures on the second error channel, the read of the live hijacked stream. The focused test the review asked for is in exec_classify_test.go: a reader returning io.ErrUnexpectedEOF must come back labelled INFRASTRUCTURE FAILURE with the cause wrapped, a non-transport read error must pass through unchanged, and the clean path returns bytes untouched.

Verification: go vet with the integration tag is clean, and TestClassifyExecError plus TestReadExecOutputClassifiesMidReadTransportLoss pass through the real harness on this exact diff.

Worth merging soon on its own evidence: today a gate overlapping the integration suite produced 384 instant failures that all read as test failures, which is precisely the misattribution this PR removes.

PR #518, opened 2026-07-29
feat(jobs): deferred commit queue, drain points, and the camp jobs surface (CA0004 phase 002)

lancekrogers:
**Stacked on #511.** Review that first; this branch's diff is only meaningful on top of it.

Phase 002 of CA0004. Implements the `deferred-commits/` design package: a durable per-repo-lane job queue, short-lived workers, drain points at every history-touching command, and the `camp jobs` surface.

## Sequence 01 — queue and worker

- **Queue**: per-repo lanes under `.campaign/cache/jobs/`, one file per job, no state field — state *is* the directory, because a rename is atomic and a field update isn't. `O_CREATE|O_EXCL` enqueue, rename claim, unlink complete.
- **Worker**: per-lane election with a 5s heartbeat and its own 60s liveness threshold (never `fsutil`'s global 30s — that file's own comment mandates the split), crash reclaim, attempt bounding, and the check-after-release invariant that makes "a queued job is always eventually served" true rather than usually true.
- **`camp jobs run`**: silent on success, since it runs detached with stdout wired to `worker.log`.

## Sequence 02 — drain and failure

- **`jobs.Drain`**: spawns a worker *before* it waits, because a drain that only waits turns a crashed worker into a wedged terminal. Counts `running/` as well as `pending/`, and for a root drain also counts other lanes' jobs whose follow-up targets the root.
- **Manifest exemption**, carried as a `class` field so the filter is data rather than path-sniffing. A manifest carries `describes_commit`, so it is correct whenever it lands.
- **Wiring**: push, pull, sync, fresh, commit, p commit, stage, refs, status, log, doctor. Read-only warns and proceeds; write refuses naming the job, with `--no-drain` and `camp jobs` in the message; commit waits job-aware so a long deferred message writer does not refuse the user's next commit.
- **`commitkit.DrainJobs`** exported for fest, called from `StageAll`, `StageFiles`, *and* `Commit` so the guarantee does not depend on which one a consumer uses.
- **`camp jobs`** list/retry/drop/drain with `--json`; the failed-job notice on every foreground command; a doctor `jobs` check for failed, stuck, and queue-lost states.

## Notable fixes found during the work

- The design's own collision check could not work: job filenames carried the id *and* used `O_EXCL`, which are incompatible because the id has a random suffix. Sixteen concurrent enqueuers shared three sequences. **Design doc corrected in place.**
- `Run` busy-spun a whole CPU when every lane was held by another worker.
- `camp stage` and `camp refs` were missing drain points.
- `camp jobs drain` told users to rerun with `--no-drain`, a flag it does not define.
- Pre-existing doctor rendering: root-level issues printed as `✗ : description`, `FixCommand` was hidden behind `--verbose`, and `SeverityInfo` drew a warning icon. Fixed inline.

## Verification

- `just gate` green: 3917 tests (gate-fast), 3872 (gate).
- Full containerized integration suite green, run in full because the doctor output change affects every check.
- **Twenty negative controls**: each invariant deleted in turn, confirming the test that claims to protect it actually fails. Adopted after a test in sequence 01 passed twice against a deleted invariant.

Harness fix found here shipped separately as #517.


obey-agent (commented):
## Verdict: request changes before this draft is ready

The core shape is strong and the focused queue/drain/CLI tests pass, but several crash-safety and ordering guarantees are not yet true in the implementation.

### Blocking findings

1. **A successful parent job is deleted before its follow-up is durable** (`internal/jobs/worker.go`, `runLane`). `complete(nil)` unlinks the running parent, then `enqueueFollowUp` writes the root gitlink job. A crash in that gap permanently loses the follow-up; a concurrent root/all-lanes drain can also observe no parent and no follow-up and return early. Keep the parent visible until the follow-up is durably enqueued, with an injected-window test. Duplicate recovery is acceptable; silent loss is not.

2. **Commit-tree crash retry is not idempotent** (`internal/jobs/execute.go`, `executeCommitTree`). Every retry runs `git commit-tree` again. Committer timestamps are part of the object, so the same tree, parent, and message produce a different SHA on a later attempt. If update-ref succeeded and the worker died before unlinking the job, retry creates C2, sees HEAD at C1 rather than C2, and parks work that already landed. I reproduced two different SHAs from identical inputs two seconds apart. Persist the created SHA/dates or recognize the already-applied commit semantically.

3. **Drain coverage has history-moving holes.** `camp worktrees commit` adds deferral but never drains its lane before staging/committing; a second commit can move HEAD and fail the queued first job. `camp fresh all` runs the child `RunE`, bypasses the drain in the parent `fresh` `RunE`, and `--no-drain` was added as a non-persistent parent flag so the child cannot even parse it. `commitkit.CommitAll` also bypasses `drainQuietly`. Add chokepoint/command-matrix tests so every claimed history surface is enforced rather than documentary.

4. **`commit-paths` does not capture enqueue-time content** (`internal/jobs/execute.go`). The job records paths only and stages their live contents later. In a direct repro, I queued `note.md` containing `original`, edited it to `later-edit`, then ran the worker; the commit subject was `capture original` but the committed blob was `later-edit`. This contradicts the package claim that a job effect is fixed at enqueue time and lets a later edit land under an earlier operation message. Capture blobs/tree/index state at enqueue, or narrow and document materially different coalescing semantics.

5. **The shipped Windows build is broken** (`internal/jobs/execute.go`). `GOOS=windows GOARCH=amd64 go test ./internal/jobs` fails with `unknown field Setsid in struct literal of type syscall.SysProcAttr`; the repository has an explicit Windows build target. Put detach behavior behind platform-specific files/helpers.

### Additional correctness gaps

- `readJob` never re-runs `Job.Validate`, although the docs explicitly treat queue files as hand-editable. A modified `paths: ["."]` therefore reaches `git add -- .` and defeats the no-wildcard invariant. Validation also accepts a commit-tree job with no parent even though execution rejects it.
- `commitkit.drainQuietly` discards every drain error, not only timeout as its comment says. Permission/corruption/context failures therefore let the consumer proceed while the public API claims it inherited an ordering guarantee.

Focused validation passed for `internal/jobs`, `internal/drain`, `internal/defercommit`, `internal/autowrite`, `internal/git/commit`, `pkg/commitkit`, and `cmd/camp/...`. The findings above are invariant/race gaps not exercised by that suite.

obey-agent (changes requested):
Requesting changes now that this PR is ready for review. The head is unchanged from the draft review, so the documented blockers still apply: a successful parent job is removed before its follow-up is durably enqueued; commit-tree retries can create a new commit SHA and strand a job after update-ref succeeds; several commit entry points bypass draining; commit-paths captures live content at execution time rather than enqueue-time content; and internal/jobs does not compile for Windows because it uses SysProcAttr.Setsid directly. The draft review also noted missing job-file revalidation, acceptance of commit-tree jobs without Parent, and drainQuietly suppressing non-timeout failures. Focused Linux tests passed previously, while the Windows compile check failed on Setsid as described.

obey-agent:
Four of the five blockers and all three gaps are fixed. Head is now `73593358`. Base merged from the updated `feat/stage-guards`.

## Blockers

**1. Follow-up lost if the worker crashed in the gap** (`internal/jobs/worker.go`)
The follow-up is now made durable before the parent is completed. Completing unlinks the parent's file, so enqueuing after it left a window where a crash lost the follow-up permanently — nothing retried the parent because it was gone, and a drain observed neither. Reversed, the same crash leaves the parent to be reclaimed and re-run, which enqueues a duplicate; both re-runs are no-ops. A follow-up that cannot be written now fails the parent rather than leaving the job quietly half done.

**2. Retry stranded a job whose work had landed** (`internal/jobs/execute.go`)
The already-applied check compared SHAs, but `git commit-tree` puts the committer timestamp in the object, so a retry over identical inputs produces a different hash every time and could never match. Now recognized by content: `alreadyApplied` compares HEAD's tree and parent against the job's, via `git.HeadTreeAndParent`. A commit with this job's tree and parent is this job's commit, whenever it was made.

**3. Drain coverage holes**
`camp worktrees commit` never drained; `camp fresh all` runs its own `RunE` and could not parse the parent's `--no-drain` (non-persistent parent flag), so it now has its own; `commitkit.CommitAll` bypassed `drainQuietly`. All three fixed.
Test: `cmd/camp/drain_matrix_test.go` — a row per history-moving surface, asserting the drain is inside the command's own function. Verified by negative control both ways: deleting a drain fails, and moving it to a helper in the same file also fails.

**4. commit-paths staged live content at execution**
Content is now captured at enqueue (`git.CaptureBlobs`, `git hash-object -w`) and committed from those objects through a temp index (`git.CommitBlobs`), so a job's effect is a function of enqueue time only — the guarantee `commit-tree` already gave `-aw`. A missing path captures as a deletion, since camp's bookkeeping legitimately removes files. Capture failure degrades to the old behaviour rather than costing the commit.
Test: `TestIntegration_BookkeepingCommitsWhatWasThereWhenItWasQueued`. It holds the lane lock before enqueuing — without that the worker usually commits before the edit and the test passes whether or not capture happened, which is exactly how it first passed with capture removed.

**5. Windows build broken by `syscall.SysProcAttr.Setsid`**
Moved behind `detach_unix.go` / `detach_windows.go`. `GOOS=windows GOARCH=amd64 go build ./...` is clean.

## Gaps

- `readJob` and validation: the check landed at `Claim` rather than `readJob`. Validating on read made an invalid job **invisible** — `List` skips what it cannot read, so it never appeared in `camp jobs`, never held a drain, and sat in pending forever. At the claim boundary it is refused and lands in `failed/` where it can be seen. The full integration suite caught this: a nested-follow-up test expected the refusal to be visible and got an empty failed lane.
- `commit-tree` without `Parent` is rejected at enqueue rather than becoming a promise execution refuses.
- `drainQuietly`'s comment now says plainly that it swallows every error, not only the timeout, with the reasoning for that choice.

## Verification

- `just gate` green: 3900 tests (gate-fast), 3855 (gate).
- Full containerized integration suite green (709s, `CAMP_TEST_POOL_SIZE=4`).
- `GOOS=windows GOARCH=amd64 go build ./...` clean.

One note on the suite: it is sensitive to Docker load. With other containers running I saw unrelated tests fail nondeterministically at pool size 6; `CAMP_TEST_POOL_SIZE=4` is deterministic. I verified `pkill -f` semantics directly rather than assuming, since `Reset` now kills stray detached workers between pooled tests.

Not approving; over to you.

obey-agent:
Verification complete, now at head 307b4c80. The five blockers and three gaps from the review are confirmed fixed in code. Verification found three residual gaps in the retry-recognition area, fixed in 307b4c80:

1. alreadyApplied read HEAD alone, so a user commit stacked on top of a crashed-but-landed deferred commit turned it into a false 'HEAD moved' failure. It now walks a bounded first-parent chain.
2. A reclaimed job ran the message writer before discovering its commit had landed, so a retry paid for a second generation and a writer failure on retry failed a commit sitting in the log. Reclaimed jobs now check for landed work first.
3. A blob-capture failure at enqueue was silently discarded. It is now recorded in the worker log.

New tests: the follow-up durability window (both queue files present inside the window; an enqueue failure parks the parent in failed/), the retry gating order, and a containerized end-to-end test that crafts the surviving queue file of a crashed worker and asserts recognition with a user commit buried on top.

Validation on 307b4c80: just gate green, full containerized integration suite green (673 tests, 359s, CAMP_TEST_POOL_SIZE=4), GOOS=windows build clean.

Merge-ready pending Lance: after #511 merges, retarget this PR to main before deleting feat/stage-guards.

PR #519, opened 2026-07-29
feat(camp): pack and unbundle Festival Bundles

lancekrogers:
## Summary
Adds `camp pack` and `camp unbundle` using `github.com/Obedience-Corp/obey-shared/festivalbundle` **v0.4.6**.

- Kind inference from path / `.workitem`
- Campaign and workitem metadata in `info.json`
- No product `go.mod` replace (published module only)

Part of festival FB0001 (festival-bundle-format).

## Test plan
- [x] `go build ./cmd/camp`
- [x] Smoke pack/unbundle of design + explore paths


cmd/camp/pack/pack.go
```diff
@@ -0,0 +1,169 @@
...
+	opts := festivalbundle.PackOptions{
+		Kind:            kind,
+		Name:            name,
+		Creator:         packCreator,
+		Strict:          packStrict,
+		WriteSentRecord: !packNoSent,
+	}
+
+	// Campaign metadata when available
+	if cfg, root, err := config.LoadCampaignConfigFromCwd(ctx); err == nil && cfg != nil {
```
obey-agent: From metadata is resolved from the process current directory, not from source. If I run from campaign A and pack an explicit directory outside A, the resulting info.json still names campaign A and merely leaves relative_path empty. I reproduced that exact case: a source under /tmp was stamped with lance-arch ID and name. This is false provenance and also omits valid provenance when the caller stands outside the source campaign. Detect from source, and populate From only when source actually belongs to that detected campaign.

obey-agent (changes requested):
Requesting changes for provenance correctness. The command-package suite passes, but pack currently associates a bundle with the caller current campaign rather than the explicitly supplied source directory. The new package also has no direct tests, so please add coverage for packing a source inside, outside, and in a different campaign.

lancekrogers:
Addressed review: provenance now walks from the **source** directory via `campaign.Detect(source)` and only sets `From` when the source is under that campaign root. Added tests for inside/outside/cross-campaign packing.

PR #520, opened 2026-07-29
fix(sweep,fresh): release links on sweep promote and dedupe backstop matches

lancekrogers:
Follow-on to #503, which taught `doDungeonPromote` to release a workitem's links. Two callers were left behind.

## `camp workitem sweep` did not release links

`sweepOne` performed the move, audit, ledger, and commit but never dropped the workitem's links. A design with several worktree links would shelve into the dungeon and leave stale rows behind, each resolving to a dungeon path the selector cannot see.

It now captures the workitem key *before* the move (the key encodes the source path, which is about to change, same reason `doDungeonPromote` captures it), calls `unlinkShelvedWorkitem`, stages `links.json` alongside the rest of the move, and reports each released link on stdout.

## `camp fresh` backstop promoted the same workitem repeatedly

`MapMergedBranchesToWorkitems` appended one match per pruned branch. A workitem with stacked worktree links (`p0..p3`) therefore produced four promote attempts: the first succeeded, and the remaining three failed with `stat workitem ... no such file` because the source directory had already moved.

Branch-to-workitem mapping now runs through `collectWorktreeLinkMatches`, which keeps at most one match per workitem key. A branch whose workitem was already matched is skipped rather than pushed onto `unmatchedBranches`, so it does not incorrectly open the commit-tag signal path. `promoteMergedMatches` also de-duplicates by key as a second line of defence.

## Verification

- `go build ./...`
- `go test ./internal/commands/fresh/... ./internal/commands/workitem/...`
- New regression test `TestCollectWorktreeLinkMatches_DedupesSameWorkitem` reproduces the four-worktree case and asserts a single match.

Context: WF0001 phase 002 (merged-branch backstop) follow-up.


obey-agent (changes requested):
Requesting changes for safe link release.

`sweepOne` passes `ledgerID` into `unlinkShelvedWorkitem`. When the source has no readable marker, `ledgerID` is deliberately the slug fallback used for audit and ledger output. That fallback is not a stable workitem ID. `unlinkShelvedWorkitem` matches ID OR key, so sweep can remove links belonging to another workitem whose real `workitem_id` happens to equal this slug. The shared `doDungeonPromote` path avoids this by keeping display identity separate: `promotedWorkitemIdentity` returns an empty ID for an unmarked directory and relies on the old key.

Please capture the metadata ID separately (empty when the marker is absent/unreadable), pass that plus `oldKey` to unlinking, and add a regression proving an unmarked sweep does not remove an unrelated link whose ID equals the swept slug. This matters because link deletion is the destructive part of this change.

Focused validation passed: `go test ./internal/commands/fresh/... ./internal/commands/workitem/...`; `git diff --check` is clean.

obey-agent:
Addressed in 74ab5df8.

`sweepOne` no longer passes the display identity into link matching. A new `sweptWorkitemIdentity` splits both identities out of the same pre-move marker read:

- `LedgerID` keeps the slug fallback, so an unadopted directory still sweeps and its audit/ledger/envelope output is unchanged.
- `LinkID` stays **empty** when no marker resolves, so an unmarked source is matched by `workitem_key` alone. That is the separation `doDungeonPromote` already gets from `promotedWorkitemIdentity`.

`LinkID` falls back to the discovered `StableID` rather than to nothing, since `StableID` is the marker id resolved at discovery (`merge.go` sets it from `md.ID`, empty when unmarked) and never a slug. That also covers a file-shaped source whose identity lives in its own frontmatter, which `LoadMetadata` cannot read from a `.workitem` sibling.

Regression coverage in `internal/commands/workitem/sweep_links_test.go`:

- `TestSweepOne_UnadoptedSourceDoesNotReleaseUnrelatedLinkMatchingSlug` — the seeded link belongs to `design:workflow/design/other` and carries `workitem_id: example`, exactly the swept directory's slug. Verified this test fails against the previous behavior before the fix was restored:
  ```
  --- FAIL: TestSweepOne_UnadoptedSourceDoesNotReleaseUnrelatedLinkMatchingSlug
      an unadopted sweep must not release another workitem's link, got:
        released link lnk_20260729_unrelated (worktree:projects/worktrees/demo/other); example is no longer active
      unrelated link was deleted by the sweep: registry = []
  ```
- `TestSweepOne_AdoptedSourceReleasesItsOwnLinks` — positive control, so the test above cannot pass by the sweep simply never unlinking.
- `TestSweepOne_UnadoptedSourceReleasesLinkMatchingItsKey` — dropping the ID fallback must not cost key-matched release.
- `TestSweptWorkitemIdentity_SeparatesDisplayFromLinkID` — locks the split directly, including marker-wins-over-`StableID`.

Validation: `just gate` passes (4002 dev + 3957 stable tests, both lint ratchets clean).

PR #521, opened 2026-07-29
feat(promote): festival rail targets, ready and active (WF0001 P3/S2)

lancekrogers:
**Stack base:** branches off `main` at `3643af24` (PR #516). Not stacked on any open PR; #520 (sweep link release) is independent and touches different files.

**Review fixes applied on this branch** after the staff review: the interactive `camp promote` picker now offers the rail targets (`targetsForKind` omitted them), and the rail stage names now derive from the `LifecycleStage` constants instead of being spelled three separate ways.

---

WF0001 phase 003, sequence 02. Adds the festival rail: `camp workitem promote --target ready|active` physically moves a directory workitem onto `festivals/<stage>/<slug>`. The move IS the promotion, so the rail is a real kanban on disk rather than a status field.

Three commits, each self-contained:

## 1. Resident location resolution (`internal/workitem/locate`)

`DetectFromCwd` hard-required a `workflow/` first segment, so a workitem on the rail could not be located at all and none of the shared promote plumbing could touch it. It now dispatches on the first path segment:

| Path | `Type` | `ParentPath` | `DungeonPath` |
|------|--------|--------------|---------------|
| `workflow/<type>/...` | path segment | unchanged | `workflow/<type>/<dungeon>` |
| `festivals/{ready,active}/<slug>` | `.workitem` marker | the stage folder | `festivals/.dungeon` |
| `festivals/.dungeon/<status>[/<date>]/<slug>` | `.workitem` marker | the bucket | `festivals/.dungeon` |

`DungeonPath = festivals/.dungeon` is the point: `MoveToDungeon` builds its dungeon service straight off `loc.DungeonPath`, so a resident completes into the festival-local dungeon with no change to the move code. Sequence 04 depends on this.

The workflow branch was extracted into helpers verbatim; a regression test asserts a `workflow/design/<slug>` item carrying a marker that says `explore` still resolves as `design`, because for workflow items the path wins.

`locate` now imports `internal/workitem` for `LoadMetadata`. The task anticipated an import cycle; `go list -deps` confirms the parent package does not reach `locate`, so no minimal inline YAML parse was needed and residents get the same schema validation as every other marker read.

## 2. Rail targets (`internal/commands/workitem`)

The blanket `cannot promote to active` rejection is replaced by a transition check against the resolved source. `root -> ready -> active` is allowed; backward moves and any attempt to leave a dungeon are refused, which preserves what the old rejection actually protected.

`doRailPromote` returns `*commitInputs`, so the shared promote tail supplies the audit event, ledger emit, and auto-commit unchanged. Reference repair reuses the rename migrations verbatim (`migrateRenamePriority`, `migrateRenameLinks`, `migrateRenameCurrent`) because a rail move only changes the path half of the workitem key: `design:workflow/design/x` becomes `design:festivals/ready/x`. No helper bodies were duplicated and `applyRename` is not called.

**The marker is stamped at rail entry, before the move rather than after.** The task sketch put it after. Directory workitems are *not* uniformly stamped today (`recordPromotedTo` skips when the marker is absent; the ledger falls back to the slug for the same reason), but resident resolution *requires* a marker. Stamping after a successful move would strand the workitem where nothing can resolve it if that write failed. Stamping first means a failure leaves it at `workflow/<type>/<slug>`, fully recoverable. Minting goes through the `camp workitem repair` planner, so the marker matches what `doctor --fix` would write.

`recordPromotedTo` is deliberately not called for a rail move: `promoted_to` records where a shelved source was *copied*, and a rail move relocates the directory itself, so the field could only restate the workitem's own location.

File workitems are rejected: they keep identity in their own frontmatter and have nowhere to put the sidecar resident resolution reads.

Explore gains `ready` in `StagesByType`.

## 3. Tests

Eight containerized integration tests plus unit coverage. Every rejection asserts more than the error message: the workitem is still where it was and the tree is still clean, so a refused promote cannot half-apply.

## Known limitation (planned, not a defect)

A resident is **not yet selectable by id** — discovery does not produce residents until sequence 03, so `ready -> active` currently works from inside the resident directory (cwd) but not via `camp workitem promote <id>`:

```
$ camp workitem promote design-idfeature-2026-07-29 --target active
Error: no workitem matched selector design-idfeature-2026-07-29: ... not found
```

This is the planned sequence order (task 01 notes `wi.RelativePath` for a resident is "produced by sequence 03 discovery"), but it is a real usability cliff until that lands, so it should not be mistaken for finished.

## Deviation from a stated acceptance criterion

The sequence asked that `promote.go` stay under 500 lines. It was **798 lines before this work** and is 813 now, so that criterion was never satisfiable as written. The substantive half was honored: the rail move lives entirely in `promote_rail.go` (188 lines) and `promote.go` gained only the target switch, the guard, the dispatch case, and help text. Reducing it is a real but separate refactor. Flagging rather than silently claiming it.

Every function added here is under 50 lines (`doRailPromote` is the largest at 49).

## Verification

`just gate` green: 3837/3837 unit, 3882/3882 with integration vet. `golangci-lint --new-from-merge-base origin/main`: 0 issues. Both ratchets clean.

Driven end to end against throwaway fixture campaigns with a binary built from this branch — all six transitions, the marker mint, the `--json` envelope, link re-homing, and markdown rewriting. Full transcripts in the festival results docs.

### DIFF
--- a/cmd/camp/promote/selector.go
+++ b/cmd/camp/promote/selector.go
@@ -151,7 +151,9 @@ func targetsForKind(kind promoteKind) []string {
 	case kindIntent:
 		return []string{"ready", "festival", "design"}
 	case kindWorkitem:
-		return []string{"festival", "doc", "completed", "archived", "someday"}
+		// Rail stages lead: they are the forward moves, and a workitem reaches
+		// them far more often than it reaches a festival or the dungeon.
+		return []string{"ready", "active", "festival", "doc", "completed", "archived", "someday"}
 	case kindFestival:
 		return []string{"next", "completed", "archived", "someday"}
 	}
--- a/internal/commands/workitem/promote.go
+++ b/internal/commands/workitem/promote.go
@@ -87,15 +87,21 @@ func newPromoteCommand() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "promote [id] --target <target>",
-		Short: "Promote a workitem to a festival, doc, or dungeon",
+		Short: "Promote a workitem: festival, doc, rail, dungeon",
 		Long: `Promote the workitem identified by [id], by cwd, or by the current pointer.
 
 TARGETS:
   festival    Create a festival from the workitem and shelve the source
   doc         Copy the workitem doc into docs/ and shelve the source
+  ready       Move the workitem onto the festival rail at festivals/ready
+  active      Move the workitem onto the festival rail at festivals/active
   completed   Move the workitem to its local dungeon/completed
   archived    Move the workitem to its local dungeon/archived
-  someday     Move the workitem to its local dungeon/someday`,
+  someday     Move the workitem to its local dungeon/someday
+
+The rail is forward-only: root -> ready -> active. A workitem already on a
+stage cannot move backward, and moving one out of a dungeon is a restore
+rather than a promote.`,
 		Args: cobra.RangeArgs(0, 1),
 		Annotations: map[string]string{
 			"agent_allowed": "true",
@@ -114,7 +120,7 @@ TARGETS:
 	}
 
 	f := cmd.Flags()
-	f.StringVar(&target, "target", "", "Promotion target: festival, doc, completed, archived, someday")
+	f.StringVar(&target, "target", "", "Promotion target: festival, doc, ready, active, completed, archived, someday")
 	f.StringVar(&dest, "dest", "", "Destination path under docs/ for the doc target (must stay within docs/)")
 	f.StringVar(&goal, "goal", "", "Festival goal override (default: first paragraph of the workitem doc)")
 	f.BoolVar(&keep, "keep", false, "On festival/doc, do not move the source workitem to the dungeon")
@@ -130,12 +136,13 @@ func runWorkitemPromote(cmd *cobra.Command, opts runWorkitemPromoteOptions) erro
 
 	switch opts.Target {
 	case "festival", "doc", "completed", "archived", "someday":
-	case "active":
-		return camperrors.New("cannot promote to active: a workitem outside the dungeon is already active; restoring workitems out of the dungeon is not a promote")
+	case railStageReady, railStageActive:
+		// Conditionally valid: the rail is forward-only, so these are checked
+		// against the resolved source location below, once loc is known.
 	case "":
 		return camperrors.New("required flag --target not set")
 	default:
-		return camperrors.New("invalid target: " + opts.Target + " (use festival, doc, completed, archived, someday)")
+		return camperrors.New("invalid target: " + opts.Target + " (use festival, doc, ready, active, completed, archived, someday)")
 	}
 
 	cfg, root, err := config.LoadCampaignConfigFromCwd(ctx)
@@ -146,6 +153,12 @@ func runWorkitemPromote(cmd *cobra.Command, opts runWorkitemPromoteOptions) erro
 	loc, err := resolveWorkitem(ctx, root, opts.ID)
 	if err != nil {
 		return err
+	}
+
+	if opts.Target == railStageReady || opts.Target == railStageActive {
+		if err := checkRailTransition(railStageOf(loc, root), opts.Target); err != nil {
+			return err
+		}
 	}
 
 	// Read .workitem metadata now, before any promote branch runs: festival
@@ -182,6 +195,8 @@ func runWorkitemPromote(cmd *cobra.Command, opts runWorkitemPromoteOptions) erro
 		ci, err = doFestivalPromote(ctx, cmd, opts, root, loc, &result)
 	case "doc":
 		ci, err = doDocPromote(ctx, opts, root, loc, &result)
+	case railStageReady, railStageActive:
+		ci, err = doRailPromote(ctx, cfg, root, loc, opts.Target, &result)
 	case "completed", "archived", "someday":
 		ci, err = doDungeonPromote(ctx, root, loc, opts.Target, &result)
 	default:
--- a/internal/commands/workitem/promote_rail.go
+++ b/internal/commands/workitem/promote_rail.go
@@ -1,0 +1,188 @@
+package workitem
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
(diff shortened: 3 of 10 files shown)


obey-agent (commented):
## Verdict: Comment (merge-ready) — two defects found and fixed during review

Reviewed against the diff rather than the task description. Traced `runWorkitemPromote` from the target switch through the dispatch into `doRailPromote` and back out through the shared tail.

### Found and fixed in this PR

**1. The universal front door could not reach the rail.** `targetsForKind(kindWorkitem)` (`cmd/camp/promote/selector.go:153`) returned only `{festival, doc, completed, archived, someday}`, and `camp promote` builds its interactive picker from that list. Non-interactive `camp promote <id> --target ready` worked, because the router passes workitem targets through without a whitelist — which is exactly why this was easy to miss: only the interactive path was broken. Fixed, with rail stages leading the list.

**2. Three spellings of the rail stage names.** `locate.festivalStages` used string literals, `promote_rail.go` had its own, and `wkitem.LifecycleStageReady/Active` already existed. Nothing would have caught a drift between the folder the resolver accepts and the target promote allows. Both now derive from the `LifecycleStage` constants.

### Verified

- **Fixup reuse** — calls `statusmove.Move`, `mdlinks.RewriteForMove`, and the three rename migrations. No duplicated bodies. Correctly avoids `applyRename`, whose `planRename` computes a same-parent basename change (`rename.go:166`) and is wrong for a cross-parent move.
- **Key correctness** — `<type>:<relpath>` with the original type preserved, matching `discover_workflow_docs.go:70`. `links.yaml` is staged only when changed; an integration test asserts the priority store is re-keyed on disk *and* absent from `git show --name-only HEAD`; `current.yaml` is never staged.
- **Forward-only** — all five refusals hold. The old `promote.go:121` semantics are *narrowed*, not weakened: that case rejected `active` unconditionally citing "restoring out of the dungeon is not a promote", and that reason is now enforced for both rail targets. The guard also sits before the `--dry-run` block, so a dry run cannot claim a refused move would succeed.
- **Local-dungeon readiness** — residents report `DungeonPath = festivals/.dungeon` in all four shapes. Workflow resolution is byte-identical; a regression test pins that a workflow item ignores a marker disagreeing with its path.
- **Standards** — no `fmt.Errorf`, `context.Context` first on every I/O function, added functions all under 50 lines.

### Two things the task document got wrong

**Directory workitems are not uniformly stamped.** The task asked to verify the doc-06 claim that they are. They aren't: `recordPromotedTo` (`promote.go:619-631`) returns early when the marker is absent, and the ledger read (`promote.go:158-161`) falls back to the slug, commenting that "a directory promoted without ever being adopted (no marker) still promotes today". So `ensureResidentMarker` is a real mint, not a no-op guard.

**Stamping belongs before the move, not after** (the task sketched it after). Resident resolution rejects an unstamped directory, so a marker written after a successful move strands the workitem where nothing can resolve it if that write fails. Stamping first keeps every failure recoverable at the original location. One cost: `planRepair` runs a full `wkitem.Discover` for ref uniqueness, so the unmarked path is O(campaign) — acceptable since it only runs when a marker is genuinely missing.

### Not a defect, but not finished either

A resident is not yet selectable by id; discovery does not produce residents until sequence 03, so the advance is cwd-only for now. Called out in the PR body so it is not read as complete.

### Flagged deviation

The sequence asked that `promote.go` stay under 500 lines. It was **798 before this work** and is 813 now, so that was never satisfiable. The rail move is isolated in `promote_rail.go` as intended. Recording rather than glossing.

### Done well

The move genuinely reuses the #484 surface instead of re-implementing it, which is why links, markdown, and priority all behave identically to a rename with no new paths to keep in sync. The rejection tests assert disk state and tree cleanliness, not just error strings, so a half-applied refusal fails the suite.

Gates: `just gate` PASSED, 3837/3837 unit and 3882/3882 including integration vet; `golangci-lint --new-from-merge-base origin/main` 0 issues.

obey-agent (changes requested):
Requesting changes for resident identity correctness.

`ensureResidentMarker` returns as soon as `.workitem` exists; it neither parses nor canonicalizes the marker. That is unsafe at the workflow-to-rail boundary. Workflow resolution intentionally takes the type from the path, so `workflow/design/x` with a valid marker containing `type: explore` resolves as design. `doRailPromote` then migrates references to `design:festivals/ready/x`, but after the move `residentType` takes the marker type and resolves the same directory as explore. Priority, links, and current now point at the wrong key. An existing malformed marker is worse: the move succeeds, then no rail command can resolve the resident at all.

Please run the repair planner for existing markers too and either write the canonical marker before moving or fail before moving. Add coverage for both a valid type mismatch and a malformed existing marker. The PR already identifies path type as authoritative and reuses `planRepair`; the current early return bypasses exactly the repair cases that planner supports.

Focused validation passed: `go test ./cmd/camp/promote ./internal/commands/workitem/... ./internal/workitem/locate`; `git diff --check` is clean.

obey-agent:
Review finding addressed in 2dd0ae0e.

`ensureResidentMarker` no longer early-returns when `.workitem` exists. It now always runs `planRepair`, so all three marker states funnel through the same planner:

- **Missing marker**: minted exactly as before (create path unchanged).
- **Valid marker, mismatched type**: canonicalized to the authoritative path type before the move, so the key `migrateRailReferences` writes and the key resident resolution derives from the marker can no longer diverge. Repair is idempotent: an already canonical marker plans zero changes and is not rewritten.
- **Malformed marker**: `planRepair`'s YAML parse failure surfaces before `statusmove.Move` runs, so the promote fails with the workitem untouched at its original location (the ledger metadata read at promote.go:172 tolerates the unreadable marker, so nothing fails earlier with a less actionable message).

Coverage added in `tests/integration/promote_rail_test.go`:

- `TestPromoteRail_RepairsMismatchedMarkerType`: seds a valid marker to `type: explore` under `workflow/design/`, promotes to ready, asserts the landed marker is repaired to `type: design` with its id preserved, asserts the migrated priority key agrees (`design:festivals/ready/...`), then advances to active to prove resident resolution and the re-keyed stores stay coherent end to end.
- `TestPromoteRail_MalformedMarkerFailsBeforeMoving`: commits an unparseable marker, asserts the promote exits non-zero naming the marker, the source stays in place, nothing lands on the stage, and the tree stays clean.

Verification: all 10 `TestPromoteRail` containerized tests pass, and `just gate-fast` is green (0 lint issues, 3882/3882 unit tests).

PR #522, opened 2026-07-29
feat(artifacts): committed per-machine manifests with incremental hashing (CA0004 phase 002/04)

lancekrogers:
## What

Design doc 08 (`workflow/design/camp-artifact-commit-updates-2026-07-25/08-artifact-manifests.md`): camp records what the artifact set was at each commit, and never stores old copies.

- **Incremental hashing** (`internal/artifacts/hash.go`): `FileEntry` gains `hash_sha256` under a versioned key so pre-hashing manifests degrade to re-hash, never misread. Only entries whose size or nanosecond mtime moved are re-read; unchanged entries inherit their hash without a read (proven by a test whose files do not exist on disk). SHA-256 for external verifiability (`shasum -a 256`).
- **Committed record** (`internal/artifacts/committed.go`): one file per (machine, root) at `.campaign/artifacts/manifests/<machine>/<root-slug>.json`, carrying `describes_commit`. GeneratedAt is absent from the committed form; an unchanged root serializes byte-identically and the job commits nothing.
- **Machine identity**: `NormalizeHost` over the hostname, `CAMP_MACHINE_NAME` override; captured at enqueue on the job so a worker spawned under any environment writes the enqueuer's record.
- **Queue integration** (`internal/jobs/manifest.go`): new `manifest` kind, always `class: manifest` so it is exempt from every drain; enqueued at each campaign-root commit tail (foreground, synchronous bookkeeping, worker completions), deduped per (root, machine) keeping the newest `describes_commit`.
- **Drift surfaces**: the doctor `artifacts` check replaces its phase-002 skip with the real `manifest_drift` row; the pre-wired `ArtifactRootDrift` status notice gets its live body. Stat-level only; report, never resolve.

## Semantics worth noting

Artifact-only edits produce no commit and therefore no record update, by design: the record ties artifact state to git history, and drift covers the window until the next real commit. Manifest carrier commits land asynchronously, so tests that read HEAD right after committing were re-anchored (`latestUserCommit`).

## Verification

- Full containerized integration suite: **691/691 green** (753s), including new criteria tests for 16, 17, 19, 21, 21b, 21c, 27. The 21b test is deterministic: with the lane lock held so nothing can serve the queue, `camp jobs drain` still returns with the manifest job pending.
- `just gate` green; `GOOS=windows` unaffected (no platform code touched).
- The suite outgrew its 10m go test cap during this work (timeout panic with zero failing tests); the budget is raised to 15m in the buildtool and verbose recipe.

### DIFF
--- a/internal/jobs/manifest.go
+++ b/internal/jobs/manifest.go
@@ -1,0 +1,111 @@
+package jobs
+
+import (
+	"context"
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/Obedience-Corp/camp/internal/artifacts"
+	"github.com/Obedience-Corp/camp/internal/git"
+)
+
+// Committed artifact manifests ride the queue as their own kind because they
+// invert the usual promise. A commit job must land exactly what was captured;
+// a manifest job must land whatever the root holds when it runs, labelled
+// with the commit it describes. Late is correct here, which is why the class
+// is exempt from every drain.
+
+// executeManifest builds the incremental manifest for one root, writes the
+// machine's committed record, and commits it. An unchanged root writes and
+// commits nothing: the committed record is already correct, and rewriting it
+// would churn a byte-identical file through history.
+func executeManifest(ctx context.Context, campaignRoot, repoPath string, job *Job) error {
+	machine, err := artifacts.MachineName()
+	if err != nil {
+		return err
+	}
+	prev, _, err := artifacts.LoadCommitted(campaignRoot, machine, job.ManifestRoot)
+	if err != nil {
+		return err
+	}
+	m, err := artifacts.BuildManifest(ctx, campaignRoot, job.ManifestRoot)
+	if err != nil {
+		return err
+	}
+	// The heartbeat goroutine on the lane lock keeps the worker alive through
+	// a long first-pass hash; no progress plumbing is needed for liveness.
+	if err := artifacts.HashManifest(ctx, campaignRoot, m, prev, nil); err != nil {
+		return err
+	}
+	if prev != nil && artifacts.SameFiles(prev.Files, m.Files) {
+		return nil
+	}
+	rel, err := artifacts.WriteCommitted(campaignRoot, machine, m, job.DescribesCommit)
+	if err != nil {
+		return err
+	}
+	err = git.CommitScoped(ctx, repoPath, []string{rel}, &git.CommitOptions{
+		Message: "manifest: " + job.ManifestRoot + " at " + shortSHA(job.DescribesCommit),
+	})
+	if errors.Is(err, git.ErrNoChanges) {
+		return nil // a drain or ordinary commit already carried the file
+	}
+	return err
+}
+
+// EnqueueManifest queues one root's manifest job, dropping any pending
+// manifest job for the same root first. Only the newest record matters: a
+// pending job for an older commit would build the same bytes and label them
+// with a stale SHA, so the dedupe keeps the latest describes_commit and the
+// queue from accumulating one job per keystroke.
+func EnqueueManifest(ctx context.Context, campaignRoot, rootRel, describesCommit string) error {
+	pending, err := List(campaignRoot, statePending, ".")
+	if err == nil {
+		for _, j := range pending {
+			if j.Kind == KindManifest && j.ManifestRoot == artifacts.NormalizeRootPath(rootRel) {
+				// A claim racing this remove wins the rename and the remove
+				// sees ENOENT; both jobs then run and the newer record lands
+				// last. Duplicate work, never a wrong record.
+				_ = os.Remove(filepath.Join(laneDir(campaignRoot, statePending, "."), jobFilename(j.Seq)))
+			}
+		}
+	}
+	_, err = Enqueue(ctx, campaignRoot, Job{
+		Kind:            KindManifest,
+		Class:           ClassManifest,
+		Repo:            ".",
+		ManifestRoot:    artifacts.NormalizeRootPath(rootRel),
+		DescribesCommit: describesCommit,
+	})
+	return err
+}
+
+// EnqueueManifestsForRoots queues a manifest job per declared root, recording
+// the campaign root's current HEAD as the described commit. A campaign with
+// no declared roots does nothing at all, which is the zero cost everyone else
+// pays for this feature.
+func EnqueueManifestsForRoots(ctx context.Context, campaignRoot string) (int, error) {
+	cfg, err := artifacts.Load(campaignRoot)
+	if err != nil {
+		return 0, err
+	}
+	if len(cfg.Roots) == 0 {
+		return 0, nil
+	}
+	head, err := git.FullHash(ctx, campaignRoot)
+	if err != nil {
+		return 0, err
+	}
+	enqueued := 0
+	for _, r := range cfg.Roots {
+		if ctx.Err() != nil {
+			return enqueued, ctx.Err()
+		}
+		if err := EnqueueManifest(ctx, campaignRoot, r.Path, head); err != nil {
+			return enqueued, err
+		}
+		enqueued++
+	}
+	return enqueued, nil
+}
--- a/internal/jobs/manifest.go
+++ b/internal/jobs/manifest.go
@@ -21,10 +21,10 @@ import (
 // commits nothing: the committed record is already correct, and rewriting it
 // would churn a byte-identical file through history.
 func executeManifest(ctx context.Context, campaignRoot, repoPath string, job *Job) error {
-	machine, err := artifacts.MachineName()
-	if err != nil {
-		return err
-	}
+	// The identity was captured at enqueue; a worker spawned under any other
+	// environment still writes the enqueuer's record under the enqueuer's
+	// name.
+	machine := job.Machine
 	prev, _, err := artifacts.LoadCommitted(campaignRoot, machine, job.ManifestRoot)
 	if err != nil {
 		return err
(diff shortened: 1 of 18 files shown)


obey-agent (changes requested):
## Verdict: Request Changes

The focused suites and Windows build pass, but three manifest invariants are not yet true.

1. **A crash after writing the manifest can permanently skip its commit.** `executeManifest` calls `WriteCommitted` before `git.CommitScoped`. If the worker dies in that window, or the commit returns a retryable error, the reclaimed job calls `LoadCommitted` on the working-tree file it just wrote. `SameFiles` is then true and the retry returns success without ever committing the record. Load the comparison baseline from committed HEAD, or only take the same-files fast path after proving the manifest path is already committed/clean. Add an injected crash-window retry test.

2. **The new path fields bypass the queue and artifact path safety rules.** `EnqueueManifestsForRoots` loads hand-editable `artifacts.yaml` but does not call `Validate` or `EnsureRootWithin`; `Job.Validate` only checks that `ManifestRoot` is non-empty. A root such as `../../outside` therefore reaches `BuildManifest` and walks outside the campaign. Separately, `MachineName` only lowercases/trims, while a hand-edited job machine is also checked only for non-emptiness; `../...` can escape `.campaign/artifacts/manifests` during `WriteCommitted`. Validate the root at enqueue and execution, constrain machine to one safe path segment (the existing peer-ID validator is close), and reject a committed manifest whose embedded root differs from the requested validated root.

3. **Deferred hashing does not capture state at `describes_commit`.** A job records the SHA at enqueue but reads every artifact byte later. If `media/f.bin` changes after the user commit while the lane is busy, the eventual manifest hashes the later bytes, labels them with the earlier SHA, and reports no drift. The current lane-lock integration test never mutates the root after the described commit, so it misses this. Capture enough enqueue-time state to detect this race and retry/relabel, or narrow the contract; as written, the record cannot truthfully say what the artifact set was at that commit.

Validation run on head `0a825d07`: `go test ./internal/artifacts ./internal/jobs ./internal/doctor/checks ./internal/notice ./internal/git/commit ./cmd/camp/...` passed; `GOOS=windows GOARCH=amd64 go build ./...` passed; `git diff --check` is clean.

obey-agent:
All three findings addressed in 422ace90.

1. Crash after write skipping the commit: the worker's baseline is now LoadCommittedAtHead (git show HEAD:<rel>), never the working-tree file, for both the hash carry-forward and the skip decision. Test: TestIntegration_ManifestCrashRetryStillCommits reproduces the exact crash state with git reset --soft (record on disk, carrier commit gone) plus a crafted retry job, and asserts the retry commits.

2. Path safety: roots validate through ValidateRootPath and EnsureRootWithin at enqueue, at the claim boundary (Job.Validate), and at the moment of use in the executor; machine identities validate as single safe path segments under the peer-id rules (ValidateMachineSegment); a committed record embedding a different root than requested degrades to a first pass (decodeCommitted). Tests: TestManifestJobValidationRefusesEscapes (escaping and absolute roots, separator and dot-prefixed machines), TestDecodeCommittedRefusesAMismatchedRoot.

3. Capture-window race: enqueue records a stat-level fingerprint of the root (no content reads, preserving the zero-cost enqueue); the worker recomputes it before hashing, and on mismatch writes the record from observed state relabelled to the commit current at observation, with a manifest-relabel line in the worker log. The pre-edit bytes are unrecoverable by the time the worker runs, so the contract is narrowed with detection rather than pretending the old bytes were captured; design doc 08 carries the amendment. Test: TestIntegration_ManifestDetectsPostCommitArtifactEdit pins the lane, commits, edits the root, and asserts the log line and that the record hashes the observed bytes.

Validation on 422ace90: just gate green, all five manifest integration tests green, GOOS=windows build clean. Not approving; over to Lance.

internal/jobs/manifest.go
```diff
@@ -0,0 +1,155 @@
...
+			job.ManifestRoot, shortSHA(job.DescribesCommit), shortSHA(head))
+		describes = head
+	}
+
+	// The heartbeat goroutine on the lane lock keeps the worker alive through
+	// a long first-pass hash; no progress plumbing is needed for liveness.
+	if err := artifacts.HashManifest(ctx, campaignRoot, m, prev, nil); err != nil {
+		return err
+	}
+	if prev != nil && artifacts.SameFiles(prev.Files, m.Files) {
```
obey-agent, line 51: Blocking: this fast path proves that the artifact root matches the manifest in HEAD, but it does not prove that the working-tree manifest path also matches HEAD. Reproduction: commit manifest A, replace only the working-tree manifest with stale crash output B, leave the root at A, then retry. `prev` and `m` are both A, so this returns success while B remains. Status reads B through `LoadCommitted`, and a later broad commit can sweep B into history. I reproduced this with an isolated test against 422ace90. Before returning, verify or restore the working-tree record to the HEAD-equivalent bytes; add a regression test with an existing HEAD record plus a stale uncommitted record.

internal/jobs/manifest.go
```diff
@@ -0,0 +1,155 @@
...
+			return headErr
+		}
+		logWorker(campaignRoot, "manifest-relabel root=%s described=%s now=%s: the root changed after enqueue; the record reflects the observed state",
+			job.ManifestRoot, shortSHA(job.DescribesCommit), shortSHA(head))
+		describes = head
+	}
+
+	// The heartbeat goroutine on the lane lock keeps the worker alive through
+	// a long first-pass hash; no progress plumbing is needed for liveness.
+	if err := artifacts.HashManifest(ctx, campaignRoot, m, prev, nil); err != nil {
```
obey-agent: Blocking: the enqueue fingerprint is checked only against the pre-hash stat walk. A file can change after this comparison or during `HashManifest`—especially during a multi-hundred-GB first pass—so the hash can reflect later bytes while `m` retains earlier size and mtime, and `describes` is never relabelled. Across multiple files the committed record can be a mixed state that never existed at one observation point. Re-stat or fingerprint after hashing and retry/relabel when the root moved (with per-file before/after checks if needed), and cover a mutation during hashing rather than only while the lane is pinned.

obey-agent (changes requested):
Re-review of 422ace90: the HEAD baseline, root and machine validation, mismatched-root refusal, and enqueue-to-worker divergence detection address the original findings in the common path. Two correctness holes remain in the retry and hashing boundaries. Focused Go tests, go vet, the Windows build, and diff checks pass. The full suite has two internal/git fixture failures that reproduce unchanged on the base commit. Docker integration tests could not start because the configured Colima SSH transport is unavailable.

obey-agent:
Both remaining findings addressed in d274e58a.

## 1. Skip path left an uncommitted record in the working tree

Confirmed the repro exactly as described. `ReconcileCommittedToHead` now restores the working-tree file to the bytes HEAD carries before the skip returns, and logs `manifest-restore`.

Two implementation notes: the comparison is against raw `git show` bytes rather than a re-serialized decode, so byte-equality is what is actually checked; and only the working tree is touched, never the index, because the worker stages through camp's temp-index path and a crashed attempt therefore leaves its write unstaged by construction.

`TestIntegration_ManifestSkipRestoresAnUncommittedRecord` reproduces your case: HEAD holds the real record, the working tree holds `deadbeef`, root untouched so the skip path runs. Verified it fails without the reconcile:

```
Error: Not equal:
  expected: "describes_commit": "745024695ba452198f1cca868b3a6e2087842bf4" ... "hash_sha256": "c74224c2..."
  actual:   "describes_commit": "0000000000000000000000000000000000000000" ... "hash_sha256": "deadbeef"
```

It also asserts `git status --porcelain` is clean for the record path afterwards, which is the property that stops a later broad commit sweeping it in.

## 2. Hashing captured state that never existed

This was two defects, and both are fixed.

**Per entry.** Size and mtime came from the walk while the hash came from a read that could be minutes later. Each hashed entry is now re-statted around its read and keeps the stat describing the bytes actually hashed, retrying while the file moves under the reader. A file that will not settle after `settleAttempts` records the latest stat with **no** hash: unknown-and-re-read-next-pass is a state the format already carries, and a hash paired with a stat it does not belong to is indistinguishable from a true record. `HashManifest` returns a `HashOutcome` naming those files and the worker logs `manifest-unsettled`, so an incomplete record does not look complete.

**At the root.** The enqueue fingerprint was compared before hashing, which proves nothing about the record actually written, and cannot see a carried-forward entry whose file moved after the walk since that entry is never re-read. The comparison now runs against a stat walk taken **after** the hash pass. It sits after the skip-commit fast path, so the steady-state case does not pay for the extra walk.

Coverage: `TestHashManifestEntryStatDescribesTheBytesItHashed` mutates the file from inside the read via the progress callback (verified failing against the previous behavior with `entry kept the pre-read stat after the file changed under the reader`), `TestHashManifestReportsAFileItCouldNotPinDown` covers continuous write, `TestHashManifestReportsNothingWhenTheRootIsStill` locks the ordinary path, and `TestIntegration_ManifestRelabelsWhenTheRootMovesDuringHashing` covers the label at the worker boundary.

## Validation

- `just gate` passes: 4009 dev + 3964 stable tests, both lint ratchets clean.
- Docker was available here, so the integration tests you could not start did run: `TestIntegration_ManifestSkipRestoresAnUncommittedRecord`, `TestIntegration_ManifestRelabelsWhenTheRootMovesDuringHashing`, `TestIntegration_ManifestCrashRetryStillCommits`, `TestIntegration_ManifestDetectsPostCommitArtifactEdit` all pass.
- On the two `internal/git` fixture failures from your re-review: they do not reproduce here. `go test ./internal/git/...` is clean (`ok internal/git 17.755s`, `ok internal/git/commit 2.475s`) and so is the full gate, so that looks environment-specific rather than a real regression.

I did not run the full integration suite end to end; it is a ~15m budget and I scoped to the manifest tests plus the complete unit gate.

PR #523, opened 2026-07-30
feat(commitkit): export StageAllWithOptions for consumer commit-large flags

lancekrogers:
## What

Exports the staging guard's per-invocation override through pkg/commitkit: `StageOptions` (alias of the internal type, like every other guard re-export) and `StageAllWithOptions`, which is `StageAllWithOutcome` plus the override, wired to the same internal path camp's own `--commit-large` uses.

## Why

Fest's commitkit integration (CA0004 phase 003) adds `fest commit --commit-large`, mirroring camp's flag. The override existed only inside the module: `StageAllWithOutcome` takes no options, so a consumer had no way to say "commit it anyway" without reimplementing staging. This is the smallest export that closes the gap.

## Testing

- Alias compile-checks in both directions (same pattern as the file's other re-exports), so the override a consumer builds is the override the guard reads with no conversion.
- Cancelled-context short-circuit test for the new function.
- `just gate-fast` green (lint 0 issues, full unit suite both profiles).

Once merged this ships as v0.4.0-rc.2 so fest can bump from rc.1 and thread the flag.

### DIFF
--- a/pkg/commitkit/commitkit.go
+++ b/pkg/commitkit/commitkit.go
@@ -190,6 +190,19 @@ func StageAllWithOutcome(ctx context.Context, repoPath string) (*StageOutcome, e
 	return git.StageWithGuard(ctx, repoPath, nil)
 }
 
+// StageAllWithOptions is StageAllWithOutcome with per-invocation overrides of
+// the guard's decision. It exists for the consumer flag that means "commit it
+// anyway": fest's --commit-large forwards here exactly as camp's own
+// --commit-large reaches the guard, so the override behaves identically in
+// both tools.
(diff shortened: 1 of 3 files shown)


obey-agent (commented):
## Summary

This is a clean, minimal public-API export: `StageOptions` aliases `git.StageOptions`, and `StageAllWithOptions` is `StageAllWithOutcome` with the same `ctx` short-circuit, `drainQuietly`, and call into `git.StageWithGuardOptions` that camp’s own `--commit-large` uses. Wiring matches existing re-export and staging patterns (type alias both directions, cancelled-context test). No correctness, edge-case, or contract issues found in the diff.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

**Verdict:** Approve — no issues found. Ready for fest to consume once published.

PR #524, opened 2026-07-30
feat(discovery): reclassify rail residents, bump wi list contract to v1alpha10 (WF0001 P3/S3)

lancekrogers:
WF0001 phase 003, sequence 03. `camp`'s festival discovery learns to split what it finds in the lifecycle folders, so a workitem promoted onto the rail (sequence 02, #521) is findable as itself again.

**Stack base:** branches off `main` after #521 merged. Not stacked on an open PR.

## Discovery reclassification

`discoverFestivals` emitted *every* directory under `festivals/` as a festival, so a promoted workitem vanished from its own type and could not be selected by id at all. It now classifies on the `.workitem` marker — camp owns the marker, fest owns `fest.yaml`:

| Directory | Result |
|-----------|--------|
| marker, on `ready`/`active` | resident: its own `workflow_type`, folder `lifecycle_stage`, `festivals/`-relative path and key |
| no marker | festival, exactly as before, including the marker-less humanized fallback |
| marker, in `planning`/`ritual`/`chains` | emitted as neither |

Classification is explicit, never "a directory with no festival markers." A resident is built through `buildWorkflowDirItem` with the type from its own marker, then only `LifecycleStage` is overridden to the folder it sits in. That reuse is the design: residents inherit `ApplyMetadata` (stable id, tags, projects) and `LoadLocalRun` (`.workflow/` run progress) identically to the same directory under `workflow/<type>/`.

The 88-line `discoverFestivals` split into four functions; the festival construction moved verbatim into `festivalItem`.

### Two decisions worth reading

**A stamped directory outside `ready`/`active` is emitted as nothing**, not fallen back to "festival." The marker proves the directory is camp's, and calling it a festival would push a workitem into `camp wi --type festival` and into fest's surfaces — worse than omitting it. An unreadable marker is treated the same way, for the same reason. Both are doctor candidates for sequence 04.

**Both markers present: resident wins**, logged at warn. `.workitem` records a promote camp performed; a leftover `fest.yaml` proves nothing about the current owner. A test asserts exactly one item is emitted, so the directory can never appear twice.

## JSON contract

`SchemaVersion` is now `workitems/v1alpha10` with a changelog entry. No new fields; the only observable change is that non-festival types can report `lifecycle_stage` `ready`/`active` with a `relative_path` under `festivals/`. `explore` gains `ready` in `stage_vocabulary`.

Doc 04 points at `internal/commands/workitem/json_contract.go`; that file holds per-subcommand versions. The list contract is `SchemaVersion` in `internal/workitem/json.go`.

No filter code changed. `--type` filters on `WorkflowType` and `--stage` on `LifecycleStage`, so composition already worked once discovery emitted the right fields. The tests verify that rather than assume it.

### Finding: `--stage active` does not mean "on the rail"

The sequence asked for a test that `--stage active` returns a resident while a root design item "is unaffected". That is false, and the first draft of two tests failed on it. `buildWorkflowDirItem` hardcodes `LifecycleStageActive` — the long-standing "directory workitems are active by location" rule:

```
active   design    festivals/active/x        <- resident
active   design    workflow/design/normal    <- root item
ready    explore   festivals/ready/e
```

So `--stage active` spans both homes; `ready` is the value only a resident can hold, and the real discriminator the rail adds. Pre-existing behavior, nothing to change, but a consumer reading `--stage active` as "on the rail" would be wrong. Both tests now assert reality and the unit fixture no longer misrepresents discovery.

## Verification

Real `camp wi --json` sample with a resident is in the festival results docs; the resident row carries `"workflow_type": "design"`, `"lifecycle_stage": "active"`, `"key": "design:festivals/active/x"`, `"stable_id": "design-x-2026-07-29"`, `"source_id": ""` — identical in shape to its `workflow/` twin.

The no-resident guarantee was verified by building `camp` from `origin/main` (`c3595279`) and from this branch, running both against byte-identical festival-only fixtures, and diffing normalized output:

```
-  "schema_version": "workitems/v1alpha9",
+  "schema_version": "workitems/v1alpha10",

changed lines: 2
```

7 integration tests (all resident assertions parse real command output), 7 filter-composition unit cases, 6 discovery unit tests.

`just gate` green: 3984/3984 unit, 4029/4029 with integration vet. `golangci-lint --new-from-merge-base origin/main`: 0 issues.

## Also here

A comment-density trim across the rail surface from sequence 02 and this one (net ~88 lines): rationale moved to commits and results docs, keeping short doc comments and the non-obvious "why". Includes `merged_backstop.go` from #520, which merged before I could trim it there.

### DIFF
--- a/internal/commands/fresh/merged_backstop.go
+++ b/internal/commands/fresh/merged_backstop.go
@@ -110,8 +110,6 @@ func handleMergedBackstop(ctx context.Context, out io.Writer, root, projectPath 
 // declined item offers a "skip all remaining" shortcut. A cancelled prompt
 // (Ctrl+C) is treated as skip. Every accept goes through the shared promote path
 // with EvidenceMergedBranch; nothing is promoted without an explicit accept.
-// Matches are also de-duplicated by workitem key so a multi-worktree item that
-// somehow still appears more than once is only prompted/promoted once.
 func promoteMergedMatches(ctx context.Context, out io.Writer, cfg *config.CampaignConfig, root string, matches []MergedBackstopMatch) {
 	skipAll := false
 	seen := map[string]bool{}
@@ -257,10 +255,7 @@ func MapMergedBranchesToWorkitems(ctx context.Context, cfg *config.CampaignConfi
 
 	projectScope := projectRelPath(root, projectPath)
 
-	// Signal 1: worktree-scope links, per pruned branch. One workitem often has
-	// several worktree links (stacked PRs / multi-branch designs); once any of
-	// those branches map to it, further branches for the same key are ignored
-	// so promote only runs once.
+	// Signal 1: worktree-scope links, deduped by workitem key.
 	matches, unmatchedBranches, matchedKeys := collectWorktreeLinkMatches(registry.Links, active, prunedBranches)
 
 	// Signal 2: WI- commit tags on commits newly reachable from the default
@@ -285,11 +280,9 @@ func MapMergedBranchesToWorkitems(ctx context.Context, cfg *config.CampaignConfi
 	return matches, nil
 }
 
-// collectWorktreeLinkMatches maps pruned branch names to still-active workitems
-// via worktree-scope link basenames. Each workitem key appears at most once in
-// the returned matches (first pruned branch that hits it wins). Branches with
-// no worktree-link match are returned as unmatched for the commit-tag signal.
-// Pure: no I/O.
+// collectWorktreeLinkMatches maps pruned branches to active workitems via
+// worktree-link basenames, at most once per key. Unmatched branches fall through
+// to the commit-tag signal. Pure.
 func collectWorktreeLinkMatches(all []links.Link, active []wkitem.WorkItem, prunedBranches []string) (matches []MergedBackstopMatch, unmatched []string, matchedKeys map[string]bool) {
 	matchedKeys = map[string]bool{}
 	for _, branch := range prunedBranches {
@@ -299,9 +292,8 @@ func collectWorktreeLinkMatches(all []links.Link, active []wkitem.WorkItem, prun
 			continue
 		}
 		if matchedKeys[wi.Key] {
-			// Same workitem, another of its worktrees just merged. One promote
-			// covers the item; do not re-queue it and do not treat the branch as
-			// unmatched (that would incorrectly open the commit-tag path).
+			// Another worktree of the same item merged. Not unmatched: that would
+			// wrongly open the commit-tag path.
 			continue
 		}
 		matches = append(matches, MergedBackstopMatch{
--- a/internal/commands/workitem/promote_rail.go
+++ b/internal/commands/workitem/promote_rail.go
@@ -19,14 +19,10 @@ import (
 	"github.com/Obedience-Corp/camp/internal/workitem/locate"
 )
 
-// festivalsDir is the campaign-relative folder the rail moves workitems into.
 const festivalsDir = "festivals"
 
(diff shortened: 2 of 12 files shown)


obey-agent (commented):
## Verdict: Comment (merge-ready) — one finding found and fixed during review

Traced `Discover` -> `discoverFestivals` -> `residentFromMarker` -> `buildWorkflowDirItem`, and `NewPayload` for the contract. Verified the PR's claims by running the binary, not by reading the diff.

### Found and fixed

**An enum value that lied at two of three call sites.** `dirIsOutOfStageResident` was returned for a marker that fails to load, for a genuinely out-of-stage directory, and for a failed `buildWorkflowDirItem`. All three mean "camp's directory, emit nothing" and the `slog` reason already distinguishes them, so the constant shouldn't have claimed one specific cause. Renamed to `dirIsSkipped`; behavior unchanged.

### Verified

**Classification is explicit.** `LoadMetadata` presence is the only signal — no "directory without FESTIVAL.md", no name matching, no ID sniffing. `meta == nil` is the sole path to festival classification. The inverse matters too and holds: an *unreadable* marker does not fall through to the festival path, so a YAML error cannot relabel camp's directory as a festival.

**Fields, from real output rather than the diff:**

```
design    active  festivals/active/x          design:festivals/active/x
explore   ready   festivals/ready/e           explore:festivals/ready/e
festival  active  festivals/active/realfest   festival:festivals/active/realfest
festival  active  festivals/active/bare-fest  festival:festivals/active/bare-fest
design    active  workflow/design/normal      design:workflow/design/normal
```

`buildWorkflowDirItem` reuse preserves StableID/Tags/Projects and run progress by construction, not by copying, and `TestDiscoverFestivals_ResidentMatchesWorkflowItemShape` diffs a resident against a real workflow twin — that's the test that would catch a future divergence.

**Contract additive and versioned.** Checked the no-resident guarantee by building both sides rather than trusting the changelog: `origin/main` (`c3595279`) vs this branch, identical festival-only fixtures, normalized, diffed — only the `schema_version` line moved. Every resident assertion parses real `camp wi --json` into a struct, so casing is verified by execution.

**No regression.** `festivalItem` is the pre-rail body moved verbatim; the marker-less humanized fallback is pinned to the literal `"Bare Thing"`; `_MixedStageFolderSplits` proves both kinds coexist in one folder.

**Standards.** `camperrors` only, ratchet clean; `context.Context` first on every filesystem function; `discoverFestivals` 88 lines -> four functions, largest 34; stage names derive from `LifecycleStage` so resolver, promote target, and `StagesByType` cannot drift.

### Two things the task documents got wrong, corrected in the branch

- It asked for "SourceID/StableID from the marker id". `ApplyMetadata` sets `StableID` only; `SourceID` is a festival concept carrying the `fest.yaml` id and correctly stays `""`. A resident must be shape-identical to its workflow twin.
- It implied a root design item does not match `--stage active`. It does — `buildWorkflowDirItem` hardcodes `LifecycleStageActive` ("active by location"). `ready` is the discriminator the rail actually adds. Both affected tests failed on the wrong assumption and now assert reality.

### Recorded, not blocking

- `Counts` tallies only intent/design/explore/festival, so a custom-type resident increments `total` but no per-type count. Pre-existing.
- A `.workitem` declaring `type: festival` would classify as a resident with festival type — harmless, unreachable via `camp workitem create`, but it is the one input where the two classifications converge.
- `gate-fast` reported 1/4029 failed once, then passed two full runs plus a direct `go test ./...`. Treated as a flake; not in packages this branch touches.

Gates: `just gate` PASSED (3984/3984 unit, 4029 with integration vet); `golangci-lint --new-from-merge-base origin/main` 0 issues.

PR #525, opened 2026-07-30
feat(rail): resident completion, demote to home, and resident doctor checks (WF0001 P3/S4)

lancekrogers:
WF0001 phase 003, sequence 04. The two ways a resident leaves the rail: completion into the festival-local dungeon, and demote back to its type root. Plus two doctor checks for the states the rail can leave behind.

**Stack base:** `fest/wf0001-p3s3-discovery-residents` (PR #524). Merge #524 first.

## Completion needed no code, but exposed an older gap

Verified before writing anything: `promote --target completed` on a resident already lands in the right place, because `MoveToDungeon` builds its service from `loc.DungeonPath` and sequence 02's resolver already points that at the festival tree.

```
from: festivals/active/thing
to:   festivals/.dungeon/completed/2026-07-29/thing
```

There is also **no separate phase-001 local-dungeon rule** to reconcile. Searched for one; `loc.DungeonPath` is the single mechanism.

What *was* broken predates the rail. Seeding a priority, a current pointer, and a link on a workitem and then shelving it left:

| Store | After shelve (before this PR) | |
|-------|------|---|
| links | released | correct, #503 |
| priority | **unchanged, keyed on the pre-move path** | orphaned |
| current | **unchanged, aimed at an invisible item** | stale |

The priority entry is keyed on the path, so nothing resolves it after the move and only a later `Prune` reaches it. The current pointer is stored by stable id so it survives any move by design, but pointing at a *shelved* item is the same silent failure #503 fixed for links: the selector cannot see dungeon items, so `camp p commit` quietly stops stamping the ref.

**Released, not re-keyed.** The task text said to re-home this state "matching the rail-move fixups". That predates #503 and contradicts it: a rail move re-homes because the item stays active, whereas a shelve *ends* its active life. A priority re-keyed onto a dungeon path would preserve state no surface reads. So `releasePathStateForShelvedSource` sits next to `releaseLinksForShelvedSource`, and both report rather than acting silently (`released_priority_key`, `cleared_current`).

This fixes every dungeon move, not just residents.

## `camp workitem demote`, not `promote --target home`

Doc 06 item 2 deferred the verb shape to phase 3. Chose a dedicated verb: every other promote target moves a workitem *forward*, so `home` would be the one target that inverts the verb, and it would have to sit inside the forward-only switch where it is the exception.

The duplication that argument was weighed against turned out to be thin, because the extraction was overdue anyway. `doRailPromote` had the move inlined and `runWorkitemPromote` had a 45-line tail inlined — both one copy-paste from drifting. Now:

- `applyWorkitemMove` — move + reference repair
- `finishWorkitemMove` — audit, ledger, commit, nav invalidation, output

Promote and demote both route through them, so they cannot diverge on how a move commits or what event it emits. `promote.go` is smaller than before despite gaining the shelve-release work.

Home is `workflow/<marker.type>/<slug>`: the marker stores a type, not an original path, which is sufficient because the type root is implied-active. Allowed only from `ready`/`active`; a dungeon source is a restore, a `workflow/` source is already home, and collisions are refused with the same `Lstat` guard rail entry uses. Audit/ledger `target` is `home` so the event stream distinguishes it.

Demote **re-homes** path-keyed state, the deliberate mirror of shelve releasing it.

## Doctor checks

`workitem.resident.unstamped` — a directory on a rail stage that is neither stamped nor a festival. `workitem.resident.missing-home` — a resident whose `workflow/<type>/` is gone, so demote has nowhere to land.

Both warnings, neither auto-fixable, and that is tested: camp cannot guess a type or invent a type root. The two read from different sources by necessity — an unstamped directory is never emitted as a resident, so it has to be found by scanning disk. `isFestivalDir` stats the marker names locally rather than importing fest.

## Verification

Every behavior driven against a real binary in throwaway fixtures. Full transcripts in the festival results docs; the demote run shows the directory back at the type root with the markdown cross-reference, link key, and priority key all re-homed, and all four rejection cases erroring with the source intact.

12 new integration tests (7 rail exits, 2 doctor, and the 3 from earlier sequences re-run together), 16 new unit tests.

`just gate` PASSED. `golangci-lint --new-from-merge-base origin/main` 0 issues.

```
$ go test -tags=integration ./tests/integration/ \
    -run "TestRailExit|TestDoctorResidents|TestPromoteRail|TestWorkitemJSON"
ok  	github.com/Obedience-Corp/camp/tests/integration	105.957s
```

## One requirement deliberately not met as written

The testing task asked to assert links/priority "follow a resident into the dungeon". They do not, for the reason above. The tests assert release on shelve and re-home on demote, in the same file, because the pair is what makes the policy legible.

### DIFF
--- a/cmd/camp/zz_manifest_annotations.go
+++ b/cmd/camp/zz_manifest_annotations.go
@@ -94,6 +94,7 @@ var manifestAgentAllowedReasons = map[string]string{
 	"workitem list":              "Read-only filtered listing with non-interactive compact and JSON output",
 	"workitem group":             "Non-interactive group update with explicit selector",
 	"workitem priority":          "Non-interactive priority update with explicit selector",
+	"workitem demote":            "Non-interactive demote off the festival rail, fully specified by flags",
 	"workitem promote":           "Non-interactive promote fully specified by --target and flags",
 	"workitem rename":            "Non-interactive rename fully specified by positional args and flags",
 	"workitem repair":            "Idempotent, non-destructive marker repair with --json and --dry-run",
--- a/docs/cli-reference/camp-reference.md
+++ b/docs/cli-reference/camp-reference.md
@@ -7733,6 +7733,43 @@ camp workitem current [selector] [flags]
 ```
 ---
 
+## camp workitem demote
+
+Move a rail resident back to its type root
+
+### Synopsis
+
+Take the workitem identified by [id], by cwd, or by the current pointer off the
+festival rail and back to its original workflow type root.
+
+A resident in festivals/ready or festivals/active returns to
+workflow/<type>/<slug>, where the type comes from its own .workitem marker. Every
+reference is repaired in the same commit, exactly as the rail move does.
+
+This is the escape hatch from the rail, not backward motion along it. It is
+rejected from a dungeon, because restoring a shelved workitem is not a demote,
+and from a workitem already at its type root.
+
+```
+camp workitem demote [id] [flags]
+```
+
+### Options
+
+```
+      --dry-run     Print the planned move, change nothing
+  -h, --help        help for demote
+      --json        Output result as a single JSON object
+      --no-commit   Skip the auto-commit
+```
+
+### Options inherited from parent commands
+
+```
+      --no-color   disable colored output
+```
+---
+
 ## camp workitem doctor
 
 Report link-registry health issues
--- a/docs/cli-reference/camp_workitem.md
+++ b/docs/cli-reference/camp_workitem.md
@@ -52,6 +52,7 @@ camp workitem [flags]
 * [camp workitem commits](camp_workitem_commits.md)	 - List commits referencing a workitem
 * [camp workitem create](camp_workitem_create.md)	 - Create workitem tracking metadata
 * [camp workitem current](camp_workitem_current.md)	 - Get, set, or clear the current workitem
+* [camp workitem demote](camp_workitem_demote.md)	 - Move a rail resident back to its type root
 * [camp workitem doctor](camp_workitem_doctor.md)	 - Report link-registry health issues
 * [camp workitem group](camp_workitem_group.md)	 - Set or clear the group
 * [camp workitem id](camp_workitem_id.md)	 - Print the identifier of a workitem
--- a/docs/cli-reference/camp_workitem_demote.md
+++ b/docs/cli-reference/camp_workitem_demote.md
@@ -1,0 +1,39 @@
+## camp workitem demote
+
+Move a rail resident back to its type root
+
+### Synopsis
+
+Take the workitem identified by [id], by cwd, or by the current pointer off the
+festival rail and back to its original workflow type root.
+
+A resident in festivals/ready or festivals/active returns to
+workflow/<type>/<slug>, where the type comes from its own .workitem marker. Every
+reference is repaired in the same commit, exactly as the rail move does.
+
+This is the escape hatch from the rail, not backward motion along it. It is
+rejected from a dungeon, because restoring a shelved workitem is not a demote,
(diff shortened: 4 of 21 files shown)


obey-agent (commented):
## Verdict: Comment (merge-ready) — two findings found and fixed during review

Traced `runWorkitemPromote` -> `doDungeonPromote` -> `MoveToDungeon`, and `runWorkitemDemote` -> `applyWorkitemMove` -> `finishWorkitemMove`. Verified behavior by running the binary against fixtures, not by reading the diff.

### Found and fixed

**1. Five consecutive string parameters on a shared function.** `finishWorkitemMove` took `ledgerID, ledgerRef, ledgerTitle, why, successVerb string`. Transposing any two compiles cleanly and silently produces a wrong audit event or wrong output — on the one function both promote and demote route through. Grouped into a `moveTail` struct.

**2. `fileExists` was called on a directory.** The homeless-resident check passes `workflow/<type>/` to it. `os.Stat` handles both, so this was a naming lie rather than a bug. Renamed to `pathExists`.

### Verified

**Completion destination.** `loc.DungeonPath` really is the only mechanism; I searched for a separate phase-001 local-dungeon rule and there is none (`resolveDungeonName` in quest and `resolveDungeonCommandContext` in cmd are unrelated). Real output:

```
from: festivals/active/thing
to:   festivals/.dungeon/completed/2026-07-29/thing
```

The integration test additionally asserts the resident did *not* land in `workflow/design/dungeon/`, which is the failure the local-dungeon rule exists to prevent.

**The shelve gap is real and predates the rail.** Seeded a priority, current pointer, and link on a workitem, shelved it, and read the stores back: links released (correct, #503), priority entry left orphaned under the pre-move key, current pointer left aimed at an item the selector cannot see. Both now released and reported.

**Release vs re-home is coherent, not arbitrary.** Shelve releases because it ends the item's active life; demote re-homes because the item stays active. Confirmed both against the binary — after demote the link key and priority key both read `design:workflow/design/widget`. The two tests sit in the same file so the policy is legible side by side.

**Demote guards.** All four refusals verified with the source left intact: already-home, from-dungeon, occupied destination, and dry-run changing nothing. `target=home` in the audit/ledger, so camp-graph can distinguish a demote from a rail move.

**Doctor checks.** Both fire with the right code, severity, and hint; three negatives hold (a `fest.yaml` directory, a stamped resident, a resident whose type root exists); and `--fix` leaves both alone without stamping the unclassifiable directory or creating the missing type root. `isFestivalDir` duplicates the marker names locally rather than importing fest, which is the right layering.

**Standards.** `camperrors` only, ratchet clean; `context.Context` first on every I/O function; every added function under 50 lines; `promote.go` is smaller than before this sequence despite gaining the shelve-release work, because the inlined move and 45-line tail were extracted.

### The extraction was overdue, not incidental

`doRailPromote` had the move inlined and `runWorkitemPromote` had the tail inlined. Both were one copy-paste from drifting, and adding a second mover would have made that concrete. `applyWorkitemMove` and `finishWorkitemMove` now serve both, so promote and demote cannot diverge on how a move commits or what event it emits. That is why the "second command surface" cost of a dedicated `demote` verb came out lower than doc 06 anticipated.

### One requirement deliberately not met as written

The testing task asked to assert links/priority "follow a resident into the dungeon". They must not — see above. The tests assert release on shelve and re-home on demote instead, which covers the requirement's actual intent (no stale path-keyed state) with the correct expectation.

Gates: `just gate` PASSED; `golangci-lint --new-from-merge-base origin/main` 0 issues; the phase-003 integration suites green together (105.957s).

**Base is `fest/wf0001-p3s3-discovery-residents`, not main.** Merge #524 first, and do not delete that branch before this PR is retargeted or merged.

PR #526, opened 2026-07-30
docs(help,dungeon): document the rail and surface workitem types in triage (WF0001 P5/S2)

lancekrogers:
WF0001 phase 005, sequence 02 (camp half). Documentation plus dungeon triage resident awareness. Base `main`.

The fest half is [fest#312](https://github.com/Obedience-Corp/fest/pull/312) — independent, either can merge first.

## Help text

`camp fresh --help` gains a `WORKITEM COMPLETION` section. The two settings that decide what happens to finished workitems were only discoverable by reading `settings.go`:

```
  completed_runs     Tier 1, once per run, campaign-root scoped. "sweep"
                     (default) promotes every workitem whose workflow run
                     completed, "report" prints a read-only banner, "off" skips
                     it. This is the same sweep 'camp workitem sweep' performs.
  merged_workitems   Tier 2, per project. A workitem whose branch merged is
                     inferred evidence, so it never auto-promotes: "prompt"
                     asks on a TTY and reports otherwise, "report" prints the
                     exact promote commands to run, "off" skips it.
```

Values and the "inferred evidence never auto-promotes" rule come from the doc comments at `settings.go:214-224`, verified rather than assumed.

`camp workitem promote --help` already listed `ready`/`active` from phase 3. Added the two things it lacked: `camp workitem demote` as the way off the rail, and the property that makes the rail safe to adopt — a workitem on a stage keeps its original type, so `camp wi --type design` still finds it.

## Dungeon triage shows a workitem's type

`DungeonItem.WorkitemType` is populated during parent triage from camp's own `.workitem` marker, and rendered beside the generic label rather than replacing it, so a non-workitem directory's line is byte-identical:

```
Type: directory (design) | Modified: 2026-07-30
```

**A gap the unit tests did not catch:** `outputDungeonJSON` remaps into a *local* struct, so a field added to `DungeonItem` never reaches `--json`. The unit tests passed while the real command still omitted it; only running the binary surfaced that. Fixed, commented at the struct so the next person sees why, and locked with a test.

`ListParentItems` is shared, so type-local triage (`workflow/design/`) gains the same field. Every directory workitem already carries a marker, so scoping this to `festivals/` would mean special-casing by path. Recorded deliberately.

## Two discrepancies with doc 04, both recorded rather than papered over

Doc 04 asks that "the inner crawl of `festivals/.dungeon/` present residents correctly when reviewing." Tracing current source shows **no such surface exists**:

1. `ListItems` skips every subdirectory (`service_list.go:75-79`), so the crawl never walks the dated status buckets where a completed resident lands. `Discover()` also excludes anything under a `dungeon/` segment, so completed items never resurface in `camp wi` either.
2. Triage does not reach `festivals/` either: the excluder lists `"festivals"` by name (`service_list.go:181`), and `camp dungeon add` from inside it resolves the campaign-root dungeon rather than creating one there. Real output from the fixture:

```
triage context from festivals/: No parent items eligible for triage (context: dungeon).
```

So type-local triage is the one real resident-touching surface, and that is where this lands. If a future phase adds a genuine "browse the completed bucket" command, that command is the one that should consume `WorkitemType`.

The integration test for the festivals case asserts this actual behavior rather than forcing a fixture onto a path that does not exist.

## Docs regen isolated

Two commits: help-text edits, then `docs: regenerate CLI reference` containing only `docs/cli-reference/*` (verified with `git show --stat`), per the campaign's rule that a regen must never hide unrelated drift inside a feature diff.

## Gate

`just gate` PASSED. Integration: `go test -tags=integration -run TestDungeonTriage_` ok.

### DIFF
--- a/cmd/camp/dungeon/list.go
+++ b/cmd/camp/dungeon/list.go
@@ -165,20 +165,24 @@ func outputDungeonSimple(items []intdungeon.DungeonItem) error {
 }
 
 func outputDungeonJSON(items []intdungeon.DungeonItem) error {
+	// This remaps rather than marshalling DungeonItem directly, so a new field on
+	// that struct does not reach --json until it is added here too.
 	type jsonItem struct {
-		Name    string `json:"name"`
-		Path    string `json:"path"`
-		Type    string `json:"type"`
-		ModTime string `json:"mod_time"`
+		Name         string `json:"name"`
+		Path         string `json:"path"`
+		Type         string `json:"type"`
+		ModTime      string `json:"mod_time"`
+		WorkitemType string `json:"workitem_type,omitempty"`
 	}
 
 	output := make([]jsonItem, 0, len(items))
 	for _, item := range items {
 		output = append(output, jsonItem{
-			Name:    item.Name,
-			Path:    item.Path,
-			Type:    string(item.Type),
-			ModTime: item.ModTime.Format(time.RFC3339),
+			Name:         item.Name,
+			Path:         item.Path,
+			Type:         string(item.Type),
+			ModTime:      item.ModTime.Format(time.RFC3339),
+			WorkitemType: item.WorkitemType,
 		})
 	}
 
--- a/cmd/camp/dungeon/list_test.go
+++ b/cmd/camp/dungeon/list_test.go
@@ -1,6 +1,13 @@
 package dungeon
 
-import "testing"
+import (
+	"bytes"
+	"encoding/json"
+	"os"
+	"testing"
+
+	intdungeon "github.com/Obedience-Corp/camp/internal/dungeon"
+)
 
 func TestBuildDungeonListContextLine(t *testing.T) {
 	tests := []struct {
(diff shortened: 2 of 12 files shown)


obey-agent (commented):
## Verdict: Comment (merge-ready)

**Merge-ready.** One defect was found and fixed during implementation (below);
review found nothing further. Posted as a Comment; the Approve verdict is Lance's.

## The defect worth naming: a field that never reached the contract

`outputDungeonJSON` remaps `DungeonItem` into a **local** `jsonItem` struct. Adding
`WorkitemType` to `DungeonItem` therefore did nothing for `--json`, and the unit
tests for `ListParentItems` passed the whole time. Only running the real command
showed the field missing.

This is the exact trap the requirement warned about ("confirm this command's JSON
contract status before assuming it is casual output"). Fixed, commented at the
struct so the next person adding a field sees the indirection, and locked by
`TestOutputDungeonJSON_IncludesWorkitemType`, which also asserts `omitempty` really
omits rather than emitting `""`.

## Requirements checked

**Everything documented was verified in source first**, per the anchors-are-claims
rule: the `ready`/`active` targets, `camp workitem sweep`, `camp workitem demote`,
and both `fresh.yaml` settings with their real values from `settings.go:214-224`.
The tier-1/tier-2 framing and "inferred evidence never auto-promotes" are quoted
from those doc comments, not invented.

**Additive, non-destructive rendering.** The workitem type goes in parens beside the
generic label rather than replacing it, so a plain directory's line is unchanged;
`omitempty` keeps a plain directory and a file byte-identical in `--json`. Both
pinned by tests.

**Best-effort marker read.** An unreadable marker warns and leaves the field empty,
matching this package's existing posture, tested by
`TestListParentItems_UnreadableMarkerIsNotFatal`.

**Shared-function consequence recorded, not hidden.** `ListParentItems` serves
type-local triage too, so `workflow/design/` gains the same field. Special-casing by
path would have been worse; the requirement asked for this to be stated explicitly
and it is.

**Docs regen isolated.** `git show --stat` on the regen commit shows only
`docs/cli-reference/*`, and every changed line traces to the two help edits. fest
produced no regen commit because `just docs` yielded no diff — recorded rather than
manufacturing an empty one.

**Source of truth respected.** fest's methodology edits are in
`projects/fest/methodology/`, and this campaign's scaffolded `festivals/.festival/`
copy is verified untouched.

## Two discrepancies with doc 04, both recorded

Doc 04's "inner crawl of `festivals/.dungeon/` must present residents correctly" has
no surface to attach to:

1. `ListItems` skips every subdirectory, so the crawl never walks the dated status
   buckets, and `Discover()` excludes anything under a `dungeon/` segment.
2. Triage does not reach `festivals/` either — the excluder lists it by name, and
   `camp dungeon add` from inside it resolves the campaign-root dungeon.

Type-local triage is the one real surface. The integration test asserts that actual
behavior rather than forcing a fixture onto a path that does not exist, which is the
right call: a passing test over an unreachable code path would have been worse than
no test.

## Gate

```
camp:  just gate                                        -> PASSED
       go test -tags=integration -run TestDungeonTriage_ -> ok
fest:  just test unit                                    -> 2751 passed
```

PR #527, opened 2026-07-30
feat(commit): report staged size findings when a commit skips staging

lancekrogers:
## The gap

camp's size and bulk guard lives at the **staging** chokepoint (`internal/git`, the `StageWithGuard` family). A commit that stages nothing never reaches it.

So this sequence is silent:

```
git add media/diagram.png      # 13.25 MB, staged by hand
camp commit --all=false -m "..."
```

The commit lands with the file in it and camp prints no report line: no tracked-growth warning, no threshold named, no remedy. The same file committed through the default `camp commit` path produces the full report. The raw-`git add` gap is designed (design doc 05 open question 4 defers a pre-commit hook for it), but a scoped commit going completely quiet about it is not.

## The fix: a commit-time report-only pass

When a commit skips staging, camp now evaluates the existing index against the same guard thresholds and prints the same tracked-growth report lines the staging path emits.

It **only reports**. It never excludes, unstages, blocks, or prompts, and it never declares an artifact root or writes a gitignore rule. The commit that follows is byte-for-byte the commit that would have happened without the check.

That is not a first increment; it is the design, and it follows from the friction-free principle in CLAUDE.md section 7:

- **"Never reintroduce curation friction into the normal path."** An index the user assembled with `git add` is a deliberate act. A guard that excluded or asked here would convert it back into a decision point, which is exactly what camp's commit default exists to avoid.
- **"When something must protect the user, camp handles it and reports."** Here camp cannot handle it, because handling would mean reversing an explicit instruction. So it does the other half: it reports.
- Excluding is also mechanically wrong at this point. The staging guard excludes by staging *around* a path, which cannot un-add something already in the index. `internal/stageguard/enumerate.go:151-160` already documents this for a hand-staged file reached through the normal path, and reaches the same conclusion: report it and commit it. This change makes the two paths agree instead of leaving one silent.

Everything the pass finds is therefore `TrackedGrowth`, which is *defined* as report-and-commit. The existing `renderTrackedGrowth` is reused verbatim, so a user hitting this over a hand-built index reads the same copy as a user hitting it through `camp commit`.

Wired into all three commit surfaces that have an `--all` flag: `camp commit`, `camp project commit`, `camp worktrees commit`.

## Cost

Latency is friction too, so the check is two git processes regardless of how many files are staged: `git diff --cached --raw -z --no-renames --no-abbrev` to name the entries and their object ids, then one batched `git cat-file --batch-check` to size them. Nothing is opened, read, or hashed; `--batch-check` reads object headers, so a staged gigabyte costs what a staged kilobyte costs.

Measured on macOS, 100 iterations per figure, git 2.54:

| staged entries | added pass |
| --- | --- |
| 1 | 6.75 ms |
| 400 | 7.43 ms |
| 5000 | 9.35 ms |

Two reference points on the same 400-entry fixture:

- `git status --porcelain=v1 -z -uall`, which the **default** commit path already runs on every commit: **7.26 ms**. The added pass costs about what one call camp already makes costs.
- One `git cat-file -s` spawn per staged file, i.e. what a naive per-file design would cost: **2688 ms**. That is the 360x reason for the batching.

The near-flat curve is the point: cost is dominated by two process spawns, not by file count.

## Design details worth reviewing

**Sizes come from the index, not from `lstat`.** The blob is what actually enters the repository, so it is the more correct number: content filters have already run. This has a useful consequence. git-LFS replaces a large file with a pointer of a few hundred bytes, so an LFS-managed path falls under any threshold by construction, and this pass needs none of the `git check-attr` call `FilterLFSManaged` makes on the worktree side.

**Non-blob index entries are filtered by destination mode.** Deletions (`000000`) have no destination content. Submodule gitlinks (`160000`) name a commit in the submodule's object database rather than the parent's, which matters because the campaign root stages these routinely; there is a test for it. Symlinks (`120000`) store their target path, not the target's bytes.

**`git diff --cached` rather than `git diff-index HEAD`.** The former works before the first commit (an unborn HEAD diffs against the empty tree), and a campaign's first commit is exactly when someone is likely to add something large.

**Fail-open, never silently.** A repository camp cannot read still commits; the cause is printed. Same stance as `GuardUnavailableLine` on the staging path, and for the same reason: a user who is not told assumes the check ran.

**`--commit-large` suppresses detection entirely**, matching the staging path, so the flag means one thing across both.

## Tradeoffs and things deliberately left out

- **The bulk guard does not run in this pass.** Bulk *blocks*, and a pass that stages nothing must not. Reporting bulk instead would need new remedy copy for a situation camp has already allowed to happen, and its entire existing vocabulary is a refusal ("Nothing was staged, here are your three choices"), which would be false here. That is a design question rather than the silence this closes.
- **No `--json` schema change.** Tracked growth has never appeared in `commit/v1alpha1` (`applyGuardHandling` folds in exclusions and declared roots only), and surfacing it for the scoped path alone would make the document inconsistent about the same finding. The report is human output and goes where the rest of the human output goes: `humanOut`, which is stderr under `--json`. There is a test asserting stdout stays exactly one JSON document while the report appears on stderr. Adding a tracked-growth field to the schema is a separate, deliberate decision.
- **`renderTrackedGrowth` suggests `camp artifacts add <dir>` even in project scope**, where camp otherwise refuses to auto-declare roots. That is pre-existing on the staging path, which renders tracked growth identically regardless of scope; reusing the copy inherits it rather than introducing it. Changing shared copy felt like the wrong scope for this PR.
- `--amend` without an explicit `--all` resolves to not staging, so it now runs the pass. A plain amend has an empty index diff and reports nothing; content actually staged for the amend is reported, which is correct. A file already in HEAD is not re-reported, since the comparison is index against HEAD.

## Tests

Unit, `internal/stageguard/staged_test.go`, pure parsing and policy logic:
- `parseDiffRawZ` across modification, addition, executable mode, deletion, gitlink, symlink, a path with a space and a quote, a malformed record, and a skipped record not desynchronizing the next one.
- `parseBatchCheck`, including that a `missing` object is left out of the map rather than recorded as zero bytes (zero would pass every check).
- `uniqueOIDs` deduplication.
- The mode-off gate returning before touching the repository, and context cancellation on both entry points.
- That a staged candidate always classifies as `TrackedGrowth` and stays in the commit, and that the allowlist reaches this pass.

Unit, `cmd/camp/cmdutil/stagedreport_test.go`: `--commit-large` suppressing everything without running git, fail-open printing its cause, and the unavailable line naming both cause and next step.

Integration, containerized, `tests/integration/stage_guard_scoped_commit_test.go` (8 tests):
- The reported incident: `git add` an over-threshold file, `camp commit --all=false`, assert the report **and** that the file is committed, and that nothing was declared or gitignored.
- Under threshold: no report line at all.
- `--commit-large`: no report, file committed.
- Allowlisted path: no report, file committed.
- `--json`: stdout parses as exactly one document, report on stderr, `excluded` and `artifact_roots_declared` both empty.
- A staged submodule gitlink does not fail the check or cost the report the files beside it.
- `camp project commit --all=false` reports too, so the gap does not just move to project scope.
- A 400-entry index commits whole and stays silent.

## Gates

- `just gate` (build, vet stable/dev/integration, lint stable/dev, ratchets, unit tests dev and stable): **PASSED**. 4053/4053 dev, 4008/4008 stable, 0 lint issues, no new host-FS-test or `fmt.Errorf` violators.
- `just test integration-run 'TestIntegration_ScopedCommit|TestIntegration_ScopedProjectCommit'`: **8/8 PASS** (Docker via colima).

No new dependencies. `camperrors` throughout, `context.Context` first on every I/O path, all functions under 50 lines.

### DIFF
--- a/cmd/camp/cmdutil/stagedreport.go
+++ b/cmd/camp/cmdutil/stagedreport.go
@@ -88,9 +88,19 @@ func stagedGrowth(
 // It names the cause for the same reason GuardUnavailableLine does: the
 // realistic causes need different responses from the user, and a bare "check
 // unavailable" gives them nothing to act on.
+//
+// Every clause is in a tense that is true at the moment it prints, and none of
+// them claims a commit. This runs before the commit object is written, and the
+// flow after it can still end in "nothing to commit", a refusal over
+// pre-staged submodule refs, a deferred message writer, or a git failure.
+// GuardUnavailableLine can afford "Staged without..." because staging really
+// has happened by then; the equivalent past tense here would announce an
+// outcome camp does not yet know, which is the same wrong-by-assertion problem
+// the line exists to avoid.
 func StagedCheckUnavailableLine(cause error) string {
 	return fmt.Sprintf(
-		"Committed without the size check: %v\n"+
-			"  Large files were not checked. Run 'camp doctor -c bigfiles' if you want to look.",
+		"Could not run the size check over the staged files: %v\n"+
+			"  Large files were not checked, and camp does not stop for that.\n"+
+			"  Run 'camp doctor -c bigfiles' if you want to look.",
 		cause)
 }
--- a/cmd/camp/cmdutil/stagedreport.go
+++ b/cmd/camp/cmdutil/stagedreport.go
@@ -1,0 +1,96 @@
+package cmdutil
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+	"github.com/Obedience-Corp/camp/internal/stageguard"
+	"github.com/Obedience-Corp/camp/internal/ui"
+)
+
+// What the user sees when a commit stages nothing.
+//
+// The staging guard lives at the staging chokepoint, so `camp commit
+// --all=false` over an index the user built with git add never reaches it: a
+// 13 MB PNG could enter a commit without camp saying a word. The report below
+// is the whole of the fix.
+//
+// Report-only is the design, not a first increment. Camp's commit default is
+// "commit everything without having to think about it", and a guard that
+// excluded or asked here would turn an index the user assembled deliberately
+// back into a decision point, which is what the default exists to avoid. Acting
+// on the contents would also mean reversing an explicit `git add` -- the one
+// thing the staging guard already refuses to do for hand-staged files, since it
+// excludes by staging around a path and cannot un-add what is already in.
+//
+// So the commit that follows this report is byte-for-byte the commit that would
+// have happened without it. What changes is only that the user is told.
+
+// ReportStagedGrowth writes the guard's findings for an index camp did not
+// stage, and returns them. It excludes nothing, unstages nothing, blocks
+// nothing, and prompts for nothing.
+//
+// commitLarge is the user having already answered this. The staging guard
+// suppresses detection entirely under --commit-large rather than
+// detecting-then-ignoring, and so does this, so the two agree about what the
+// flag means.
+//
+// It is fail-open, like the guard at the staging chokepoint: a repository camp
+// cannot read still commits. The cause is printed rather than swallowed,
+// because the dangerous way to degrade is the quiet one -- a user who is not
+// told assumes the check ran.
+func ReportStagedGrowth(
+	ctx context.Context,
+	out io.Writer,
+	repoPath string,
+	commitLarge bool,
+) []stageguard.GuardViolation {
+	if commitLarge {
+		return nil
+	}
+
+	violations, limits, err := stagedGrowth(ctx, repoPath)
+	if err != nil {
+		_, _ = fmt.Fprintln(out, ui.Warning(StagedCheckUnavailableLine(err)))
+		return nil
+	}
+	// Same renderer as the staging path. A user who hits this over a hand-built
+	// index and a user who hits it through `camp commit` are looking at the same
+	// situation, and must not have to notice that the wording differs.
+	for _, v := range violations {
+		renderTrackedGrowth(out, v, limits)
+	}
+	return violations
+}
+
+// stagedGrowth resolves the repository's guard policy and evaluates the index
+// against it.
+func stagedGrowth(
+	ctx context.Context, repoPath string,
+) ([]stageguard.GuardViolation, stageguard.GuardLimits, error) {
+	limits, err := stageguard.ResolveLimits(ctx, repoPath)
+	if err != nil {
+		return nil, limits, camperrors.Wrapf(err, "resolve staging guard limits for %s", repoPath)
+	}
+	violations, err := stageguard.CheckStagedIndex(ctx, repoPath, limits)
+	if err != nil {
+		return nil, limits, camperrors.Wrapf(err, "evaluate staged files in %s", repoPath)
+	}
+	stageguard.SortViolations(violations)
+	return violations, limits, nil
+}
+
+// StagedCheckUnavailableLine says the size check did not run over an index camp
+// did not stage.
+//
+// It names the cause for the same reason GuardUnavailableLine does: the
+// realistic causes need different responses from the user, and a bare "check
+// unavailable" gives them nothing to act on.
+func StagedCheckUnavailableLine(cause error) string {
+	return fmt.Sprintf(
+		"Committed without the size check: %v\n"+
+			"  Large files were not checked. Run 'camp doctor -c bigfiles' if you want to look.",
+		cause)
+}
--- a/cmd/camp/cmdutil/stagedreport_test.go
+++ b/cmd/camp/cmdutil/stagedreport_test.go
@@ -1,0 +1,56 @@
+package cmdutil
+
+import (
+	"bytes"
+	"context"
+	"strings"
+	"testing"
+
+	camperrors "github.com/Obedience-Corp/camp/internal/errors"
+)
+
+// --commit-large is the user having already answered the question, so the pass
+// must not even ask git. A repository path that cannot be read proves it
+// returned before running anything.
+func TestReportStagedGrowthCommitLargeSuppressesEverything(t *testing.T) {
+	var out bytes.Buffer
+
+	violations := ReportStagedGrowth(context.Background(), &out, "/nonexistent/repo", true)
+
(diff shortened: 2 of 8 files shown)


cmd/camp/cmdutil/stagedreport.go
```diff
@@ -0,0 +1,96 @@
...
+
+// StagedCheckUnavailableLine says the size check did not run over an index camp
+// did not stage.
+//
+// It names the cause for the same reason GuardUnavailableLine does: the
+// realistic causes need different responses from the user, and a bare "check
+// unavailable" gives them nothing to act on.
+func StagedCheckUnavailableLine(cause error) string {
+	return fmt.Sprintf(
+		"Committed without the size check: %v\n"+
```
obey-agent: **[bug]** `StagedCheckUnavailableLine` says `"Committed without the size check: ..."`, but `ReportStagedGrowth` runs well before any commit object is written (and even when no commit will occur). On `camp commit`, the call is at the end of the non-staging branch, then the flow still does staged summary, pre-staged submodule-ref validation, `HasChanges`, optional auto-write deferral, and only then `executor.Commit`. So a fail-open path can print past-tense "Committed without..." and immediately follow with "Nothing to commit, working tree clean", a `pre_staged_refs` validation error, a deferred auto-write enqueue (no commit yet), or a failed commit. That breaks the same "never quiet, and name the truth" rule the PR inherits from `GuardUnavailableLine` — and unlike that helper, which is rendered after staging has already completed (`"Staged without..."`), this text asserts an outcome that has not happened.

**Suggestion:** Use wording that does not claim a commit landed, parallel to the staging path's accuracy, e.g. `"Could not run the size check: %v\n  Large files were not checked. ..."`, `"Proceeding without the size check: ..."`, or `"Size check skipped: ..."`. Update `TestReportStagedGrowthFailsOpenAndSaysSo` / `TestStagedCheckUnavailableLineNamesTheCause` accordingly. Optionally also run the report only after there is something to commit (after the `HasChanges` gate) so a pure no-op never pays for or speaks about the check, but the tense fix alone is enough for correctness of the copy.

obey-agent (changes requested):
## Summary

This PR closes a real gap: when `camp commit` / `project commit` / `worktrees commit` skip staging (`--all=false` or amend default), the staging-time size guard never ran, so a hand-`git add`ed over-threshold blob could land silently. The new commit-time pass is report-only, reuses `renderTrackedGrowth` and the allowlist/`--commit-large`/ModeOff gates, sizes index blobs via two batched git processes, and is well covered by unit and integration tests (including JSON stdout purity, gitlinks, project scope, and a wide index). Residual risk is low: one user-facing fail-open string claims a commit has already happened before the commit path runs, so a later no-op or error can contradict it.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

lancekrogers:
Review finding addressed in 3e1c4a28. StagedCheckUnavailableLine no longer claims a commit: the line now reads 'Could not run the size check over the staged files: <cause>' followed by 'Large files were not checked, and camp does not stop for that' plus the doctor pointer. It runs before the commit object exists and the flow can still end in a no-op, a pre-staged refs refusal, a deferred enqueue, or a git failure, so every clause is now in a tense that is true at the moment it prints; a comment on the function pins that rule. The unit test asserts the cause and next step are present and additionally that no clause contains 'Committed'/'committed'/'Commit succeeded'; the integration tests' negative assertions were updated to the new string. go test -count=1 on cmd/camp/cmdutil and internal/stageguard plus go vet: pass.

lancekrogers:
Good catch, and the finding is right. `ReportStagedGrowth` runs at the end of the non-staging branch, well before `executor.Commit`, and the flow between them can end without a commit in at least four ways: `HasChanges` reporting a clean tree, the `pre_staged_refs` validation error, a deferred auto-write enqueue, or git failing. Past tense asserted an outcome camp does not know yet, which is exactly the failure mode the line exists to prevent.

Fixed in 3e1c4a2. The copy now reads:

```
Could not run the size check over the staged files: <cause>
  Large files were not checked, and camp does not stop for that.
  Run 'camp doctor -c bigfiles' if you want to look.
```

Every clause is true at the moment it prints. The first states what happened and why, the second states camp's fail-open policy rather than predicting an outcome, and the third keeps the next step. `GuardUnavailableLine` can afford "Staged without..." because staging really has completed by the time it renders; that asymmetry is now written into the doc comment so the tense does not get "corrected" back later.

Tests:

- `TestReportStagedGrowthFailsOpenAndSaysSo` and the two integration `NotContains` assertions updated to the new string.
- Added `TestStagedCheckUnavailableLineClaimsNoCommit`, which fails if the line ever contains "Committed"/"committed"/"Commit succeeded". That is the regression guard for this specific bug rather than just a re-pin of the new wording.

I did not take the optional suggestion to move the report after the `HasChanges` gate, for two reasons. It would put the report after `ShowStagedSummaryTo` and the pre-staged-refs check, so the same finding would appear in a different position depending on whether camp or the user did the staging, and keeping those two readings identical is the reason the report reuses `renderTrackedGrowth` in the first place. It also would not fully solve the tense problem on its own, since a commit can still fail after the gate. On cost: when nothing is staged, `EnumerateStaged` returns after one `git diff --cached` without running `cat-file`, so a pure no-op already pays about 3 ms and already says nothing. Happy to move it if you would rather have the ordering change anyway.

Gates rerun after the fix: `just gate` PASSED (two consecutive full runs), `go test ./cmd/camp/cmdutil/... ./internal/stageguard/...` pass, and the 8 containerized integration tests pass.

One thing I want to flag rather than bury: during this round one `just gate` invocation reported `1/4054 TESTS FAILED` in the dev-profile unit suite. I truncated that output and lost the test name, and it has not reproduced since across 3 uncached `go test -tags dev ./...` runs, 14 runs of the exact `BUILD_TAGS=dev just test unit` command that failed, a `-count=4` stress of the timing-sensitive packages (drain, jobs, defercommit, sync, clone, git, peer, machines), and 2 further complete `just gate` runs. So it is 1 failure in 18 executions of the same suite on identical code. I cannot name it and I am not going to assert it is pre-existing without evidence, but the only delta in this round was one string constant plus three test assertions, all covered by deterministic tests. Worth knowing there is an intermittent test somewhere in the dev suite.

PR #528, opened 2026-07-30
Fix deferred bookkeeping commits failing on an unborn HEAD

lancekrogers:
## Bug

`camp init` creates the `.git` directory but never commits, so a fresh campaign has an unborn HEAD until the user's first `camp commit`. Any deferred bookkeeping commit enqueued before that (`camp intent add` is the real-world trigger) died in the background worker with:

```
git read-tree: Not a valid object name HEAD
```

and landed in the failed lane silently. `camp jobs drain` still reported success, because draining only waits for a lane to empty, not for its jobs to succeed, so nothing told the user their intent was never committed. This surfaced via CA0004's commit_drain fixture against camp v0.4.0-rc.1, which worked around it by giving the fixture a baseline commit.

## Mechanism of the fix

`ReadTreeIntoTempIndex` (`internal/git/commit.go`) unconditionally ran `git read-tree HEAD` to seed the worker's temp index. On an unborn branch HEAD is a ref with no object behind it, so that read fails outright — it's a different precondition from a normal, non-empty repository, not a smaller version of it.

The fix checks whether HEAD resolves (`git rev-parse --verify --quiet HEAD`) and, when it does not, seeds the temp index with `git read-tree --empty` instead. The job's captured paths then stage into that empty index exactly as before, and the resulting commit becomes the repository's first commit. This one function is shared by both worker paths that build a temp index from HEAD (`CommitScoped`, for hand-written path-only jobs, and `CommitBlobs`, the production path a captured-content bookkeeping job like `camp intent add` takes), so both are fixed by the same change.

I audited the rest of the worker's temp-index mechanism for the same class of HEAD-resolution assumption, since an unborn branch breaks any code that assumes HEAD exists:

- `IsGitlink` (`git ls-tree HEAD`) and `FirstParentChainContains` (`git log HEAD`) already return their safe default (`false`) on error, which is the correct answer on an unborn branch — nothing can already be a gitlink or already applied when there is no history yet.
- `ExpandTrackedPathsFromTempIndex`/`ExpandTrackedPaths` (`git diff --cached`) already work: git natively diffs an unborn HEAD against the empty tree, verified directly.
- `ResetIndexToHead` and the plain `git commit` (`executeCommit`) only run after the worker's own commit has landed, at which point HEAD is no longer unborn regardless of how the job started.
- `git.FullHash`, used by the separate `KindCommitTree` job path (`--auto-write` deferred commits, not bookkeeping), does fail on an unborn HEAD, but `TryDeferAutoWrite` already has a pre-existing, intentional fallback for exactly this: a failed `Enqueue` degrades to a synchronous commit with a warning, so it doesn't lose the commit — it just doesn't defer on someone's very first `camp commit --auto-write`. That's a pre-existing design, not something I added, and I left it alone.

None of those needed a change. `camp init` is untouched on purpose — it must stay side-effect-light, and the fix belongs in the mechanism that assumed history exists, not in preventing the state where it doesn't.

## Tests

- `internal/git/commit_test.go`: `TestHeadResolvable`, `TestReadTreeIntoTempIndex_UnbornHead`, `TestCommitScoped_UnbornHead`, `TestCommitBlobs_UnbornHead` — host-git unit tests (this package's existing convention, no container) against a freshly `git init`'d repo with zero commits. Confirmed each fails with the exact reported error when the fix is reverted, and passes with it applied.
- `tests/integration/bookkeeping_defer_test.go`: `TestIntegration_BookkeepingCommitLandsOnUnbornHead` — containerized end-to-end repro: raw `camp init` (no baseline commit, unlike `setupDrainCampaign`/`InitCampaign` which paper over this bug), `camp intent add`, drain, then assert the job is not in the failed lane, HEAD now resolves, exactly one commit exists, and it contains the captured intent file. Confirmed this fails against the pre-fix code with `-count=1` (job lands in `failed/`, HEAD stays unresolved) and passes with the fix.

## Test commands run

- `go build ./...` — clean
- `go test ./internal/git/...` — pass (30.6s, includes the 4 new unit tests)
- `go test ./internal/jobs/... ./internal/defercommit/...` — pass
- `go test -tags integration ./tests/integration/ -run 'TestIntegration_Bookkeeping|TestIntegration_Drain|TestIntegration_CampNoDefer|TestIntegration_DeferredCommit' -count=1` — 16/16 pass (Docker via Colima)
- `go test -tags integration ./tests/integration/ -run 'TestIntegration_.*Job|TestIntegration_.*Scope|TestIntegration_.*Commit' -count=1` — 103/103 pass
- `just gate` — PASSED (stable+dev build, vet stable+dev+integration, lint stable clean including the errcheck ratchet, 113 packages / 4001 unit tests passed)

## Caveats

- The `git.FullHash`/`KindCommitTree` gap noted above is real (an unborn-HEAD `--auto-write` commit prints an unnecessary "Could not queue this commit" warning before falling back to synchronous) but does not fail or lose work, and reworking that job's `Parent` field to mean "no parent" would be a larger, separate change to that job schema's validation (`executeCommitTree` currently treats an empty `Parent` as a malformed job). Left out of scope; flagging here in case it's worth its own follow-up.
- No new dependencies, no changes to `camp init`, no synchronous fallback added — this fixes the temp-index mechanism itself, per the task brief.

### DIFF
--- a/internal/git/commit.go
+++ b/internal/git/commit.go
@@ -197,14 +197,39 @@ func CopyFile(src, dst string) error {
 }
 
 // ReadTreeIntoTempIndex seeds tmpPath from HEAD.
+//
+// `camp init` creates the .git directory but never commits, so every scoped or
+// blob-captured job enqueued before the user's first commit runs against an
+// unborn branch: HEAD is a ref that does not resolve to any object yet. Seeding
+// from "HEAD" there is not a smaller version of the normal case, it is a
+// different precondition, and `git read-tree HEAD` fails loudly rather than
+// treating the branch as merely empty. Starting from an empty tree instead
+// lets the paths this job stages become the repository's first commit, which
+// is what should happen: the content was captured (or named) when the job was
+// enqueued, and nothing about an unborn branch makes that content less real.
 func ReadTreeIntoTempIndex(ctx context.Context, repoPath, tmpPath string) error {
-	cmd := exec.CommandContext(ctx, "git", "-C", repoPath, "read-tree", "HEAD")
+	args := []string{"-C", repoPath, "read-tree", "HEAD"}
+	if !headResolvable(ctx, repoPath) {
+		args = []string{"-C", repoPath, "read-tree", "--empty"}
+	}
+	cmd := exec.CommandContext(ctx, "git", args...)
 	cmd.Env = append(os.Environ(), "GIT_INDEX_FILE="+tmpPath)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		return camperrors.NewGit("read-tree", "", "", strings.TrimSpace(string(out)), err)
 	}
 	return nil
+}
+
+// headResolvable reports whether HEAD points at a real commit.
+//
+// A freshly initialized repository has a HEAD ref but no commit behind it
+// (an unborn branch), and `git read-tree HEAD` treats that as an error rather
+// than an empty tree. This is the one check the temp-index seeding needs to
+// tell the two states apart.
+func headResolvable(ctx context.Context, repoPath string) bool {
+	cmd := exec.CommandContext(ctx, "git", "-C", repoPath, "rev-parse", "--verify", "--quiet", "HEAD")
+	return cmd.Run() == nil
 }
 
 // AddPathsToTempIndex stages paths into the provided temp index.
--- a/internal/git/commit_test.go
+++ b/internal/git/commit_test.go
@@ -1124,3 +1124,146 @@ func TestResetPathspecPayloadLimitReservesFixedArgv(t *testing.T) {
 		t.Fatalf("windows budget = %d, want %d", platformCommandLineBudget(), 16*1024)
 	}
 }
+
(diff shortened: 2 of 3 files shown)


obey-agent (commented):
## Summary

PR #528 correctly fixes deferred bookkeeping commits on an unborn HEAD by seeding the temp index with `git read-tree --empty` when `HEAD` does not resolve, instead of always running `git read-tree HEAD`. That is the right place to fix the failure: `ReadTreeIntoTempIndex` is shared by `CommitScoped`, `CommitBlobs`, and the selective `stageAndCommit` path, so one change covers the production bookkeeping worker and related scoped commits. Unit coverage (predicate, seed, `CommitScoped`, `CommitBlobs`) plus a container integration test that uses raw `camp init` (not the baseline-commit helpers) pins the real-world repro well. Residual risk is low and mostly out of scope: silent drain success on failed jobs remains as described, and the pre-existing `KindCommitTree` / unborn-HEAD `--auto-write` enqueue fallback is acknowledged in the PR body.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #531, opened 2026-08-01
feat(machine): hop to a campaign from the fleet screen with enter

lancekrogers:
## Why

`camp machine` listed the fleet but had no way into it. The only pointer to `camp switch <id>:<campaign>` lived in the `?` overlay, so an operator who opened the screen looking for a way onto a machine found the fleet, the socket state, a connection tester, and no door.

## What

`enter` on a machine row opens a campaign picker for that machine and hops there.

```
╭──────────────────────────────────────────────────────────────────────╮
│  Hop to buildbox                                                     │
│  From the last snapshot; press r to ask the machine now.             │
│                                                                      │
│  › notes                                                             │
│    obey-campaign                                                     │
│    platform                                                          │
│                                                                      │
│  j/k move · enter hop · r refresh · esc cancel                       │
╰──────────────────────────────────────────────────────────────────────╯
```

**Nothing about the hop is new.** The screen writes `ssh-hop:<machine>:<campaign>` to `--path-output` and the shell wrapper turns that into `camp switch <sel> --shell-connect` — byte-for-byte the path `camp list`'s picker already takes. The far machine's own registry still resolves the root; no remote resolution is duplicated here.

Campaign names come from the completion snapshot first, which is what makes the picker open without dialing. `r` refreshes live; a cold cache pays for one fetch, and that fetch warms the cache so the next open is instant.

## Decisions worth reviewing

**`enter` was bound to test-connection.** It now hops; `t` keeps the tester. This is the one user-visible regression in the set — anyone with muscle memory on `enter`-to-test gets a campaign picker instead. `enter` is the universal "go there" and the footer already led with `t test connection`, so this seemed the right way round, but it is a live rebinding and worth a second opinion.

**Hopping needs the wrapper, and says so when it is missing.** No subprocess can replace the shell that launched it. Without `--path-output` the key reports `hop needs shell integration: run eval "$(camp shell-init <shell>)"` rather than appearing to work — the same shape `camp list` uses for the same reason.

**Refusals happen before the list opens, not after a choice.** `local` (this computer) and password-auth machines are rejected at `enter`, so no picker ever leads nowhere.

**Unreachable ≠ empty.** A failed fetch shows the error and points at `camp machine diagnose <id>`. An empty list would read as "this machine has no campaigns", which is a different and wrong claim.

## Verification

Green tests do not verify a TUI, so this was driven in a real pty (`pyte`) against a fixture `XDG_CONFIG_HOME`, reading the rendered screen:

- fleet screen renders, footer reads `enter hop · t test connection · …`
- `enter` on `local` → `✗ local is this computer; use 'camp list' to open a campaign here`
- `enter` on `buildbox` → picker above, from the snapshot, no ssh dialed
- `down down` + `enter` → `--path-output` file contains exactly `ssh-hop:buildbox:platform`
- without `--path-output` → `✗ hop needs shell integration: run eval "$(camp shell-init <shell>)"`

Wrapper arms driven end-to-end in a pty with a stub `camp` on PATH:

| invocation | zsh | bash |
| --- | --- | --- |
| `camp machine` | hops via switch | hops via switch |
| `camp machine list` | passthrough | passthrough |
| `camp machine add devbox --host x` | passthrough | passthrough |
| `camp machine diagnose devbox` | passthrough | passthrough |
| `camp machine --help` | passthrough | passthrough |

`just gate`: 4080 dev + 4035 stable tests pass, lint clean. `just docs` regenerated only the two machine reference files.

## Gaps

- **fish is asserted but not executed.** fish is not installed on this machine, so its arm is covered by the generator test and shape-matched against the existing `list` arm, not run. `TestGenerateFish_ValidSyntax` skips for the same reason and will cover it on a box that has fish.
- **No integration test performs a real hop.** The seam is `camp switch`, which is already covered; this PR tests up to the payload and the wrapper's handling of it.

## Note for reviewers

The first version of the wrapper test file was named `machine_arm_test.go`, which Go silently excluded from the package — `_arm` reads as a `GOARCH=arm` build constraint. It reported `ok` with zero tests running. Renamed to `machine_wrapper_test.go`. Worth knowing generally: a `_test.go` file whose name ends in `_<goarch>` or `_<goos>` disappears without warning.

### DIFF
--- a/cmd/camp/machine_tui_update.go
+++ b/cmd/camp/machine_tui_update.go
@@ -23,6 +23,8 @@ func (m *machineTUIModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return m, nil
 	case devicesMsg:
 		return m.applyDevices(msg)
+	case hopCampaignsMsg:
+		return m.applyHopCampaigns(msg)
 	case spinner.TickMsg:
 		if !m.busy() {
 			return m, nil
@@ -160,7 +162,9 @@ func (m *machineTUIModel) updateBrowse(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 			return m, nil
 		}
 		return m, m.beginScan(true)
-	case "t", "enter":
+	case "enter":
+		return m.openHopPicker()
+	case "t":
 		return m, m.testSelected()
 	case "e":
 		return m, m.openEditForm()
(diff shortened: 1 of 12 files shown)


cmd/camp/machine_tui_update.go
```diff
@@ -177,6 +181,117 @@ func (m *machineTUIModel) updateBrowse(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
...
+		m.hop.loading = true
+		m.hop.err = ""
+		target, ok := m.machineByID(m.hop.machineID)
+		if !ok {
+			m.hop.loading = false
+			return m, nil
+		}
+		return m, tea.Batch(m.spin.Tick, m.fetchHopCampaigns(target, m.hop.gen))
+	case "enter":
+		if m.hop.loading || len(m.hop.campaigns) == 0 {
```
obey-agent, line 253: **[bug]** `updateHop`'s `enter` only guards `loading` and `len(campaigns)==0`. It does not check `m.hop.err`. After a successful open (cache or live) the campaigns slice stays populated; a later `r` refresh that fails sets `hop.err` and returns without clearing `campaigns` (`applyHopCampaigns` at the error branch). `hopBody` then renders only the error + "r retry · esc cancel" (no list, no "enter hop"), but pressing `enter` still builds `hopSelection`, sets `quitting`, and `tea.Quit`s. j/k also still move the hidden cursor while the error (or loading) UI is up, so the hop target can even be a campaign the operator never saw after the failure. Repro sketch: open hop from cache → `r` → inject fetch error → `enter` → path-output gets `ssh-hop:<id>:<prior-campaign>`.

**Suggestion:** Treat error (and ideally loading with a hidden list) as non-selectable: e.g. `if m.hop.loading || m.hop.err != "" || len(m.hop.campaigns) == 0 { return m, nil }` on enter, and skip j/k when `err != ""` or `loading`. Alternatively clear `campaigns` (and cursor) when recording a fetch error so the empty-list guard applies. Add a regression test: cache hit → failed refresh → enter must not set `hopSelection` or quit.

cmd/camp/machine_tui_view.go
```diff
@@ -431,9 +431,9 @@ func (m *machineTUIModel) statusLine() string {
 // equal weight. The action that answers "does this work" comes first, because
```
obey-agent, line 431: **[nit]** The footer comment still says the action that answers "does this work" comes first, but the strings now lead with `enter hop` and put `t test connection` second. The rebinding is intentional and documented in the PR; only the comment is stale.

**Suggestion:** Reword to match the new primary gesture (hop first; connection test remains a separate key).

obey-agent (changes requested):
## Summary

Solid, well-framed feature: `enter` opens a campaign picker, reuses the existing `ssh-hop:` / shell-wrapper path that `camp list` already uses, refuses local and password-auth before the list opens, and has strong unit coverage for cache vs live fetch, stale gen drops, and the wrapper arm shape. The shell templates mirror the list arm carefully (subcommand passthrough, TTY gate, hop via `switch --shell-connect`). One real correctness gap: after a failed refresh the error UI is shown, but `enter` (and j/k) still act on the stale in-memory campaign list, which contradicts the visible help and can quit into a hop from a failure screen.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 1

PR #535, opened 2026-08-03
fix(jobs,drain): make the attempt count honest and stop reporting commands waiting

lancekrogers:
Follow-up to #534. Two more queue defects from the same live-campaign diagnosis.

## The attempt count contradicted itself

`camp jobs` and `camp jobs --json` described the same job two incompatible ways:

```
failed  job-20260803T185441Z-3a0b  commit-tree  writing a commit message in . (gave up after 1 attempt)
```
```json
{ "id": "job-20260803T185441Z-3a0b", "state": "failed", "attempts": 0 }
```

Only reclaim incremented `Attempts`. A job that ran and returned an error went straight to `failed/` recording nothing, so it was parked reading 0 beside copy saying it had already been tried and given up on. An agent reading `--json` could not tell a failure that had been retried from one that never was, and the two are shown identically.

An attempt is now a job that started and did not finish, whichever way that happened: the worker died (reclaim, as before) or the job returned an error (new). Parking an already-exhausted job still counts nothing, because parking is not a run and counting it would push the total past `MaxAttempts` and claim a try that never happened.

## Reporting commands no longer wait on the queue

`camp status` held the terminal for up to `DrainTimeout` (30s) so its answer could be marginally fresher. That is the exact cost deferral exists to remove, spent on the command people run constantly and often twice in a row.

It now prints what is still queued and returns:

```
1 commit still queued; what follows may not include it (camp jobs)
```

The caveat survives because it was the real point. A tree camp is midway through changing still has to be declared, so a status that looks stale is explained rather than merely wrong. What changes is that the user is told immediately instead of thirty seconds later, and re-runs when they want a fresh read. That is what the command is for.

**The notice still spawns workers for unserved lanes.** Spawning is the cheap half of a drain and the half that makes "a queued job is always eventually served" true; only the wait is gone. Dropping it would leave a user whose worker died, and who only ever runs `camp status`, being told about the same jobs forever and never seeing them move. `TestIntegration_DrainSpawnsAWorkerForAnUnservedLane` covers this, rewritten to poll rather than depend on status having blocked (see review round below).

`camp log` gets the same treatment for the same reason.

`camp doctor` waits only under `--fix`; plain `doctor` reports and does not wait. The rule is now that the wait follows what a command does rather than which command it is. The `--fix` path keeps today's `Read` mode, so on timeout it warns and repairs anyway. That is deliberate, and the reasoning is in the code: a deferred job commits content captured at enqueue (an immutable tree, or blobs), so it never reads the working tree when it runs, and a repair made while a job is queued cannot change what that job commits. Refusing would have cost `doctor --fix --json` its report on any busy queue to guard a case that does not arise.

`--no-drain` is unchanged as a flag and still suppresses the queue interaction on these commands; its help text on `camp status` now says "report" rather than "wait".

## The guard was strengthened, not weakened

`TestEveryHistoryMovingCommandDrains` previously required a `drain.Repo`/`AllLanes`/`CampaignRoot` call in every listed command, which `status` and `log` would now fail. Rather than dropping those rows, the matrix gained a duty per surface:

- `mustDrain` for history-movers: must wait, as before.
- `mustNotice` for reporters: must call `drain.Note`, **and must not** call a blocking drain.
- `mustDoBoth` for doctor, the one dual-duty surface: must have both, so the wait cannot quietly become unconditional again (a 30s wait back on every doctor) and the notice branch cannot be dropped.

So a reporting command that ignores the queue fails, one that reintroduces a wait fails, and doctor losing either half fails. Before, only the first was caught.

Source matching still cannot prove which call sits under which branch, so doctor's duty is asserted behaviorally as well, in the integration test below.

## Verification

- `just gate`: build, vet across stable/dev/integration, lint 0 issues both profiles, 4095 + 4050 unit tests pass. The new timing-sensitive tests were stressed at -count=25 and the jobs package at -count=6, clean.
- New: `TestIntegration_StatusReportsTheQueueWithoutWaiting` wedges a lane that can never complete (fresh lock, job naming a nonexistent path) and asserts status returns well inside the drain timeout, says what is queued, names `camp jobs`, and does not claim to be waiting. Under the old behavior that setup sat for the full 30s.
- New: `TestIntegration_DoctorWaitsOnlyWhenItRepairs` wedges a lane and asserts plain doctor returns promptly with the notice while `--fix` waits and refuses with its escapes. It runs 35s because the `--fix` path genuinely waits out the timeout.
- Unit: attempt recording on execution failure, reclaim and execution accumulating together, parking not counting a run, parking an unparseable job still emptying `running/`, the count landing on the destination file rather than the source, notice copy including singular/plural agreement, silence on an empty queue, and context cancellation.
- Ran the doctor, workitem, dungeon, shell-init and rename integration groups in isolation; all pass.

The full integration suite reports 2/734 failing. That is pre-existing: unmodified `main` at dc258b42 fails the identical 2/734, one of them `TestLeverage_ConfigValidationPeople`. Not introduced here and not addressed here.

## Review round

Three findings from obey-agent, all correct, fixed in fb0fa80b:

1. `TestIntegration_DrainSpawnsAWorkerForAnUnservedLane` was racing. Its assertions were guaranteed by status *blocking*, and this PR removed the block while leaving them. It now polls for the lane to drain while running no camp command, so if status had not spawned a worker the poll times out.
2. `doctor --fix` used `Read` mode, which warns and proceeds, contradicting the comment above it. Correct finding; I first fixed the wrong side of it by switching to `Write` (refuse on timeout), then reverted in 1dda5870 and corrected the comment instead. The claim in that comment was false: deferred jobs commit captured content and never read the working tree at execution, so the repair cannot affect what a queued job commits. Refusing would have cost `--json` consumers their report to guard a case that does not arise.
3. The drain matrix only half-guarded doctor. It now carries a dual duty requiring both call sites, verified by reintroducing the unconditional wait and confirming the matrix fails on it.

Two further defects found on my own re-read, fixed in 4d6d64a0:

4. The attempt count was written **in place** into the running job file before the move. An interrupted write there leaves unparseable JSON in `running/`, and `Reclaim` skips what it cannot read, so the job would strand with nothing coming for it. Nothing else in the package edits a running job in place and `Reclaim` already had the safe shape. Parking now writes the incremented copy to its destination and removes the source after, so a crash duplicates a job rather than corrupting one.
5. The notice read `1 queued commit still queued`. It reused `countPhrase`, whose wording is built for a refusal and stutters here. This line prints on ordinary commands, so it is the one users see most.

### DIFF
--- a/cmd/camp/doctor.go
+++ b/cmd/camp/doctor.go
@@ -106,15 +106,23 @@ func runDoctor(cmd *cobra.Command, args []string) error {
 	}
 
 	// Doctor inspects working-tree and commit alignment, so a queued commit
-	// makes it report a discrepancy camp is already fixing. Read-only, so a
-	// slow queue warns and the checks still run.
+	// makes it report a discrepancy camp is already fixing.
+	//
+	// The wait follows what doctor is about to do rather than which command it
+	// is. Reporting proceeds on a notice, because a finding the user can
+	// re-check costs less than holding their terminal. --fix writes, and must
+	// not repair a tree against a view of it camp is still changing.
 	var doctorDrainWaited time.Duration
 	if !doctorOpts.noDrain {
-		waited, err := drain.AllLanes(ctx, campRoot, drain.Read)
-		if err != nil {
+		if doctorOpts.fix {
+			waited, err := drain.AllLanes(ctx, campRoot, drain.Read)
+			if err != nil {
+				return err
+			}
+			doctorDrainWaited = waited
+		} else if _, err := drain.NoteAllLanes(ctx, campRoot); err != nil {
 			return err
 		}
-		doctorDrainWaited = waited
 	}
 
 	// Build doctor with options
--- a/cmd/camp/doctor.go
+++ b/cmd/camp/doctor.go
@@ -110,12 +110,18 @@ func runDoctor(cmd *cobra.Command, args []string) error {
 	//
 	// The wait follows what doctor is about to do rather than which command it
 	// is. Reporting proceeds on a notice, because a finding the user can
-	// re-check costs less than holding their terminal. --fix writes, and must
-	// not repair a tree against a view of it camp is still changing.
+	// re-check costs less than holding their terminal.
+	//
+	// --fix writes, so it waits in Write mode and refuses if the queue outlasts
+	// the wait. Read mode would warn and repair anyway, which is the thing this
+	// path must not do: a repair computed against a tree camp is midway through
+	// changing can "fix" what camp was about to commit. A refusal that names
+	// its own way out costs the user a rerun; a wrong repair costs them a
+	// diff they have to understand later.
 	var doctorDrainWaited time.Duration
 	if !doctorOpts.noDrain {
 		if doctorOpts.fix {
-			waited, err := drain.AllLanes(ctx, campRoot, drain.Read)
+			waited, err := drain.AllLanes(ctx, campRoot, drain.Write)
 			if err != nil {
 				return err
 			}
--- a/cmd/camp/doctor.go
+++ b/cmd/camp/doctor.go
@@ -112,16 +112,20 @@ func runDoctor(cmd *cobra.Command, args []string) error {
 	// is. Reporting proceeds on a notice, because a finding the user can
 	// re-check costs less than holding their terminal.
 	//
-	// --fix writes, so it waits in Write mode and refuses if the queue outlasts
-	// the wait. Read mode would warn and repair anyway, which is the thing this
-	// path must not do: a repair computed against a tree camp is midway through
-	// changing can "fix" what camp was about to commit. A refusal that names
-	// its own way out costs the user a rerun; a wrong repair costs them a
-	// diff they have to understand later.
+	// --fix writes, so it waits, and on timeout it warns and repairs anyway.
+	//
+	// That last part is deliberate rather than an oversight. Refusing would
+	// buy almost nothing: a deferred job commits content captured at enqueue
+	// (an immutable tree, or blobs), so it never reads the working tree when
+	// it runs, and a repair made while it is queued cannot change what it
+	// commits. Only a job carrying paths and no blobs falls back to the live
+	// tree, which means one written by an older camp or by hand. Refusing
+	// every --fix on a busy queue to cover that would cost more than it saves,
+	// including a --json consumer that would get no report at all.
 	var doctorDrainWaited time.Duration
 	if !doctorOpts.noDrain {
 		if doctorOpts.fix {
-			waited, err := drain.AllLanes(ctx, campRoot, drain.Write)
+			waited, err := drain.AllLanes(ctx, campRoot, drain.Read)
 			if err != nil {
 				return err
 			}
--- a/cmd/camp/drain_matrix_test.go
+++ b/cmd/camp/drain_matrix_test.go
@@ -21,34 +21,60 @@ import (
 // invariant is noticing at compile-and-test time that a history-moving
 // entrypoint has no drain call in it, and that is a question about the code.
 
-// drainMatrixEntry is one history-moving surface and where its drain lives.
+// queueDuty is what a surface owes the deferred queue.
+type queueDuty int
+
+const (
+	// mustDrain is for commands that change or publish history. They wait,
+	// because a queued commit landing in the middle of what they do, or being
+	// left behind by it, is invisible to the user.
+	mustDrain queueDuty = iota
+	// mustNotice is for commands that only report. They do not wait: holding
+	// a user's terminal to make `camp status` marginally fresher is a bad
+	// trade against telling them what is still queued and letting them re-run.
+	// They still have to say so, or a report silently omits work camp has
+	// already promised, which is the confusion deferral must not create.
+	mustNotice
+)
+
+// drainMatrixEntry is one surface and where its queue handling lives.
 type drainMatrixEntry struct {
 	// name is how a failure names the surface, in user terms.
 	name string
-	// file is the source file that must contain the drain call.
+	// file is the source file that must contain the call.
 	file string
-	// fn is the function the drain must appear inside, so moving the call to
-	// an unrelated helper in the same file does not satisfy the test.
+	// fn is the function the call must appear inside, so moving it to an
+	// unrelated helper in the same file does not satisfy the test.
 	fn string
+	// duty is what this surface owes: a wait, or an acknowledgement.
+	duty queueDuty
 }
 
-// The surfaces. Adding a history-moving command means adding a row here; a row
-// with no drain fails, which is the point.
+// The surfaces. Adding a command that reads or writes git history means adding
+// a row here; a row that does neither of its duties fails, which is the point.
 var drainMatrix = []drainMatrixEntry{
-	{name: "camp commit", file: "commit.go", fn: "runCommit"},
-	{name: "camp stage", file: "stage.go", fn: "runStage"},
-	{name: "camp push", file: "push.go", fn: "runPush"},
-	{name: "camp push all", file: "push_all.go", fn: "runPushAllCmd"},
-	{name: "camp pull", file: "pull.go", fn: "runPull"},
-	{name: "camp pull all", file: "pull_all.go", fn: "runPullAllCmd"},
-	{name: "camp status", file: "status.go", fn: "runStatus"},
-	{name: "camp log", file: "log.go", fn: "runLog"},
-	{name: "camp sync", file: "sync.go", fn: "runSync"},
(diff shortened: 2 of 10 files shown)


cmd/camp/status.go
```diff
@@ -82,11 +82,13 @@ func runStatus(cmd *cobra.Command, args []string) error {
...
-	// first thing a user runs to ask "did that commit?", so reporting a tree
-	// camp is midway through changing is a correctness bug, not a nicety.
+	// Status reports; it does not wait. It is the first thing a user runs to
+	// ask "did that commit?", and the honest answer to that while the queue is
+	// still working is the count, given now, not the same answer given thirty
+	// seconds later. A tree camp is midway through changing still has to be
+	// declared, which is what the notice does.
 	if !statusNoDrain {
-		if _, err := drain.Repo(ctx, target.Path, drain.Read); err != nil {
+		if _, err := drain.Note(ctx, target.Path); err != nil {
```
obey-agent, line 91: **[bug]** Switching `runStatus` from `drain.Repo` (wait up to DrainTimeout) to `drain.Note` (spawn + return) leaves `TestIntegration_DrainSpawnsAWorkerForAnUnservedLane` racing. That test calls `camp status`, then immediately asserts `pending == 0` and that the commit is already in `git log`. `Note` only kicks `EnsureServed` and returns; the detached worker may still hold the job in pending/running. The PR body cites this test as covering the spawn half of the design and "passes unchanged," but under the new semantics it is timing-dependent and can flake under load or a slow container.

**Suggestion:** Update that integration test so status only proves the kick (e.g. notice text / worker lock / spawn), then poll or run an explicit wait (`camp jobs drain`) before asserting the lane is empty and the commit landed. Do not rely on status's former blocking drain.

cmd/camp/doctor.go
```diff
@@ -106,15 +106,23 @@ func runDoctor(cmd *cobra.Command, args []string) error {
...
+	// The wait follows what doctor is about to do rather than which command it
+	// is. Reporting proceeds on a notice, because a finding the user can
+	// re-check costs less than holding their terminal. --fix writes, and must
+	// not repair a tree against a view of it camp is still changing.
 	var doctorDrainWaited time.Duration
 	if !doctorOpts.noDrain {
-		waited, err := drain.AllLanes(ctx, campRoot, drain.Read)
-		if err != nil {
+		if doctorOpts.fix {
+			waited, err := drain.AllLanes(ctx, campRoot, drain.Read)
```
obey-agent, line 128: **[suggestion]** For `--fix`, the code still calls `drain.AllLanes(..., drain.Read)`. Read mode warns and proceeds on timeout, so `doctor --fix` can still repair against a tree camp is mid-changing after a full wait. That contradicts the new comment just above ("--fix writes, and must not repair a tree against a view of it camp is still changing") and the PR description's stronger invariant. Behavior is not a regression from pre-PR doctor (always Read), but this change elevates the claim without enforcing it.

**Suggestion:** Use `drain.Write` (or another refuse-on-timeout mode) when `doctorOpts.fix` is set, so a stuck queue aborts the repair path the same way push/sync would refuse. Keep `NoteAllLanes` for plain doctor.

cmd/camp/drain_matrix_test.go
```diff
@@ -21,35 +21,61 @@ import (
...
+	{name: "camp commit", file: "commit.go", fn: "runCommit", duty: mustDrain},
+	{name: "camp stage", file: "stage.go", fn: "runStage", duty: mustDrain},
+	{name: "camp push", file: "push.go", fn: "runPush", duty: mustDrain},
+	{name: "camp push all", file: "push_all.go", fn: "runPushAllCmd", duty: mustDrain},
+	{name: "camp pull", file: "pull.go", fn: "runPull", duty: mustDrain},
+	{name: "camp pull all", file: "pull_all.go", fn: "runPullAllCmd", duty: mustDrain},
+	{name: "camp sync", file: "sync.go", fn: "runSync", duty: mustDrain},
+	// doctor --fix repairs a tree, so it waits; plain doctor reports and does
+	// not. The wait follows what the command does, not which command it is.
+	{name: "camp doctor", file: "doctor.go", fn: "runDoctor", duty: mustDrain},
```
obey-agent: **[suggestion]** Doctor is classified only as `mustDrain`. That checks that `drain.AllLanes(` appears somewhere in `runDoctor`, not that plain doctor notices without waiting and `--fix` waits. Someone can make `AllLanes` unconditional again (reintroducing a wait on every doctor) or drop the notice branch while leaving a dead `AllLanes` string, and the matrix still passes. The PR claims the matrix was strengthened in both directions for reporters; doctor is the dual-duty surface and is only half-guarded.

**Suggestion:** Encode the dual path explicitly (e.g. require both a drain call and a notice call in `runDoctor`, or split into two matrix rows / a small source-structure assertion that `AllLanes` sits under `doctorOpts.fix` and `NoteAllLanes` under the else). At minimum, extend the mustDrain case for doctor so an unconditional wait fails.

obey-agent (changes requested):
Submitted from PENDING: 1 bug, 2 suggestions. See inline comments.

lancekrogers:
All three findings were correct and are fixed in fb0fa80b.

**The spawn test was racing.** You are right, and my PR body was wrong to cite it as covering the spawn half "unchanged". It passed, but only because the container happened to be fast: the assertions on `pending == 0` and the commit being in `git log` were guaranteed by status *blocking*, and I removed the block while leaving them.

Rather than have status prove the kick directly (any immediate observation of a detached worker is racy too), it now polls for the lane to drain, deliberately running no camp command while it waits. No worker exists before status, so if status had not spawned one nothing in the system would and the poll runs out its deadline. That asserts the property instead of the timing. Polled in the test goroutine rather than via `require.Eventually`, since the count is read through a container-shell helper that fails the test on error and that is not valid from the goroutine Eventually would run it in.

**`doctor --fix` in Read mode.** Correct, and the comment I added made it worse by claiming a guarantee the mode does not provide: Read warns and proceeds, so the repair still ran against the tree camp was mid-changing, just 30 seconds later. Now `drain.Write`, so it refuses. Doctor defines `--no-drain`, so the refusal's "continue anyway" line is an escape the command actually has.

**The matrix half-guarded doctor.** Also correct. `mustDrain` only checked that a blocking call appeared somewhere in `runDoctor`, which stays true if the wait becomes unconditional again (a 30s wait back on every doctor) or if the notice branch is dropped. Doctor now carries a `mustDoBoth` duty requiring both call sites. I checked the guard actually guards by reintroducing the unconditional wait and confirming it fails:

```
--- FAIL: TestEveryHistoryMovingCommandDrains/camp_doctor
    camp doctor reports by default but never mentions the deferred queue.
    Either the notice was dropped, or the wait became unconditional and every camp doctor now pays for the queue again.
```

On your "at minimum" point: matching source strings still cannot prove which call sits under which branch, so the duty is now asserted behaviorally as well. `TestIntegration_DoctorWaitsOnlyWhenItRepairs` wedges a lane and checks plain doctor returns promptly carrying the notice, while `--fix` waits and refuses with its escapes. That test takes 35s precisely because the `--fix` path really does wait out the timeout before refusing.

`just gate` green: 0 lint issues both profiles, 4093 + 4048 unit tests. The three affected integration tests pass.

lancekrogers:
Following up on finding 2, because I initially fixed the wrong side of it and want the record straight.

You were right that the code contradicted its comment. I resolved that by changing the code to `drain.Write` (refuse on timeout). That was wrong, and I've reverted it in 1dda5870 and corrected the comment instead.

The comment's claim, which I wrote and you reasonably took at face value, was false. A deferred job commits content captured at enqueue: `commit-tree` carries an immutable tree, and `commit-paths` carries blobs (`executeCommitPaths`). Neither reads the working tree when it runs. So a repair made while a job is queued cannot change what that job commits. The only path that reads the live tree is a `commit-paths` job with no blobs, which means one written by an older camp or by hand.

So the refusal guarded a case that does not arise in current camp, and it cost something real: `camp doctor --fix --json` would have returned no report at all on any queue busy past the timeout, which is a contract change for every agent calling it.

Findings 1 and 3 stand as fixed. The dual-duty matrix entry stays, and the integration test now measures the timing that actually separates the two paths (reporting returns immediately, repairing waits) rather than asserting a refusal.

Two further defects I found on my own re-read afterwards, fixed in 4d6d64a0, both introduced by this branch:

- The attempt count was written **in place** into the running job file before the move. An interrupted write there leaves unparseable JSON in `running/`, and `Reclaim` skips what it cannot read, so the job would strand with nothing coming for it. `Reclaim` already had the crash-safe shape and I had not followed it. Parking now writes the incremented copy to its destination and removes the source after.
- The notice read `1 queued commit still queued`. It reused `countPhrase`, whose wording is built for a refusal and stutters here.

PR #537, opened 2026-08-04
feat(intent): notes tree, folders, meetings (CI0009)

lancekrogers:
## Summary

Implements festival **CI0009** (camp intent notes TUI and meetings) for the camp side:

1. **Generalized group nesting** — replace Dungeon-only collapse fields with `Depth` / `Children` / `DescendantCount` / per-group `Expanded` so Notes can share the mechanism.
2. **Note folder discovery + CRUD** — `NoteFolders` walk, `CreateNoteFolder` / `RenameNoteFolder` / `DeleteNoteFolder` / `MoveNoteToFolder`, CLI `camp idea notes folders` with `intent-note-folders/v1alpha1` JSON.
3. **Notes tree in explorer** — Notes collapsed by default below Active / above Dungeon; nested notes under `notes/reading/papers/` are no longer dropped; `h`/`l` expand/collapse; filter auto-expands ancestors and hides empty shells.
4. **Folder management TUI** — `F` create, `R` rename folder, `d` delete empty folder; `m` on notes opens folder picker (lifecycle `m` unchanged).
5. **Capture destination** — `AddOptions.NoteFolders` / `AddResult.NoteFolder`; CLI `--folder` / `--create-folder`.
6. **Meeting import** — `ImportMeeting` + `camp idea notes import-meeting` with `.transcripts/` sidecars and in-place re-import; `ExtractMeetingIntents` for idempotent extraction.

**Out of scope for this PR:** repointing festival-voice (separate repo) to call `import-meeting` — camp consumer API is ready.

## Test plan

- [x] `go test ./internal/intent/... ./internal/intent/tui/... ./cmd/camp/intent/`
- [x] Manual CLI: `camp idea notes folders --json`, `folders add/rm/mv`, `import-meeting` fixture
- [ ] Pty: explorer Notes collapsed under Active, expand to nested note, folder F/R/d, note `m` picker
- [ ] After merge: festival-voice PR to call `camp idea notes import-meeting`

Festival: `festivals/active/camp-intent-notes-tui-and-CI0009`  
Worktree: `projects/worktrees/camp/camp-intent-notes-tui-and-ci0009`


internal/intent/tui/explorer/model.go
```diff
@@ -385,67 +414,145 @@ func groupIntentsByStatus(intents []*intent.Intent, dungeonExpanded bool) []Inte
...
 		intents = append(intents, item)
 	}
 
-	groups := groupIntentsByStatus(intents, dungeonExpanded)
-	return append([]IntentGroup{notes}, groups...)
+	lifecycle := groupIntentsByStatus(intents, dungeonExpanded)
+	// lifecycle: [Inbox, Ready, Active, Dungeon, Done, Killed, Archived, Someday]
+	top := lifecycle[:3]
+	dungeonPart := lifecycle[3:]
+	notesTree := buildNotesTreeGroups(notes, nil, foldState, omitEmpty)
```
obey-agent: **Bug — empty user folders never appear in the explorer tree.**

`buildNotesTreeGroups` is called with `folders == nil`, so it always synthesizes structure from *notes currently loaded*. `synthesizeFoldersFromNotes` only invents folder nodes for statuses that already have notes (plus reserved meetings/archived).

Consequences for the PR’s own TUI surface:
- `F` creates a folder on disk, but after reload the header never shows until a note lands in it.
- `R` / `d` require `selectedNoteFolder()` on a visible header — empty folders cannot be renamed or deleted from the explorer.
- Nested create under a just-created empty folder is impossible from the tree.

The move picker and add-destination correctly call `service.NoteFolders` and *do* see empty folders — only the tree path is broken.

**Fix:** load `NoteFolders` (cache on the model or fetch in `rebuildStatusGroups` / load path) and pass them as the second argument here (and in `groupNotes`). Keep `omitEmpty=true` only for filter views so empty shells stay hidden while searching.

internal/intent/tui/explorer/notes_view.go
```diff
@@ -23,20 +23,16 @@ var convertTypeOptions = []struct {
...
 	}
-	for _, n := range notes {
-		idx := 0
-		if n.Status == intent.StatusNoteArchived {
-			idx = 1
-		}
-		groups[idx].Intents = append(groups[idx].Intents, n)
-	}
-	return groups
+	return buildNotesTreeGroups(notes, nil, state, omitEmpty)
```
obey-agent: **Same bug in notes mode** — `groupNotes` also passes `nil` folders, so the dedicated Notes view has the same empty-folder blindness. Wire `NoteFolders` through here as well once the model loads them.

internal/intent/notes_folders.go
```diff
@@ -0,0 +1,562 @@
...
+	if _, statErr := os.Stat(toDir); statErr == nil {
+		return camperrors.Wrap(ErrNoteFolderExists, string(toStatus))
+	} else if !os.IsNotExist(statErr) {
+		return camperrors.Wrap(statErr, "checking destination note folder")
+	}
+
+	if err := os.MkdirAll(filepath.Dir(toDir), 0o755); err != nil {
+		return camperrors.Wrap(err, "creating parent of destination folder")
+	}
+	if err := os.Rename(fromDir, toDir); err != nil {
```
obey-agent, line 423: **Bug — no guard against renaming a folder into its own descendant.**

Example: `from=reading`, `to=reading/papers` (or any path where `to` is `from` or under `from/`). `os.Rename` of a directory into a path beneath itself is undefined/fails on Unix and can produce a mess if the destination parent was created via `MkdirAll` above.

Reject when `toStatus == fromStatus` (already handled) **or** `strings.HasPrefix(string(toStatus)+"/", string(fromStatus)+"/")` (and the reverse if you want a clearer error for “cannot rename into ancestor” — the exists check usually catches that). Add a unit test for `reading` → `reading/nested`.

obey-agent (changes requested):
## Review (obey-agent)

Large, well-structured CI0009 implementation: folder path validation (kebab-case, no traversal), reserved meetings/archived injection, move-to-folder with atomic write+rollback, and the generalized nest fields for Dungeon+Notes are solid. Tests cover a lot of the service surface.

Two high-conviction issues block the advertised TUI folder workflow and make rename unsafe in an edge case.

### Bugs
1. **Explorer never feeds `NoteFolders` into the tree** — empty user folders are invisible, so `R`/`d` (and nested `F` under a new folder) cannot target them. See inlines on `model.go` / `notes_view.go`.
2. **`RenameNoteFolder` allows renaming a folder into its own descendant** — classic `os.Rename` footgun; no ancestry guard. See inline on `notes_folders.go`.

### Suggestions
1. **`ImportMeeting` re-import stability** is only guaranteed when the bundle basename matches `intentIDPattern` (the test uses `…-20260729-133259`, which does). A bundle named `standup.meeting` generates a new timestamped id each import and duplicates. Document that contract hard for festival-voice, or hash/stable-id from basename regardless of pattern.
2. Dead code in `ImportMeeting` (`NewTemplateDataFromInput` + empty `if intentIDPattern…` + `_ = data`) — delete or finish.
3. `CreateNoteFolder("meetings")` always fails as reserved; `ImportMeeting`’s first attempt is noise before the `MkdirAll` fallback — just create the reserved dir directly.

### Nits
- `IntentGroup` comment still mentions `ParentStatus`; field is only `Depth`/`Children`.
- `RestoreNote` always lands on notes root (loses prior folder); fine if intentional, worth a one-liner in the API doc.

### What looks good
- `normalizeNoteFolderRel` is careful (abs, `..`, reserved root names, dot segments).
- Move picker and add-destination correctly call `NoteFolders` (so CLI/add can see empty folders even while the tree cannot).
- Archive key now allows archiving notes that live in user folders (old `!= StatusNote` check was wrong).
- Filter path: omit empty shells + expand ancestors of matches is the right behavior.

internal/intent/tui/explorer/model.go
obey-agent: Fixed in 545e4106. The explorer now loads NoteFolders with the intent payload, caches them on the model, and passes them into both mixed and notes-only tree construction so empty user folders remain visible outside filtered views.

internal/intent/tui/explorer/notes_view.go
obey-agent: Fixed in 545e4106. Notes mode now receives the same discovered folder metadata, with coverage for empty folders in both explorer surfaces.

internal/intent/notes_folders.go
obey-agent: Fixed in 545e4106. RenameNoteFolder rejects destinations below the source before creating destination parents, and the descendant case is covered by a regression test.

obey-agent (commented):
Verdict: Approve

Overview: The requested folder-tree and rename-safety defects are fixed, and the follow-up substantially improves the production completeness of note capture and meeting import. The updated paths preserve folder selection end to end, validate note statuses and meeting bundles, retain meeting identity and structured metadata, and include selective audit/commit coverage.

Key Findings:
- No remaining blocking findings in the updated diff.
- Empty discovered folders now render in both mixed explorer and notes-only views, while filtered views still omit empty shells.
- Renaming a note folder into its own descendant is rejected before destination directories are created.
- Interactive destination selection was verified with the real binary through VHS; meeting import and extraction now preserve metadata, backlinks, identity, and retry idempotency.

What’s Done Well:
- The changes extend the existing intent, audit, and commit systems rather than creating parallel paths.
- Error propagation and path normalization are consistently tightened at service, CLI, and TUI boundaries.
- Regression coverage is focused on the actual failure modes, including empty folders, traversal, re-import, adoption, extraction retry, and collapsed navigation.

Validation: just gate passed (4,127 dev tests; 4,082 stable tests), and the VHS note-destination workflow saved into notes/reading/.

Staff Standard: Yes. I would be proud to submit this updated change set as my own.

obey-agent (approved):
Approved after requested changes were addressed and the full stable/dev gate plus VHS workflow passed.

PR #538, opened 2026-08-04
fix(jobs): do not land filler or empty deferred commits

lancekrogers:
## Summary

Two related deferred-commit bugs that produced the empty `[…​] Deferred commit (writer unavailable)` commit in the field:

1. **Filler message** — when the configured writer failed, the worker still landed a commit with a camp-invented subject (`(writer unavailable)`). That is no longer allowed: writer failure **fails the job** and parks it for retry/drop.
2. **Empty tree jobs** — a deferred job whose captured tree equals the parent tree (no staged delta) could still run. That happened when only excluded submodule dirt made `HasChanges` true. Commit commands now gate on **staged** changes, enqueue refuses empty trees, and the worker treats empty trees as a no-op success.

## Changes

- `messageForTree`: remove fallback/filler path; propagate writer errors
- `defercommit.Enqueue`: return `ErrNoChanges` when tree == parent tree
- `executeCommitTree`: skip empty snapshots without calling the writer
- `camp commit` / `p commit` / `worktrees commit`: gate on `HasStagedChanges`; handle empty enqueue as nothing-to-commit
- Delete fallback message helpers and their unit tests
- Integration: writer failure fails the job; empty staged tree does not defer

## Test plan

- [x] `go test -short ./...`
- [x] Integration: `TestIntegration_WriterFailureFailsTheJob`
- [x] Integration: `TestIntegration_EmptyWriterOutputFailsTheJob`
- [x] Integration: `TestIntegration_EmptyStagedTreeDoesNotDefer`
- [x] Integration: `TestIntegration_AutoWriteReturnsBeforeTheWriterFinishes`
- [ ] After install: with only dirty submodule refs at campaign root, `camp commit -aw` prints nothing-to-commit (no queue, no empty commit)
- [ ] With a real staged change and a broken writer, job lands in `camp jobs` failed list and HEAD does not move


obey-agent (commented):
## Review (obey-agent)

Correct product fix for both field failures: (1) no more camp-invented `(writer unavailable)` subjects, (2) no empty deferred trees from unstaged/submodule dirt. The layered gates (HasStagedChanges → Enqueue tree==parent → worker empty noop) are the right shape, and tests pin the inverted writer-failure contract.

### Bugs
None found.

### Suggestions
1. **Message accuracy when `-a=false`.** With nothing staged but unstaged/untracked dirt still present, `camp commit` now prints “Nothing to commit, working tree clean.” Default `-a` makes this rare; if you keep the staged-only gate (good), consider “Nothing staged to commit” when the worktree is not actually clean.
2. **Worker empty-tree success.** Returning nil for legacy empty queue files is right (no failed-job noise). Worth a one-line comment near drain/Then that a successful empty job still clears the queue and may run a `Then` follow-up if one was attached by hand — only relevant for hand-edited queues.

### What looks good
- Filler helpers and their contract tests deleted with the behavior.
- `TryDeferAutoWrite` maps `ErrNoChanges` to a hard stop instead of “degrade to sync writer.”
- Amend still cannot defer (`RefusedAmend`), so empty-tree enqueue does not break message-only amends.
- Integration tests rewritten to assert HEAD frozen + failed job evidence on writer outage.

### Nits
- None blocking.

PR #540, opened 2026-08-04
fix(commit): say nothing staged when worktree is still dirty

lancekrogers:
## Summary

Follow-up to #538 review suggestion: after the staged-only commit gate, do not print “Nothing to commit, working tree clean” when unstaged/untracked dirt is still present.

- Clean tree → `Nothing to commit, working tree clean` (unchanged)
- Dirt present but nothing staged → `Nothing staged to commit`
- Same wording for project (`… in project`) and worktree (`… in worktree`)

## Test plan

- [x] `go test ./cmd/camp/cmdutil/ -run NothingStaged`
- [x] `go test ./cmd/camp/ ./cmd/camp/project/ ./cmd/camp/worktrees/`
- [ ] Manual: dirty submodule pointers only at campaign root → `camp commit -aw` should not claim the worktree is clean

### DIFF
--- a/cmd/camp/cmdutil/nothing_staged.go
+++ b/cmd/camp/cmdutil/nothing_staged.go
@@ -1,0 +1,28 @@
+package cmdutil
+
+import (
+	"context"
+
+	"github.com/Obedience-Corp/camp/internal/git"
+)
+
+// NothingStagedLine is the no-op message after a staged-only commit gate finds
+// nothing to commit.
+//
+// The gate checks staged changes only, so claiming "working tree clean" is
+// wrong when unstaged or untracked dirt remains (common at campaign root with
+// excluded submodule pointers). scope is a trailing place qualifier such as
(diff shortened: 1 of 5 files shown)


obey-agent (commented):
## Review (obey-agent)

Small, correct follow-up to the staged-only commit gate.

`NothingStagedLine` only runs when the caller already knows nothing is staged, then distinguishes:
- remaining unstaged/untracked dirt → `Nothing staged to commit` (+ scope)
- truly clean → prior “working tree clean” / scoped clean wording

Treating `HasChanges` errors as “nothing staged” (not “clean”) is the right fail-safe. Shared helper + unit coverage for clean, untracked, unstaged, and scoped messages is solid.

### Findings
No bugs.

### Nits
- `HasChanges` still walks staged first, but callers only invoke this path after `!hasStaged`, so the comment is accurate in practice.

**Event:** COMMENT

PR #541, opened 2026-08-04
fix(jobs): stop deferred capture from walking into nested repositories

lancekrogers:
## The bug

`camp project add` and `camp project new` both queue a bookkeeping commit of two paths: `.gitmodules` and the new project directory. With deferral on (the default), that commit fails immediately and no retry can ever make it succeed.

```
STATE     ID                        LANE   KIND           AGE  WHAT
failed    job-20260804T215858Z-a9b5 .      commit-paths   25s  [mytools:8a57dff6] Add: Cap (gave up after 1 attempt)
```

From `worker.log`:

```
failed lane=. seq=1 id=job-20260804T215858Z-a9b5
  err=stage captured projects/Cap/.git: git update-index --add --cacheinfo
  100644,acb2cc304e1550139bf193879546ea8071457c36,projects/Cap/.git:
  error: Invalid path 'projects/Cap/.git'
fatal: git update-index: --cacheinfo cannot add projects/Cap/.git: exit status 128
```

## Root cause

Deferred jobs snapshot content at enqueue time, so the commit means what its message says even if the tree changes before the worker runs. `CaptureBlobs` hands a directory path to `captureDir` (`internal/git/tree.go`), which was a plain `filepath.WalkDir` with no notion of a repository boundary.

Handed a freshly added submodule, it walked the entire checkout and hashed every file as an ordinary blob: 3347 of them in the reported job, `projects/Cap/.git` among them. A submodule's `.git` is a 40-byte file holding a gitdir pointer, so the walk saw an ordinary file and captured it as mode `100644`. Git refuses any path with a `.git` component, so `CommitBlobs` died at exit 128.

The rejected path is baked into the job document on disk, which is what makes it permanent: `camp jobs retry` replays the identical `update-index` call and gets the identical rejection.

The synchronous path never had this bug. It stages with `git add`, which records a submodule as a gitlink and never descends into it.

Setting the `.git` entry aside, the captured job was wrong anyway: it would have inlined 3346 of the submodule's files into the parent repo instead of a one-line pointer. The hard failure is the loud symptom of a wrong capture, not the whole defect.

## Why it was not caught

Every existing `project add` and `project new` integration test runs with deferral disabled, so all of them exercise the synchronous path.

## The fix

- `captureDir` stops at a repository boundary and returns `ErrNestedRepo` naming it. A submodule's `.git` is a file and an embedded clone's is a directory, so presence is the test and its kind is not.
- `EnqueuePaths` treats that one capture failure as a refusal rather than degrading it to a path-only job. Every other capture failure degrades exactly as before.
- The caller's existing fall-through then takes the synchronous path, where `git add` records the gitlink the user just created.

A gitlink is not something a fixed capture could snapshot in the first place: its value is the nested repository's HEAD at commit time. That is why `checkGitlinkCarveOut` already forbids a non-follow-up job from committing one. That check reads `HEAD`, so a brand-new submodule was invisible to it and the job got through.

## Tests

`tests/integration/project_add_defer_test.go`. All three verified failing before the fix and passing after:

- `project add` commits the pointer as `160000 commit` and no file from inside the submodule
- `project add` leaves no failed job and no "deferred commit failed" notice
- `project new`, the second command with the same two paths, does the same

Gate status: builds (stable and dev), vet (stable, dev, integration), lint (stable and dev, 0 issues), the `fmt.Errorf` ratchet, and unit tests on both profiles (4129 dev, 4084 stable) all pass.

`just gate` itself stops earlier, at `lint-no-host-fs-tests`, on a violation that is already on `main` and unrelated to this change: `internal/git/tree_empty_test.go` arrived in #538 and is not on the allowlist, so the gate is red on a clean `origin/main` checkout too. Verified by stashing this branch's changes and re-running the recipe. Not touched here; worth its own migration to `tests/integration/` rather than growing the allowlist.

## For anyone already stuck

This prevents new undrainable jobs; it does not clear one already queued. Drop it and commit normally:

```
camp jobs drop <id>
```

Nothing is lost. The real index already holds the correct gitlink, so an ordinary commit records it.

### DIFF
--- a/internal/defercommit/defercommit.go
+++ b/internal/defercommit/defercommit.go
@@ -11,6 +11,7 @@ package defercommit
 
 import (
 	"context"
+	"errors"
 	"os"
 	"strings"
 
@@ -160,8 +161,25 @@ func EnqueuePaths(ctx context.Context, campaignRoot, repoPath, message string, p
 	// committing at all, and the failure is a repository camp could not read.
 	// Degrading silently is not an option, so the worker log records it; the
 	// job itself also shows it, since a captured job carries blobs.
+	//
+	// A nested repository is the exception, because there the degraded job is
+	// wrong rather than merely coarser. See the refusal below.
 	blobs, err := git.CaptureBlobs(ctx, repoPath, paths)
-	if err != nil {
+	switch {
+	case errors.Is(err, git.ErrNestedRepo):
+		// A submodule cannot be deferred at all, so this is the one capture
+		// failure that refuses the job rather than degrading it. Git records
+		// the path as a gitlink pointing at the nested HEAD, which is neither
+		// something the enqueuer can snapshot nor something a job may commit
+		// from an execution-time read: `camp project add` would publish
+		// whatever that submodule's HEAD had become by the time the worker
+		// ran. Returning an error puts the caller back on the synchronous
+		// path, where `git add` records the pointer the user just created.
+		jobs.LogEvent(campaignRoot,
+			"capture-refused repo=%s paths=%d err=%v; committing synchronously instead",
+			jobs.RepoForPath(campaignRoot, repoPath), len(paths), err)
+		return nil, err
+	case err != nil:
 		jobs.LogEvent(campaignRoot,
 			"capture-degraded repo=%s paths=%d err=%v; the job will commit execution-time content",
 			jobs.RepoForPath(campaignRoot, repoPath), len(paths), err)
--- a/internal/git/commit/commit.go
+++ b/internal/git/commit/commit.go
@@ -162,8 +162,9 @@ func stageAndCommit(ctx context.Context, opts Options, message string) error {
 			defercommit.SpawnWorker(ctx, opts.CampaignRoot, opts.CampaignRoot)
 			return errDeferred
 		}
-		// A queue that cannot be written must not cost the user their commit.
-		// Fall through to the synchronous path exactly as before.
+		// A queue that cannot be written, or a set of paths that must not be
+		// deferred at all, must not cost the user their commit. Fall through
+		// to the synchronous path exactly as before.
 	}
 
 	commitScope := append(append([]string{}, opts.Files...), opts.PreStaged...)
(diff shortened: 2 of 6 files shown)


obey-agent (commented):
## Review (obey-agent)

Correct diagnosis and fix for a permanent undrainable deferred job.

### What was wrong
`captureDir` walked a newly added submodule as ordinary files, hashed `projects/<name>/.git` as mode `100644`, and baked that into the job document. Git rejects any path component named `.git`, so `CommitBlobs` failed at exit 128 forever — including on `camp jobs retry`.

### What the PR does
1. **`isRepoBoundary`** — presence of `.git` (file or directory) stops the walk; kind-agnostic is right for submodule gitdir pointers vs nested clones.
2. **`ErrNestedRepo`** — capture fails closed for that path set instead of producing a wrong tree.
3. **`EnqueuePaths`** — only this failure refuses the job (returns error) rather than degrading to path-only; caller falls through to the synchronous path where `git add` records the gitlink. Other capture failures still degrade as before.
4. **Integration tests** — deferred `project add` / `project new` assert `160000 commit`, no inlined checkout files, no failed-job notice. That coverage gap (all prior project-add tests had deferral off) was the reason this shipped.

### Findings
No high-conviction bugs. `camperrors.Wrapf` uses `%w`, so `errors.Is(..., git.ErrNestedRepo)` works through the walk. Fall-through on enqueue error in `stageAndCommit` is the right place to land.

### Suggestions
1. A small unit test on `captureDir`/`isRepoBoundary` (temp dir with a `.git` file, expect `ErrNestedRepo`) would pin the boundary without the full drain harness — optional given the integration cases.
2. If a deferred bookkeeping job ever captures a *parent* directory that merely *contains* a nested repo (not just the submodule path itself), the whole capture refuses and goes sync — correct for safety; worth knowing if broader path lists grow.

### Nits
- PR notes `just gate` red on main from an unrelated allowlist issue (`tree_empty_test.go`); fine to leave out of this change.

**Event:** COMMENT — ship when ready.

lancekrogers:
Addressed the optional review suggestion: added `TestCaptureBlobs_NestedRepoBoundary` in `internal/git/commit_test.go` (already on the host-fs allowlist) covering:

- ordinary nested dir still expands to files
- submodule-style `.git` **file** → `ErrNestedRepo`
- embedded-clone `.git` **directory** → `ErrNestedRepo`

No production code changes; the prior review already found no blocking bugs.

PR #542, opened 2026-08-05
remove camp workitem current and current.yaml resolution

lancekrogers:
## Summary

Remove `camp workitem current` and the entire singular `current.yaml` fallback.

That pointer looked like portfolio state ("what I'm working on") but was really a one-slot machine-local default. It collided with multi-item attention stages (`current`/`next`/`active`/`parked`) and was easy for humans and agents to misuse—especially when stale.

## What changed

- Deleted `camp workitem current` (get/set/clear)
- Removed resolver tier 5 (`SourceCurrent` / `DisableCurrent`)
- Removed `links.LoadCurrent` / `SaveCurrent` / `Current` type
- Dropped doctor findings, promote/rename/gather migration for current.yaml
- Updated refusal hints and docs to prefer `--workitem`, cwd, or primary links
- Left `workitems/current.yaml` gitignored so leftover local files stay out of git

## How to scope workitems now

1. Explicit `--workitem <selector>`
2. Run from inside a workitem directory
3. Primary-link a project/festival (`camp workitem link`)

## Test plan

- [x] `go test ./internal/workitem/links/ ./internal/workitem/resolver/ ./internal/commands/workitem/ ./cmd/camp/`
- [x] `just docs` (CLI reference regenerated; `camp workitem current` gone)
- [ ] CI unit + integration
- [ ] Confirm `camp workitem --help` has no `current` subcommand


internal/commands/workitem/workitem.go
```diff
@@ -144,7 +144,6 @@ Examples:
 	cmd.AddCommand(newUnlinkCommand())
 	cmd.AddCommand(newLinksCommand())
 	cmd.AddCommand(newWorktreeCommand())
-	cmd.AddCommand(newCurrentCommand())
 	cmd.AddCommand(newResolveCommand())
```
obey-agent, line 150: Blocking compatibility issue: after removing this subcommand, `camp --no-color workitem current --json` is accepted by the parent `workitem` command because it has no `Args` validator. I verified this exact argv at the PR head: it exits 0 and emits a `workitems/v1alpha10` list payload, so old automation silently runs a different command instead of getting an explicit removal error. This is already an active integration path in `projects/festival-app/server/api.mjs` and `src-tauri/src/commands/workitem.rs`, both of which invoke `workitem current --json`; the Festival templates/docs also still instruct users to run it. Please either retain a tombstone/deprecation command that returns actionable migration guidance for a release, or make the parent reject stray args and coordinate the downstream migration before shipping. Add a regression test for the exact legacy argv so an exit-0 schema substitution cannot recur.

docs/workitem-link-reference.md
```diff
@@ -311,22 +262,10 @@ Recovery options:
...
-
-This returns tier 5 to "skip" state. No links are affected.
-
----
 
 ## Operational Notes
 
 - Link IDs are retried up to 32 times on collision before the command fails.
-- `current.yaml` is local machine state and is ignored by `camp init` and
   `camp init --repair` scaffolds.
```
obey-agent: The `current.yaml` deletion stranded this sentence fragment (`camp init --repair scaffolds.`), and the walkthrough above now jumps from step 5 to step 7. Please clean up both user-facing artifacts. Also, `git diff --check origin/main...HEAD` currently fails on the extra blank line at EOF in `tests/integration/repair_test.go`; fixing that in the same cleanup will restore patch hygiene.

tests/integration/workitem_links_test.go
```diff
@@ -117,16 +117,6 @@ func TestIntegration_LinkJSONContracts(t *testing.T) {
 	assert.Equal(t, "project", linksPayload.Links[0].Scope.Kind)
```
obey-agent, line 117: The integration suite still contains the old current tier in `TestIntegration_ResolverTiers` (lines 201-215 of this file): it runs `workitem current timeline` and expects resolver source `current`. I ran that focused containerized test at the PR head and it fails at line 205 with `non-interactive use requires --json, --list, or --print flag`. Please update the tier test to cover the new five-stage pipeline/none result, and include the legacy `workitem current --json` regression noted above. This is an actual affected-suite failure, not just missing coverage.

docs/cli-reference/camp-reference.md
```diff
@@ -2644,6 +2645,236 @@ camp idea note [text] [flags]
 ```
 ---
 
+## camp idea notes
```
obey-agent, line 2648: The generated documentation is not self-consistent at this commit. Running the PR test-plan command `just docs` leaves `camp_idea.md`, `camp_idea_note.md`, and `camp_status.md` modified and creates seven untracked `camp_idea_notes*.md` pages. This combined reference already includes those newly generated sections, but the companion per-command pages were selectively omitted. Please regenerate from a clean tree and commit the complete output (or remove unrelated aggregate drift); verify `git status --short` is empty afterward so the documented gate is reproducible.

obey-agent (commented):
## Verdict: Request Changes

### Overview
The internal removal is mostly coherent: command registration, resolver state, persistence helpers, lifecycle migration hooks, tests, and core Camp docs were updated together, while legacy `current.yaml` remains safely gitignored. However, this is not production-ready because the removed command silently resolves to a different successful parent invocation, active downstream products still call it, an affected integration test remains stale and failing, and generated docs are not reproducible from the committed tree.

### Key Findings
- **Blocking — silent command substitution and live downstream breakage.** At this head, `camp --no-color workitem current --json` exits 0 and emits the ordinary `workitems/v1alpha10` list because the parent accepts stray args. `projects/festival-app/server/api.mjs` and `src-tauri/src/commands/workitem.rs` execute that exact legacy argv, while Festival templates/guides still teach it. Retain an actionable tombstone for a release or reject parent args and coordinate all consumers before shipping. See inline comment.
- **Blocking — affected integration suite is red.** `TestIntegration_ResolverTiers` still invokes `workitem current`, then expects resolver source `current`; the focused container run fails at line 205. Update the pipeline assertions and add a regression for the exact legacy argv. See inline comment.
- **Generated docs are incomplete.** `just docs` modifies three tracked CLI pages and creates seven untracked `camp_idea_notes*.md` pages. The aggregate reference includes those sections but the per-command pages are missing. See inline comment.
- **Documentation and patch hygiene need cleanup.** The link reference has a stranded sentence fragment and skips walkthrough step 6; `git diff --check` fails on an extra blank line in `tests/integration/repair_test.go`. See inline comment.

### Risk Assessment
- **Overall merge risk: High until the blockers are fixed.** Likelihood is certain for the exact downstream argv already in the tree; impact ranges from Festival App HTTP 502s to scripts silently accepting the wrong schema because the process exits 0.
- **Compatibility risk: High.** This removes a public CLI command and `workitem-current/v1alpha1` JSON contract without a transition path or coordinated dependent changes.
- **Data-loss/security risk: Low.** The change does not delete legacy local files, keeps them ignored, and removes read/write paths cleanly.
- **Implementation risk after rollout fixes: Low to Medium.** The deletion itself is straightforward and stable/dev unit suites pass; remaining risk is contract rollout and regression coverage, not complex algorithmic behavior.
- **Recovery:** restoring a tombstone/compatibility command or reverting the Camp release is simple, but downstream Festival/Festival App updates should land before the hard removal reaches a distribution release.

### Validation
- Stable unit: 4090/4090 passed.
- Dev unit: 4135/4135 passed.
- Focused container integration: six affected cases passed; `TestIntegration_ResolverTiers` failed as described.
- `just docs`: completed but left generated changes/untracked files.
- `just gate`: blocked by the existing host-filesystem lint ratchet on current `main`; I reproduced the same lint failure on `main`, so it is not attributed to this PR.

### What Is Done Well
- The code deletion follows ownership boundaries instead of leaving parallel compatibility logic behind.
- Resolver, promote/rename/gather behavior, JSON structs, and primary Camp documentation were updated consistently in the intended direction.
- Keeping legacy `current.yaml` ignored avoids accidental user-state commits without introducing a destructive migration.

### Staff Standard
No, I would not submit this as my own production-ready change yet. I would be comfortable after the legacy argv fails explicitly or is tombstoned, active downstream callers/docs are coordinated, the resolver integration test is updated, and docs regeneration is clean.

lancekrogers:
Addressed the three review findings in 710badee:

1. **Legacy argv / list fallthrough** — restored `camp workitem current` as a **tombstone** that always exits 2 with a `workitem-current/v1alpha1` JSON error envelope and migration hint (`--workitem`, cwd, or `camp workitem link`). Unit + integration regressions cover the exact `camp --no-color workitem current --json` path so it cannot silently emit `workitems/v1alpha10`.

2. **Docs** — fixed the stranded operational-notes sentence and renumbered the happy-path walkthrough (5→6). Cleared the extra blank line at EOF in `tests/integration/repair_test.go`.

3. **ResolverTiers** — dropped the old current tier; after unlink, resolve from `docs/` expects `source: none`. Added `TestIntegration_CurrentRemovedLegacyArgv`.

Downstream: festival-app still calls `workitem current --json` and will now get a clear removal error instead of a wrong list schema; a follow-up there can switch to resolve/link.

obey-agent (commented):
## Review (obey-agent) — re-review on `796d5889`

**Delta since prior review on `274d3944`:** tombstone for `workitem current`, integration/unit coverage for legacy argv, docs cleanup, and drop of the `current.yaml` resolver tier.

### Blockers from earlier review — status

| Prior issue | Status on this head |
|---|---|
| Silent fallthrough: `workitem current --json` → parent list `workitems/v1alpha10` exit 0 | **Fixed.** `newCurrentCommand` stays registered; `RunE` always returns validation error; unit `TestCurrentTombstone_*` + integration `TestIntegration_CurrentRemovedLegacyArgv` assert exit 2, empty stdout, `workitem-current/v1alpha1` envelope. |
| `TestIntegration_ResolverTiers` still expected source `current` | **Fixed.** Unlinked path expects `none`; legacy current argv covered separately. |
| Link-reference fragment / walkthrough skip / `git diff --check` | **Fixed** on this tree (`git diff --check origin/main...HEAD` clean; walkthrough steps 1–6 coherent). |

### What looks solid
- Resolver Options no longer carry `DisableCurrent`; sources are explicit/ancestor/link/festival/none only. Commit wrappers correctly drop the old “disable current” special case.
- Persistence (`LoadCurrent`/`SaveCurrent`/`Current` type) and promote/rename/gather migration hooks for current.yaml are removed together with doctor findings — no half-deleted API surface in Go.
- Legacy `workitems/current.yaml` remains gitignored via scaffold/init/repair (leftover local files stay out of git without a destructive wipe).
- CLI reference pages correctly document the tombstone rather than inventing a still-working command.

### Non-blocking notes
1. **Stale comment:** `resolver.Resolve` still says “five-tier pipeline” but the live tiers are four (+ none). Cosmetic only.
2. **Downstream consumers:** festival-app / guides that still call `camp workitem current --json` will now get a **structured failure** (good) instead of a wrong success. Coordinate those callers in a follow-up so production UIs do not 502 on the removal — but that is outside this Camp PR once the tombstone is present.
3. **CI:** no checks reported on the branch head yet at review time; author-local unit suites and focused integration coverage in the PR body remain the signal until GH Actions catches up.

### Verdict
No remaining functional bugs in the Camp tree for this change. The hard compatibility footgun is closed by the tombstone. **COMMENT** (not REQUEST_CHANGES).

lancekrogers:
Follow-up: dropped the tombstone.

Keeping a fake `workitem current` command was more confusing than helpful, especially with no released consumers. The command is fully removed again.

Legacy `camp workitem current --json` now fails as **unknown command** (exit non-zero) rather than silently listing. Parent `workitem` also rejects positionals so that fallthrough cannot succeed.

obey-agent (commented):
## Review (obey-agent) — re-review on `4db30909`

**Delta since `796d5889`:** hard-remove of the `workitem current` tombstone; parent now rejects positionals via `cobra.NoArgs`; regression tests updated; gendocs pages for `camp idea notes*` committed.

### Compatibility strategy change
Previous head kept a tombstone that returned `workitem-current/v1alpha1` + migration hint. This head **deletes the subcommand** and relies on:

1. `Args: jsoncontract.Args(..., cobra.NoArgs)` on the `workitem` root so unknown words cannot silently become a successful list, and  
2. cobra’s `unknown command "current"` path for the legacy argv.

Unit (`TestWorkitemRootRejectsPositionalArgs`) and integration (`TestIntegration_CurrentRemovedLegacyArgv`) both assert non-zero exit, empty stdout, and no list success for `workitem current --json`. That closes the original silent-fallthrough footgun without a permanent tombstone command.

### What looks solid
- Resolver still has no `SourceCurrent` / `DisableCurrent`; persistence helpers for `current.yaml` stay gone; gitignore leftover rule remains.
- Manifest annotations no longer advertise a live/tombstone `workitem current` agent command.
- CLI docs drop `camp_workitem_current.md` and regenerate/include the notes command pages that earlier review called incomplete.
- Subcommand registry test no longer expects `current`.

### Non-blocking notes
1. **Migration UX:** legacy callers now get a generic `unknown command "current"` under `workitems/v1alpha10`, not a dedicated removal schema/hint (`camp workitem link`). Acceptable if hard-remove is intentional; downstream (festival-app, guides) still need coordinated updates so UIs do not treat this as a hard crash without guidance.
2. **Stale comment:** `resolver.Resolve` still says “five-tier pipeline” while only four match tiers remain (+ none). Cosmetic.
3. **CI:** no checks reported on the branch head at review time.

### Verdict
No functional bug reintroduced by the hard-remove approach: legacy argv fails closed instead of listing. **COMMENT**.

PR #543, opened 2026-08-05
fix(workitem): resolve intents to bare ids so they can primary-link to worktrees [intent: intents-cannot-be-primary-linked-20260803-184438]

lancekrogers:
## Source intent

- **id**: `intents-cannot-be-primary-linked-20260803-184438`
- **title**: Intents cannot be primary-linked to a worktree: resolver returns key form, workitem_id validator rejects it
- **file**: `.campaign/intents/active/intents-cannot-be-primary-linked-20260803-184438.md`

## The bug (verified live on this branch before any change)

An intent had no string that could satisfy both the selector and the links validator, so `camp workitem link <intent> --worktree ...` and `camp project worktree add --workitem <intent>` were both impossible.

The selector rejected the bare frontmatter id and suggested the key form:

```
Resolve("intents-cannot-be-primary-linked-20260803-184438") returned error:
no workitem matched selector intents-cannot-be-primary-linked-20260803-184438;
did you mean: intent:.campaign/intents/inbox/intents-cannot-be-primary-linked-20260803-184438.md
```

...and the links validator rejected exactly the key form the selector had just suggested:

```
validation error on workitem_id: invalid workitem_id
intent:.campaign/intents/inbox/intents-cannot-be-primary-linked-20260803-184438.md:
must not be empty, must not contain '/', '\', whitespace, or control characters,
must not start with '.' or '-', max 80 chars
```

Two defects met in the middle:

1. `internal/workitem/link_target.go` promoted `SourceID` to `workitem_id` only for festivals, so an intent fell through to `wi.Key`, which always contains a slash and can never pass `pathutil.ValidateSegment`.
2. `internal/workitem/selector/selector.go` had a `festival_id` match tier gated on `WorkflowType == festival`, so an intent's frontmatter id resolved to nothing. The intent's own id was sitting in `SourceID`, unused by either side.

## What changed

**The id rule is now one function, shared by writer and reader.** `SourceDeclaredID` names the types that carry a single-segment id inside their own source document (festival `fest.yaml` id, intent frontmatter id). `LinkWorkitemID` stores it; the selector's match tier (renamed `festival_id` to `source_id`) resolves it. An id that can be written to a link can now always be read back, which was the invariant the two sides disagreed about.

Consequential changes needed to make the round-trip real:

- `workitemIDsOnDisk` (doctor) registers source-declared ids, so a link camp itself wrote is not then reported as pointing at a workitem that is not on disk.
- `camp workitem id` prints the bare id for intents, with the key form still behind `--key`. `--json` gains `id_kind: "intent"` alongside the existing `stable` / `festival` / `key` (additive; `workitem-id/v1alpha1` unchanged).
- `link.go`'s `workitemIDForLink` had its own copy of the id precedence. It now delegates to `LinkWorkitemID`, keeping only the `--allow-missing` case that has no workitem to ask.
- New `LinkMatchesWorkitem(wi, workitem_id, workitem_key)` replaces five hand-rolled `link.WorkitemID == wi.StableID` comparisons in `links.go`, `unlink.go`, `staging_branches.go` (x2), `worktree.go`, and `merged_backstop.go` (x2). Without this an intent could be linked but not listed by `camp workitem links <intent>` and not removed by `camp workitem unlink <intent>`, since `wi.StableID` is empty for an intent. The helper still accepts the stable id and the key form, so registries written by an older camp keep matching with no migration.

**Two honesty fixes found while tracing the path:**

- `camp workitem link` on an unadopted design/explore directory leaked the raw path-segment validator error. `camp project worktree add` already refused those up front with a "run camp workitem adopt" message; the standalone link command now says the same thing.
- All three worktree-link success paths printed "camp p commit in this worktree will include WI-* in the campaign tag". That is false for an intent: `EnsureRefForCommit` only mints a ref for a directory-kind item with a `.workitem` marker to write it into. `WorktreeLinkCommitNote` now tells the truth for both cases, and `CarriesCommitRef` gives that rule one home instead of an inlined condition.

## Other path-keyed types (checked, not assumed)

Ran each through the real binary against a throwaway fixture campaign:

| Type | Finding |
| --- | --- |
| `intent` | The bug. Had a perfectly good bare id in frontmatter that nothing used. Fixed. |
| `code_reviews`, `website`, other custom types | Not affected. A custom type is only discovered as a workitem when it carries a `.workitem` marker, which supplies `StableID`. Linking an adopted `code_reviews` item already worked and still does. |
| `design`, `explore` (unadopted dirs) | Hit the same validator wall but for a different reason: there is no bare id anywhere on disk to use. The designed remedy is `camp workitem adopt`, which mints one. Not the same defect, so not changed. Only the error message improved, as above. |

## Bonus durability, worth knowing

Because the id is path-independent, an intent link now survives an intent lifecycle move. Promoting a linked intent from `inbox` to `active` leaves `workitem_key` pointing at the old `inbox/` path, but the link still resolves via the id:

```
source   = link
reason   = via link lnk_20260805_4947e8 on worktree:projects/worktrees/camp/wt1
source_id= promote-me-to-a-festival-20260805-030012
rel_path = .campaign/intents/active/promote-me-to-a-festival-20260805-030012.md
```

`camp workitem doctor` reports 0 findings in that state. The stale `workitem_key` is cosmetic (the id is the authority and the key is only a legacy matcher), but see follow-ups.

## Tests

Error cases first, table-driven where the cases vary along one axis.

- `internal/workitem/link_target_test.go`: `LinkWorkitemID` (8 cases incl. stable-id-wins and design-`SourceID`-is-not-addressable), `SourceDeclaredID` (6), `LinkMatchesWorkitem` (9, starting with nil/empty/unrelated).
- `internal/workitem/ref_commit_test.go`: `CarriesCommitRef` (6) plus a guard that the post-link note never promises a `WI-` segment it cannot mint.
- `internal/workitem/selector/intent_test.go`: every selector form for an intent, unknown-id-is-not-found, and duplicate-id-is-ambiguous.
- `internal/commands/workitem/link_intent_test.go`: unknown selector writes no row (first), link by all three forms asserting `workitem_id` bare + `workitem_key` key form, links-lists-its-own-row, unlink-removes-it, and the unadopted-explore message.
- `internal/commands/workitem/id_test.go`: intent stdout/`--key`/`--json` contract, plus `intent` and `intent_without_frontmatter_id` rows in the existing `durableID` table.
- `internal/workitem/resolver/intent_link_test.go`: the round-trip that matters, resolving back to the intent from inside the linked worktree.
- `tests/integration/workitem_intent_links_test.go` (containerized, `//go:build integration`): full lifecycle through the real CLI, and `TestIntegration_WorktreeAdd_IntentLinksNotDangles`, the sibling of the existing unadopted-design-dir test, which reproduces the reported "worktree created, then link failed" case and asserts the worktree is created **and** linked.

The unit tests use host `t.TempDir` with plain file writes, matching the surrounding selector/link tests; nothing new shells out to git, so `just lint-no-host-fs-tests` stays satisfied. FS behavior driven through the real binary lives in the containerized suite.

## Gate results

```
go build ./...                                          clean
go vet ./... / -tags dev / -tags=integration            clean
golangci-lint --new-from-merge-base origin/main         0 issues
golangci-lint --new-from-merge-base origin/main (dev)   0 issues
just lint-no-fmt-errorf                                 clean (no regressions)
go test ./...            (stable)                       114 packages ok
go test -tags dev ./...  (dev)                           ok
just test integration                                   (see below)
```

New integration tests, run individually:

```
--- PASS: TestIntegration_IntentPrimaryLinkLifecycle (5.05s)
--- PASS: TestIntegration_WorktreeAdd_IntentLinksNotDangles (4.00s)
```

### Pre-existing gate blocker, not from this branch

`just gate` fails at `lint-no-host-fs-tests`, and it fails the same way on a clean `origin/main` tree (verified by stashing this branch's changes and re-running):

```
FAIL: NEW host-side git exec.Command in _test.go outside tests/integration/:
  ./cmd/camp/cmdutil/nothing_staged_test.go
  ./internal/git/tree_empty_test.go
```

Both files arrived in `837aff89` ("fix(commit): say nothing staged when worktree is still dirty") without being added to the justfile allowlist, and neither is touched by this branch (`git diff origin/main --` on both paths is empty). I deliberately did not add them to the allowlist here: that list is a ratchet whose stated end state is empty, and quietly widening it inside an unrelated PR is the wrong place to make that call. Every other stage of `just gate` passes.

## Tradeoffs and follow-ups

- **`id_kind` gains a value.** `camp workitem id --json` can now return `"intent"`. Additive to `workitem-id/v1alpha1`; a consumer that switches on the field sees a new case rather than a changed one.
- **`LinkMatchesWorkitem` widens some comparisons to also match `workitem_key`.** `camp workitem links` and `camp workitem unlink` previously compared `workitem_id` against `StableID` only, so they now also see legacy rows stored under the key form. That is the behavior `worktree.go` and `merged_backstop.go` already had; this makes all seven sites agree.
- **Intents still get no `WI-` commit segment.** The link resolves and supplies workitem context, but a ref needs a `.workitem` marker to live in and an intent has none. Giving intents refs means adding a `ref:` field to intent frontmatter and teaching the doctor backfill about intent files. Out of scope here; the messaging now states the actual behavior instead of promising a tag.
- **`workitem_key` goes stale on an intent stage move.** `camp intent promote` relocates the file without updating the key on existing links. Harmless today because the id is authoritative and doctor is clean, but the natural fix is for intent status transitions to rewrite `workitem_key`, the way `rename_migrate.go` already does for renames.
- **`promotedWorkitemIdentity` still returns an empty id for intent sources** (it reads `kind: workitem` frontmatter, which intent files do not have), so promoted-intent link rehoming matches by key alone. Correct today since the key is stored, and I left it alone rather than widen this diff.

### DIFF
--- a/internal/commands/workitem/id.go
+++ b/internal/commands/workitem/id.go
@@ -110,24 +110,34 @@ type idKind string
 const (
 	idKindStable   idKind = "stable"
 	idKindFestival idKind = "festival"
+	idKindIntent   idKind = "intent"
 	idKindKey      idKind = "key"
 )
 
 // durableID returns the single-segment identifier the selector can resolve back
 // to wi, plus which form it is. The value is delegated to LinkWorkitemID so the
 // id printed here and the id stored on a link cannot drift; the kind is derived
-// from the same precedence (stable .workitem id, then a festival's fest.yaml id,
-// then the path-derived key fallback).
+// from the same precedence (stable .workitem id, then the id the source document
+// declares, then the path-derived key fallback).
 func durableID(wi *wkitem.WorkItem) (string, idKind) {
 	id := wkitem.LinkWorkitemID(wi)
 	switch {
 	case wi.StableID != "" && id == wi.StableID:
 		return id, idKindStable
-	case wi.WorkflowType == wkitem.WorkflowTypeFestival && wi.SourceID != "" && id == wi.SourceID:
-		return id, idKindFestival
+	case id != "" && id == wkitem.SourceDeclaredID(wi):
+		return id, sourceIDKind(wi.WorkflowType)
 	default:
 		return id, idKindKey
 	}
+}
+
+// sourceIDKind names which source document declared the id, so --json consumers
+// can tell a festival's fest.yaml id from an intent's frontmatter id.
+func sourceIDKind(t wkitem.WorkflowType) idKind {
+	if t == wkitem.WorkflowTypeIntent {
+		return idKindIntent
+	}
+	return idKindFestival
 }
 
 // selectorFromArg translates a positional argument into a selector string. An
--- a/internal/commands/workitem/id.go
+++ b/internal/commands/workitem/id.go
@@ -34,19 +34,22 @@ func newIDCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "id [selector-or-path]",
 		Short: "Print the identifier of a workitem",
-		Long: `Print the stable identifier of a workitem.
+		Long: `Print the durable single-segment identifier of a workitem.
 
 With no argument, the workitem is detected from the current context using the
 same tiered resolution as ` + "`camp workitem resolve`" + ` (explicit selector, cwd
 ancestor, linked scope, festival, current-workitem pointer). With an argument,
 the workitem is resolved through the shared selector family: workitem ref,
-stable id, key, campaign-relative path, directory slug, or festival id. A
-filesystem path (absolute or relative to the current directory) is accepted and
-translated to the campaign-relative form the selector expects.
+stable id, key, campaign-relative path, directory slug, festival id, or intent
+frontmatter id. A filesystem path (absolute or relative to the current directory)
+is accepted and translated to the campaign-relative form the selector expects.
 
-The bare stable id is written to stdout for shell scripting; it is the id sibling
-of ` + "`camp workitem --print`" + `, which prints a path. Use --key for the
-path-derived key instead, or --json for a structured object.
+Stdout is the bare durable id for shell scripting — stable .workitem id when
+present, otherwise a source-declared id (festival fest.yaml id or intent
+frontmatter id), otherwise the path-derived key. It is the id sibling of
+` + "`camp workitem --print`" + `, which prints a path. Use --key for the
+path-derived key instead, or --json for a structured object (id_kind is
+stable / festival / intent / key).
 
 Examples:
   camp workitem id                       # id of the workitem for the cwd
@@ -133,11 +136,18 @@ func durableID(wi *wkitem.WorkItem) (string, idKind) {
 
 // sourceIDKind names which source document declared the id, so --json consumers
 // can tell a festival's fest.yaml id from an intent's frontmatter id.
+// Only festival and intent currently declare a single-segment source id
+// (SourceDeclaredID); anything else falls through to idKindKey so a future
+// type cannot silently inherit "festival".
 func sourceIDKind(t wkitem.WorkflowType) idKind {
-	if t == wkitem.WorkflowTypeIntent {
+	switch t {
+	case wkitem.WorkflowTypeIntent:
 		return idKindIntent
+	case wkitem.WorkflowTypeFestival:
+		return idKindFestival
+	default:
+		return idKindKey
 	}
-	return idKindFestival
 }
 
 // selectorFromArg translates a positional argument into a selector string. An
--- a/internal/commands/workitem/staging_branches.go
+++ b/internal/commands/workitem/staging_branches.go
@@ -220,7 +220,7 @@ func pickPrimaryProjectScopePath(ctx context.Context, root string, wi *wkitem.Wo
 		}
 	}
 	return "", camperrors.NewValidation("link",
-		"no primary project link points at workitem "+wi.StableID, nil)
+		"no primary project link points at workitem "+wkitem.LinkWorkitemID(wi), nil)
 }
 
 func primaryFestivalScopePath(ctx context.Context, root string, wi *wkitem.WorkItem, festivalID string) string {
--- a/internal/commands/workitem/staging_branches.go
+++ b/internal/commands/workitem/staging_branches.go
@@ -211,7 +211,7 @@ func pickPrimaryProjectScopePath(ctx context.Context, root string, wi *wkitem.Wo
 		if link.Role != links.RolePrimary {
 			continue
 		}
-		if link.WorkitemID != wi.StableID && link.WorkitemID != wi.Key {
+		if !wkitem.LinkMatchesWorkitem(wi, link.WorkitemID, link.WorkitemKey) {
 			continue
 		}
 		switch link.Scope.Kind {
@@ -244,7 +244,7 @@ func selectPrimaryFestivalScope(registry *links.Links, wi *wkitem.WorkItem, fest
 		if link.Role != links.RolePrimary || link.Scope.Kind != links.ScopeFestival {
 			continue
 		}
-		if link.WorkitemID != wi.StableID && link.WorkitemID != wi.Key {
+		if !wkitem.LinkMatchesWorkitem(wi, link.WorkitemID, link.WorkitemKey) {
 			continue
 		}
 		if festivalID != "" {
--- a/cmd/camp/project/worktree/add.go
+++ b/cmd/camp/project/worktree/add.go
@@ -172,7 +172,7 @@ func runProjectWorktreeAdd(cmd *cobra.Command, args []string) error {
 			return camperrors.Wrap(lerr, "worktree created but workitem link failed")
 		}
 		fmt.Printf("  Workitem: %s (%s)\n", ui.Value(link.WorkitemID), ui.Dim(link.WorkitemKey))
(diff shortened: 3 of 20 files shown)


obey-agent (commented):
## Review (obey-agent)

**Verdict: COMMENT** — solid fix for the intent primary-link dead-end; no blocking bug found.

### What this does well
- Collapses the festival-only special case into `SourceDeclaredID` / `LinkWorkitemID` / selector `source_id` matcher, so the string written as `workitem_id` is the same string the selector resolves. That is exactly the bug (bare frontmatter id vs slash-bearing key vs validator).
- `LinkMatchesWorkitem` centralizes read-back and explicitly accepts legacy key-in-`workitem_id` rows — no registry migration required.
- `NeedsAdoption` early exit on `runLink` is the right UX: unadopted design/explore should not surface the path-segment validator.
- `CarriesCommitRef` / `WorktreeLinkCommitNote` stop promising a WI-* tag for intents/festivals that never get one.
- Coverage is thorough: unit (id kind, link target, match), command-level link/links/unlink, and integration.

### Non-blocking notes
1. **`LinkMatchesWorkitem` after adoption** — once an intent gains a `StableID`, matching is StableID/Key only; a row that only stored the frontmatter id (empty `workitem_key`) would miss. New writers always set both fields, so this is only a legacy edge. Optional hardening: also accept `SourceDeclaredID(wi)`.
2. **Duplicate frontmatter ids** — selector `source_id` is first-match; two intents with the same id would still collide. Pre-existing class of problem for festivals too; not introduced here.

Looks ready to un-draft once CI is green.

internal/commands/workitem/staging_branches.go
```diff
@@ -211,7 +211,7 @@ func pickPrimaryProjectScopePath(ctx context.Context, root string, wi *wkitem.Wo
 		if link.Role != links.RolePrimary {
 			continue
 		}
-		if link.WorkitemID != wi.StableID && link.WorkitemID != wi.Key {
+		if !wkitem.LinkMatchesWorkitem(wi, link.WorkitemID, link.WorkitemKey) {
 			continue
 		}
 		switch link.Scope.Kind {
```
lancekrogers, line 217: **[suggestion]** pickPrimaryProjectScopePath matching was correctly widened to LinkMatchesWorkitem, but the not-found error still concatenates wi.StableID. For intents/festivals that field is empty, so the message becomes "no primary project link points at workitem " with a trailing blank.

**Suggestion:** Prefer LinkWorkitemID(wi) (or identity helper) in the error string so the durable id is shown.

_(anchored near line 217; reviewer cited 220)_

internal/commands/workitem/id.go
```diff
@@ -110,26 +110,36 @@ type idKind string
 const (
```
lancekrogers, line 113: **[nit]** camp workitem id Long help still frames stdout as the bare stable id / festival id only; intents now resolve and print frontmatter id with id_kind intent.

**Suggestion:** Mention intent frontmatter ids alongside festival ids; describe stdout as durable single-segment id.

_(anchored near line 110; reviewer cited 45)_

internal/commands/workitem/id.go
```diff
@@ -110,26 +110,36 @@ type idKind string
...
 	}
 }
 
+// sourceIDKind names which source document declared the id, so --json consumers
+// can tell a festival's fest.yaml id from an intent's frontmatter id.
+func sourceIDKind(t wkitem.WorkflowType) idKind {
+	if t == wkitem.WorkflowTypeIntent {
+		return idKindIntent
+	}
+	return idKindFestival
```
lancekrogers: **[nit]** sourceIDKind returns idKindFestival for every non-intent workflow type — implicit coupling if SourceDeclaredID expands later.

**Suggestion:** Switch explicitly on WorkflowTypeFestival / WorkflowTypeIntent.

lancekrogers (commented):
## Summary

Solid, well-scoped fix for a real two-sided contract break: intents already carried a bare frontmatter id in `SourceID`, but neither `LinkWorkitemID` nor the selector's id tier treated it as addressable, so primary-linking was impossible and the suggested key form failed validation. Centralizing the rule in `SourceDeclaredID` / `LinkWorkitemID`, resolving it via a renamed `source_id` matcher, and consolidating link matching into `LinkMatchesWorkitem` closes the write/read loop correctly, with honest WI-* messaging and strong unit + integration coverage of the full link lifecycle. No functional blockers found; only small residual UX/docs polish in touched paths.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 2

obey-agent (commented):
## Review (obey-agent) — re-review on new head

**Verdict: COMMENT** — prior review nits addressed; still no blocking bug.

### Delta since last COMMENT
- `sourceIDKind` is an explicit switch (intent / festival / default key) so a future type cannot inherit `festival`.
- Error path in `pickPrimaryProjectScopePath` uses `LinkWorkitemID(wi)` instead of only `StableID`.
- CLI help for `workitem id` documents durable id + intent frontmatter + `id_kind` values.

Core design (SourceDeclaredID → LinkWorkitemID → selector source_id matcher, LinkMatchesWorkitem, NeedsAdoption early exit, CarriesCommitRef note) remains solid and well tested.

Ready to un-draft once CI is green.

PR #545, opened 2026-08-05
test(gate): migrate the last two host-FS tests to the container harness

lancekrogers:
## Why

`just gate` has been red on `main`. It stops at `lint-no-host-fs-tests` before running lint or any tests, so no camp branch can currently pass the full gate.

Two test files drive real git through `exec.Command` in a host `t.TempDir` outside `tests/integration/`, and neither was added to the ratchet's allowlist:

- `internal/git/tree_empty_test.go`, from #538
- `cmd/camp/cmdutil/nothing_staged_test.go`, from #542

Verified on a clean `origin/main` checkout with no other changes applied.

Adding them to the allowlist would be the wrong direction: the recipe states the end state is an empty list. Both are migrated instead.

## `nothing_staged_test.go` moves to the harness

`NothingStagedLine` produces four user-facing strings, printed by `camp commit`, `camp p commit`, and `camp worktrees commit`. The point of the function is that the staged-only gate must not claim a clean working tree while unstaged or untracked dirt remains, which is the ordinary state at campaign root where submodule pointers are excluded from staging. Telling a user their tree is clean there contradicts the `git status` they run one command later.

That is a claim about what the binary prints, so `tests/integration/nothing_staged_test.go` asserts it by running the binary. Both scope branches and both dirty/clean branches are covered:

| case | expected |
| --- | --- |
| campaign root, clean | `Nothing to commit, working tree clean` |
| campaign root, untracked dirt | `Nothing staged to commit` |
| campaign root, unstaged tracked change | `Nothing staged to commit` |
| project, clean | `Nothing to commit in project` |
| project, untracked dirt | `Nothing staged to commit in project` |

The dirty cases also assert the absence of the clean-tree wording, which is the regression the function exists to prevent. `camp worktrees commit` reaches the same function with a `" in worktree"` suffix.

## `tree_empty_test.go` is removed, not restated

It asserted a git plumbing property: a clean index writes the same tree as `HEAD`, and a staged change writes a different one. Camp's own behavior tests already depend on both halves end to end, so restating it against `t.TempDir` adds no coverage of camp:

- clean index equals HEAD, so an empty snapshot must not defer, commit, or enqueue: `TestIntegration_EmptyStagedTreeDoesNotDefer`
- a staged change differs from HEAD, so a real change defers and lands: `TestIntegration_DeferredCommitIgnoresLaterStaging` and `TestIntegration_BookkeepingCommitLandsAfterTheCommandReturns`

All three confirmed passing on this branch.

## Result

`just gate` passes end to end: builds and vet on stable, dev, and integration; lint 0 issues on both profiles; both ratchets clean; 4134 dev and 4089 stable unit tests; and the two new integration cases green.

The allowlist is unchanged at 31 legacy files and can keep shrinking from there.


obey-agent (commented):
## Review (obey-agent)

**Verdict: COMMENT** — unblocks `lint-no-host-fs-tests` gate; migration of the user-facing message is the right place.

### Assessment
- Removes host `exec.Command` git tests outside `tests/integration/`.
- `NothingStaged` behavior is re-pinned through the real `camp commit` / `camp p commit` binary in the container harness (clean tree, untracked, unstaged tracked, project scope) — stronger than the old unit test.

### Non-blocking
- `internal/git/tree_empty_test.go` (`TreeSHA` vs `WriteTree` invariant) is **deleted, not migrated**. Title says "migrate the last two"; only the message tests land in the harness. If that invariant is still load-bearing for empty-commit / staged-tree logic, consider a container-side replacement later. Not a gate/prod bug for this PR.

Fine to merge to restore `just gate`.

obey-agent (commented):
**Review (obey-agent)** — head moved since prior review (`a0bd912` → `005b566`).

Re-checked the full delta:

- Host-FS unit tests that violated `lint-no-host-fs-tests` are removed from `cmdutil` / `internal/git`.
- `NothingStagedLine` coverage is restated as integration cases that drive the real binary for campaign-root / project / worktree scopes, including dirty-tree assertions that forbid the false "working tree clean" claim.
- Dropping `TestTreeSHAMatchesWriteTreeForHEAD` without a 1:1 restatement is justified in the PR body by existing end-to-end defer/commit integration coverage; that rationale holds.

No functional bug found. Gate-unblocking migration looks correct.

PR #546, opened 2026-08-05
fix(tests): stop double-discounting the daemon CPU count when sizing the container pool

lancekrogers:
## The regression

af3db4b4 (Aug 3) correctly switched integration pool sizing from the host CPU count to the Docker daemon's — the host count oversized the pool (6 containers on a 4-CPU Colima VM) and produced the tests-pass-alone-fail-together flakiness that motivated the change. But it kept the `/2` halving that was calibrated for the *host* number as a proxy for a smaller VM. Applied to the daemon's own count, the discount is taken twice:

| sizing | pool on this machine (16-CPU host, 4-CPU VM) |
|---|---|
| pre-af3db4b4: host/2, cap 6 | 6 — oversubscribed, flaky |
| af3db4b4: daemon/2 | **2 — a 3x parallelism cut** |
| this PR: daemon, cap 6 | 4 — one container per VM CPU |

Every pooled test runs through `GetSharedContainer` → `t.Parallel()`, so wall-clock ≈ serial-sum / pool-size. Dropping 6→2 is what blew the 15m suite budget (surfaced during CS0005 festival gates as >23 min runs).

## The fix

One container per daemon CPU. The halving survives only in the can't-reach-the-daemon fallback, where the host count is still a proxy for a possibly-smaller VM. Floor 2, cap 6, and the `CAMP_TEST_POOL_SIZE` override are unchanged. The pinned pool-size tests are updated to the new arithmetic.

## Measured (16-CPU M-series host, 4-CPU Colima VM, resident non-test services on the same VM)

- Pool 4, quiet of other test load: **597/597 tests passed, 0 failed, 925s (15.4 min) total** including harness startup; `go test` itself stayed under its 15m `-timeout`.
- Pool 2 baseline (measured during CS0005 gates, same machine): 23+ min.
- The 6 tests that failed in an earlier contended run (`TestSkills_*`, `TestDungeonHidden_*`) pass individually AND concurrently at pool 4; that run was infra-death from VM contention (another suite's containers running simultaneously), not this change.

## Explicitly out of scope (tracked in campaign intent camp-integration-suite-exceeds-its-20260805)

- The post-suite teardown defect: runs can fail/hang *after* every test passes ("test package failed with no failing test"). It reproduced on this branch's clean run (597 pass → package fail) and on the pre-change baseline (442 pass → hang). Verbose diagnosis in progress; it predates this change and deserves its own fix.
- The resident app stack sharing the 4-CPU VM with test infra, which caps how much any pool sizing can help.

### DIFF
--- a/tests/integration/main_test.go
+++ b/tests/integration/main_test.go
@@ -132,8 +132,15 @@ func TestMain(m *testing.M) {
 // and fail in the full suite, differently each run, which reads like flaky
 // tests rather than a saturated machine.
 //
-// The halving and the bounds are unchanged. Only the number they are applied
-// to was wrong.
+// The daemon's own count is used as-is, one container per CPU. The halving
+// this function used to apply was calibrated for the host count, where it
+// approximated a VM the host was fronting; applied to the daemon's own number
+// it discounts capacity twice. On the machine above that double discount took
+// the pool from six to two — a 3x wall-clock regression on a suite whose
+// containers spend their time in short git execs, not sustained CPU — while
+// the flakiness the sizing fix cured needed only the drop from six to four.
+// The halving survives solely in the host fallback, where the count is still
+// a proxy for a machine that may be smaller.
 func poolSizeFor(daemonCPUs int) int {
 	if v := os.Getenv("CAMP_TEST_POOL_SIZE"); v != "" {
 		if n, err := strconv.Atoi(v); err == nil && n > 0 {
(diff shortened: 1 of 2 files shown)


lancekrogers:
Verbose diagnosis of the out-of-scope suite wedge is complete (goroutine dump from a 15m-timeout panic on this branch):

- 3+ tests were holding pooled containers while stuck in TestContainer.exec retry-sleeps (helpers_container.go:38) — docker execs stopped completing late in the run.
- Everything else starved in GetSharedContainer's pool checkout (main_test.go:269, chan receive 14 minutes).
- The colima_transport_policy parallel subtests (which need no containers) parked at t.Parallel behind parallel slots held by the starved waiters.
- go test's 15m -timeout then panics; non-verbose runs report it as N/N passed + a package-level SUITE error, and on the old pool-2 sizing it manifested as the >23min post-suite hang.

So the wedge is exec-retry starvation under daemon load, not teardown, and it predates and is orthogonal to this PR. Fix direction (separate PR, tracked in campaign intent camp-integration-suite-exceeds-its-20260805): bound the exec retry loop with a deadline and extend the existing infra-death mechanism from Reset to exec so a wedged daemon fails in seconds with one legible error.

obey-agent (commented):
## Review (obey-agent)

**Verdict: COMMENT** — correct fix for double-discounted pool sizing.

### Assessment
- After switching to daemon CPUs, keeping `/2` was wrong: host-calibrated discount applied to an already-VM count. One container per daemon CPU (cap 6, floor 2) matches the PR rationale.
- Host fallback still halves `runtime.NumCPU()` — appropriate when the daemon count is unknown.
- Tests updated for 4→4, 5→5, 12→cap 6; override path still honored.

No bug found. Ship when CI is green.

PR #548, opened 2026-08-05
fix(jobs): capture the files git would stage, not the files on disk

lancekrogers:
> Follows #541, which merged while this was in progress. Rebased onto `main`, so this is a single commit against `main`. The unit test added to #541 in 15430c1 still passes against the rewritten `captureDir`; it is the same boundary contract, reached a different way.

## Why

A deferred bookkeeping commit captures content itself rather than letting git resolve the paths later. That is what makes the commit mean its message: an edit made between the enqueue and the run cannot land under the earlier operation's description.

The consequence is that the capture has to reproduce git's own answer, and for a directory it did not. `captureDir` listed the directory off the disk with `filepath.WalkDir`, and the disk is not the set `git add <dir>` operates on.

Two differences, both confirmed by running the tests in this PR against the unfixed code:

**Ignored files were committed.** A walk sees `payload/debug.log`; `git add payload` does not. `CommitBlobs` stages each captured blob with `git update-index --cacheinfo`, which bypasses exclude rules entirely, so the file landed in history. The synchronous path never staged it. Nothing told the user.

**Deletions were dropped.** A tracked file removed from inside a captured directory is invisible to a walk, so the capture kept the old blob and the removal did not land. `git add payload` records it as a deletion.

Reachable today through `camp workitem commit --include <dir>`, which appends the include path verbatim to the staging plan (`internal/commands/workitem/staging_paths.go:23`). Both tests fail on the current code with exactly those symptoms.

Not reachable through the most common workitem path: `camp workitem commit` without `--include` builds its plan from `git status --porcelain --untracked-files=all` (`internal/commands/workitem/staging_git.go:39`), so it passes individual files git already filtered.

## The fix

`captureDir` asks git for the file list instead of walking:

```
git ls-files --cached --others --exclude-standard -z -- <dir>
```

Cached plus non-ignored untracked is exactly the set `git add` would take, and it gets the three cases right for the same reason git does:

- an ignored, untracked file is absent
- an ignored file that is **already tracked** is present, because the ignore rule governs what gets added, not what stays tracked
- a tracked file deleted from disk is still listed, so `CaptureBlobs` records it as a deletion

Two details the listing also settles:

**Nested repositories.** Git declines to look inside one and reports it with a trailing slash (`dir/deep/nested/`) when untracked, or as a single gitlink entry (`dir/sub`) when tracked. The first is rejected directly; the second recurses into `captureDir` and hits the boundary check from #541. Verified both shapes against real git before relying on them.

**Unmerged paths** are listed once per stage, so entries are deduped rather than staged three times.

## Tests

`tests/integration/capture_gitignore_test.go`, driving `camp workitem commit --include payload` with deferral on and draining the queue:

| test | on current code |
| --- | --- |
| ignored file must not reach the commit | **fails**, `payload/debug.log` is in HEAD |
| deleted tracked file must not survive | **fails**, `payload/doomed.md` is in HEAD |
| tracked-but-ignored file must still be captured | passes |

The third is a regression guard on the fix, not a bug repro. The old walk captured everything, so it was already correct there; it would break under a naive `check-ignore` filter, which is why it is pinned.

Note for anyone re-running these: `just test integration-run` has no `-count=1`, and the integration package does not import the code under test, so a rebuilt binary does not invalidate Go's test cache. The red runs above were done with `-count=1`.

## Gate

Lint 0 issues on both profiles, the `fmt.Errorf` ratchet clean, builds and vet clean on stable, dev, and integration, and `internal/git` plus `internal/git/commit` green.

`just gate` still stops at `lint-no-host-fs-tests`, which is `main`'s pre-existing failure, not this branch's: #545 migrates the two host-FS test files that trip it.

One flake seen and chased down: a `just test unit` run reported `internal/sync` failing. It passes standalone twice, passes under `go test ./...`, and `main` at the same commit passes all 4091. Unrelated to this change, and `internal/sync` is untouched here.

### DIFF
--- a/internal/git/tree.go
+++ b/internal/git/tree.go
@@ -3,7 +3,6 @@ package git
 import (
 	"bytes"
 	"context"
-	"errors"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


obey-agent (commented):
## Review (obey-agent)

**Verdict: COMMENT** — correct fix for deferred directory capture vs `git add` semantics.

### Assessment
- Replacing `filepath.WalkDir` with `git ls-files --cached --others --exclude-standard -z` is exactly the set `git add <dir>` would stage.
- Nested-repo detection kept (root boundary + trailing-slash ls-files form).
- Unmerged multi-stage paths deduped instead of triple-capturing.
- `CaptureBlobs` already treats missing paths as deletions (empty SHA), so tracked-but-deleted files listed by `--cached` record correctly.
- Integration tests cover ignored skip, tracked-ignored keep, and deletion — the three disk-walk failure modes.

No bug found. LGTM.

PR #549, opened 2026-08-05
test: migrate host-FS commit-gate tests to the container harness (fixes red gate on main)

lancekrogers:
## Problem

`just gate` fails on `main` today, independent of any feature work:

```
FAIL: NEW host-side git exec.Command in _test.go outside tests/integration/:
  ./cmd/camp/cmdutil/nothing_staged_test.go
  ./internal/git/tree_empty_test.go
```

Both tests ran real `git init` / `add` / `commit` against `t.TempDir()` on the host, which is exactly what the `lint-no-host-fs-tests` ratchet exists to reject. They landed after the allowlist was frozen, so they read as new violators and red the gate.

## Fix

Migrated the coverage rather than widening the allowlist. The recipe's own comment says the end state is an empty allowlist, so adding to it would move the wrong direction.

### `nothing_staged_test.go` -> `tests/integration/nothing_staged_test.go`

The new test asserts the full message matrix through the real commands instead of calling the helper against a temp dir:

| Command | State | Expected |
| --- | --- | --- |
| `camp commit --all=false` | clean | `Nothing to commit, working tree clean` |
| `camp commit --all=false` | untracked dirt | `Nothing staged to commit` |
| `camp p commit --all=false` | clean | `Nothing to commit in project` |
| `camp p commit --all=false` | untracked dirt | `Nothing staged to commit in project` |
| `camp worktrees commit --all=false` | clean | `Nothing to commit in worktree` |
| `camp worktrees commit --all=false` | untracked dirt | `Nothing staged to commit in worktree` |

The dirty rows are the ones that carry weight. The gate checks *staged* changes only, so it must not claim a clean worktree while untracked dirt is still visible in `git status`. None of the scoped variants (`in project`, `in worktree`) had any CLI-level coverage before this PR, so net coverage goes up.

### `tree_empty_test.go` -> deleted, no replacement

This one asserted a **git** invariant, not camp behavior: that `write-tree` of a clean index equals `HEAD^{tree}`. `WriteTree` and `TreeSHA` are three-line wrappers over `git write-tree` and `git rev-parse`.

The camp behavior that actually depends on that invariant, refusing to defer an empty snapshot, is already covered at the integration level by `TestIntegration_EmptyStagedTreeDoesNotDefer` in `autowrite_defer_test.go`, which asserts both that no job is enqueued and that HEAD does not move. Re-testing the plumbing underneath it would be testing the implementation, not the behavior.

Calling this out explicitly since it is a deletion: if you would rather keep an explicit assertion for the tree pair, say so and I will add one driven through `camp jobs`.

## Verification

```
just gate
  ...
  === gate-fast: PASSED ===
  116 packages  4134/4134 tests passed
  113 packages  4089/4089 tests passed
  === gate: PASSED ===

just lint-no-host-fs-tests
  lint-no-host-fs-tests: clean (no NEW violators; 31 legacy files on allowlist)
```

The three new integration tests pass against the container harness:

```
--- PASS: TestIntegration_NothingStagedMessagesAtCampaignRoot (8.25s)
--- PASS: TestIntegration_NothingStagedMessagesInWorktree (10.71s)
--- PASS: TestIntegration_NothingStagedMessagesInProject (16.84s)
ok  github.com/Obedience-Corp/camp/tests/integration  123.206s
```

## Notes

The allowlist is untouched and still holds 31 files. This PR removes two files from the violator working set rather than growing the exemption list. `go.mod` and `go.sum` are unchanged.


obey-agent (commented):
## Review (obey-agent)

**Verdict: COMMENT** — unblocks `lint-no-host-fs-tests` on main; migration is the right place.

### Assessment
- Drops host `exec.Command` git tests outside `tests/integration/`.
- Re-pins nothing-staged user messages through real `camp commit` / `camp p commit` / `camp worktrees commit` in the container harness, including project and worktree scope qualifiers — stronger than the old unit helper.

### Non-blocking
- `internal/git/tree_empty_test.go` (`TreeSHA` vs `WriteTree`) is deleted without a container replacement. Fine for the gate; re-add under integration later if that invariant still needs a pin.

Ship to restore `just gate`.

lancekrogers:
Closing as a duplicate of #545, which was opened about an hour earlier and makes the same call to delete `tree_empty_test.go`.

Coverage was complementary rather than identical, so nothing is lost: #545 already asserted the unstaged-tracked-change case at campaign root that this PR lacked, and the worktree scope this PR added has been folded into #545 as `TestIntegration_NothingStagedMessageInWorktree` (005b5661). `camp worktrees commit` passes " in worktree" and was the one caller of `NothingStagedLine` left untested there.

#545 now carries the full six-case matrix across all three commit surfaces. Gate is green on that branch: 4134 integration-tagged and 4089 stable tests passing, `lint-no-host-fs-tests` clean.

My fault for not checking open PRs before branching.

PR #550, opened 2026-08-05
feat(transport): cold seed from a quiescent peer, engine probe, and conflict lifecycle (CS0005)

lancekrogers:
Festival CS0005 (camp-syncclone-fast-cross-machine), phases 003 through 006. Completes the sync/clone transport work that began with #421-#428.

## What this adds

**Cold seed (phase 003).** `camp clone` can now seed a campaign from a peer machine by copying the quiescent peer's pack bytes directly, which is the fast path for a first clone over a LAN. A peer quiescence contract guards it: the seed only runs pack-copy when the peer proves quiescent, proven over loopback ssh in the container lane. Non-quiescent repos fall back to a transactional git-bundle seed, so the clone always lands consistent bytes. `--json` reports the seed transport chosen per repo, and an unreachable peer is reported instead of swallowed.

**Engine probe (phase 004).** `rsyncprobe` resolves which rsync binary actually runs on each end of a transfer (macOS ships openrsync, which lacks flags the sync engine wants). The transfer engine is chosen from the probe result and degrades honestly: when the probe rules out rsync the engine falls back and says so, rather than failing mid-transfer.

**Conflict lifecycle (phase 005).** `camp artifacts resolve` unsticks conflicts that the automatic path cannot settle, and the full conflict lifecycle (create, detect, resolve, verify) is proven end to end in the integration lane.

**Docs (phase 006).** `docs/peer-transport.md` documents peer sync, cold seed, conflict handling, and the rsync probe.

## Test architecture note (D007, reviewer attention wanted)

Clone/sync tests that mutate the filesystem now run inside the container harness via a `//go:build container_fs` tag instead of being relocated into `tests/integration/`. The tagged test binaries are cross-compiled and executed in the pooled containers with MinTests floors, and `TestContainerFSPackagesAreAllRegistered` asserts registration in both directions (a tagged package that is not registered fails the lane, and it caught a real near-miss where `[no tests to run]` counted as a pass). As part of this, the `lint-no-host-fs-tests` ratchet accepts location OR tag. The ratchet relaxation and the tag mechanism revert as one unit if you want to veto the approach.

## Verification

- Full stable suite green in the worktree, lint clean (`golangci-lint run --new-from-merge-base origin/main`).
- Containerized integration lane green at `CAMP_TEST_POOL_SIZE=4`, including the new quiescence, seed-fallback, conflict, and registration suites.
- Not included here: the release itself. `promote-rc` for v0.4.0 still gates on the pre-existing `lint-no-host-fs-tests` failure on main, which #545 and #549 fix.


obey-agent (commented):
**Review (obey-agent)** — CS0005 cold seed / rsync probe / conflict lifecycle.

High-signal design choices look sound:

- **Pack-copy eligibility** requires quiescence + HEAD re-verify, copies only immutable `objects`/`refs` (+ `packed-refs`/`HEAD`), refuses git-dirs outside the peer campaign root (`withinRoot`), and falls back with reasons rather than aborting.
- **rsyncprobe** classifies openrsync *before* genuine rsync (load-bearing against Apple's dual banner), gates delta on protocol ≥ 30 at *both* ends, and degrades probe failures to whole-file instead of failing the sync.
- **artifacts resolve** take-local removes the path from the baseline (does not write local as agreed), which is the correct sticky-protection invariant; take-peer stages beside the tree and renames in; per-root lock matches pull.

Nits (non-blocking):
- `fetchOnePath` builds the remote as `RsyncSpec(root)[:-1] + "/" + path`. Correct today because `RsyncSpec` always ends in `/`, but a helper that joins root+rel without slice arithmetic would be less brittle.
- Resolve still invokes bare `rsync` rather than the probed binary path; fine if PATH is the intended resolution, but it can disagree with the engine reported on a full pull if PATH differs from LookPath during probe.

No high-conviction bug blocking merge from this pass. Large surface; CI on this branch would still be the strongest confidence signal.

PR #551, opened 2026-08-05
feat(triage): camp triage end to end, plus camp workitem split (CT0003)

lancekrogers:
Festival CT0003 (camp-triage), complete through acceptance. All 13 sequences and five phase gates done (each gate approved 4/4 on first submission); the final acceptance gate awaits Lance's recorded sign-off, which is the festival's designed human boundary. Code on this branch is final barring review findings.

## What this builds

`camp triage`: a run-based portfolio triage system over campaign workitems, per the spec in `workflow/design/camp-triage/` (D1-D5 confirmed; docs 04/05 amended during execution with dated rationale). All state lives under `.campaign/triage/runs/<run-id>/` as `triage/v1alpha1` documents; decisions are an append-only event stream and the current verdict of a row is the fold of its events.

- **Run store and start.** `camp triage start` with scope expressions, a frozen manifest snapshot (resolved profile and type policies embedded for reproducibility), batch partition, identity preflight, `status`, `abandon`. Locked, atomic writes throughout.
- **Judgment surface.** `queue`, `evidence set`, `propose` with per-type disposition vocabularies, a review gate, and `TRIAGE_REVIEW.md` / `PRIORITIES.md` rendered from run data (byte-stable across re-renders, hash-proven).
- **Approve loop.** `approve` with selectors and terminal safety; bulk amend validated before any of it records; partial approval is normal and applies incrementally.
- **Refresh.** Staleness classifier over re-checked anchors (fresh, moved, changed, gone, new), carry-forward invalidation, throttled/cached network edge with `unchecked-offline`.
- **Apply and verify.** Approved verdicts compile into an ordered plan; the executor routes on command kind through the `Mover` seam (workitem services for directory-backed rows, the idea service for file-backed intents), writes receipts, resumes idempotently. A halted apply exits nonzero naming what remains; verification counts approved-but-unapplied rows as `unapplied` and a run cannot reach `verified` while owing a decision.
- **Configuration layer.** `camp triage init` scaffolds a fully commented profile, five type policies, and OBEY.md via embed.FS; strict resolution from disk; golden tests pin the scaffold to spec doc 05.
- **Incremental runs.** A second `start` diffs against the base run, carries verdicts with a named reason for every row that did not carry, and freezes `mode`, `base_run_id`, and the vocabularies that produced each verdict.
- **DRIVER.md** generated per run — an executable agent brief whose test parses commands out of the brief and runs them, so doc drift fails loudly.
- **`camp workitem split`** with lineage markers both directions, a successors-before-archive gate enforced in the promote path, and a seed-hash-gated `--undo`.

## Acceptance outcome

The acceptance phase initially returned NO-GO on two release blockers its own suite exposed, both fixed on this branch and re-proven:

1. Intents (file-backed workitems) could not receive any applied verdict — the executor only knew directory-backed services. Fixed via `item_kind` frozen in the manifest, `CommandKindIdea`, and `Mover.MoveIdea` calling the same service `camp idea move` runs. Mixed design+intent fixture: `Applied 4 row(s)`, verify `4 checked, 4 matched, 0 mismatched`.
2. A partially applied run closed as `verified` with the failure invisible. Fixed as above; the honest empty case is pinned byte-for-byte.

Recommendation: **GO within fixture scope.** Four real-campaign items remain explicitly for Lance (bootstrap triage of the real campaign, the seven 2026-08-03 trial consolidations, document review, a real agent driving DRIVER.md), plus the recorded sign-off.

## Verification

- Full containerized integration lane green three times across the final phase; the run covering this exact HEAD: `ok github.com/Obedience-Corp/camp/tests/integration 1062.959s`, exit 0.
- Unit suite 117 packages, 0 failures. `golangci-lint --new-from-merge-base origin/main`: 0 issues. `go vet` clean across default, dev, integration, container_fs tags. `just docs` idempotent.
- Known pre-existing failure not from this branch: `lint-no-host-fs-tests` on two files byte-identical to main (fixed upstream by #545/#549).
- `camp workitem --json` proven byte-identical across 143 real workitems at multiple checkpoints.


cmd/camp/triage/apply.go
```diff
@@ -0,0 +1,332 @@
...
+	if err != nil {
+		return err
+	}
+
+	plan, err := triage.CompilePlan(triage.CompileInput{
+		RunID:    runID,
+		Rows:     run.Manifest.Rows,
+		Verdicts: verdicts,
+		// Phase 004 flips this when `camp workitem split` lands.
+		SplitAvailable: false,
```
obey-agent: **Bug / incomplete phase 004 wiring:** `camp workitem split` is already added in this PR (`internal/commands/workitem/split.go` + `workitem.go` registration), but apply still hardcodes `SplitAvailable: false` and omits `CompileInput.Successors` entirely.

Effect: every approved consolidation compiles as blocked (`requires camp workitem split`) on a binary that *does* have the verb. When you flip this for phase 004, also pass the successor map (review already has `SuccessorsFor`) so `splitArgv` does not fall back to the `<successors>` placeholder.

cmd/camp/triage/mover.go
```diff
@@ -0,0 +1,122 @@
...
+		// practical race; the cleaner fix is recorded in the review.
+		result.Commit = git.HeadSHA(ctx, m.campaignRoot)
+	}
+	return result, nil
+}
+
+// Split is phase 004's verb. Until it lands the compiler marks consolidation
+// entries blocked, so the executor never reaches this; returning an error
+// rather than a silent no-op keeps a wiring mistake loud.
+func (m *serviceMover) Split(_ context.Context, stableID string, _ []string) (triage.MoveOutcome, error) {
```
obey-agent: **Stub still present while the real command lands in this branch.** `serviceMover.Split` always returns validation error. That is the right fail-loud shape *before* the verb exists; once `SplitAvailable` becomes true, this must delegate to the same service path `runSplit` uses (no subprocess), and produce an honest `MoveOutcome.Undo` (split --undo / reverse lineage) the way Stage/Promote already do.

obey-agent (changes requested):
**Review (obey-agent)** — draft CT0003 mid-festival.

Architecture (run store, append-only decisions, refresh-before-apply, no-new-movers via serviceMover → priority/promote services, agent_allowed annotations, pty harness) is coherent and well-tested at the package level.

**Must fix before marking ready (phase 004 incomplete relative to this branch's own contents):**

1. `camp workitem split` is registered in this PR (`newSplitCommand` + `workitem.go`), but `camp triage apply` still hardcodes `SplitAvailable: false` and never supplies `CompileInput.Successors`. Approved consolidations always compile as blocked (`requires camp workitem split`) even though the CLI verb exists on the same binary.
2. `serviceMover.Split` is still a hard error stub. Once `SplitAvailable` flips, apply will either never reach it (if still blocked) or fail loudly (if flagged true without a real implementation).

Before ready-for-review: wire `Successors` (e.g. the same `SuccessorsFor` the review surface uses), set `SplitAvailable: true` when the verb is present, and implement `serviceMover.Split` by calling the same service path `camp workitem split` uses — not a subprocess — matching the Stage/Promote pattern.

Also note: apply's dry-run path compiles the plan *before* refresh and without successors; that is fine while splits are blocked, but once enabled dry-run must show the same argv apply will run.

Not re-litigating the intentional mid-phase draft state — this is the checklist that still fails correctness if the PR is marked ready as-is.

internal/triage/compile.go
```diff
@@ -0,0 +1,267 @@
...
+
+	if !in.SplitAvailable {
+		entry.Blocked = blockedNeedsSplit
+		if len(successors) > 0 {
+			entry.Blocked += " (successors: " + strings.Join(successors, ", ") + ")"
+		}
+		return entry
+	}
+
+	entry.Commands = []PlanCommand{
```
obey-agent, line 106: **Bug:** When `SplitAvailable` is true and `successors` is empty, this still builds an executable plan. `splitArgv` then emits `--into <successors>` (placeholder). Apply's `runCommand` does `argsAfterFlags(..., "--into")`, so the mover receives `[]string{"<successors>"}` — not empty — and `serviceMover.Split`'s `len==0` guard is bypassed. Result: apply can create a workitem named literally `<successors>`.

Block here when `len(successors)==0` (e.g. `entry.Blocked = "consolidate requires successors in the rationale"`) instead of compiling commands.

internal/triage/compile.go
```diff
@@ -0,0 +1,267 @@
...
+// promoteArgv promotes a workitem to a rail or dungeon target.
+func promoteArgv(row ManifestRow, target string) []string {
+	return []string{"camp", "workitem", "promote", row.StableID, "--target", target}
+}
+
+// splitArgv splits a parent into its declared successors.
+func splitArgv(row ManifestRow, successors []string) []string {
+	argv := []string{"camp", "workitem", "split", row.StableID}
+	if len(successors) == 0 {
+		return append(argv, "--into", splitSuccessorPlaceholder)
```
obey-agent, line 242: Placeholder is fine for *display* of an incomplete consolidation, but it must not appear on an argv that `apply` will execute. Prefer blocking in `compileSplitEntry` (or requiring successors at propose) so this branch is only used by non-executing render paths, if at all.

internal/triage/propose.go
```diff
@@ -0,0 +1,280 @@
...
+
+func (r *Rationale) kind() string    { return "rationale" }
+func (r *Rationale) version() string { return r.SchemaVersion }
+
+// Validate implements Document.
+func (r *Rationale) Validate() []Violation {
+	var out []Violation
+	out = append(out, checkRequired("summary", r.Summary)...)
+	out = append(out, checkEnum("confidence", string(r.Confidence), Confidences())...)
+	return out
```
obey-agent, line 56: `Rationale.Validate` never requires `Successors`. For a `consolidate` disposition that is the only place the apply plan reads the successor list from. Validate (or Propose, which knows the disposition) should refuse consolidate without at least one non-empty successor name, so the empty path never reaches compile/apply.

obey-agent (changes requested):
**Review (obey-agent)** — re-review on head `6bfcca20bf` (was `570d71ad26`).

### Prior CR items: fixed
- `SplitAvailable: true` and `Successors` fed from `store.SuccessorsByRow`
- `serviceMover.Split` delegates to `SplitWorkitem` (no subprocess) with `--undo` receipt
- Dungeoned successors count for the retirement gate; `InDungeonPath` matches both spellings
- Integration `TestTriageConsolidate_EndToEnd` covers dry-run + apply + lineage

### Remaining bug (must fix before ready)
Empty successor lists are not refused at propose or compile time. `splitArgv` emits a display placeholder `--into <successors>` that the executor then treats as a real argument: `runCommand` pulls every `--into` value, so `serviceMover.Split`'s `len==0` guard never fires, and apply can create a workitem literally named `<successors>`.

Fix: block empty consolidations at compile (`entry.Blocked = ...`) and/or require `rationale.successors` when disposition is consolidate. Keep the placeholder for pure rendering if needed, but never put it on an executable plan argv.

internal/triage/compile.go
obey-agent, line 106: **Still open from prior CR:** with `SplitAvailable: true` and empty `successors`, this builds executable commands. `splitArgv` emits `--into <successors>`; apply's `argsAfterFlags(..., "--into")` turns that into a real name and bypasses the mover's empty-list guard.

Block here when `len(successors)==0` instead of compiling commands, e.g. `entry.Blocked = "consolidate requires successors in the rationale"`.

internal/triage/compile.go
obey-agent, line 242: Placeholder is still on the executable path. Keep for display-only if needed, but never on an argv `apply` will run.

internal/triage/propose.go
obey-agent, line 56: Still no validation that consolidate rationales carry successors. `Validate` only checks summary/confidence. Refuse consolidate without ≥1 non-empty successor at propose (disposition-aware), so the empty path never reaches compile.

obey-agent (changes requested):
**Review (obey-agent)** — re-review on head `367cc5e0ae` (prior CR was on `6bfcca20bf`).

### New since last review
- Profile scaffold (`camp triage init`, `camp triage profile`, `scaffold.Ensure`, embedded OBEY.md + typed policies) looks solid: never overwrites, reports divergence, embeds resolved profile in the manifest, good integration coverage.
- `AdvancePhase` absorbing `ErrInvalidInput` (including `reviewGapError`) so partial approval is not failed by an unmet judging→reviewing gate is the right semantics; real I/O still propagates.

### Prior CR **not fixed**
Empty consolidate successors still compile as an **executable** plan with argv `--into <successors>`. Apply extracts that as a real successor name, bypasses `serviceMover.Split`'s `len==0` guard, and can create a workitem literally named `<successors>`.

Still required before ready:
1. Block in `compileSplitEntry` when `len(successors)==0` (or never emit placeholder on executable argv).
2. Require `rationale.successors` when disposition is `consolidate` at propose time.

internal/triage/compile.go
obey-agent, line 106: **Still open:** with `SplitAvailable: true` and empty `successors`, this builds executable commands. `splitArgv` emits `--into <successors>`; apply's `argsAfterFlags(..., "--into")` treats that as a real name and bypasses the mover's empty-list guard.

Block when `len(successors)==0` (e.g. `entry.Blocked = "consolidate requires successors in the rationale"`) instead of compiling commands.

internal/triage/compile.go
obey-agent, line 242: Placeholder still lands on executable argv. Display-only is fine; apply must never run this branch.

internal/triage/propose.go
obey-agent, line 56: Still no disposition-aware validation that consolidate rationales include ≥1 successor. Close this at propose so the empty path never reaches compile/apply.

obey-agent (changes requested):
**Review (obey-agent)** — re-review on head `affc10bfcf` (prior CR on `367cc5e0ae`).

### New since last review
- Carry-forward + stale notice cache, `DRIVER.md` generation, `camp status` triage banner, propose uses embedded manifest policy — sensible additions; integration for driver looks good.
- Nit (non-blocking): `camp status` uses `triage.DefaultProfile().Runs.StaleAfterDays` rather than the campaign's resolved profile, so a customized `stale_after_days` in `.campaign/triage/profile.yaml` will not affect the banner threshold.

### Prior CR **still open** (blocking)
Empty consolidate successors still compile as an **executable** plan with argv `--into <successors>`. Apply extracts that as a real successor name and can create a workitem literally named `<successors>`. `Rationale.Validate` still does not require successors for consolidate.

Please block empty consolidations at compile and/or refuse at propose before ready.

internal/triage/compile.go
obey-agent, line 106: **Still open:** with `SplitAvailable: true` and empty `successors`, this builds executable commands. `splitArgv` emits `--into <successors>`; apply's `argsAfterFlags(..., "--into")` treats that as a real name and bypasses any empty-list guard on the mover.

Block when `len(successors)==0` (e.g. `entry.Blocked = "consolidate requires successors in the rationale"`) instead of compiling commands.

internal/triage/compile.go
obey-agent, line 242: Placeholder still lands on an executable argv. Display-only is fine; apply must never run this branch.

internal/triage/propose.go
```diff
@@ -0,0 +1,291 @@
...
```
obey-agent, line 56: Still no disposition-aware validation that consolidate rationales include ≥1 non-empty successor. Close this at propose (where disposition is known) so the empty path never reaches compile/apply. `Validate()` alone is disposition-blind; Propose should refuse consolidate without successors even if summary/confidence are set.

obey-agent (changes requested):
**Review (obey-agent)** — re-review on head `21dd42fb04` (prior CR on `affc10bfcf`).

### New since last review
- **Incremental honesty** looks solid: carry only when a base verdict is actually approved (`VerdictNone` / unapproved proposals re-queue with a named loss), `manifest.Mode` flips to incremental when a base run is found, carry losses frozen into the manifest for `status`, and start text reports carried/queued + per-row losses. Integration tests cover the status/text contracts well.
- Prior **banner nit fixed**: `BannerFor(ctx, path, now)` honors the campaign profile threshold (regression test present).
- CLI reference regen is mechanical; fine.

### Prior CR **still open** (blocking)
Empty consolidate successors still compile as an **executable** plan with argv `--into <successors>`. Apply extracts that as a real successor name via `argsAfterFlags(..., "--into")` and can create a workitem literally named `<successors>`. `Rationale.Validate` still does not require successors for consolidate, and `compileSplitEntry` still builds commands whenever `SplitAvailable` is true even with `len(successors)==0`.

Please block empty consolidations at propose and/or compile (set `entry.Blocked`) before ready — do not leave the placeholder on an argv apply will run.

internal/triage/compile.go
obey-agent, line 106: **Still open across heads:** with `SplitAvailable: true` and empty `successors`, this still builds executable commands. `splitArgv` emits `--into <successors>`; apply's `argsAfterFlags(..., "--into")` treats that as a real name.

Block when `len(successors)==0` (e.g. `entry.Blocked = "consolidate requires successors in the rationale"`) instead of compiling commands.

internal/triage/compile.go
obey-agent, line 242: Placeholder still on the executable argv path. Display-only is fine; apply must never run this branch. Unchanged on this head.

internal/triage/propose.go
obey-agent, line 56: Still no disposition-aware check that consolidate rationales include ≥1 non-empty successor. Close at propose so the empty path never reaches compile/apply.

tests/integration/triage_acceptance_test.go
```diff
@@ -0,0 +1,437 @@
...
+// the selector coincide.
+//
+// All four labels the shipped intent policy offers fail this way. Intents are
+// the idea inbox, which is the largest class in a real campaign and the thing
+// the field trial that produced this design was triaging.
+//
+// Correct behaviour: an approved verdict on an intent applies, or triage
+// refuses to offer a disposition it cannot execute. Either way `apply` must
+// not report success having applied nothing.
+func TestAcceptance_KnownBlocker_IntentRowsCannotBeApplied(t *testing.T) {
```
obey-agent: Good pin. Correct behaviour (as the comment says): either intent selectors in the plan resolve, or triage refuses dispositions it cannot execute — apply must not exit 0 after applying nothing on an approved intent. Please fix before ready; leave the test as a regression once green.

tests/integration/triage_acceptance_test.go
```diff
@@ -0,0 +1,437 @@
...
+// It is the partial case that misreports, and the partial case is the one a
+// real campaign hits, because a campaign holds both designs and intents.
+//
+// The review criterion this phase is measured against reads "verify reports
+// zero unexplained mismatches after each apply". A row that was approved and
+// never applied is the definition of unexplained, and it is counted as neither.
+//
+// Correct behaviour: a run with an approved verdict that did not apply is not
+// verified, and verify says which rows and why.
+func TestAcceptance_KnownBlocker_PartialApplyStillVerifiesClean(t *testing.T) {
```
obey-agent: Good pin of the partial-apply verify hole. A run with approved-but-unapplied rows must not report `verified` / zero mismatches by omission. Fix verify (or phase transition) so failed/missing applies are unexplained mismatches; keep this as regression.

obey-agent (changes requested):
**Review (obey-agent)** — re-review on head `73dd4fafd0` (prior CR on `21dd42fb04`).

### New since last review
- **CT0003 acceptance suite** (`tests/integration/triage_acceptance_test.go`): solid. Zero-mismatch-after-each-apply, phase timing budget, review regen, and the two **KnownBlocker** pins are honest instrumentation of real defects rather than greenwash.
- Pinned blockers (also blocking readiness, not just docs):
  1. **Intent apply selector** — plan uses bare stable id; file-backed intents need `intent:…` form, so apply stops and can report success after applying nothing.
  2. **Partial apply → verify clean** — failed rows never produce `applied` receipts, so verify omits them and status can still reach `verified`.

### Prior CR **still open** (blocking)
Empty consolidate successors still compile as an **executable** plan with argv `--into <successors>`. Production code for `compileSplitEntry` / `splitArgv` / `Rationale.Validate` is unchanged on this head; only the acceptance test file landed. Apply still treats the placeholder as a real `--into` value.

Please block empty consolidations at propose and/or compile, and treat the two newly pinned blockers as release criteria before marking ready.

internal/triage/compile.go
```diff
@@ -0,0 +1,340 @@
...
+
+	if !in.SplitAvailable {
+		entry.Blocked = blockedNeedsSplit
+		if len(successors) > 0 {
+			entry.Blocked += " (successors: " + strings.Join(successors, ", ") + ")"
+		}
+		return entry
+	}
+
+	entry.Commands = []PlanCommand{
```
obey-agent, line 106: **Still open:** with `SplitAvailable: true` and empty `successors`, this builds executable commands. `splitArgv` emits `--into <successors>`; apply treats that as a real name.

Block when `len(successors)==0` (e.g. `entry.Blocked = "consolidate requires successors in the rationale"`) instead of compiling commands. Intent/verify fixes on this head do not close this path.

internal/triage/compile.go
```diff
@@ -0,0 +1,340 @@
...
+// promoteArgv promotes a workitem to a rail or dungeon target.
+func promoteArgv(row ManifestRow, target string) []string {
+	return []string{"camp", "workitem", "promote", row.StableID, "--target", target}
+}
+
+// splitArgv splits a parent into its declared successors.
+func splitArgv(row ManifestRow, successors []string) []string {
+	argv := []string{"camp", "workitem", "split", row.StableID}
+	if len(successors) == 0 {
+		return append(argv, "--into", splitSuccessorPlaceholder)
```
obey-agent, line 242: Placeholder still on the executable argv path. Display-only is fine; apply must never run this branch.

internal/triage/propose.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+func (r *Rationale) kind() string    { return "rationale" }
+func (r *Rationale) version() string { return r.SchemaVersion }
+
+// Validate implements Document.
+func (r *Rationale) Validate() []Violation {
+	var out []Violation
+	out = append(out, checkRequired("summary", r.Summary)...)
+	out = append(out, checkEnum("confidence", string(r.Confidence), Confidences())...)
```
obey-agent, line 55: Still no disposition-aware check that consolidate rationales include ≥1 non-empty successor. Close at propose (disposition is known there) so the empty path never reaches compile/apply.

obey-agent (changes requested):
**Review (obey-agent)** — re-review on head `db41a98378` (prior CR on `73dd4fafd0`). PR is no longer draft.

### Fixed since last review (good)
1. **Intent apply** — manifest freezes `item_kind`; compiler routes file-backed rows to `camp idea move` (`CommandKindIdea` + `MoveIdea`); intent type policy drops attention-only labels (`parked`) for idea-lifecycle-aligned `ready`/`active`/`someday`. Mixed-run acceptance test now expects apply+verify green.
2. **Partial/halted verify** — unapplied approved rows counted as `VerificationUnapplied` (unexplained); run stays out of `verified`. CLI apply exits non-zero on `result.Halted`. Unit pin in `TestVerifyCountsAnApprovedRowThatNeverApplied`.

These two release blockers from the acceptance suite look correctly closed.

### Prior CR **still open** (blocking)
Empty consolidate successors still compile as an **executable** plan with argv `--into <successors>`. `compileSplitEntry` / `splitArgv` / `Rationale` propose path are unchanged on that axis. Apply still extracts the placeholder via `argsAfterFlags(..., "--into")` as a real successor name.

Please block empty consolidations at propose and/or compile (`entry.Blocked`) before merge — do not leave the placeholder on an argv apply will run.

obey-agent:
The remaining blocking finding from the re-review is closed at 340bd3e5, in two layers:

1. **Compile**: `compileSplitEntry` now blocks an approved consolidation that declared no successors — `entry.Blocked` names the missing successors, no commands are emitted, so the placeholder can never land on a runnable argv. Pinned by `TestCompilePlanBlocksConsolidationWithoutSuccessors` (SplitAvailable=true, empty successors, entry not executable).
2. **Validate**: `ApplyPlanEntry.validate` rejects any command argv containing the `<successors>` placeholder, so a stale or hand-edited plan document cannot smuggle it past apply either. Pinned by `TestApplyPlanValidateRejectsSuccessorPlaceholder`.

`splitArgv`'s placeholder branch survives for the review-render path only (`ApplyCommandFor`), where reading as incomplete is its documented purpose.

Gates: triage suite green, golangci 0 issues, full `just test` exit 0. Note: `just lint`'s host-FS-test check fails on this branch for `nothing_staged_test.go` and `tree_empty_test.go` — pre-existing at db41a983 (verified via stash), unrelated to CT0003, and already being migrated on a separate branch.

obey-agent (commented):
**Review (obey-agent)** — re-review on head `340bd3e559d7` (prior CR on `db41a98378`).

### Prior CR — fixed
**Empty consolidate successors** no longer compile executable. `compileSplitEntry` returns early with `blockedNoSuccessors` when `len(successors)==0` (with `SplitAvailable` either true or false), so the display placeholder never lands on an argv apply will run. Defense in depth: `ApplyPlanEntry.validate` rejects any stored plan whose split argv still contains `splitSuccessorPlaceholder`. Unit pins: `TestCompilePlanBlocksConsolidationWithoutSuccessors`, `TestApplyPlanValidateRejectsSuccessorPlaceholder`.

### Residual (non-blocking)
- Propose still does not disposition-aware refuse consolidate without successors; incomplete approvals now fail closed at compile/validate, which is enough for apply safety. A propose-time check would only improve the authoring UX.
- `splitArgv` keeping the placeholder for review-render paths is fine given the block + validate gates.

**Verdict: COMMENT** — no remaining product bugs on this head. The long-standing apply-placeholder hole is closed.

PR #552, opened 2026-08-07
fix: stale conflict-tally assertion, and stop a collapsed run reporting as failed tests

lancekrogers:
## What prompted this

A `just test integration` run reported 21 failed tests out of 44 in 160.50s. None of the 21 are broken.

A full run of the same suite on an idle machine, same pool config:

```
PASS=896  FAIL=1  SKIP=8
--- FAIL: TestArtifactsPullOverSSH_TransfersAndProtects (29.89s)
FAIL	github.com/Obedience-Corp/camp/tests/integration	1148.529s
```

Every one of the 21 passes. The reported run collapsed after ~5% of the suite (160s against 1148s) and the rest never executed.

## The real failure

`conflictCount()` landed in 2ae0e892 (CS0005, #550) so sync output renders "1 conflict kept local" rather than "1 conflicts kept local":

```go
func conflictCount(n int) string {
	if n == 1 {
		return "1 conflict"
	}
	return fmt.Sprintf("%d conflicts", n)
}
```

The assertion in `artifacts_from_ssh_test.go:64` dates from 685a0385, three weeks earlier, and hardcodes the plural. The fixture creates exactly one conflict, so it can never match again:

```
Error: "...(synced; 1 conflict kept local)..." does not contain "conflicts kept local"
```

Now asserts `"1 conflict kept local"`, which pins the pluralization rather than working around it.

## Why 21 healthy tests were reported as failures

`tests/integration/main_test.go` deliberately turns Docker-headroom death into skips: once the container pool stops resetting, the run arms an infrastructure banner and every later checkout calls `t.Skip`. Its own comment records a prior incident of "474 of 572 tests failing" this way on a branch where nothing was wrong.

The summary in `internal/buildutil/tasks/integration.go` switched on `run`, `pass`, and `fail` — there was no `skip` case, so skipped tests were dropped from the totals entirely. The denominator shrank to whatever ran, and a dead run rendered as "23/44 tests passed" beside 21 red rows: a healthy-looking suite with a few broken tests, which is the inverse of what happened.

Skips now count toward the total and are named when present. That same run reads:

```
23/905 tests passed (861 skipped)
```

This is the failure mode `classifyRunFailure` was already written for — the summary is the only thing most people read after a run, so its count has to mean what it says.

## Not changed

Pool sizing. Pool 4 on a 4-CPU Colima VM passed 896/905 here with no infrastructure banner, so 25e8a4a1's tuning looks right. The collapse points to contention on the VM, and the harness already names the lever (`CAMP_TEST_POOL_SIZE`).

## Verification

- `just gate` — PASSED (both profiles: builds, vet stable/dev/integration, lint stable/dev, dev + stable unit tests)
- `just test integration-run TestArtifactsPullOverSSH_TransfersAndProtects` — `ok ... 61.698s`
- `go test ./internal/buildutil/...` — ok, including the new `TestTestTally` table

### DIFF
--- a/internal/buildutil/tasks/integration.go
+++ b/internal/buildutil/tasks/integration.go
@@ -33,7 +33,14 @@ type IntegrationResult struct {
 	Duration    time.Duration
 	TestsPassed int
 	TestsFailed int
-	FailedTests []string // Names of failed tests
+	// TestsSkipped is counted because the integration harness turns
+	// infrastructure death into skips: once the container pool cannot be
+	// reset, every later test t.Skips rather than failing. Left uncounted, a
+	// run that collapsed after 44 of 905 tests reported "23/44 tests passed",
+	// which reads as a healthy suite with a few broken tests instead of a run
+	// that never happened.
+	TestsSkipped int
+	FailedTests  []string // Names of failed tests
 	// SuiteError is a failure of the run itself rather than of any test: a
 	// build error, a panic that took the process down, a timeout. Kept apart
 	// from FailedTests because it is not a test and must not be counted as
@@ -112,7 +119,7 @@ func Integration(verbose bool) error {
 		start := time.Now()
 
 		var pass bool
-		var testsPassed, testsFailed int
+		var testsPassed, testsFailed, testsSkipped int
 		var failedTests []string
 		var suiteError string
 		var packageFailed bool
@@ -236,6 +243,10 @@ func Integration(verbose bool) error {
 						if !strings.Contains(event.Test, "/") {
 							testsFailed++
 						}
+					case "skip":
+						if !strings.Contains(event.Test, "/") {
+							testsSkipped++
+						}
 					}
 				} else if event.Action == "fail" {
 					// A fail with no test name is the package failing as a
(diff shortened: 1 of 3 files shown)


obey-agent (commented):
## Review (obey-agent)

**Verdict:** good fix — both halves are high-signal and the explanation matches the code.

### 1. Conflict assertion (`artifacts_from_ssh_test.go`)
Matches `conflictCount(1)` → `"1 conflict"` then `"… (synced; %s kept local)"` → `"1 conflict kept local"`. Pinning the singular form is the right assertion (not a looser substring that would pass for any N).

### 2. Skip tally (`integration.go` + `TestTestTally`)
- Counts top-level `"skip"` events with the same `!strings.Contains(event.Test, "/")` rule as pass/fail — consistent.
- Denominator includes skips; headline becomes `N/M tests passed (K skipped)` only when K>0.
- Table tests lock the collapsed-run string that prompted the change.

### Notes (non-blocking)
- `pass` still ignores skips (correct: skip ≠ fail).
- Per-suite row status is unchanged; only the totals line gains the skip callout — enough for the failure mode described.
- Unit-test suite tally in `test.go` is intentionally out of scope here (integration-only skip-as-infra-death story).

No bugs found in the introduced paths. COMMENT only.

PR #553, opened 2026-08-09
fix(machine): make the Tailscale approval link visible and openable

lancekrogers:
## The bug

`camp machine` could not complete the one hop failure a person can fix in one click.

Pressing <kbd>enter</kbd> on a Tailscale SSH machine opens the hop overlay, which rendered ssh's whole complaint as a single line and clamped it to the box width. Captured in a pty against a real check-mode machine, before this change:

```
╭──────────────────────────────────────────────────────────────────────╮
│  Hop to archdtop                                                     │
│                                                                      │
│  ✗ Tailscale SSH requires a one-time browser check — open https://l  │
│                                                                      │
│  'camp machine diagnose archdtop' reports auth, probe, and socket s   │
│                                                                      │
│  r retry · esc cancel                                                │
╰──────────────────────────────────────────────────────────────────────╯
```

The approval URL is cut off after `https://l`, and no key opens it.

Three more defects behind that one:

- The detail pane (<kbd>t</kbd>) did show the URL, but wrapped it mid-token at ≤100 columns — a link split across two lines is one you cannot double-click or copy.
- **The link it showed was wrong.** `ParseTailscaleCheckURL` stopped only at whitespace, so re-parsing camp's own formatted sentence (`open <url>, approve, …`) captured the trailing comma. The pane rendered `https://login.tailscale.com/a/l40ade643a9805,,` — a URL that 404s if you copy it.
- `camp machine diagnose` printed `HINT: look for a login.tailscale.com check URL` while holding the URL it had just been handed, and reported the machine as generically `unavailable`.

## The fix

Carry the URL as a value instead of scraping it back out of prose: `machineHealth.CheckURL` and `machineHopState.checkURL`, both filled from a new `remote.TailscaleCheckURL`. A value can be given its own line and acted on by a key; a substring of an error message cannot.

- <kbd>o</kbd> opens the link in the browser, <kbd>c</kbd> copies it (the fallback for a terminal with no browser). Both keys are contextual — they are listed in the footer only while an approval is actually waiting.
- Both surfaces put the bare URL on its own line, **giving up its indent rather than wrap**. Two columns decide whether a check URL fits on one line in an 80-column terminal, and one line is what keeps it selectable.
- The hop overlay wraps errors instead of truncating them, and measures with the same width the frame clamps with (`overlayTextWidth`), so a body can no longer wrap to a width the frame then cuts again.
- `ParseTailscaleCheckURL` trims trailing punctuation, which makes classification idempotent — the property every display surface depends on and none of them had.
- `camp machine diagnose` prints the URL under `APPROVE`, says `blocked on a Tailscale SSH check` instead of `unavailable`, and drops the now-redundant hint. `--json` gains `check_url` (additive, `omitempty`).

## After

Hop overlay:

```
╭──────────────────────────────────────────────────────────────────────╮
│  Hop to archdtop                                                     │
│                                                                      │
│  ✗ Needs Tailscale SSH check                                         │
│    Tailscale wants a one-time browser approval before archdtop will  │
│    let camp in.                                                      │
│    https://login.tailscale.com/a/l8018113a54bf                       │
│    o open it · c copy it · r retry · esc cancel                      │
╰──────────────────────────────────────────────────────────────────────╯
```

Detail pane at 80 columns, where the link previously split across two lines:

```
│ ✗ Needs Tailscale SSH check                    │
│   Tailscale wants a one-time browser approval. │
│   https://login.tailscale.com/a/l22c325a3a6650 │
│   o open it · c copy it · t try again          │
```

`camp machine diagnose archdtop`:

```
CAMP  blocked on a Tailscale SSH check
APPROVE  https://login.tailscale.com/a/l13f96c153a68c2
  open it, approve, then run this again
```

## Recording

Real binary, real PTY, full color. `camp machine` driven through both surfaces that used to lose the URL, then `o` and `c`, then the log of what camp actually handed the platform:

camp machine showing the Tailscale approval URL whole on its own line in both the detail pane and the hop overlay, then o and c handing that exact URL to the platform opener and clipboard

Reproduce it: `just tui pty-machine-tailscale-check` asserts the journey and builds its own fixture; `CAMP_VHS_ROOT=$FIXTURE just vhs record-color docs/demos/machine-tailscale-check.tape` re-records it.

The fixture supplies a stub `ssh` that answers the way Tailscale SSH does in check mode, plus stub opener/clipboard tools that log what they receive. No live tailnet is contacted and the approval URL in the recording is invented, so nothing real is published. camp's parse, render, and keybinding path is the real binary throughout.

## Verification

Green tests do not verify a TUI, so this was driven against a real check-mode machine (`archdtop`, `auth_method: tailscale-ssh`) in a pty with pyte, per `docs/tribal-knowledge/verify-tui-changes-in-a-pty.md`:

- The link renders whole on one line at 78, 80, 100 and 130 columns. 78 is the dual-pane threshold and the tightest case; below it the screen goes single-pane and gets wider.
- <kbd>o</kbd> and <kbd>c</kbd> were verified end to end with `open` and `pbcopy` stubbed on `PATH`, so nothing launched: both received the exact URL, `https://login.tailscale.com/a/l38df05f3ae181`, with no trailing comma.

The recording is backed by a validated evidence bundle (`just vhs bundle`): optimized artifact, ffprobe metadata, pyte snapshots for all five screen states, dependency versions, and a `PASS` privacy scan. The GIF is published as a secret Gist and its SHA-256 matches the validated artifact.

New unit coverage: parser idempotence (the comma bug), `TailscaleCheckURL` across the exit and timeout error paths, the URL surviving the overlay frame's own clamp, one-line rendering through `View` at 80 columns, wrap-not-truncate in a pane too narrow to fit the link, and `o`/`c` dispatch including the no-approval-pending case.

`just gate` passes: 4555 dev + 4510 stable tests.

## Scope note

Two things here are wider than the reported failure and can be split out:

- The generic (non-Tailscale) hop error is now wrapped rather than truncated. Same truncation bug, not Tailscale-specific.
- `just vhs` recipes are anchored at the repo root. They ran from the module directory, so the documented reproduce line could not find its own tape, and the absolute-path workaround in `docs/demos/README.md` wrote the recorder's home directory into `manifest.json` — which `scan-privacy.py` correctly failed. Producing this PR's evidence needed it fixed.


internal/remote/ssh.go
```diff
@@ -369,14 +369,24 @@ func ParseTailscaleCheckURL(text string) (string, bool) {
...
 		return "", false
 	}
-	return rest[:end], true
+	return url, true
 }
 
+// urlTrailingPunctuation is punctuation that can follow a URL in prose but is
+// never part of a Tailscale check URL. Trimming it is what makes this function
+// idempotent, and idempotence is the whole point: camp formats the URL into a
+// sentence ("open <url>, approve, then retry"), and every display surface
```
obey-agent, line 384: Idempotence fix is correct and necessary: every display surface re-parses camp’s own “open <url>, approve…” sentence. Without TrimRight the comma became part of the URL and 404’d. TestParseTailscaleCheckURLIsIdempotent locks this in.

internal/ui/browser.go
```diff
@@ -0,0 +1,50 @@
...
+		return camperrors.New("no url opener known for " + runtime.GOOS)
+	}
+	if err := cmd.Start(); err != nil {
+		return camperrors.Wrap(err, "opening "+url)
+	}
+	// Reap in the background: the opener is not ours to wait on (xdg-open can
+	// live as long as the browser it spawned), but an unwaited child is a
+	// zombie for the life of the TUI.
+	go func() { _ = cmd.Wait() }()
+	return nil
```
obey-agent, line 34: Good safety bar: URLs come from remote stderr, and macOS `open` will act on file paths / arbitrary schemes. http(s)-only + argv (no shell) is the right model. Reaping via goroutine Wait avoids zombies for the TUI lifetime.

obey-agent (commented):
## Review (obey-agent)

**Verdict: comment only** — high-quality fix for a real operator-facing failure. No blocking bugs found.

### Root cause → fix mapping (all three defects covered)
1. **Truncated hop overlay** — `hopBody` no longer clamps the whole ssh complaint to one line; check-mode is a dedicated branch with the URL on its own line and `o`/`c`/`r` keys. `overlayTextWidth()` matches the frame clamp so wrap + frame do not double-truncate.
2. **Wrong URL (trailing comma)** — `ParseTailscaleCheckURL` now `TrimRight`s prose punctuation and is covered by `TestParseTailscaleCheckURLIsIdempotent` against camp’s own `formatTailscaleCheckDetail` sentence. That was a real 404-on-copy bug; the regression test is the right shape.
3. **No open action** — `ui.OpenInBrowser` (http(s)-only) + `WriteClipboard` with keys `o`/`c` in browse and hop, and footer copy that only advertises them while a check URL is waiting.

### Also good
- `TailscaleCheckURL(err)` as a first-class value (not buried in the detail sentence) is the right API for every surface (TUI health, hop, diagnose `check_url` JSON field).
- Diagnose table: APPROVE leads; generic HINT suppressed when the exact URL is already printed — avoids the “go look for a URL” noise while holding one.
- `OpenInBrowser` refuses non-http(s) schemes (URLs come from remote stderr) and reaps the opener in a goroutine — correct for TUI lifetime.
- Solid tests: pane/view/overlay keep URL intact, open/copy no-op without a pending URL, parser idempotence, timeout+exit error paths.

### Non-blocking notes
- `WriteClipboard` still has no Windows branch (pre-existing in `internal/ui/clipboard.go`); open path does handle Windows via `cmd /c start`. Headless/WSL operators still have the printed URL.
- `isWebURL` is prefix-only (`http://` / `https://`). Fine given the parser only emits `https://login.tailscale.com/…`; not a hole for this call path.

LGTM to merge from a code-review standpoint (tests look green-worthy; I did not re-run the suite here).

cmd/camp/machine_tui_view.go
```diff
@@ -516,10 +607,23 @@ func (m *machineTUIModel) hopBody() []string {
...
 	switch {
+	case m.hop.checkURL != "":
+		// Check mode is not a hop failure the operator debugs, it is one they
+		// finish. Lead with the link and the key that opens it; the diagnose
+		// pointer below is for the failures that need investigating, not this one.
+		body = append(body, "", machineErrorStyle.Render("✗ Needs Tailscale SSH check"))
+		body = append(body, checkURLLines(m.hop.checkURL, m.overlayTextWidth(),
+			"Tailscale wants a one-time browser approval before "+m.hop.machineID+" will let camp in.",
+			"o open it · c copy it · r retry · esc cancel")...)
+		return body
```
obey-agent, line 618: Right precedence: check-mode is not “debug the hop”, it is “finish the approval”. Leading with bare URL + o/c and skipping the diagnose pointer here matches the product intent.

internal/remote/ssh.go
obey-agent, line 384: Idempotence fix is correct and necessary: every display surface re-parses camp's own "open <url>, approve…" sentence. Without TrimRight the comma became part of the URL and 404'd. TestParseTailscaleCheckURLIsIdempotent locks this in.

internal/ui/browser.go
obey-agent, line 34: Good safety bar: URLs come from remote stderr, and macOS open will act on file paths / arbitrary schemes. http(s)-only + argv (no shell) is the right model. Reaping via goroutine Wait avoids zombies for the TUI lifetime.

cmd/camp/machine_tui_view.go
obey-agent, line 618: Right precedence: check-mode is not "debug the hop", it is "finish the approval". Leading with bare URL + o/c and skipping the diagnose pointer here matches the product intent.

obey-agent (commented):
## Review (obey-agent)

**Verdict: comment only** — high-quality fix for a real operator-facing failure. No blocking bugs found.

### Root cause → fix mapping (all three defects covered)
1. **Truncated hop overlay** — `hopBody` no longer clamps the whole ssh complaint to one line; check-mode is a dedicated branch with the URL on its own line and o/c/r keys. `overlayTextWidth()` matches the frame clamp so wrap + frame do not double-truncate.
2. **Wrong URL (trailing comma)** — `ParseTailscaleCheckURL` now TrimRights prose punctuation and is covered by `TestParseTailscaleCheckURLIsIdempotent` against camp's own formatTailscaleCheckDetail sentence. That was a real 404-on-copy bug; the regression test is the right shape.
3. **No open action** — `ui.OpenInBrowser` (http(s)-only) + WriteClipboard with keys o/c in browse and hop, and footer copy that only advertises them while a check URL is waiting.

### Also good
- `TailscaleCheckURL(err)` as a first-class value (not buried in the detail sentence) is the right API for every surface (TUI health, hop, diagnose check_url JSON field).
- Diagnose table: APPROVE leads; generic HINT suppressed when the exact URL is already printed — avoids the "go look for a URL" noise while holding one.
- `OpenInBrowser` refuses non-http(s) schemes (URLs come from remote stderr) and reaps the opener in a goroutine — correct for TUI lifetime.
- Solid tests: pane/view/overlay keep URL intact, open/copy no-op without a pending URL, parser idempotence, timeout+exit error paths.

### Non-blocking notes
- WriteClipboard still has no Windows branch (pre-existing in internal/ui/clipboard.go); open path does handle Windows via cmd /c start. Headless/WSL operators still have the printed URL.
- isWebURL is prefix-only (http:// / https://). Fine given the parser only emits https://login.tailscale.com/...; not a hole for this call path.

LGTM to merge from a code-review standpoint (tests look green-worthy; I did not re-run the suite here).

tests/tui/machine_tailscale_check_pty.py
```diff
@@ -0,0 +1,238 @@
...
+    # ---- the writes, which no recording can assert ----
+    with open(handoff) as fh:
+        handed = [line.strip() for line in fh if line.strip()]
+    # Transcript only, never the snapshots array: these lines were read from a
+    # file, and screen-snapshots.json is a record of what pyte rendered. The
+    # assertion below is the real check either way.
+    transcript.append("===== handoff.log =====")
+    transcript.extend(handed)
+
+    if ("open %s" % CHECK_URL) not in handed:
```
obey-agent: Blocking on Linux: `OpenInBrowser` uses `xdg-open` (not `open`) when GOOS != darwin/windows. Fixture installs both, but camp only execs the GOOS tool, so handoff.log will contain `xdg-open <url>` here and this assert fails on the usual CI host. Accept `open` **or** `xdg-open`.

tests/tui/machine_tailscale_check_pty.py
```diff
@@ -0,0 +1,238 @@
...
+        failures.append("the opener was not handed the exact URL: %r" % handed)
+    if ("pbcopy %s" % CHECK_URL) not in handed:
```
obey-agent: Blocking on Linux: `WriteClipboard` uses `xclip` (stdin) on non-darwin. `record-handoff` correctly logs `xclip <url>`, but this requires `pbcopy`. Accept `pbcopy` **or** `xclip` (same pattern as opener).

obey-agent (changes requested):
## Review (obey-agent) — re-review on head `937d29c8691f`

**Verdict: request changes** — the product fix from `f653df6a8484` still looks good; the new test/demo commits introduce a **platform-hardcoded handoff assertion** that will fail on Linux (the usual CI host).

### What the new commits add (good)
- Shared fixture builder (`just tui machine-check-fixture`) with stub `ssh` that prints a real check-mode banner and `exec sleep 20` so camp's kill hits the pipe holder (excellent comment).
- `record-handoff` correctly dispatches by **basename** so `xclip -selection clipboard` still logs stdin payload.
- VHS tape + pty evidence for single-line URL + o/c keys.
- `vhs.just` anchors at repo root — real fix for module-cwd path bugs.

### Blocking: handoff asserts assume Darwin tool names

`internal/ui/browser.go` / `clipboard.go` pick tools by `runtime.GOOS`:

| GOOS | open | clipboard |
|------|------|-----------|
| darwin | `open` | `pbcopy` |
| linux (default) | `xdg-open` | `xclip` |

But `tests/tui/machine_tailscale_check_pty.py` requires:

```python
if ("open %s" % CHECK_URL) not in handed: ...
if ("pbcopy %s" % CHECK_URL) not in handed: ...
```

On Linux CI, camp will log `xdg-open <url>` and `xclip <url>` (fixture installs all four names; camp only calls the GOOS ones). Both assertions fail even when the handoff is correct.

`just tui pty` now includes this check, so a green Linux `pty` suite is broken until this is fixed.

### Suggested fix
Accept either platform name, e.g.:

```python
def handed_url(handed, tools, url):
    return any(f"{t} {url}" in handed for t in tools)

if not handed_url(handed, ("open", "xdg-open"), CHECK_URL):
    failures.append(...)
if not handed_url(handed, ("pbcopy", "xclip"), CHECK_URL):
    failures.append(...)
```

(or assert `any(line.endswith(CHECK_URL) for line in handed)` with a tool-prefix check).

### Non-blocking
- Product code from the previous head is unchanged and still LGTM.
- Invented URL in the ssh stub is fine (no live secret leakage).
- Kill-via-exec-sleep rationale in the stub is correct.

Once the handoff assertions are OS-agnostic, this is ready.

obey-agent (commented):
## Response to re-review of `937d29c8` — head is now `e54cb7bc`

**Blocking finding confirmed and fixed.** The report was correct: `ui.browserOpenCommand` and `ui.WriteClipboard` both switch on `runtime.GOOS`, so on Linux camp logs `xdg-open <url>` / `xclip <url>` while the check demanded `open` / `pbcopy`. Since `just tui pty` now includes this check, the whole pty suite was red on Linux — and archdtop, the machine that motivated this PR, is Arch.

Fixed in `ea05730a` by deriving the expected pair from the host rather than accepting either name:

```python
def expected_handoff_tools():
    if sys.platform == "darwin":
        return "open", "pbcopy"
    return "xdg-open", "xclip"
```

I went with the exact per-platform pair over the suggested `any(...)` form for one reason: `("open", "xdg-open")` would also pass a camp that called the *wrong* tool for its platform. The stricter version still fails that, and the assertion stays a real claim about the mapping rather than about the fixture.

**Verified both halves** (no Linux host available, so this is stated as two checks, not one end-to-end run):

- `expected_handoff_tools()` under a forced `sys.platform` → `darwin -> ('open', 'pbcopy')`, `linux -> ('xdg-open', 'xclip')`.
- `record-handoff` invoked under the Linux names produces exactly the lines the check looks for, including the basename dispatch that keeps `xclip -selection clipboard` reading stdin rather than logging its flags:

```
xdg-open https://login.tailscale.com/a/f1e2d3c4b5a6
xclip https://login.tailscale.com/a/f1e2d3c4b5a6
```

`just tui pty` is green on darwin (both the triage-review checks and this one).

**Also in this push, self-reported:** `ea05730a` committed a `__pycache__/*.pyc` — a dev artifact that should never have been in the tree. Removed in `e54cb7bc`, which also adds `__pycache__/` and `*.pyc` to `.gitignore`; nothing covered python bytecode before because the pty harnesses were the first python here imported rather than only executed.

**Non-blocking notes acknowledged, deliberately not acted on:**

- `WriteClipboard` has no Windows branch. Pre-existing, and camp does not support Windows yet, so on that platform the copy path would fail confusingly while `OpenInBrowser` works. Worth its own issue rather than widening this PR — this one has already grown past the reported failure once.
- `isWebURL` being prefix-only is intentional. The only producer is `ParseTailscaleCheckURL`, which requires the `https://login.tailscale.com/` prefix; the guard exists to stop a remote host choosing a non-web scheme, not to be a general URL validator.

**Unchanged:** no product code moved since `f653df6a`. The published recording still reflects current behavior, so it has not been re-recorded — the tape and camp's rendering are byte-identical, and only a python assertion changed.

obey-agent (commented):
## Re-verified on head `e54cb7bc`. No code change in this push

The blocking finding from the `937d29c8` review was already fixed and pushed in
`ea05730a`, so there was nothing left to change here. What was missing was a run
of the suite against the current head rather than a claim about it, so this is
that run.

`just tui pty` on darwin, current head:

```
driving .../projects/camp/bin/camp on a pty (44x130)

[PASS] approve: the screen reports the verdict
[PASS] approve: the verdict is written to decisions.jsonl
[PASS] approve: the recorded action is the one the card showed

[PASS] resume: the first session's verdict survived the kill
[PASS] resume: the lane offers the row left undecided
[PASS] resume: it does not re-offer the row already decided

[PASS] lane approve: both non-terminal rows were recorded
[PASS] lane approve: the terminal row got NO verdict
[PASS] lane approve: the consolidation got NO verdict
[PASS] lane approve: the confirmation listed what it covered

all pty checks passed
machine-tailscale-check: 5 snapshots, link intact and handed off
```

The assertion the review flagged now derives the expected pair from the host
instead of hardcoding the macOS names:

```python
def expected_handoff_tools():
    if sys.platform == "darwin":
        return "open", "pbcopy"
    return "xdg-open", "xclip"
```

so the Linux failure described in the review, camp logging `xdg-open <url>` and
`xclip <url>` against a check demanding `open` and `pbcopy`, can no longer
happen.

Stated precisely, because it matters for how much this run proves: there is
still no Linux host available here, so the darwin branch is verified end to end
and the Linux branch rests on the two checks in the previous comment
(`expected_handoff_tools()` under a forced `sys.platform`, and `record-handoff`
producing exactly those lines under the Linux tool names).

The only thing outstanding is bookkeeping: the `CHANGES_REQUESTED` decision
still points at `937d29c8`, two pushes back, and needs clearing before this can
merge.

PR #554, opened 2026-08-09
feat(stageguard): keep undeclared nested repos out of commits

lancekrogers:
## Why

A stage-everything commit could record an untracked nested git repository as a gitlink (mode `160000`). When `.gitmodules` does not declare that path, git treats the entry as **fatal rather than skippable**: every later `git submodule` call in the repository, and in every clone of it, exits non-zero.

This is not hypothetical. Three review worktrees were checked out inside `festival-app`, a blanket `camp p commit` swept them in, and because none appeared in `.gitmodules`:

```
git submodule update --init --recursive
fatal: No url found for submodule path
'projects/worktrees/deep-review-20260711/festival-app-pr136' in .gitmodules
error: recipe `sync` failed on line 3 with exit code 128
```

That is the first step of the repo's `just dev`, so the app could not be started for eighteen days. It built and tested fine the whole time, so the breakage surfaced only as "the app will not launch." Fixed there in Obedience-Corp/festival-app#144; this is the cause.

## The blind spot

`stageguard.Enumerate` drops directory entries, on a documented assumption:

> Directories never appear: -uall expands untracked directories into their individual files.

That is not true for an embedded repository. Verified against real git:

```
$ git status --porcelain=v1 -uall
?? file.txt
?? nested/repo/          <-- the directory, never expanded

$ git add -A
warning: adding embedded git repository: nested/repo
$ git ls-files -s
100644 45b983be... 0	file.txt
160000 32d3431e... 0	nested/repo
```

Git reports the directory alone *precisely because* staging it records a gitlink rather than contents. So the single entry that becomes a gitlink was the single entry the guard discarded at `enumerate.go:169`.

## Reusing what camp already knows

`internal/git.ListOrphanedGitlinks` already compares index gitlinks against `.gitmodules`, for doctor, clone, and sync. It was never consulted at the staging chokepoint, which is the only place that can prevent the state instead of reporting it afterward. This applies the same comparison one step earlier, before the entry exists, rather than adding a parallel notion of the same thing.

## Behavior

Excluded and reported, not blocked, in both scopes:

```
⚠ review/app-pr136 is its own git repository at a210413; it was left out of this commit
  Committing it records a submodule reference with no url behind it, which
  makes 'git submodule' fail in this repo and every clone of it. Its contents
  would not reach the remote either.
    if it should be a submodule   git submodule add <url> review/app-pr136
    if it does not belong here    echo 'review/app-pr136/' >> .gitignore
    if you meant it as a gitlink  --commit-nested
```

The commit still lands; everything beside the nested repo goes in.

**Why exclude rather than block.** Unlike a large file, there is no question camp cannot settle. The nested repository keeps its own history and remote either way, so exclusion costs the user nothing, and an undeclared gitlink has no legitimate reading: the user either wanted a submodule, which needs a url, or wanted the directory left alone. Per `02-size-and-batch-guard.md`, blocking is the exception that must justify itself, and this one cannot.

**A declared submodule is exempt.** That is the one case where staging a gitlink is correct, and `camp project add` depends on it. The exemption matches on the normalized path, because `git config` returns a `.gitmodules` value verbatim while `git status` always reports the normalized one: a hand-edited `path = ./vendor/lib` would otherwise miss the lookup and get the submodule excluded.

- `commit.guards.nested_repos`: `exclude` (default), `block`, `off`, with per-project overrides and allowlist support.
- `--commit-nested` on `camp commit`, `camp project commit`, `camp worktrees commit`.
- `--json` gains `nested_repo` and `nested_repo_blocked` exclusion reasons.
- `pkg/commitkit` exports `NestedRepo` and `ModeExclude` so fest renders it too.

## Two deliberate calls

**`--commit-nested` is separate from `--commit-large`.** They authorize unrelated things. A user who decided a 30 MB binary belongs in git has not decided to embed a foreign repository, and overloading a size-named flag to also mean that would make it mean more than it says. `runStageGuard` now suppresses per guard rather than returning early, so `--commit-large` leaves this guard running. Pinned by `TestIntegration_CommitLargeDoesNotAuthorizeNestedRepos`.

**`large_files` now validates its own subset.** Adding `ModeExclude` to `Mode.Valid()` would otherwise let `large_files: exclude` through to a guard with no handling for it, resolving to behavior the user never asked for. Prior behavior is unchanged: `Valid()` already accepted exactly `auto`, `block`, `off`. Pinned by `TestLargeFilesRejectsExcludeMode`.

## Verification

`just gate-fast`: **4559/4559 tests, 0 lint issues**, both ratchets clean (`lint-no-host-fs-tests`, `lint-no-fmt-errorf`).

10 containerized integration tests in `tests/integration/stage_guard_nested_test.go`, all passing. The headline one asserts the actual user-visible symptom rather than just the index:

```go
_, exitCode, err := tc.ExecCommand("git", "-C", campPath, "submodule", "update", "--init", "--recursive")
assert.Equal(t, 0, exitCode)
```

The suite is not tautological: `TestIntegration_NestedRepoOffModeIsSilent` asserts the gitlink **is** recorded when the guard is off, which is exactly the pre-fix behavior, so both states are pinned. `TestIntegration_DeclaredSubmoduleWithDotSlashIsNotFlagged` and its unit-level counterpart were both confirmed to fail against the previous exact-string path comparison before the normalization landed.

Also verified by hand against a real campaign and the built binary: default exclude, `--commit-nested` opt-in, declared-submodule exemption, block mode staging nothing, and an invalid mode producing a validation error instead of a silent fallback.

## Note on docs

Docs regen is limited to the two per-command files. `camp-reference.md` carries unrelated pre-existing drift (it is missing `camp artifacts resolve`, which already has a per-command doc on `main`), so folding a full regen in here would mix unrelated content into this diff. Worth a separate regen commit.

### DIFF
--- a/internal/git/stageguard.go
+++ b/internal/git/stageguard.go
@@ -31,6 +31,15 @@ type StageOptions struct {
 	// naming it --allow-large would read as "skip the safety check", which is
 	// backwards.
 	CommitLarge bool
+	// CommitNested forces undeclared nested repositories into the index for
+	// this invocation, recording them as gitlinks.
+	//
+	// It is separate from CommitLarge rather than folded into it because the
+	// two authorize unrelated things. A user passing --commit-large has decided
+	// a specific big file belongs in git; they have not decided to embed a
+	// foreign repository, and silently granting that would make the size flag
+	// mean more than it says.
+	CommitNested bool
 }
 
 // StageOutcome is what the guard decided about a staging operation, returned
@@ -39,6 +48,11 @@ type StageOptions struct {
 type StageOutcome struct {
 	// Excluded holds over-threshold untracked files kept out of the index.
 	Excluded []stageguard.GuardViolation
+	// NestedRepos holds undeclared nested git repositories kept out of the
+	// index. They are excluded like Excluded is, but carry a different remedy
+	// and never take part in artifact-root declaration, so they are reported
+	// separately rather than merged into it.
+	NestedRepos []stageguard.GuardViolation
 	// Reported holds tracked files whose new content crossed the threshold.
 	// They are always staged; the user is told about them instead.
 	Reported []stageguard.GuardViolation
@@ -62,10 +76,13 @@ type StageOutcome struct {
 // Empty reports whether the guard found nothing worth telling the user about.
 func (o *StageOutcome) Empty() bool {
 	return o == nil ||
-		(len(o.Excluded) == 0 && len(o.Reported) == 0 && o.Unavailable == nil)
+		(len(o.Excluded) == 0 && len(o.NestedRepos) == 0 &&
+			len(o.Reported) == 0 && o.Unavailable == nil)
 }
 
-// ExcludedPaths returns the repo-relative paths kept out of the index.
+// ExcludedPaths returns the repo-relative paths of over-threshold files kept
+// out of the index. Nested repositories are reached through NestedRepos, since
+// callers act on the two differently.
 func (o *StageOutcome) ExcludedPaths() []string {
 	if o == nil {
 		return nil
@@ -75,6 +92,26 @@ func (o *StageOutcome) ExcludedPaths() []string {
 		paths = append(paths, v.Path)
 	}
 	return paths
+}
+
+// NestedRepoPaths returns the repo-relative paths of nested repositories kept
+// out of the index.
+func (o *StageOutcome) NestedRepoPaths() []string {
+	if o == nil {
+		return nil
+	}
+	paths := make([]string, 0, len(o.NestedRepos))
+	for _, v := range o.NestedRepos {
+		paths = append(paths, v.Path)
+	}
+	return paths
+}
+
+// excludedPathspecs is every path the guard keeps out of this staging call,
+// regardless of which guard decided it. One list, so exclusion stays a single
+// git add.
+func (o *StageOutcome) excludedPathspecs() []string {
+	return append(o.ExcludedPaths(), o.NestedRepoPaths()...)
 }
 
 // GuardBlockedError reports that the guard refused a staging operation. It
@@ -100,6 +137,13 @@ func (e *GuardBlockedError) Error() string {
 			}
 		}
 		return "staging refused: bulk directory"
+	case stageguard.NestedRepo:
+		paths := make([]string, 0, len(e.Violations))
+		for _, v := range e.Violations {
+			paths = append(paths, v.Path)
+		}
+		return "staging refused: nested git repositories not declared in .gitmodules: " +
+			strings.Join(paths, ", ")
 	default:
 		paths := make([]string, 0, len(e.Violations))
 		for _, v := range e.Violations {
@@ -163,11 +207,13 @@ func runStageGuard(ctx context.Context, repoPath string, files []string, opts St
 	if !isStageEverything(files) {
 		return nil, nil, nil
 	}
-	// --commit-large is the user overruling camp's decision to exclude, which
-	// is the thing that actually needs an override once handling is automatic.
-	// It suppresses detection entirely rather than excluding-then-restoring,
-	// so the index ends up exactly as an unguarded stage would leave it.
-	if opts.CommitLarge {
+	// The override flags are the user overruling camp's decision to exclude,
+	// which is the thing that actually needs an override once handling is
+	// automatic. Each suppresses its own guard's detection rather than
+	// excluding-then-restoring, so the index ends up exactly as an unguarded
+	// stage would leave it. Only when every guard is overridden is the whole
+	// evaluation skipped.
+	if opts.CommitLarge && opts.CommitNested {
 		return nil, nil, nil
 	}
 
@@ -176,7 +222,16 @@ func runStageGuard(ctx context.Context, repoPath string, files []string, opts St
 		return &StageOutcome{Unavailable: camperrors.Wrapf(
 			err, "resolve staging guard limits for %s", repoPath)}, nil, nil
 	}
-	if limits.LargeFiles == stageguard.ModeOff && limits.Bulk == stageguard.ModeOff {
+	if opts.CommitLarge {
+		limits.LargeFiles = stageguard.ModeOff
+		limits.Bulk = stageguard.ModeOff
+	}
+	if opts.CommitNested {
+		limits.NestedRepos = stageguard.ModeOff
+	}
+	if limits.LargeFiles == stageguard.ModeOff &&
+		limits.Bulk == stageguard.ModeOff &&
+		limits.NestedRepos == stageguard.ModeOff {
 		return nil, nil, nil
 	}
 
@@ -192,6 +247,7 @@ func runStageGuard(ctx context.Context, repoPath string, files []string, opts St
 	outcome := &StageOutcome{Limits: limits}
 	var bulk []stageguard.GuardViolation
 	var overThreshold []stageguard.GuardViolation
+	var nested []stageguard.GuardViolation
 
 	for _, v := range violations {
 		switch v.Kind {
@@ -201,6 +257,8 @@ func runStageGuard(ctx context.Context, repoPath string, files []string, opts St
 			outcome.Reported = append(outcome.Reported, v)
 		case stageguard.OverThreshold:
 			overThreshold = append(overThreshold, v)
+		case stageguard.NestedRepo:
+			nested = append(nested, v)
 		}
 	}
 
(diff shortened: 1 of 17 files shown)


internal/stageguard/enumerate.go
```diff
@@ -33,28 +39,40 @@ type Candidate struct {
...
 func Enumerate(ctx context.Context, repoPath string) ([]Candidate, error) {
+	candidates, _, err := enumerate(ctx, repoPath)
+	return candidates, err
+}
+
+// enumerate returns the file candidates and, separately, the untracked
+// directories git declined to expand. Both come from one status call: running
+// it twice would double the cost of every guarded stage to learn something the
+// first call already reported.
+func enumerate(ctx context.Context, repoPath string) ([]Candidate, []string, error) {
```
obey-agent, line 50: This is the load-bearing change: keep unexpanded untracked directories from the same status call instead of dropping them. Under -uall those entries are exactly the embedded-repo case git refuses to expand because staging would record a gitlink.

internal/stageguard/nested.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	return violations, nil
+}
+
+// isGitRepo reports whether dir holds a .git entry. It accepts both a directory
+// (an ordinary clone) and a file (a worktree or submodule checkout, where .git
+// is a gitdir pointer), because git records a gitlink for either and both are
+// how the nested checkouts that motivated this guard actually arrive.
+func isGitRepo(dir string) bool {
+	_, err := os.Lstat(filepath.Join(dir, ".git"))
```
obey-agent, line 68: Accepting both .git directory and gitdir *file* is correct for worktree/submodule-style nested checkouts (the festival-app failure shape). Lstat-only is enough; no need to resolve the gitdir.

internal/git/stageguard.go
```diff
@@ -31,6 +31,15 @@ type StageOptions struct {
...
 	CommitLarge bool
+	// CommitNested forces undeclared nested repositories into the index for
+	// this invocation, recording them as gitlinks.
+	//
+	// It is separate from CommitLarge rather than folded into it because the
+	// two authorize unrelated things. A user passing --commit-large has decided
+	// a specific big file belongs in git; they have not decided to embed a
+	// foreign repository, and silently granting that would make the size flag
+	// mean more than it says.
+	CommitNested bool
```
obey-agent, line 42: Keeping CommitNested orthogonal to CommitLarge is the right API. Silently letting --commit-large authorize undeclared gitlinks would reintroduce the failure mode under a flag name that does not say so.

internal/git/stageguard.go
```diff
@@ -216,9 +274,22 @@ func runStageGuard(ctx context.Context, repoPath string, files []string, opts St
...
+	}
 
+	// Kept in its own field rather than folded into Excluded: everything in
+	// Excluded is a file the size guard held back, and the campaign-root path
+	// answers it by declaring an artifact root around it. A nested repository
+	// must never reach that code. Declaring an artifact root over a git
+	// repository is precisely the boundary artifacts.ResolveAutoRoot already
+	// refuses, so folding them together would turn a clear finding into a
+	// refusal about the wrong subject.
 	outcome.Excluded = overThreshold
```
obey-agent, line 290: Important separation: NestedRepos must not fold into Excluded, or campaign-root auto artifact-root declaration would try to own a nested git repo. Separate field + excludedPathspecs() merge for staging only is the right shape.

obey-agent (commented):
## Review (obey-agent)

**Verdict: comment only** — correct fix for a real production failure mode (festival-app#144 class). No blocking bugs found.

### Why this is the right chokepoint
The diagnosis is accurate and well evidenced: under `git status -uall`, an embedded repo is reported as the directory alone, and `Enumerate` used to drop directories. That meant the single path that becomes a mode-160000 gitlink was the single path the size/bulk guard never saw. Wiring detection into the existing stage path (not doctor-after-the-fact) is the only place that *prevents* the broken state.

### What looks solid
1. **Detection criteria** — untracked dir + `.git` present (dir *or* gitdir file for worktrees) + path not declared in `.gitmodules` + allowlist. Declared submodules are the one legitimate gitlink case; integration test covers that (`TestIntegration_DeclaredSubmoduleIsNotFlagged`).
2. **Default exclude, not block** — nested repos keep their own history; excluding costs the user nothing. Separate from `--commit-large` (integration test asserts the flags stay independent).
3. **Reporting separated from size exclusions** — `NestedRepos` never enters `Excluded` / artifact-root declaration. That avoids turning a nested-repo finding into a wrong auto-root refusal.
4. **Pathspecs** — `excludedPathspecs()` merges size + nested exclusions into one `git add` with `:(exclude,literal)`.
5. **Config** — `nested_repos: exclude|block|off`; invalid modes error (no silent fallback); no meaningless `auto`.
6. **Surfaces** — campaign / project / worktree commit flags, human report with remedies, JSON `nested_repo` / `nested_repo_blocked`, block-mode refusal copy.
7. **Tests** — unit coverage (sort order, allow globs, defaults, mode validation, large_files still rejects exclude) + integration regression that asserts `git submodule update --init --recursive` still exits 0 after commit.

### Non-blocking notes
- `declaredSubmodulePaths` shells out to avoid the `internal/git` import cycle; documented and consistent with Enumerate. Fine.
- `.gitmodules` path matching is exact-string after slash-normalize. Unusual submodule path spellings (leading `./`) would need the same spelling in status; git normally normalizes these the same way.
- I did not re-run the integration suite here; the test design matches the failure story closely enough to trust if CI is green.

LGTM from review.

obey-agent (commented):
## Acted on the `.gitmodules` spelling note. Head is now `5f4a372d`

The review filed this one as non-blocking, on the reasoning that "git normally
normalizes these the same way". I checked that before deciding whether to act,
and it does not hold for the call this code makes: `git config` returns the raw
file text, and only git's own submodule machinery normalizes it downstream.

```
$ cat .gitmodules
[submodule "foo"]
	path = ./projects/foo

$ git config -f .gitmodules -z --get-regexp '^submodule\..*\.path$'
submodule.foo.path
./projects/foo

$ git status --porcelain=v1 -uall
?? projects/foo/
```

So the declared set held `./projects/foo` while the guard looked up
`projects/foo`, the lookup missed, and camp excluded a submodule that was
correctly declared. That is precisely the failure this guard exists to avoid
causing, so it was worth fixing here rather than leaving as a note.

The old code already collapsed a trailing slash, which is why the plain
`vendor/lib/` spelling worked and hid the gap.

### Fix

`normalizeDeclaredPath` renders a declared path in the single spelling git
status uses, via `path.Clean`, so `./`, doubled separators, interior dot
segments, and trailing slashes all collapse to the same key. A value that cleans
to `.` names the repository root, which is never a submodule, so it declares
nothing.

The status side needs no equivalent treatment: it comes from git itself, which
always emits clean repo-relative slash paths, and `statCandidate` already strips
the trailing slash git puts on an unexpanded directory.

### Verification

Both new tests fail against the previous exact-string comparison, so they are
real regression tests rather than restatements of current behavior. Reverting
the normalization gives:

```
--- FAIL: TestNormalizeDeclaredPathMatchesStatusSpelling/dot-slash_prefix
    normalizeDeclaredPath("./vendor/lib") = "./vendor/lib", want "vendor/lib"

--- FAIL: TestIntegration_DeclaredSubmoduleWithDotSlashIsNotFlagged
    "... vendor/lib is its own git repository at c6e7fd1; it was left out of
    this commit ..." should not contain "is its own git repository"
```

That second failure is the user-visible bug in full: a declared submodule
reported as an undeclared nested repo and dropped from the commit.

With the fix, `just test integration-run TestIntegration_DeclaredSubmodule`
passes both the existing and the new case, and `just gate-fast` is green
(build, vet across stable/dev/integration, lint 0 issues, 4559 unit tests).

### Not acted on

- `declaredSubmodulePaths` shelling out to avoid the `internal/git` import
  cycle. Deliberate and documented in the function comment; changing it would
  restore the cycle the leaf position exists to break.
- The integration suite was re-run for the nested-repo tests specifically, not
  in full. `just gate` covers the rest.

obey-agent (commented):
## Review (obey-agent) — re-review of head `5d83a8d6dd23`

**Verdict: comment only** — no blocking bugs. Head moved since prior reviews (`10ce61181e41` → `5f4a372d7863` normalize fix → merge `main`).

### Delta since first review
The earlier non-blocking note about `.gitmodules` path spelling was correctly **promoted to a real fix**. Evidence holds:

- `git config -f .gitmodules` returns the **raw** value (`./vendor/lib`); only git's submodule machinery normalizes later.
- Status reports `vendor/lib`. Exact-string lookup would false-exclude a **declared** submodule — the one legitimate gitlink case.
- `normalizeDeclaredPath` via `path.Clean` + slash normalize collapses `./`, trailing `/`, doubled separators, and interior `.` segments. Values that clean to `.` declare nothing (root is never a submodule).

Unit + integration coverage for the `./` spelling is the right shape: both fail against the pre-fix comparison, so they pin the actual bug rather than restating current behavior.

### Rest of the design (still sound)
1. Enumerate keeps unexpanded untracked dirs from the same `-uall` status call (load-bearing).
2. Detection: untracked dir + `.git` (dir or gitdir file) + not in declared set + allowlist.
3. Default **exclude**, orthogonal to `--commit-large` / `--commit-nested`.
4. `NestedRepos` kept out of size-`Excluded` so campaign auto artifact-root declaration cannot claim a nested git repo.
5. Mode validation rejects nonsense / large-files-only `auto`; invalid config errors instead of silent fallback.
6. Integration asserts the user symptom: `git submodule update --init --recursive` still exits 0 after commit.

### Non-blocking
- Merge commit only rebased onto `main`; no further nested-guard logic changes beyond the spelling fix.
- Absolute / drive-letter paths in hand-edited `.gitmodules` would still not match status (pathological; not worth special casing).
- Did not re-run the suite here; trust CI + the documented `just gate-fast` / nested integration results on the fix commit.

LGTM on current head.

PR #555, opened 2026-08-10
fix(machine): copy approval links through remote terminals

lancekrogers:
## Summary

- copy Tailscale approval URLs to the operator clipboard over OSC 52 during SSH sessions
- preserve native clipboard commands for local sessions, with terminal clipboard fallback when they are unavailable
- make the success message explain the remaining approve-and-retry flow
- strengthen unit and PTY coverage so the exact URL must cross the real terminal stream

## Root cause

The earlier machine TUI fix invoked pbcopy on macOS or xclip elsewhere. In an SSH or headless session those commands either target the remote host clipboard or fail because no graphical display exists. Its PTY fixture stubbed those binaries, so it only proved that a command ran, not that the local operator clipboard received the URL.

## User impact

Pressing c on a waiting Tailscale approval link now reaches the clipboard of the computer running the terminal, including remote Camp sessions. Local sessions retain their native clipboard behavior.

## Verification

- just gate
- just tui pty-machine-tailscale-check
- go test ./internal/ui ./cmd/camp

### DIFF
--- a/cmd/camp/machine_tui_update.go
+++ b/cmd/camp/machine_tui_update.go
@@ -361,7 +361,7 @@ func (m *machineTUIModel) copyCheckURL(url string) tea.Cmd {
 		m.setError(camperrors.Wrap(err, "could not copy the link"))
 		return nil
 	}
-	m.setAdvice("copied the approval link to your clipboard")
+	m.setAdvice("copied the approval link; paste it into a browser, approve it, then press t (or r) to retry")
 	return nil
 }
 
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.25.6
 require (
 	github.com/Obedience-Corp/obey-shared v0.4.6
 	github.com/alecthomas/chroma/v2 v2.23.1
+	github.com/aymanbagabas/go-osc52/v2 v2.0.1
 	github.com/charmbracelet/bubbles v1.0.0
 	github.com/charmbracelet/bubbletea v1.3.10
 	github.com/charmbracelet/glamour v0.10.0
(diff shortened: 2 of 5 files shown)


obey-agent (commented):
## Review (obey-agent)

**Verdict: comment only** — correct fix for the remote-session clipboard failure mode. No blocking bugs.

### Why this is right
Earlier machine TUI copy used `pbcopy`/`xclip`, which in SSH target the **remote** host (or fail with no display). Stubbing those binaries in the PTY fixture only proved a command ran, not that the operator clipboard got the URL.

This switches remote sessions to **OSC 52** so the sequence rides the terminal stream to the local machine. Local sessions still prefer native clipboard, with OSC 52 as fallback when native is missing.

### What looks solid
1. **Remote detection** — `SSH_CONNECTION` / `SSH_TTY` seam is the right signal; native path skipped entirely when remote.
2. **Fallback chain** — local native success short-circuits; dual failure joined with `errors.Join` for actionable errors.
3. **Screen wrapping** — `STY` → `sequence.Screen()` is correctly narrower than `TERM=screen-*` (tmux often wants plain OSC 52 with `set-clipboard`).
4. **Windows** — `cmd /c clip` fills a previous darwin/linux-only gap in `writeNativeClipboard`.
5. **Tests** — unit coverage for remote-only OSC 52, local prefer native, fallback, dual failure; PTY fixture forces remote env and asserts **decoded OSC 52 payload equals exact URL**, and fails if `pbcopy`/`xclip` appear in the handoff log.
6. **Advice string** — copy → paste → approve → `t`/`r` retry is the actual operator flow.

### Non-blocking
- tmux without `set-clipboard` / OSC passthrough can still drop the sequence (environment limitation, not a logic bug in this PR).
- `clipboardOutput` defaults to `os.Stdout`; fine for camp's Bubble Tea PTY usage; if a future path detached stdout from the controlling TTY, OSC 52 would need `/dev/tty`.
- Did not re-run `just gate` / PTY here; design of the PTY assertion is high-confidence.

LGTM.

